PR #1, opened 2017-11-01
Initial commits for geomstats module

ninamiolane:


rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+
+    return regularized_transfo
+
+
+def inverse(transfo):
+    """
+    Compute the group inverse in SE(3).
+
+    Formula:
+    TODO(nina).
```
johmathe: remove

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+
+
+def jacobian_translation(transfo, left_or_right='left'):
+    """
+    Compute the jacobian matrix of the differential
+    of the left/right translations
+    from the identity to transfo in the Lie group SE(3).
+
+    Formula:
+    TODO(nina).
```
johmathe: syntax: TODO(nina): add formula

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+
+    prod_transfo[0:3] = rotations.rotation_vector_from_rotation_matrix(
+            prod_rot_mat)
+    prod_transfo[3:6] = np.dot(rot_mat_1, translation_2) + translation_1
+
+    prod_transfo = regularize_transformation(prod_transfo)
+    return prod_transfo
+
+
+def jacobian_translation(transfo, left_or_right='left'):
```
johmathe: what about direction instead of left_or_right?

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+
+    elif left_or_right == 'right':
+        jacobian_rot = rotations.jacobian_translation(rot_vec,
+                                                      left_or_right='right')
+        jacobian[:3, :3] = jacobian_rot
+        jacobian[3:, :3] = -rotations.skew_matrix_from_vector(translation)
+        jacobian[3:, 3:] = np.eye(3)
+
+    else:
+        raise ValueError('\'left_or_right\' needs \'left\' or \'right\'')
```
johmathe: you should probably do the argument checking at the beginning: 

assert left_or_right in ('left', 'right')

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+        ref_point = regularize_transformation(ref_point)
+        jacobian = jacobian_translation(ref_point, left_or_right='left')
+        transfo_near_id = compose(inverse(ref_point), transfo)
+        tangent_vector_from_id = group_log(transfo_near_id)
+        tangent_vector = np.dot(jacobian, tangent_vector_from_id)
+
+    return tangent_vector
+
+
+# --- Riemannian
```
johmathe: comment is kinda useless. if you really want to make it clear that these are specific, maybe put this in a riemannian.py file? more of a nit though

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+
+
+def inner_product(coef_rotations, coef_translations):
+    """
+    Compute a 6x6 diagonal matrix, where the diagonal is formed by:
+    coef_rotations * [1, 1, 1] and coef_translations * [1, 1, 1].
+
+    Formula:
+    TODO(nina): put block diagonal matrix.
```
johmathe: it seems that there is a pattern that you're not really into putting these formulaes. Honestly the code is clear enough for these small functions you don't really need the formula. Potentially the big ones. If I were you I would remove all formulas from comments :)

rigid_transformations.py
```diff
@@ -0,0 +1,611 @@
...
+    mean_transformation[0:3] = mean_rotation
+    mean_transformation[3:6] = np.dot(np.linalg.inv(matrix), translation)
+    return mean_transformation
+
+
+def riemannian_variance(ref_point, transfo_vectors, weights,
+                        left_or_right='left',
+                        inner_product=ALGEBRA_CANONICAL_INNER_PRODUCT):
+    """
+
```
johmathe: comment would be nice :)

rotations.py
```diff
@@ -0,0 +1,415 @@
+"""Computations on the Lie group of 3D rotations."""
+
+import numpy as np
+
+GROUP_IDENTITY = np.zeros(3)
```
johmathe: why is it a zero matrix here?

rotations.py
```diff
@@ -0,0 +1,415 @@
...
+    """
+    assert len(vec) == 3
+
+    skew_vec = np.array([[0, -vec[2], vec[1]],
+                         [vec[2], 0, -vec[0]],
+                         [-vec[1], vec[0], 0]])
+    return skew_vec
+
+
+# -- Conversions: rotation matrix <-> rotation vector
```
johmathe: still not a fan of comments here, they can go out of date fairly quickly

rotations.py
```diff
@@ -0,0 +1,415 @@
...
+                   left_or_right='left',
+                   ref_point=GROUP_IDENTITY):
+    """
+    Compute the Riemannian exponential at point ref_point
+    of tangent vector tangent_vec wrt the metric obtained by
+    left/right translation
+    of the inner product inner_product at the Lie algebra.
+
+    Formula:
+    R*Exp(DL(R^-1)*a)
```
johmathe: you can use latex in these comments

rotations.py
```diff
@@ -0,0 +1,415 @@
...
+        raise ValueError('Param \'left_or_right\' '
+                         'should be \'left\' or \'right\'.')
+
+    return rot_vec_log
+
+
+# -- Statistics
+
+
+def riemannian_mean(rotation_vectors, weights, epsilon=1e-5):
```
johmathe: yay

rotations.py
```diff
@@ -0,0 +1,415 @@
...
+                                       rotation_vectors,
+                                       weights)
+
+        if mean_rotations_diff < epsilon * variance:
+            break
+
+    return mean_rotation
+
+
+def riemannian_variance(ref_rotation_vector, rotation_vectors, weights):
```
johmathe: nit: since you're using the variance in the mean, define the function before

unittests.py
```diff
@@ -0,0 +1,199 @@
+"""Unit tests to check the maths functions."""
```
johmathe: yay, unit tests!

unittests.py
```diff
@@ -0,0 +1,199 @@
...
+        rot_vec_4 = rotations.regularize_rotation_vector(rot_vec_4)
+        rot_vec_4_expected = rot_vec_4
+        self.assertTrue(np.allclose(rot_vec_4, rot_vec_4_expected))
+
+        rot_vec_5 = 1e-11 * np.array([12., 1., -81.])
+        rot_vec_5 = rotations.regularize_rotation_vector(rot_vec_5)
+        rot_vec_5_expected = rot_vec_5
+        self.assertTrue(np.allclose(rot_vec_5, rot_vec_5_expected))
+
+    def test_representations_of_rotations(self):
```
johmathe: you might want to split your tests a bit more. Like one test per assert, but that's just me being picky :)

unittests.py
```diff
@@ -0,0 +1,199 @@
+"""Unit tests to check the maths functions."""
+
+import logging
+import numpy as np
+import unittest
```
johmathe: for the tests, could you do one test file per python module? It would probably create less dependencies. Here in order to run this test you end up needing matplotlib since you import visualizations

visualization.py
```diff
@@ -0,0 +1,78 @@
...
+
+import numpy as np
+from matplotlib import pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D
+
+# import rigid_transformations
+import rotations
+
+
+class Arrow3D():
```
johmathe: awesome!!

visualization.py
```diff
@@ -0,0 +1,78 @@
+"""Visualization for Geometric Statistics."""
+
+# import pylab
+
+import numpy as np
+from matplotlib import pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D
+
+# import rigid_transformations
```
johmathe: leftover comment

visualization.py
```diff
@@ -0,0 +1,78 @@
+"""Visualization for Geometric Statistics."""
+
+# import pylab
```
johmathe: leftover comment

visualization.py
```diff
@@ -0,0 +1,78 @@
...
+
+    return trihedron
+
+
+def plot_trihedron(trihedron, **kwargs):
+    "Plot the trihedron."
+    fig = plt.figure()
+    ax = fig.add_subplot(111, projection='3d')
+    trihedron.draw(ax, **kwargs)
+    plt.show()
```
johmathe: you may also want to return a figure instead of showing here. And then ppl can use the figure how they see fit (like save it etc.)

visualization.py
```diff
@@ -0,0 +1,78 @@
...
+    trihedron_vec_3 = np.dot(rot_mat, basis_vec_3)
+    trihedron = Trihedron(translation,
+                          trihedron_vec_1,
+                          trihedron_vec_2,
+                          trihedron_vec_3)
+
+    return trihedron
+
+
+def plot_trihedron(trihedron, **kwargs):
```
johmathe: kinda hate kwargs... At the very least say which function you're passing the kwards too. Something like 

plot_trihedron(trihedron, **draw_kwargs)

rigid_transformations.py
ninamiolane: I added the formula :)

rigid_transformations.py
ninamiolane: Left or Right is the actual mathematical term: it comes from the left- or right- translations associated to the group composition resp. by the left- or the right-.

rigid_transformations.py
ninamiolane: Yes, actually when I told you that I wanted to factorize code this is because the code on riemannian is general across rotations and rigid transformations. So I will indeed create a separate file riemannian.py and I want to use OO code for lie groups to use the riemannian functions.

I'll only delete the comments for now.

rotations.py
ninamiolane: Actually this is correct! This is because the group identity for the rotations is the identity, thus the identity matrix. However, the rotation vector (also called the axis-angle vector) representing the identity matrix is the null vector.

rotations.py
ninamiolane: Isn't it clearer like this?

unittests.py
ninamiolane: You mean per assert or per assert's type (corresponding to a method)?

visualization.py
ninamiolane: You advice not to use them in general? Isn't it tedious to copy-paste all plotting parameters from fn to fn?

rotations.py
johmathe: it depends if you generate the doc with readthedocs or something, you will get latex for free generated...

rigid_transformations.py
johmathe: cool thanks for the clarification

rigid_transformations.py
johmathe: you can add a todo if you want

unittests/test_rigid_transformations.py
```diff
@@ -0,0 +1,78 @@
+"""Unit tests for rigid_transformations module."""
```
johmathe: so the canonical syntax for tests is: 

in the same directory, create a file that starts with the module that you want to test and suffix _test.py

this will work with self discovery for the unit test library

unittests/test_rotations.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertTrue(np.allclose(rot_vec_1, rot_vec_1_expected))
+
+        rot_vec_2 = 1.5 * np.pi * np.array([0., 1., 0.])
+        rot_vec_2 = rotations.regularize_rotation_vector(rot_vec_2)
+        rot_vec_2_expected = np.pi / 2. * np.array([0., -1., 0.])
+        self.assertTrue(np.allclose(rot_vec_2, rot_vec_2_expected))
+
+        rot_vec_3 = 11 * np.pi * np.array([1., 2., 3.])
+        rot_vec_3 = rotations.regularize_rotation_vector(rot_vec_3)
+        fact = 0.84176874548664671 * np.pi / np.sqrt(14)
```
johmathe: where does this constant come from? can you be a bit more clear into the kind of things you're testing here?? Even though I usually hate comments, in unit tests it's fine.

unittests/test_rigid_transformations.py
```diff
@@ -0,0 +1,78 @@
...
+
+        translation_4 = np.array([4, 10, -2])
+        rot_vec_4 = (np.pi / (2 * np.sqrt(3)) *
+                     np.array([6 * 1e-8, 5.5 * 1e-7, -2 * 1e-6]))
+        transfo_4 = np.concatenate([rot_vec_4, translation_4])
+
+        all_transfos = [transfo_1, transfo_2, transfo_3, transfo_4]
+        i_debug = 1
+        for transfo in all_transfos:
+            print('at i=%d' % i_debug)
```
johmathe: so one tip for this test: here you rely on 2 things for the test to pass: your regularization function and your transformation function. I would recommend not relying on the regularization for testing your transformation - or this is not a unit test and it makes debugging harder.

rotations.py
ninamiolane: Good to know! thanks.

unittests/test_rigid_transformations.py
ninamiolane: Cool, thanks.

ninamiolane:
I wrote more and (hopefully) cleaner unit tests. 

I kept the unit tests that use several functions, as I want to test not only the code, but also the maths.

PR #3, opened 2018-01-10
Start sphere module

ninamiolane:
Add a new manifold: n-dimensional sphere, embedded in (n+1)-dimensional euclidean space.

### DIFF
--- a/geomstats/rotations.py
+++ b/geomstats/rotations.py
@@ -280,9 +280,9 @@ def riemannian_log(rot_vec,
 
     This gives a tangent vector at point ref_point.
 
-    :params rot_vec: 3D rotation vector
+    :param rot_vec: 3D rotation vector
     :param ref_point: 3D rotation vector
-    :returns a: 3D rotation vector, tangent vector at ref_point
+    :returns rot_vec_log: 3D rotation vector, tangent vector at ref_point
     """
     assert len(rot_vec) == 3 & len(ref_point) == 3
     assert left_or_right in ('left', 'right')
--- a/geomstats/sphere.py
+++ b/geomstats/sphere.py
@@ -1,0 +1,114 @@
+"""
+Computations on the n-dimensional sphere
+embedded in the (n+1)-dimensional Euclidean space.
+"""
+
+import numpy as np
+
+EPSILON = 1e-6
+
+
+def projection_to_tangent_space(ref_point, vector):
+    """
+    Project the vector v onto the tangent space:
+    T_{ref_point} S = {w | scal(w, ref_point) = 0}
+    """
+    sq_norm = np.dot(ref_point, ref_point)
+    tangent_vec = vector - np.dot(ref_point, vector) / sq_norm * ref_point
+    return tangent_vec
+
+
+def riemannian_exp(ref_point, vector, epsilon=EPSILON):
+    """
+    Compute the Riemannian exponential at point ref_point
+    of tangent vector tangent_vec wrt the metric obtained by
+    embedding of the n-dimensional sphere
+    in the (n+1)-dimensional euclidean space.
+
+    This gives a point on the n-dimensional sphere.
+
+    :param ref_point: a point on the n-dimensional sphere
+    :param vector: (n+1)-dimensional vector
+    :returns riem_exp: a point on the n-dimensional sphere
+    """
+    tangent_vec = projection_to_tangent_space(ref_point, vector)
+    norm_tangent_vec = np.linalg.norm(tangent_vec)
+
+    if norm_tangent_vec < epsilon:
+        coef_1 = (1. - norm_tangent_vec ** 2 / 2.
+                  + norm_tangent_vec ** 4 / 24.
+                  - norm_tangent_vec ** 6 / 720.
+                  + norm_tangent_vec ** 8 / 40320.)
+        coef_2 = (1. - norm_tangent_vec ** 2 / 6.
+                  + norm_tangent_vec ** 4 / 120.
+                  - norm_tangent_vec ** 6 / 5040.
+                  + norm_tangent_vec ** 8 / 362880.)
+    else:
+        coef_1 = np.cos(norm_tangent_vec)
+        coef_2 = np.sin(norm_tangent_vec) / norm_tangent_vec
+
+    riem_exp = coef_1 * ref_point + coef_2 * tangent_vec
+
+    return riem_exp
+
+
+def riemannian_log(ref_point, point, epsilon=EPSILON):
+    """
+    Compute the Riemannian logarithm at point ref_point,
+    of point wrt the metric obtained by
+    embedding of the n-dimensional sphere
+    in the (n+1)-dimensional euclidean space.
+
+    This gives a tangent vector at point ref_point.
+
+    :param ref_point: point on the n-dimensional sphere
+    :param point: point on the n-dimensional sphere
+    :returns riem_log: tangent vector at ref_point
+    """
+    norm_ref_point = np.linalg.norm(ref_point)
+    norm_point = np.linalg.norm(point)
+
+    cos_angle = np.dot(ref_point, point) / (norm_ref_point * norm_point)
+    if cos_angle > 1.0:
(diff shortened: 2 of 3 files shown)


geomstats/rotations.py
```diff
@@ -280,9 +280,9 @@ def riemannian_log(rot_vec,
 
     This gives a tangent vector at point ref_point.
 
-    :params rot_vec: 3D rotation vector
+    :param rot_vec: 3D rotation vector
     :param ref_point: 3D rotation vector
-    :returns a: 3D rotation vector, tangent vector at ref_point
+    :returns rot_vec_log: 3D rotation vector, tangent vector at ref_point
```
johmathe: What's the standard for return syntax for this? It doens't make sense to pass a param to the return argument right?

geomstats/sphere.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+
+import numpy as np
+
+EPSILON = 1e-6
+
+
+def projection_to_tangent_space(ref_point, vector):
+    """
+    Project the vector v onto the tangent space:
```
johmathe: project the vector "vector" or rename vector to v

geomstats/sphere.py
```diff
@@ -0,0 +1,114 @@
...
+    This gives a point on the n-dimensional sphere.
+
+    :param ref_point: a point on the n-dimensional sphere
+    :param vector: (n+1)-dimensional vector
+    :returns riem_exp: a point on the n-dimensional sphere
+    """
+    tangent_vec = projection_to_tangent_space(ref_point, vector)
+    norm_tangent_vec = np.linalg.norm(tangent_vec)
+
+    if norm_tangent_vec < epsilon:
```
johmathe: can you explain a little bit waht happens here? I assume this is a taylor expansion of the sine/cosine?  Maybe keeping the factorial underneath like math.factorial would be easier to read? Python will probably cache the value so no need to freak out about performance and it will make the code more readable.

geomstats/sphere.py
```diff
@@ -0,0 +1,114 @@
...
+        coef_1 = (1. - norm_tangent_vec ** 2 / 2.
+                  + norm_tangent_vec ** 4 / 24.
+                  - norm_tangent_vec ** 6 / 720.
+                  + norm_tangent_vec ** 8 / 40320.)
+        coef_2 = (1. - norm_tangent_vec ** 2 / 6.
```
johmathe: why do you divide by norm_tangent_vec at line 48 and not here?

geomstats/sphere.py
```diff
@@ -0,0 +1,114 @@
...
+
+    :param ref_point: point on the n-dimensional sphere
+    :param point: point on the n-dimensional sphere
+    :returns riem_log: tangent vector at ref_point
+    """
+    norm_ref_point = np.linalg.norm(ref_point)
+    norm_point = np.linalg.norm(point)
+
+    cos_angle = np.dot(ref_point, point) / (norm_ref_point * norm_point)
+    if cos_angle > 1.0:
```
johmathe: nit: if you put >= 1 you will avoid this function to compute the arccos.

geomstats/sphere.py
```diff
@@ -0,0 +1,114 @@
...
+        angle = 0.0
+    else:
+        angle = np.arccos(cos_angle)
+
+    if angle < epsilon:
+        coef_1 = (1. + angle ** 2 / 6.
```
johmathe: same comment as before for factorials

tests/test_sphere.py
```diff
@@ -0,0 +1,32 @@
+"""Unit tests for sphere module."""
+
+import geomstats.sphere as sphere
+
+import numpy as np
+import unittest
+
+
+class TestSphereMethods(unittest.TestCase):
+    def test_riemannian_exp_and_log(self):
```
johmathe: can you unit test the results you got when it's under espilon please?

geomstats/rotations.py
ninamiolane: You mean the stantard syntax in docstrings for the return argument? I'm not sure, but I could use Sphinx' syntax (close to what I have now) to then automatically generate documentation?
https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html
http://www.sphinx-doc.org/en/stable/index.html
Unless you've heard about another documentation generator?

geomstats/sphere.py
ninamiolane: This is because tangent_vector is not normalized, whereas ref_point is (a point on a sphere represented by) a normalized vector.

We could normalize tangent_vector and avoid this, but cos(angle) and sin(angle)/angle show the same orders in the taylor expansions so I'd rather keep them like this.

geomstats/sphere.py
ninamiolane: Yes, there are taylor expansions of cos(angle) and sin(angle)/angle.

Are you sure that python will cache them? Otherwise I could define them as constants?

ninamiolane:
Thanks for the review. 

There is a numerical issue with np.dot: the dot product of a unit vector with itself doesn't give 1 in this situation:
```
vector = np.array([1., 1., 1.])
unit_vector = vector / np.linalg.norm(vector)
```
where we normalize by a square root `np.sqrt(3)`.
Then, `np.dot(unit_vector, unit_vector)` gives: `1.0000000000000002`and
`np.linalg.norm(unit_vector)` gives `1.0`.

This comes from the square root and should be a standard issue: any idea on how this is usually addressed?

I've put an emergency fix for now.

farrell236:
`np.dot(unit_vector, unit_vector)` giving `1.0000000000000002` is due to a precision error caused by `np.linalg.norm(vector)`, as you correctly stated being `np.sqrt(3)`. Unfortunately, python does not track the variables as rational numbers, by default it's stored as doubles (64bit precision).

If the precision doesn't need to be this high, you can cast it as a 32bit float. i.e.
`np.float32(np.dot(unit_vector, unit_vector)) #gives 1.0`

geomstats/sphere.py
ninamiolane: These are the taylor expansion of x / sin(x) and x / tan(x), whose coefficients are computed *from* the factorials in the expansions of sin(x) and tan(x) using the computational rules on  taylor expansion's inversion (basically writing that the product of the taylor expansions will give 1 and identifying the coefficients).

Since the numerical factors are not exactly the inverse of factorials, and there will be sums of inverses of factorials, I don't think that it makes sense to make the factorials appear here.

See: https://www.wolframalpha.com/input/?i=1+%2F+(sin(x)) for the result.

geomstats/sphere.py
johmathe: I meant explaining in the code, not to me :) 

Feel free to use constants if you want.

geomstats/rotations.py
johmathe: I was assuming you were using the sphinx syntax already. There shouldn't be any return parameter.

geomstats/sphere.py
johmathe: I meant explaining in the code as a comment. Constants is fine for the factorial

geomstats/sphere.py
johmathe: ok, can you just make it more readable - it's just pretty awful to see plain numbers like this. Maybe you could define each coefficient for each term somewhere? Maybe in an array or a dictionnary?

Something like:

TAYLOR_COEFFS = [1, 1, 7./360., 31./15120.] ...

Same for previous function actually.

This will also have the side effect of making it clear to the reader that this is a taylor expansion.

johmathe:
To your comment - these numerical stability issues are pretty standard. Few approaches:

1. Make a check if the vector is unit vector, then return the value that is supposed to be returned.
2. Do not care about the result

geomstats/rotations.py
ninamiolane: Really? In the first link above they put some.

geomstats/sphere.py
ninamiolane: Great idea! thanks.

PR #5, opened 2018-01-12
update unittest for loss gradient checker

farrell236:


farrell236:
updated with python2 syntax to python3

tests/test_losses.py
```diff
@@ -124,6 +124,60 @@ def test_rigids_riemannian_loss_and_grad(self):
...
+            print('Forward with -dx: ',str(forward_loss_y_pred_minus_dx))
+            
+            dy_dx = (forward_loss_y_pred_plus_dx - forward_loss_y_pred_minus_dx) / (dx*2)
+            print('Computed grad backward_grad[',str(xdim),']: ',str(backward_grad[xdim]))
+            print('Calculated grad dL/dy_pred[',str(xdim),']: ',str(dy_dx))
+            
+            diff = np.abs(backward_grad[xdim]-dy_dx)
+            scale = np.max(np.fabs(np.array([backward_grad[xdim],dy_dx,1.0])))
+
+            if diff > (scale * threshold):
```
johmathe: use assert statements, not if statements

Something like self.assertAlmostEqual or something

tests/test_losses.py
```diff
@@ -124,6 +124,60 @@ def test_rigids_riemannian_loss_and_grad(self):
 
         # self.assertTrue(np.allclose(result_2, curved_expected_2))
 
+    def test_caffe_gradient_checker(self):
+
+        print('\n #####  GRADIENT CHECKER (CAFFE) ##### \n')
```
johmathe: Please avoid print statements. Use logging library if necessary. Also avoid big titles like this, it makes the code hard to maintain and the logs hard to read.

tests/test_losses.py
```diff
@@ -124,6 +124,60 @@ def test_rigids_riemannian_loss_and_grad(self):
...
+
+        print('Setting Test Parameters: ') # Adjustable Parameters
+        dx = 0.01
+        threshold = 0.01
+        y_pred = np.array([-0.662919 , -1.29015 , 1.30245 , 1.25567 , 1.09228 , 0.153399 ]) #Regularized
+        y_true = np.array([0.0349858 , 0.562573 , -1.94181 , -0.173889 , 0.127527 , -0.19467 ]) #Regularized
+
+        print('Stepsize: ', str(dx))
```
johmathe: ditto, I don't think all these are necessary. If needed, just do assertions in the code.

tests/test_losses.py
```diff
@@ -124,6 +124,60 @@ def test_rigids_riemannian_loss_and_grad(self):
...
         # self.assertTrue(np.allclose(result_2, curved_expected_2))
 
+    def test_caffe_gradient_checker(self):
+
+        print('\n #####  GRADIENT CHECKER (CAFFE) ##### \n')
+
+        print('Setting Test Parameters: ') # Adjustable Parameters
+        dx = 0.01
+        threshold = 0.01
+        y_pred = np.array([-0.662919 , -1.29015 , 1.30245 , 1.25567 , 1.09228 , 0.153399 ]) #Regularized
```
johmathe: What do you mean by "regularized?" Instead of a comment can you make things more clear in the code?

tests/test_losses.py
farrell236: I'll move these to logging

tests/test_losses.py
farrell236: 'regularised' means the vector example used is a regularised transformation, though its not completely necessary as geomstats library seems regularise all input vectors anyway.

tests/test_losses.py
farrell236: there are 6 partials to check in total, wouldn't it be better to see which partial(s) it failed on instead of fixing them in order of assertion?

johmathe: just add multiple assertions for all your prior invariants :)

farrell236:
the code has now been refactored to address above points, expected to fail checks as 3/6 gradients are not quite yet correct

PR #7, opened 2018-01-15
Nina add sym pos matrices

ninamiolane:
Start manifold of symmetric positive definite matrices.


geomstats/symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,216 @@
...
+
+def group_log(sym_mat):
+    """
+    Group logarithm of the Lie group of
+    all invertible matrices has a straight-forward
+    computation for symmetric positive definite matrices.
+    """
+    assert is_symmetric(sym_mat)
+    # Make it fully symmetric for numerics.
+    sym_mat = (sym_mat + sym_mat.transpose()) / 2
```
johmathe: maybe create a function for this? That way you don't need the comment and can re-use afterwards. Maybe a small unit test for it too.

geomstats/symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,216 @@
...
+
+def matrix_to_vector(matrix):
+    """
+    Convert the symmetric part of a symmetric matrix
+    into a vector.
+    """
+    assert is_symmetric(matrix)
+
+    dim_mat, _ = matrix.shape
+    dim_vec = dim_mat * (dim_mat + 1) / 2
```
johmathe: be careful here, while this might work with python2, python3 might behave differently with float. Make sure to always force the behavior you expect wrt integers.

geomstats/symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,216 @@
...
+    vector = np.zeros(dim_vec)
+
+    idx = 0
+    for i in range(dim_mat):
+        for j in range(i + 1):
+            if i == j:
+                vector[idx] = matrix[j, i]
+            else:
+                vector[idx] = matrix[j, i] / np.sqrt(2.)
+            idx += 1
```
johmathe: maybe this would help:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.triu_indices.html

Then you should be able to do this in one single numpy operation and avoid the double for loop.

geomstats/symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,216 @@
...
+    dim_vec = dim_mat * (dim_mat + 1) / 2
+    vector = np.zeros(dim_vec)
+
+    idx = 0
+    for i in range(dim_mat):
+        for j in range(i + 1):
+            if i == j:
+                vector[idx] = matrix[j, i]
+            else:
+                vector[idx] = matrix[j, i] / np.sqrt(2.)
```
johmathe: why are you normalizing here?

tests/test_symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,54 @@
+"""Unit tests for sphere module."""
```
johmathe: nice!

geomstats/symmetric_positive_definite_matrices.py
```diff
@@ -0,0 +1,216 @@
...
+    dists_between_iterates = []
+    n_sym_matrices = len(sym_matrices)
+
+    riem_mean = sym_matrices[0]
+
+    dim = len(riem_mean)
+
+    dist = math.inf
+    it = 0
+    while it < n_max_iterations and dist > epsilon:
```
johmathe: at some point when you want to deal with performance it might be interesting to profile this code and check where you're getting the biggest performance hit. Not for this code review but just out of curiosity :)

PR #17, opened 2018-01-31
Nina example geodesic

ninamiolane:
I got to see where we need vectorization :) I should vectorize in the next CR.

### DIFF
--- a/examples/plot_geodesics.py
+++ b/examples/plot_geodesics.py
@@ -1,6 +1,6 @@
 """
 Plot a geodesic of SE(3) equipped
-with its left-invariant canonical metric.
+with its left-invariant canonical METRIC.
 """
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


examples/plot_geodesics.py
```diff
@@ -0,0 +1,25 @@
...
+metric = se3_group.left_canonical_metric
+
+initial_point = se3_group.identity
+initial_tangent_vec = np.array([1.8, 0.2, 0.3, 3., 3., 1.])
+geodesic = metric.geodesic(initial_point=initial_point,
+                           initial_tangent_vec=initial_tangent_vec)
+
+points = np.vstack([geodesic(t)
+                    for t in np.linspace(-0.7, 0.7, 20)])
+
```
johmathe: Create a main method and the usual if __main__ ...

examples/plot_geodesics.py
```diff
@@ -0,0 +1,25 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+from geomstats.special_euclidean_group import SpecialEuclideanGroup
+import geomstats.visualization as visualization
+
+
+se3_group = SpecialEuclideanGroup(n=3)
+metric = se3_group.left_canonical_metric
```
johmathe: https://www.python.org/dev/peps/pep-0008/#constants

geomstats/visualization.py
```diff
@@ -70,9 +68,26 @@ def trihedron_from_rigid_transformation(transfo):
     return trihedron
 
 
-def plot_trihedron(trihedron, **trihedron_draw_kwargs):
-    "Return the figure with the plotted trihedron."
-    fig = plt.figure()
-    ax = fig.add_subplot(111, projection='3d')
-    trihedron.draw(ax, **trihedron_draw_kwargs)
-    return fig
+def plot_points(points, ax=None, **point_draw_kwargs):
```
johmathe: Plot points seems a bit redundant to me if you have a parameter called points.

PR #19, opened 2018-02-10
v1.1

ninamiolane:


johmathe:
You should create tags on specific commits for a given version that you publish

PR #23, opened 2018-02-20
Nina geodesics h2

ninamiolane:
Add plot of geodesics for the hyperbolic space in 2D, with Poincare disk visualization.
Add unit tests to check that the geodesics belong to the space for all spaces.

### DIFF
--- a/examples/plot_geodesics_h2.py
+++ b/examples/plot_geodesics_h2.py
@@ -13,20 +13,18 @@ METRIC = H2.metric
 
 
 def main():
-    initial_point = np.array([np.sqrt(10), 3., 0.])
+    initial_point = np.array([np.sqrt(2), 1., 0.])
     assert H2.belongs(initial_point)
     initial_tangent_vec = H2.projection_to_tangent_space(
-                                        vector=np.array([1., 2., 0.8]),
+                                        vector=np.array([3.5, 0.6, 0.8]),
                                         base_point=initial_point)
     geodesic = METRIC.geodesic(initial_point=initial_point,
                                initial_tangent_vec=initial_tangent_vec)
 
-    n_steps = 100
+    n_steps = 10
     t = np.linspace(0, 1, n_steps)
 
     points = geodesic(t)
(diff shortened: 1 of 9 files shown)


tests/test_hyperbolic_space.py
```diff
@@ -303,6 +303,21 @@ def test_exp_and_dist_and_projection_to_tangent_space(self):
         expected_1 = math.sqrt(sq_norm)
         self.assertTrue(np.allclose(result_1, expected_1))
 
+    def test_geodesic_and_belongs(self):
+        # TODO(nina): this tests fails when geodesic goes "too far"
```
johmathe: that's gonna be interesting to fix :)

examples/plot_geodesics_h2.py
```diff
@@ -0,0 +1,35 @@
+"""
+Plot a geodesic on the hyperbolic space h2,
```
johmathe: could you potentially add one or a few renderings to this or a subsequent CL? I'm curious :)

geomstats/visualization.py
```diff
@@ -153,20 +180,34 @@ def plot(points, ax=None, space=None, **point_draw_kwargs):
...
         trihedrons = convert_to_trihedron(points, space=space)
         for t in trihedrons:
             t.draw(ax, **point_draw_kwargs)
+
     elif space is 'S2':
         sphere = Sphere()
         sphere.add_points(points)
         sphere.draw(ax)
 
+    elif space is 'H2':
```
johmathe: are you raising some ValueError if space is not in the 4 possibilities that you're passing?

examples/plot_geodesics_h2.py
ninamiolane: screen shot 2018-02-21 at 11 39 52 am

The curvature increases next to the disk center and the points get closer as the geodesic approches the disk boundary, which makes sense as the disk's boundary represents the infinity, the boundary is never reached.

PR #29, opened 2018-03-16
Nina ndim checks

ninamiolane:
Factorization of the code using expand_dims as you suggested in a previous CR


geomstats/vectorization_utils.py
```diff
@@ -0,0 +1,12 @@
+"""
```
johmathe: not a huge fan on the "utils" naming scheme. Sounds a bit like laziness to me when I see something with a utils suffix.

geomstats/special_orthogonal_group.py
```diff
@@ -2,7 +2,9 @@
 
 import numpy as np
 
+# TODO(nina): Rename modules to make imports cleaner?
```
johmathe: +1 :)

PR #41, opened 2018-04-12
Nina geodesic two points

ninamiolane:
Adapt of geodesic's code to define geodesic between an initial point and an end point + Add example in H2.


examples/plot_geodesics_h2.py
```diff
@@ -14,20 +14,30 @@
 
 
 def main():
-    initial_point = np.array([np.sqrt(2), 1., 0.])
+    initial_point = H2.intrinsic_to_extrinsic_coords(np.array([0, 0.]))
```
johmathe: no need for np.array

examples/plot_geodesics_h2.py
```diff
@@ -14,20 +14,30 @@
...
+    end_point = H2.intrinsic_to_extrinsic_coords(np.array([1.5, 1.5]))
     assert H2.belongs(initial_point)
+    assert H2.belongs(end_point)
     initial_tangent_vec = H2.projection_to_tangent_space(
-                                        vector=np.array([3.5, 0.6, 0.8]),
+                                        vector=np.array([3.5, 0.6, 0]),
```
johmathe: ditto

examples/plot_geodesics_h2.py
```diff
@@ -14,20 +14,30 @@
...
                                         base_point=initial_point)
-    geodesic = METRIC.geodesic(initial_point=initial_point,
+
+    geodesic_with_initial_tangent_vec = METRIC.geodesic(
```
johmathe: for clarity purposes can we do the 2 point plot first and then the point + vector plot? I think this would make things clearer. Maybe with 2 subfunctions?

Issue #38, opened 2018-04-12
Impossible to go more than 2.xx


johmathe:
Framework for some reason doesn't allow me to go further than 2.xxx in the intrinsic coordinate system of H2. See the square example to reproduce.

johmathe:
Also I've noticed it happens when you have both coordinates going further than 0

PR #43, opened 2018-04-13
Nina bug h2

ninamiolane:
Taking steps towards solving issue #38 
- add a regularize function
- use regularize function to systematically regularize outputs of the exponential
- assess if a point belongs to H by checking if its squared Minkowski norm is close to -1, at the scale of its Euclidean norm (relative tolerance, not absolute tolerance.)


johmathe:
Build is failing:

Traceback (most recent call last):
  File "/home/travis/build/ninamiolane/geomstats/tests/test_manifold.py", line 26, in test_regularize
    self.assertTrue(np.allclose(point), self.manifold.regularize(point))
TypeError: allclose() missing 1 required positional argument: 'b'

PR #45, opened 2018-04-18
add h2 grid example

johmathe:
add h2 grid example 
add corresponding image


examples/plot_h2_grid.py
```diff
@@ -0,0 +1,47 @@
+"""
```
ninamiolane: rename the file for consistency with the current file namings: plot_grid_h2 (what we plot, THEN, on which space we plot it).

examples/plot_h2_grid.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+from geomstats.hyperbolic_space import HyperbolicSpace
+import geomstats.visualization as visualization
+
+H2 = HyperbolicSpace(dimension=2)
+METRIC = H2.metric
+
+c = H2.intrinsic_to_extrinsic_coords
```
ninamiolane: maybe find a better (longer, more unique) name for the function if possible? it's easier for characters' replacements using regex in vim.

examples/plot_h2_grid.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def main():
+    starts = []
+    ends = []
+    left = -128
```
ninamiolane: Should these be constants and at the top of the file?

examples/plot_h2_grid.py
```diff
@@ -0,0 +1,47 @@
...
+    n_steps = 512
+    for p in np.linspace(left, right, grid_size):
+        starts.append(np.array([top, p]))
+        ends.append(np.array([bottom, p]))
+    for p in np.linspace(top, bottom, grid_size):
+        starts.append(np.array([p, left]))
+        ends.append(np.array([p, right]))
+    starts = [c(s) for s in starts]
+    ends = [c(e) for e in ends]
+    for i, _ in enumerate(starts):
```
ninamiolane: how about `for start, end in zip(starts, ends)`?

PR #47, opened 2018-04-23
Johmathe gradient descent

johmathe:
- Update viz library
- Add gradient descent module

### DIFF
--- a/examples/gradient_descent.py
+++ b/examples/gradient_descent.py
@@ -5,12 +5,11 @@ Gradient descent on a sphere.
 import numpy as np
 import matplotlib
 matplotlib.use("Agg")  # NOQA
-import matplotlib.pyplot as plt
-import matplotlib.animation as manimation
-
-from geomstats.hypersphere import Hypersphere
-import geomstats.visualization as visualization
 import geomstats.vectorization as vectorization
+import geomstats.visualization as visualization
+import matplotlib.animation as manimation
+import matplotlib.pyplot as plt
+from geomstats.hypersphere import Hypersphere
 
 SPHERE2 = Hypersphere(dimension=2)
 METRIC = SPHERE2.metric
@@ -23,12 +22,14 @@ def gradient_descent(start,
                      lr=0.05,
                      max_iter=128,
                      precision=1e-5):
+    """Operate a gradient descent on a given manifold until either max_iter or
+    a given precision is reached."""
     x = start
     for i in range(max_iter):
         x_prev = x
-        euclidian = - lr * grad(x)
-        tangent_vec = manifold.projection_to_tangent_space(vector=euclidian,
-                                                           base_point=x)
+        euclidian_grad = - lr * grad(x)
+        tangent_vec = manifold.projection_to_tangent_space(
+                vector=euclidian_grad, base_point=x)
         x = manifold.metric.exp(base_point=x, tangent_vec=tangent_vec)[0]
         if np.abs(loss(x) - loss(x_prev)) <= precision:
             print('x: %s' % x)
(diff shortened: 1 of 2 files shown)


examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+
+import numpy as np
+import matplotlib
+matplotlib.use("Agg")  # NOQA
+import matplotlib.pyplot as plt
+import matplotlib.animation as manimation
+
+from geomstats.hypersphere import Hypersphere
+import geomstats.visualization as visualization
+import geomstats.vectorization as vectorization
```
ninamiolane: import order

ninamiolane: orders* :)

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+                     loss,
+                     grad,
+                     manifold,
+                     lr=0.05,
+                     max_iter=128,
+                     precision=1e-5):
+    x = start
+    for i in range(max_iter):
+        x_prev = x
+        euclidian = - lr * grad(x)
```
ninamiolane: euclidean*. why not euclidean_grad?

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+        tangent_vec = manifold.projection_to_tangent_space(vector=euclidian,
+                                                           base_point=x)
+        x = manifold.metric.exp(base_point=x, tangent_vec=tangent_vec)[0]
```
ninamiolane: hm, the [0] does not look great. Is there a way to avoid this? Or should I change the vectorization strategy by enforcing the outputs of methods to be of the same ndim and shapes as the inputs (-> feature request)?

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+        if np.abs(loss(x) - loss(x_prev)) <= precision:
+            print('x: %s' % x)
```
ninamiolane: Are you sure that we should use print over logging?

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+            print('reached precision %s' % precision)
+            print('iterations: %d' % i)
+            break
+        yield x, loss(x)
+
+
+def animate(geodesics, size=20, fps=0, dpi=103, out='out.mp4', color='red'):
+    FFMpegWriter = manimation.writers['ffmpeg']
+    metadata = dict(title='Movie Test', artist='Matplotlib',
```
ninamiolane: Change title and comment?

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+                    comment='Movie support!')
+    writer = FFMpegWriter(fps=fps, metadata=metadata)
+    fig = plt.figure(figsize=(size, size))
+    ax = fig.add_subplot(111, projection='3d', aspect='equal')
+    sphere = visualization.Sphere()
+    points = vectorization.to_ndarray(geodesics[0], to_ndim=2)
```
ninamiolane: Same as above. How could we avoid this [0]?

examples/gradient_descent.py
johmathe: originally to limit space - but was waiting for your comment to check if it was super horrible to read :)

examples/gradient_descent.py
johmathe: Agreed - would be great if we didn't have to do this.

examples/gradient_descent.py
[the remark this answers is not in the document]
johmathe: no - the code is right. This is a double array - a list of geodesics made of points. Each frame is made of one geodesic curve.

examples/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
...
+    sphere.draw(ax, color=color, marker='.')
+    with writer.saving(fig, 'foo.mp4', dpi=dpi):
+        for points in geodesics[1:]:
+            points = vectorization.to_ndarray(points, to_ndim=2)
+            sphere.add_points(points)
+            sphere.draw_points(ax, color=color, marker='.')
+            writer.grab_frame()
+
+
+def main():
```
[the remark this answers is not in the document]
johmathe: We do have a docstring at the beginning of the file. You want another one here?

examples/gradient_descent.py
johmathe: Since this is an example I don't want to make it overly complex.

examples/gradient_descent.py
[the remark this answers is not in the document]
ninamiolane: I was thinking adding more details about what we are trying to maximize / minimize, and how it boils down to an optimization on the sphere + printing it.

examples/gradient_descent.py
```diff
@@ -23,12 +22,14 @@ def gradient_descent(start,
...
                      precision=1e-5):
+    """Operate a gradient descent on a given manifold until either max_iter or
+    a given precision is reached."""
     x = start
     for i in range(max_iter):
         x_prev = x
-        euclidian = - lr * grad(x)
-        tangent_vec = manifold.projection_to_tangent_space(vector=euclidian,
-                                                           base_point=x)
+        euclidian_grad = - lr * grad(x)
```
ninamiolane: people tend to use "euclidean" instead of "euclidian" https://en.wikipedia.org/wiki/Euclidean

examples/gradient_descent.py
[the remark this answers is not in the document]
ninamiolane: Ah, I see!

examples/gradient_descent.py
```diff
@@ -0,0 +1,90 @@
...
+
+import numpy as np
+import matplotlib
+matplotlib.use("Agg")  # NOQA
+import geomstats.vectorization as vectorization
+import geomstats.visualization as visualization
+import matplotlib.animation as manimation
+import matplotlib.pyplot as plt
+from geomstats.hypersphere import Hypersphere
+
```
ninamiolane: imports:

matplotlib.*
numpy

geomstats.*
?

PR #53, opened 2018-04-24
Heatmap colored sphere in visualisation 

johmathe:


examples/gradient_descent_s2.py
```diff
@@ -18,11 +18,14 @@
...
 import geomstats.vectorization as vectorization
 import geomstats.visualization as visualization
 
+
 SPHERE2 = Hypersphere(dimension=2)
 METRIC = SPHERE2.metric
+np.random.seed(1984)
 
 
 def gradient_descent(start,
```
ninamiolane: Just a thought. This works for any manifold embedded in a euclidean space. Later we should probably divide geomstats' code into: 'geometry', 'statistics', 'visualization', 'optimization', 'tests' folders.

geomstats/visualization.py
```diff
@@ -98,6 +98,46 @@ def draw_points(self, ax, **scatter_kwargs):
...
+            y_vals.append(y)
+            z_vals.append(z)
+
+        x_vals = [(self.radius * i) for i in x_vals]
+        y_vals = [(self.radius * i) for i in y_vals]
+        z_vals = [(self.radius * i) for i in z_vals]
+
+        return np.array([x_vals, y_vals, z_vals])
+
+    def plot_heatmap(self, ax, loss, n_points=16000, alpha=0.2, cmap='jet'):
```
ninamiolane: instead of `loss`, which is a vocabulary optimization's specific, could we use `scalar_field` or `scalar_function`? We might want to plot scalar fields on the sphere, and later on other manifolds, unrelated to any optimization problem.

geomstats/visualization.py
```diff
@@ -98,6 +98,46 @@ def draw_points(self, ax, **scatter_kwargs):
...
+        """Plot a heatmap defined by a loss on the sphere."""
+        x = self.fibonnaci_points(n_points)
+        intensity = np.array([loss(x_) for x_ in x.T])
```
ninamiolane: why the notation `x_` (why the underscore?)?

examples/gradient_descent_s2.py
```diff
@@ -72,9 +77,15 @@ def plot_and_save_vid(geodesics,
             writer.grab_frame()
 
 
+def generate_well_behaved_matrix():
```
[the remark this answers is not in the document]
johmathe: yes I actually just added it to make the code more clear and for the end user to understand why we use an SPD matrix.

geomstats/visualization.py
johmathe: originally just to avoid collision with x - but that's pretty horrible naming indeed :)

PR #54, opened 2018-04-24
Add first functions for SO(n) with n!=3

ninamiolane:
Add first functions for the implementation of SO(n) for n!=3.
Prepare the generalization of unit tests  of SO(n) for n!=3


geomstats/special_orthogonal_group.py
```diff
@@ -103,7 +131,7 @@ def belongs(self, rot_vec):
 
     def regularize(self, rot_vec):
         """
-        Regularize the norm of the rotation vector,
+        In, 3D, regularize the norm of the rotation vector,
```
johmathe: wanna add a little todo for the comment about the nd case here?

PR #55, opened 2018-04-25
Implement all functions but the jacobian of translations for SO(n), with unit tests

ninamiolane:


geomstats/special_orthogonal_group.py
```diff
@@ -264,107 +272,125 @@ def rotation_vector_from_matrix(self, rot_mat):
         quaternion to axis-angle:
          r = angle * v / |v|, where (w, v) is a unit quaternion.
 
-        :param rot_mat: 3x3 rotation matrix
-        :return rot_vec: 3d rotation vector
+        In nD, the rotation vector store the n(n-1)/2 values of the
+        skew-symmetric matrix representing
```
johmathe: ?

PR #65, opened 2018-05-02
add numpy backend

johmathe:


geomstats/backend/numpy_backend.py
```diff
@@ -0,0 +1,90 @@
+"""Numpy based computation backend."""
+
+import numpy as np
+
+pi = np.pi
```
ninamiolane: Should this be a constant PI?

geomstats/hypersphere.py
```diff
@@ -5,8 +5,8 @@
 
 import logging
 import math
-import numpy as np
 
+import geomstats.backend as gs
```
ninamiolane: Yessss

tests/test_hypersphere.py
```diff
@@ -342,6 +342,7 @@ def test_geodesic_and_belongs(self):
 
     def test_variance(self):
         point = self.space.random_uniform()
+        print(point)
```
ninamiolane: ?

PR #72, opened 2018-05-03
first steps towards tensorflow-enabled tests

johmathe:


geomstats/backend/__init__.py
```diff
@@ -18,8 +20,8 @@
 elif _BACKEND == 'tensorflow':
     sys.stderr.write('Using tensorflow backend\n')
     from .tensorflow import *  # NOQA
-    from . import tensorflow_random as random
-    from . import tensorflow_linalg as linalg
+    from . import tensorflow_random as random  # NOQA
```
ninamiolane: imports order

geomstats/backend/__init__.py
```diff
@@ -10,6 +10,8 @@
 
 _BACKEND = _backend
```
ninamiolane: Out of curiosity: why the _ before BACKEND?

geomstats/backend/common.py
```diff
@@ -0,0 +1,3 @@
+import numpy as np
```
ninamiolane: why this new file named common?

geomstats/backend/numpy.py
```diff
@@ -2,8 +2,6 @@
 
 import numpy as np
 
-pi = np.pi
-
 
 def array(val):
```
ninamiolane: should we order the functions by alphabetic order? it would be easier to have a look through it later, or to add new functions.

geomstats/backend/tensorflow_random.py
```diff
@@ -3,5 +3,5 @@
 import tensorflow as tf
 
 
-def rand(shape):
-    return tf.random_uniform(shape)
+def rand(*args):
```
ninamiolane: you use `rand` and not `random_uniform` to match with the other backends?

geomstats/backend/__init__.py
johmathe: private

geomstats/backend/common.py
johmathe: common variables/computations across frameworks

geomstats/backend/numpy.py
johmathe: yes absolutely - right now I'm adding them at the end but I'll do this refactoring (or find a way to automate it) later

geomstats/backend/tensorflow.py
```diff
@@ -4,8 +4,6 @@
 
 import tensorflow as tf
 
-pi = tf.pi
-
 
 def abs(x):
```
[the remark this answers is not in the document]
johmathe: ditto

PR #86, opened 2018-05-10
Nina parallel hypersphere

ninamiolane:


tests/test_hypersphere.py
```diff
@@ -140,23 +174,100 @@ def test_exp_vectorization(self):
...
+                    one_tangent_vec[i, j], n_base_points[i, j])
+        self.assertTrue(gs.allclose(result, expected))
+
+        n_tangent_vecs = self.space.projection_to_tangent_space(
+            n_vecs, base_point=n_base_points)
+        result = self.metric.exp(n_tangent_vecs, n_base_points)
+        self.assertTrue(
+            gs.allclose(result.shape, (n_samples, depth, dim + 1)))
+        expected = gs.zeros((n_samples, depth, dim + 1))
+        for i in range(n_samples):
```
johmathe: can you do this w/o the two nested for loops?

tests/test_hypersphere.py
```diff
@@ -140,23 +174,100 @@ def test_exp_vectorization(self):
         one_tangent_vec = self.space.projection_to_tangent_space(
             one_vec, base_point=one_base_point)
         result = self.metric.exp(one_tangent_vec, one_base_point)
-        self.assertTrue(gs.allclose(result.shape, (1, dim + 1)))
+        self.assertTrue(
+            gs.isclose(result.shape, dim + 1),
+            'result.shape = {}'.format(result.shape))
```
johmathe: I feel that yoouv'e been using this pattern a fair amount - is there a way you could refactor this (not necessarily in this PR) in a test_util file or somtg?

tests/test_examples.py
```diff
@@ -31,6 +32,10 @@ def test_plot_geodesics_so3(self):
     # def test_plot_grid_h2(self):
     #    plot_grid_h2.main()
 
+    # TODO(johmathe): this test fails
```
johmathe: on it

johmathe:
Please update the ordering

tests/test_hypersphere.py
ninamiolane: Yes, good idea.

tests/test_hypersphere.py
ninamiolane: Yes, we could vectorize it. But we want to unit test the vectorization here, so I think we should keep it?

PR #92, opened 2018-05-14
update backend and add tensorflow test

johmathe:
update backend and add tensorflow test


geomstats/backend/tensorflow.py
```diff
@@ -5,6 +5,11 @@
 import tensorflow as tf
 
 
+def eval(x):
+    return x
+    return x.eval()
```
ninamiolane: what is this?

PR #93, opened 2018-05-14
Nina ellipsis and tf

ninamiolane:


examples/gradient_descent_s2.py
```diff
@@ -42,7 +42,7 @@ def gradient_descent(start,
         euclidean_grad = - lr * grad(x)
         tangent_vec = manifold.projection_to_tangent_space(
                 vector=euclidean_grad, base_point=x)
-        x = manifold.metric.exp(base_point=x, tangent_vec=tangent_vec)
+        x = manifold.metric.exp(base_point=x, tangent_vec=tangent_vec)[0, 0]
```
johmathe: mmmh why do we still need double indexing here?

geomstats/backend/tensorflow.py
```diff
@@ -5,6 +5,10 @@
 import tensorflow as tf
 
 
+def tile(*args, **kwargs):
```
johmathe: can we avoid passing args and kwargs in the tensorflow backend? I've been passing converting the numpy arguments explicitely so far.

geomstats/backend/tensorflow_random.py
```diff
@@ -4,7 +4,7 @@
 
 
 def rand(*args):
-    return tf.random_uniform(shape=args)
+    return tf.random_uniform(args)
```
johmathe: I don't think you want to change this - this was a bugfix

geomstats/euclidean_space.py
```diff
@@ -19,22 +19,22 @@ def belongs(self, point):
...
+    def random_uniform(self, n_samples=1):
         """
         Sample a vector uniformly in the Euclidean space,
         with coordinates each between -1. and 1.
         """
-        if depth is None:
-            size = (n_samples, self.dimension)
-        else:
-            size = (n_samples, depth, self.dimension)
+        size = (n_samples, self.dimension)
```
johmathe: phew much better

geomstats/hyperbolic_space.py
```diff
@@ -65,45 +65,69 @@ def belongs(self, point, tolerance=TOLERANCE):
         Note: point must be given in extrinsic coordinates.
         """
         point = gs.to_ndarray(point, to_ndim=2)
-        _, point_dim = point.shape
+        point = gs.to_ndarray(point, to_ndim=3, axis=1)
```
johmathe: why do we still need to go to dimension 3 here?

geomstats/minkowski_space.py
```diff
@@ -57,12 +61,20 @@ def exp(self, tangent_vec, base_point):
         """
         The Riemannian exponential is the addition in the Minkowski space.
         """
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=2)
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3, axis=1)
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
```
johmathe: why do we still need to go to rank 3 for these?

geomstats/vectorization.py
```diff
@@ -10,5 +10,5 @@ def to_ndarray(element, to_ndim, axis=0):
 
     if element.ndim == to_ndim - 1:
         element = np.expand_dims(element, axis=axis)
-    assert element.ndim == to_ndim
+    assert element.ndim >= to_ndim
```
johmathe: is this still neede?

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+    def assert_random_uniform_and_belongs_vectorization(self,
+                                                        space,
+                                                        n_samples):
+        n_samples = self.n_samples
+
+        points = self.space.random_uniform(n_samples=n_samples)
+
+        self.assertTrue(gs.all(self.space.belongs(points)))
+
+    def check_shape_exp_vectorization(self, space, n_samples, dim):
```
johmathe: do we still need tests that are so complex now that we've reduced the complexity?

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+        self.assertScalar(result, n_samples)
+
+        result = self.metric.dist(one_point_a, n_points_b)
+        self.assertScalar(result, n_samples)
+
+        result = self.metric.dist(n_points_a, n_points_b)
+        self.assertScalar(result, n_samples)
+
+
+def to_scalar(expected):
```
johmathe: the logic in this function is weird - you say to_scalar and you make this a rank 2 tensor

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+            '\nexpected = {}'.format(
+                result.shape,
+                expected.shape,
+                result,
+                expected))
+
+    def check_shape_belongs(self, space):
+        point = space.random_uniform()
+        belongs = space.belongs(point)
+        print('belongs={}'.format(belongs))
```
johmathe: do we want this print statement? Why not logging?

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+            '\nresult.shape = {} instead of {}'.format(
+                result.shape, (n_samples, 1)))
+
+    def assertVector(self, result, n_samples=1, dim=1):
+        return self.assertTrue(
+            gs.allclose(result.shape, (n_samples, dim)),
+            '\nresult.shape = {} instead of {}'.format(
+                result.shape, (n_samples, dim)))
+
+    def assertAllClose(self, result, expected):
```
johmathe: we should actually be using a backend function for this - that works like 

np.testing.assert_almost_equal(x, y)

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
 
+import unittest
+
+import geomstats.backend as gs
+
+
+class TestGeomstatsMethods(unittest.TestCase):
+    def assertScalar(self, result, n_samples=1):
+        return self.assertTrue(
+            gs.allclose(result.shape, (n_samples, 1)),
```
johmathe: what you're doing here is really weird with the testing of the string. You're basically relying on something that's not necessarily documented. Things could fail if formatting changes or someone messes with streams..?

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+
+        self.assertScalar(belongs)
+
+    def check_shape_belongs_vectorization(self, space, n_samples):
+        n_samples = self.n_samples
+
+        points = self.space.random_uniform(n_samples=n_samples)
+        belongs = self.space.belongs(points)
+
+        self.assertScalar(belongs, n_samples)
```
johmathe: should be in the backend too... https://docs.scipy.org/doc/numpy-1.10.4/reference/generated/numpy.isscalar.html

tests/helper.py
```diff
@@ -2,6 +2,195 @@
...
+
+    def check_shape_random_uniform(self, space, vec_dim):
```
johmathe: not a huge fan of this pattern... Basically this seems like too much logic for tests IMHO.  tests should be clear and super dumb. Here the logic for the testing is itself a bit too complex.

tests/helper.py
ninamiolane: I'm using the backend function `gs.allclose` but I wanted to encompass it as a unittest's function.

Should we use the other backend function that you suggest? Or not encompassing it as a unittest' function?

tests/helper.py
ninamiolane: I don't understand, I'm not testing the string.

tests/helper.py
ninamiolane: Yes, it should be: to_our_convention_for_scalar, which is a (1, 1) tensor.

tests/helper.py
ninamiolane: I want to assert that the element is a scalar, according to our convention, which is a (1, 1)-tensor.

PR #94, opened 2018-05-15
Add the get_dimension getter on the manifold

johmathe:
get_dimension getter on the manifold

### DIFF
--- a/tests/test_manifold.py
+++ b/tests/test_manifold.py
@@ -11,7 +11,7 @@ class TestManifoldMethods(unittest.TestCase):
 
     def setUp(self):
         gs.random.seed(1234)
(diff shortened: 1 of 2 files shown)


tests/test_manifold.py
```diff
@@ -28,6 +28,10 @@ def test_regularize(self):
         point = gs.array([1, 2, 3])
         self.assertTrue(gs.allclose(point, self.manifold.regularize(point)))
 
+    def test_get_dimension(self):
+        manifold = Manifold(10)
+        self.assertEquals(manifold.get_dimension(), 10)
```
ninamiolane: I think we should use `assertEqual` now, bcs `assertEquals` is deprecated.

PR #111, opened 2018-07-11
Nina point representation

ninamiolane:
First CR on adding the point representation of an element of SOn (either as a rotation vector or a rotation matrix) and performing the corresponding computations in this representation.


geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
+
+                angle = gs.linalg.norm(point, axis=1)
+                angle = gs.expand_dims(angle, axis=1)
+
+                coef_1 = gs.zeros([n_points, 1])
+                coef_2 = gs.zeros([n_points, 1])
+
+                mask_0 = gs.isclose(angle, 0)
+                mask_0 = gs.squeeze(mask_0, axis=1)
+                coef_1[mask_0] = (1 - angle[mask_0] ** 2 / 12
```
johmathe: didn't we have these coefficients somewhere? We need to define these.

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
+        if point_representation == 'vector':
+            if self.n == 3:
+                point = self.regularize(
+                    point, point_representation=point_representation)
+                n_points, _ = point.shape
+
+                angle = gs.linalg.norm(point, axis=1)
+                angle = gs.expand_dims(angle, axis=1)
+
+                coef_1 = gs.zeros([n_points, 1])
```
johmathe: not clear why this is n_points and 1?

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
-            inv_rot_mat = gs.linalg.inv(rot_mat)
-            inv_point = self.rotation_vector_from_matrix(inv_rot_mat)
+
+        if point_representation is None:
+            point_representation = self.point_representation
+
+        if point_representation == 'vector':
+            if self.n == 3:
+                inv_point = -self.regularize(point)
+                return inv_point
```
johmathe: not clear why you return here - why is the inverse minus the regularized version?

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
+                                  - angle[mask_0] ** 4 / 720
+                                  - angle[mask_0] ** 6 / 30240)
+                coef_2[mask_0] = (1 / 12 + angle[mask_0] ** 2 / 720
+                                  + angle[mask_0] ** 4 / 30240
+                                  + angle[mask_0] ** 6 / 1209600)
+
+                mask_pi = gs.isclose(angle, gs.pi)
+                mask_pi = gs.squeeze(mask_pi, axis=1)
+                delta_angle = angle[mask_pi] - gs.pi
+                coef_1[mask_pi] = (- gs.pi * delta_angle / 4
```
johmathe: ditto

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
+                coef_1[mask_else] = ((angle[mask_else] / 2)
+                                     / gs.tan(angle[mask_else] / 2))
+                coef_2[mask_else] = ((1 - coef_1[mask_else])
+                                     / angle[mask_else] ** 2)
+
+                jacobian = gs.zeros((n_points, self.dimension, self.dimension))
+
+                for i in range(n_points):
+                    if left_or_right == 'left':
+                        jacobian[i] = (
```
johmathe: maybe just define a skew_sign and assign -1 or 1 depending on the direction and do one line for the jacobian after? Or the first part of the equation and add or substract the skew matrix based on the left_or_right?

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
+
+        if point_representation is None:
+            point_representation = self.point_representation
+
+        if point_representation == 'vector':
+            if self.n == 3:
+                inv_point = -self.regularize(point)
+                return inv_point
+            else:
+                point = self.matrix_from_rotation_vector(point)
```
johmathe: shouldn't the case n == 3 be covered by this?

geomstats/special_orthogonal_group.py
```diff
@@ -112,29 +112,49 @@ class SpecialOrthogonalGroup(LieGroup, EmbeddedManifold):
     i.e. the Lie group of rotations.
     """
 
-    def __init__(self, n):
+    def __init__(self, n, point_representation=None):
```
johmathe: would just "representation" instead of point_representation work?

if not would point_type work?

just trying to make things shorter.

geomstats/special_orthogonal_group.py
```diff
@@ -480,165 +526,243 @@ def matrix_from_quaternion(self, quaternion):
...
         """
         Compute the group exponential barycenter in SO(n), which is the
         Frechet mean of the canonical bi-invariant metric on SO(n).
         """
-        n_points = points.shape[0]
-        assert n_points > 0
+        if point_representation is None:
+            point_representation = self.point_representation
+
+        if point_representation == 'vector':
```
johmathe: elif?

tests/test_special_orthogonal_group.py
```diff
@@ -717,7 +756,12 @@ def test_inverse_vectorization(self):
             n_samples = self.n_samples
             points = group.random_uniform(n_samples=n_samples)
             result = group.inverse(points)
-            self.assertTrue(result.shape == (n_samples, group.dimension))
+
+            if n == 3:
+                self.assertTrue(result.shape == (n_samples, group.dimension))
```
johmathe: why is this a special case? Shouldn't it also be n, n?

geomstats/special_orthogonal_group.py
```diff
@@ -144,114 +164,140 @@ def regularize(self, point):
...
+                                              point_representation)
+
+                regularized_tangent_vec = gs.dot(tangent_vec_at_id,
+                                                 gs.transpose(jacobian,
+                                                              axes=(0, 2, 1)))
+                regularized_tangent_vec = gs.squeeze(regularized_tangent_vec,
+                                                     axis=1)
+            else:
+                # TODO(nina): is regularization needed in nD?
+                regularized_tangent_vec = tangent_vec
```
johmathe: do you want to make a copy here? because this is not a copy.

geomstats/special_orthogonal_group.py
ninamiolane: We had them defined for the hypersphere and hyperbolic space for not here. My bad. It's corrected now.

geomstats/special_orthogonal_group.py
ninamiolane: coef_1 is a coefficient, thus a scalar. For vectorization purposes, coef_1 becomes a column vector where each line represents a point. For ease of following computations, I make coef_1 a gs.array of ndim=2. Should I change anything?

geomstats/special_orthogonal_group.py
ninamiolane: if you take a rotation represented by a rotation vector (in 3D), then its inverse is: minus the rotation vector.

Regularizing before or after the minus doesn't change anything, (operations are commutative in this case). I chose to first regularize, then take the minus.

geomstats/special_orthogonal_group.py
ninamiolane: Ah yes, good idea.

geomstats/special_orthogonal_group.py
ninamiolane: Yes, but I distinguish the case n==3 for efficiency (instead of converting to matrix and back). Shouldn't I?

geomstats/special_orthogonal_group.py
ninamiolane: Not here, but at a lot of other places: yes.

tests/test_special_orthogonal_group.py
ninamiolane: Because the default point_type for n==3 is 'vector', and the default point_type for n!=3 is 'matrix'.

geomstats/special_orthogonal_group.py
ninamiolane: hm right now I don't know since this 'else' doesn't do much. but let's write gs.copy for later. thanks!

ninamiolane:
PTAL.

PR #112, opened 2018-07-24
Clean-up on the use of n

ninamiolane:
In SO(n), we test if n==3 or not, as the implementation differs in these two cases. 

I've cleaned the code so that we consistently use the variable 'n' for these ifs, and not the length of the input vector.


geomstats/special_orthogonal_group.py
```diff
@@ -36,8 +36,9 @@ def closest_rotation_matrix(mat):
 
     n_mats, mat_dim_1, mat_dim_2 = mat.shape
     assert mat_dim_1 == mat_dim_2
+    n = mat_dim_1
```
johmathe: Convert functions to methods

PR #122, opened 2018-08-14
Nina debug tait bryan

ninamiolane:
Add unit tests for the tait bryan angles in SO3.

### DIFF
--- a/geomstats/special_orthogonal_group.py
+++ b/geomstats/special_orthogonal_group.py
@@ -676,6 +676,10 @@ class SpecialOrthogonalGroup(LieGroup, EmbeddedManifold):
         assert self.n == 3, ('The Tait-Bryan angles representation'
                              ' does not exist'
                              ' for rotations in %d dimensions.' % self.n)
+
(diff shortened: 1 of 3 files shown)


geomstats/special_orthogonal_group.py
```diff
@@ -672,23 +678,31 @@ def matrix_from_tait_bryan_angles(self, tait_bryan_angles,
...
-                       or (extrinsic_or_intrinsic == 'intrinsic'
-                           and order == 'xyz'))
-
-        extrinsic_xyz_or_intrinsic_zyx = (
-                       (extrinsic_or_intrinsic == 'extrinsic'
-                        and order == 'xyz')
-                       or (extrinsic_or_intrinsic == 'intrinsic'
-                           and order == 'zyx'))
-        if extrinsic_zyx_or_intrinsic_xyz:
+        extrinsic_zyx = (extrinsic_or_intrinsic == 'extrinsic'
```
johmathe: could you do a check like assert order in ('zyx', 'xxx')
extrinsic or intrinsic in ('foo', 'bar')

PR #125, opened 2018-08-16
Ajout d'un test unitaire pour random_vonMisesFisher

AliceLeBrigant:


tests/test_hypersphere.py
```diff
@@ -396,6 +396,34 @@ def test_mean_and_belongs(self):
         result = self.metric.mean([point_a, point_b, point_c])
         self.assertTrue(self.space.belongs(result))
 
+    def test_random_vonMisesFisher(self):
```
ninamiolane: Update method's name: random_von_mises_fisher

ninamiolane: And update variable names as well (see PR #124). Thanks!

tests/test_hypersphere.py
```diff
@@ -396,6 +396,34 @@ def test_mean_and_belongs(self):
...
+        sphere = Hypersphere(dim)
+
+        # check mean value for concentrated distribution
+        K = 10
+        points = sphere.random_vonMisesFisher(K, n_points)
+        S = gs.sum(points, axis=0)
+        mu_hat = S/gs.linalg.norm(S)
+        mu = gs.array([0, 0, 1])
+        error = sphere.metric.dist(mu, mu_hat)
+        assert error < 1e-2
```
ninamiolane: Define the constant: `TOLERANCE = 1e-2` (could we do better? like 1e-6?) at the beginning of the file.

Then:  
```
result = mean_estimate
expected = mean 
self.assertTrue(gs.allclose(result, expected, atol=TOLERANCE))
``` 
in order to use the method of the `unittest.TestCase` class.

tests/test_hypersphere.py
```diff
@@ -396,6 +396,34 @@ def test_mean_and_belongs(self):
...
+        """
+        dim = 2
+        n_points = 1000000
+        sphere = Hypersphere(dim)
+
+        # check mean value for concentrated distribution
+        K = 10
+        points = sphere.random_vonMisesFisher(K, n_points)
+        S = gs.sum(points, axis=0)
+        mu_hat = S/gs.linalg.norm(S)
```
ninamiolane: could we write `mean` instead of `mu`? the code is more readable if the variable can be interpreted in english without refering to notation convention (especially for future contributors).

ninamiolane: or `mean_estimate`?

tests/test_hypersphere.py
```diff
@@ -396,6 +396,34 @@ def test_mean_and_belongs(self):
...
+        mu = gs.array([0, 0, 1])
+        error = sphere.metric.dist(mu, mu_hat)
+        assert error < 1e-2
+
+        # check concentration parameter for dispersed distribution
+        K = 0.5
+        points = sphere.random_vonMisesFisher(K, n_points)
+        S = gs.sum(points, axis=0)
+        R = gs.linalg.norm(S)/n_points
+        K_hat = R*(dim + 1 - R**2)/(1 - R**2)
```
ninamiolane: Maybe use flake8 to clean the style here.

tests/test_hypersphere.py
```diff
@@ -396,6 +396,34 @@ def test_mean_and_belongs(self):
...
+        error = gs.abs(K-K_hat)/K
+        assert error < 1e-2
```
ninamiolane: Same as above for the test.

tests/test_hypersphere.py
AliceLeBrigant: If we only test the mean value, we can lower the tolerance to 1e-6 by sampling from a much more concentrated distribution (concentration parameter K=100000). See the commit https://github.com/AliceLeBrigant/geomstats/commit/794b49dc7e41927f2a86c43650cdc6d1ff10f330. 

However I could not lower the tolerance below 1e-3 for the estimation of the concentration parameter. What would you recommend ?

PR #129, opened 2018-08-21
Addition and test of random_von_mises_fisher method

AliceLeBrigant:


geomstats/hypersphere.py
```diff
@@ -135,6 +135,22 @@ def random_uniform(self, n_samples=1):
...
+        centered in the north pole.
+        """
+        if self.dimension != 2:
+            raise NotImplementedError(
+                    'Sampling from the von Mises Fisher distribution'
+                    'is only implemented in dimension 2.')
+        angle = 2 * gs.pi * gs.random.rand(n_samples)
+        unit_vector = gs.vstack((gs.cos(angle), gs.sin(angle)))
+        scalar = gs.random.rand(n_samples)
+        coord_z = 1 + 1/kappa*gs.log(scalar + (1-scalar)*gs.exp(-2*kappa))
```
ninamiolane: Answering here to your comment on PR #124 : Thanks! I didn't know about the lowest priority rule :) !

tests/test_hypersphere.py
```diff
@@ -396,6 +398,24 @@ def test_mean_and_belongs(self):
...
+        """
+        dim = 2
+        n_points = 1000000
+        sphere = Hypersphere(dim)
+
+        kappa = 1000000
+        points = sphere.random_von_mises_fisher(kappa, n_points)
+        sum_points = gs.sum(points, axis=0)
+        mean = gs.array([0, 0, 1])
+        mean_estimate = sum_points/gs.linalg.norm(sum_points)
```
ninamiolane: Here there should be a space round the operator, right?

AliceLeBrigant: Yes that is completely right !

AliceLeBrigant:
I had forgotten a line in the random_von_mises_fisher method... sorry about that ! It should work better now :)

AliceLeBrigant:
I have also re-integrated the estimation of the concentration parameter kappa with a higher tolerance (1e-3) than the mean estimation tolerance. It is still a little better than the first version (1e-2) thanks to a Newton scheme.

tests/test_hypersphere.py
```diff
@@ -3,12 +3,16 @@
 """
 
 import unittest
+import scipy.special
```
ninamiolane: import order: since s < u scipy should come first.

https://www.python.org/dev/peps/pep-0008/#imports

PR #133, opened 2018-08-29
Add closest_neighbor and diameter to riemannian_metric

AliceLeBrigant:
...and an "unbounded" version of `random_uniform` to `hypersphere.py`, that I use in the `plot_quantization_s1.py` example. I couldn't find a better name...

### DIFF
--- a/geomstats/backend/numpy.py
+++ b/geomstats/backend/numpy.py
@@ -164,6 +164,10 @@ def rand(*args, **largs):
     return np.random.rand(*args, **largs)
 
 
+def randint(*args, **kwargs):
+    return np.random.randint(*args, **kwargs)
+
+
 def isclose(*args, **kwargs):
     return np.isclose(*args, **kwargs)
 
@@ -298,3 +302,15 @@ def nonzero(x):
 
 def copy(x):
     return np.copy(x)
+
+
+def ix_(*args):
+    return np.ix_(*args)
+
+
+def arange(*args, **kwargs):
+    return np.arange(*args, **kwargs)
+
+
+def prod(a, axis=None):
+    return np.prod(a, axis=axis)
--- a/geomstats/backend/numpy.py
+++ b/geomstats/backend/numpy.py
@@ -272,6 +272,10 @@ def argmax(*args, **kwargs):
     return np.argmax(*args, **kwargs)
 
 
+def argmin(*args, **kwargs):
+    return np.argmin(*args, **kwargs)
+
+
 def diagonal(*args, **kwargs):
     return np.diagonal(*args, **kwargs)
 
--- a/geomstats/hypersphere.py
+++ b/geomstats/hypersphere.py
@@ -124,7 +124,7 @@ class Hypersphere(EmbeddedManifold):
 
         return point_intrinsic
 
-    def random_uniform(self, bound=0.5, n_samples=1):
+    def random_uniform(self, n_samples=1, bound=0.5):
         """
         Sample in the Hypersphere with the uniform distribution.
         """
--- a/geomstats/hypersphere.py
+++ b/geomstats/hypersphere.py
@@ -124,14 +124,25 @@ class Hypersphere(EmbeddedManifold):
 
         return point_intrinsic
 
-    def random_uniform(self, n_samples=1):
+    def random_uniform(self, bound=0.5, n_samples=1):
         """
         Sample in the Hypersphere with the uniform distribution.
         """
         size = (n_samples, self.dimension)
-        point = gs.random.rand(*size) - 0.5
 
-        point = self.intrinsic_to_extrinsic_coords(point)
+        if bound is None:
+            spherical_coord = gs.random.rand(*size) * gs.pi
+            spherical_coord[:, -1] *= 2
+
+            point = gs.zeros((n_samples, self.dimension+1))
+            for i in range(self.dimension):
+                point[:, i] = gs.prod(gs.sin(spherical_coord[:, :i]), axis=1)\
+                            * gs.cos(spherical_coord[:, i])
+            point[:, -1] = gs.prod(gs.sin(spherical_coord), axis=1)
+
+        else:
+            point = bound * (2*gs.random.rand(*size) - 1)
+            point = self.intrinsic_to_extrinsic_coords(point)
 
         return point
 
--- a/geomstats/riemannian_metric.py
+++ b/geomstats/riemannian_metric.py
@@ -327,17 +327,16 @@ class RiemannianMetric(object):
 
         for i in range(n_points-1):
             dist_to_neighbors = self.dist(points[i, :], points[i+1:, :])
-            dist_to_farthest_neighbor = dist_to_neighbors.max()
-            diameter = gs.maximum(diameter, dist_to_farthest_neighbor * 2)
+            dist_to_farthest_neighbor = gs.amax(dist_to_neighbors)
+            diameter = gs.maximum(diameter, dist_to_farthest_neighbor)
 
         return diameter
 
-    def closest_neighbor(self, point, neighbors):
+    def closest_neighbor_index(self, point, neighbors):
         """
         Closest neighbor of point among neighbors.
         """
         dist = self.dist(point, neighbors)
-        index_closest_neighbor = dist.argmin()
+        closest_neighbor_index = gs.argmin(dist)
 
-        return index_closest_neighbor
-
+        return closest_neighbor_index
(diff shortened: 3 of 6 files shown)


geomstats/backend/numpy.py
```diff
@@ -8,10 +8,6 @@
 float64 = np.float64
 
 
-def flip(*args, **kwargs):
```
ninamiolane: Can you rebase on the latest version of master? This function has been added in a recent pull request.

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

geomstats/backend/numpy.py
```diff
@@ -294,3 +294,15 @@ def nonzero(x):
...
+def ix_(*args):
+    return np.ix_(*args)
+
+
+def arange(*args, **kwargs):
+    return np.arange(*args, **kwargs)
+
+
+def prod(a, axis=None):
+    return np.prod(a, axis=None)
```
ninamiolane: `np.prod(a, axis=axis)`

geomstats/hypersphere.py
```diff
@@ -135,6 +135,24 @@ def random_uniform(self, n_samples=1):
 
         return point
 
+    def random_uniform_unbounded(self, n_samples=1):
```
ninamiolane: How about merging this into the function `random_uniform` as:
```
random_uniform(self, bound=None, n_samples=1)
```
and distinguishing the case when there is no bound (`bound` is None) and when there is a bound?

geomstats/riemannian_metric.py
```diff
@@ -295,3 +295,28 @@ def tangent_pca(self, points, base_point=None):
...
+        Distance between the two points that are farthest away from each other
+        in points.
+        """
+        diameter = 0.0
+        n_points = points.shape[0]
+
+        for i in range(n_points-1):
+            dist_to_neighbors = self.dist(points[i, :], points[i+1:, :])
+            dist_to_farthest_neighbor = dist_to_neighbors.max()
+            diameter = gs.maximum(diameter, dist_to_farthest_neighbor * 2)
```
ninamiolane: Why `* 2`?

geomstats/riemannian_metric.py
```diff
@@ -295,3 +295,28 @@ def tangent_pca(self, points, base_point=None):
...
+        """
+        Distance between the two points that are farthest away from each other
+        in points.
+        """
+        diameter = 0.0
+        n_points = points.shape[0]
+
+        for i in range(n_points-1):
+            dist_to_neighbors = self.dist(points[i, :], points[i+1:, :])
+            dist_to_farthest_neighbor = dist_to_neighbors.max()
```
ninamiolane: Could we do something like `gs.max`? I'm worried that the syntax `dist_to_neightbors.max()` might not translate well with the other backends (pytorch, tf).

geomstats/riemannian_metric.py
```diff
@@ -295,3 +295,28 @@ def tangent_pca(self, points, base_point=None):
...
+            diameter = gs.maximum(diameter, dist_to_farthest_neighbor * 2)
+
+        return diameter
+
+    def closest_neighbor(self, point, neighbors):
+        """
+        Closest neighbor of point among neighbors.
+        """
+        dist = self.dist(point, neighbors)
+        index_closest_neighbor = dist.argmin()
```
ninamiolane: Same as above: something like `gs.argmin`?

tests/test_hypersphere.py
```diff
@@ -400,6 +400,34 @@ def test_mean_and_belongs(self):
         result = self.metric.mean([point_a, point_b, point_c])
         self.assertTrue(self.space.belongs(result))
 
+    def test_diameter(self):
```
ninamiolane: Yes! Thanks for the unit test!

tests/test_hypersphere.py
```diff
@@ -400,6 +400,34 @@ def test_mean_and_belongs(self):
...
+        n_points = 10
+        points = self.space.random_uniform(n_points)
+        result = self.metric.diameter(points)
+        gs.testing.assert_allclose(result.size, 1)
```
ninamiolane: Could you also test the value of the result (not only its size) with a handmade set of points (maybe 3 points?). Thank you.

geomstats/riemannian_metric.py
```diff
@@ -295,3 +295,28 @@ def tangent_pca(self, points, base_point=None):
...
+
+        return index_closest_neighbor
```
ninamiolane: either return the closest neighbor or rename the function: `closest_neighbor_index`?

tests/test_hypersphere.py
```diff
@@ -400,6 +400,34 @@ def test_mean_and_belongs(self):
...
+        Check that the closest neighbor is one of neighbors.
+        """
+        n_points = 10
+        points = self.space.random_uniform(n_points)
+        point = points[0, :]
+        neighbors = points[1:, :]
+        index = self.metric.closest_neighbor(point, neighbors)
+        closest_neighbor = points[index, :]
+        result = False
+        for i in range(n_points):
```
ninamiolane: Is there a numpy builtin function that could do this? Something like `np.isin` (-> `gs.isin`)?

geomstats/hypersphere.py
```diff
@@ -135,6 +135,24 @@ def random_uniform(self, n_samples=1):
...
+        """
+        Sample in the whole of the Hypersphere with the uniform
+        distribution.
+        """
+        size = (n_samples, self.dimension)
+        spherical_coord = gs.random.rand(*size) * gs.pi
+        spherical_coord[:, -1] *= 2
```
ninamiolane: Matrix multiplication with gs.pi * gs.array([1, 2]) would make the code slightly more readable.

geomstats/backend/numpy.py
```diff
@@ -294,3 +294,15 @@ def nonzero(x):
 
 def copy(x):
     return np.copy(x)
+
+
+def ix_(*args):
```
ninamiolane: Did you add this? Where do you use it? Same for randint and arange.

geomstats/backend/numpy.py
ninamiolane: (it will solve the unit test failure)

geomstats/backend/numpy.py
ninamiolane: Nvm, I found them in the examples!

geomstats/backend/numpy.py
AliceLeBrigant: It took me a while to figure it out but it's done ! Hope it worked.

geomstats/riemannian_metric.py
AliceLeBrigant: You are right, there is no *2 necessary. Corrected.

geomstats/hypersphere.py
AliceLeBrigant: Since it depends on the dimension, it would require to add ```* gs.hstack([gs.ones(self.dimension - 1), 2])``` which seems more involved. But if you still think it is more readable (I trust your experience !) I can change it.

tests/test_hypersphere.py
AliceLeBrigant: It seems that ```np.isin``` works element-wise, whereas here I want to find an entire row. But you are right, I found a way to do it with numpy built-in functions :)

tests/test_hypersphere.py
```diff
@@ -403,6 +408,31 @@ def test_mean_and_belongs(self):
...
+        sphere = Hypersphere(dim)
+        point_a = [0., 0., 1.]
+        point_b = [1., 0., 0.]
+        point_c = [0., 0., -1.]
+        result = sphere.metric.diameter(gs.vstack((point_a, point_b, point_c)))
+        expected = gs.pi
+        gs.testing.assert_allclose(result, expected)
+        gs.testing.assert_allclose(result.size, 1)
+
+    def test_closest_neighbor(self):
```
ninamiolane: NIT: test_closest_neighbor_index

ninamiolane:
@AliceLeBrigant there's one unit test failing for plotting the point on S2: could you have a look?

Otherwise it's good to go! Thank you!

AliceLeBrigant:
@ninamiolane which unit test is failing exactly ? I am running ```test_hypersphere.py``` but I am not getting any failures.

Also I don't think I have writing access to this repo and so I don't seem to be able to merge pull requests.

AliceLeBrigant:
Never mind, I found it and it should be fixed ! I still don't seem to have the authorization to merge though.

ninamiolane:
Awesome, thank you!

PR #137, opened 2018-09-17
 ENH: Support optional dependencies, geomstats.__version__ 

### LINKED ISSUE #136: Split out dependencies into extras
A project that wants to use `geomstats` as a dependency is going to pull in several further dependencies that are not strictly necessary to use the package:

* For testing: `codecov`, `coverage`, `nose2`
* Extra backends: `tensorflow`, `torch`

`matplotlib` is mostly used for examples and a self-contained visualization module, so that could go either way, depending on how integral you think the visualizations are to the functionality of the package.

The way to handle these is to use the `extras_requires` key in `setup.py`, e.g.:

```Python
install_requires = [
    'autograd',
    'numpy>=1.14.1',
    'scipy',
    'matplotlib',
    ]

extras_requires = {
    'test': ['nose2', 'coverage', 'codecov'],
    'tensorflow': ['tensorflow>=1.8'],
    'torch': ['torch==0.4.0'],
    }
extras_requires['all'] = sum(extras_requires.values(), [])

setup(name='geomstats',
      version='1.11',
      install_requires=install_requires,
      extras_requires=extras_requires,
      description='Geometric statistics on manifolds',
      url='http://github.com/geomstats/geomstats',
      author='Nina Miolane',
      author_email='ninamio78@gmail.com',
      license='MIT',
      packages=find_packages(),
      zip_safe=False)
```

Then you can install with `pip install "geomstats[tensorflow]"` to get tensorflow capabilities, or `pip install "geomstats[all]"` if you want full functionality.

This will help with nipy/nipype#2607, which is at least partially held up because tensorflow does not yet support Python 3.7. It will also help keep geomstats lightweight for people who are only using it for a function or two, and do not need tensorflow or pytorch.

effigies:
This PR adds "extra" installs, which allow optional functionality to be installed only if requested. It also adds a `geomstats.__version__` string, which is generally a good practice.

I use the `<package>/__about__.py` convention to store this metadata, so it's available both within the package and without.

Happy to rework if this doesn't fit with your development workflow.

Closes #136.

### DIFF
--- a/geomstats/__about__.py
+++ b/geomstats/__about__.py
@@ -1,0 +1,15 @@
+__version__ = '1.11-dev'  # Remove -dev before releasing
+
+install_requires = [
+    'autograd',
+    'numpy>=1.14.1',
+    'scipy',
+    'matplotlib',
+    ]
+
+extras_require = {
+    'test': ['nose2', 'coverage', 'codecov'],
+    'tf': ['tensorflow>=1.8'],
+    'torch': ['torch==0.4.0'],
+    }
+extras_require['all'] = sum(extras_require.values(), [])
--- a/geomstats/__about__.py
+++ b/geomstats/__about__.py
@@ -1,15 +1,18 @@
-__version__ = '1.11-dev'  # Remove -dev before releasing
+# Remove -dev before releasing
+__version__ = '1.11-dev'
+
+from itertools import chain
 
 install_requires = [
     'autograd',
+    'matplotlib',
     'numpy>=1.14.1',
     'scipy',
-    'matplotlib',
     ]
 
 extras_require = {
-    'test': ['nose2', 'coverage', 'codecov'],
+    'test': ['codecov', 'coverage', 'nose2'],
     'tf': ['tensorflow>=1.8'],
     'torch': ['torch==0.4.0'],
     }
-extras_require['all'] = sum(extras_require.values(), [])
+extras_require['all'] = list(chain(*extras_require.values()))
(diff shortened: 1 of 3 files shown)


geomstats/__about__.py
```diff
@@ -0,0 +1,15 @@
+__version__ = '1.11-dev'  # Remove -dev before releasing
+
+install_requires = [
+    'autograd',
```
johmathe: alphabetical order?

geomstats/__about__.py
```diff
@@ -0,0 +1,15 @@
...
+    'scipy',
+    'matplotlib',
+    ]
+
+extras_require = {
+    'test': ['nose2', 'coverage', 'codecov'],
+    'tf': ['tensorflow>=1.8'],
+    'torch': ['torch==0.4.0'],
+    }
+extras_require['all'] = sum(extras_require.values(), [])
```
johmathe: nit - python doesn't recommend summing other things than numbers [1]. Also it makes the code harder to read for people not familiar with the internals of summation of lists (and is one of the least efficient methods for concatenating lists of strings - but it doens't matter here with the size).

Maybe just something like 

`list(itertools.chain(*extras_require.values())) `



[1] https://docs.python.org/3/library/functions.html#sum

effigies:
Thanks for the review and merge. If you don't have any must-have PRs in the pipeline, could I request a release soon (ideally within the next week)? It would be nice to depend on this without pulling in tensorflow.

PR #151, opened 2018-09-29
Nina gln tf

ninamiolane:


tests/test_general_linear_group_tensorflow.py
```diff
@@ -0,0 +1,104 @@
...
+        inv_mat = self.group.inverse(mat)
+
+        result = self.group.compose(mat, inv_mat)
+        expected = self.group.identity
+        expected = helper.to_matrix(expected)
+
+        with self.test_session():
+            self.assertAllClose(gs.eval(result), gs.eval(expected))
+
+        # 2. Compose transformation by its inverse on the left
```
johmathe: you should split this test in two parts - *unit* tests :)

PR #152, opened 2018-09-30
Add quantization as method of RiemannianMetric

AliceLeBrigant:
I have changed the name ```optimal_quantization``` to ```riemannian_quantization``` and added it as a method of the ```RiemannianMetric``` class as we previously discussed. The test is in ```test_hypersphere.py``` and I moved the examples to https://github.com/geomstats/geomstats/tree/master/examples. 

If this is to be merged eventually, it would mean deleting the quantization folder in https://github.com/geomstats/examples.


AliceLeBrigant:
I have changed my mind about the name. I figured every method of ```RiemannianMetric``` is Riemannian... :)

PR #157, opened 2018-10-24
Create DiscretizedCurvesSpace

AliceLeBrigant:
I have implemented two different metrics for curves:
- the L2 metric
- the Square Root Velocity (SRV) metric (see e.g. https://ieeexplore.ieee.org/abstract/document/5601739)

The SRV framework is implemented only for curves in a Euclidean space for starters (I have to figure out how to test the fact that the base manifold is Euclidean), because the formulas are considerably simpler in that setting. 

I needed the ```inner_product_matrix``` method of ```Hypersphere``` so I added it.

### DIFF
--- a/geomstats/discretized_curves_space.py
+++ b/geomstats/discretized_curves_space.py
@@ -20,14 +20,14 @@ R3 = EuclideanSpace(dimension=3)
 
 class DiscretizedCurvesSpace(EmbeddedManifold):
     """
-    Space of parameterized curves sampled at points in embedding_manifold.
+    Space of discretized curves sampled at points in embedding_manifold.
     """
     def __init__(self, embedding_manifold):
         super(DiscretizedCurvesSpace, self).__init__(
                 dimension=math.inf,
                 embedding_manifold=embedding_manifold)
-        self.L2_metric = L2Metric(self.embedding_manifold)
-        self.SRV_metric = SRVMetric(self.embedding_manifold)
+        self.l2_metric = L2Metric(self.embedding_manifold)
+        self.square_root_velocity_metric = SRVMetric(self.embedding_manifold)
 
     def belongs(self, point):
         return gs.all(self.embedding_manifold.belongs(point))
@@ -35,144 +35,153 @@ class DiscretizedCurvesSpace(EmbeddedManifold):
 
 class L2Metric(RiemannianMetric):
     """
-    L2 Riemannian metric on the space of parameterized curves or surfaces.
+    L2 Riemannian metric on the space of discretized curves.
     """
     def __init__(self, embedding_manifold):
         super(L2Metric, self).__init__(
-                dimension=embedding_manifold.dimension,
-                signature=(embedding_manifold.dimension, 0, 0))
+                dimension=math.inf,
+                signature=(math.inf, 0, 0))
+        self.embedding_manifold = embedding_manifold
         self.embedding_metric = embedding_manifold.metric
 
-    def inner_product(self, tangent_vec_a, tangent_vec_b, base_point):
+    def inner_product(self, tangent_vec_a, tangent_vec_b, base_curve):
         """
-        Inner product between two tangent vectors at a base point.
+        Inner product between two tangent vectors at a base curve.
         """
         assert tangent_vec_a.shape == tangent_vec_b.shape
-        assert tangent_vec_a.shape == base_point.shape
-        n_coord = base_point.shape[-1]
-        n_sampling_points = base_point.shape[-2]
-        shape_cumprod = np.cumprod(base_point.shape)
-        tangent_vec_a = tangent_vec_a.reshape(shape_cumprod[-2], n_coord)
-        tangent_vec_b = tangent_vec_b.reshape(shape_cumprod[-2], n_coord)
+        assert tangent_vec_a.shape == base_curve.shape
+        tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=3)
+        tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
+
+        n_curves = tangent_vec_a.shape[0]
+        n_sampling_points = tangent_vec_a.shape[1]
+        n_coords = tangent_vec_a.shape[2]
+
+        new_dim = n_curves * n_sampling_points
+        tangent_vec_a = tangent_vec_a.reshape(new_dim, n_coords)
+        tangent_vec_b = tangent_vec_b.reshape(new_dim, n_coords)
+        base_curve = base_curve.reshape(new_dim, n_coords)
 
         inner_prod = self.embedding_metric.inner_product(
-                tangent_vec_a, tangent_vec_b, base_point)
-        inner_prod = inner_prod.reshape(base_point.shape[:-1])
+                tangent_vec_a, tangent_vec_b, base_curve)
+        inner_prod = inner_prod.reshape(n_curves, n_sampling_points)
         inner_prod = gs.sum(inner_prod, -1) / n_sampling_points
 
         return inner_prod
 
-    def dist(self, point_a, point_b):
+    def dist(self, curve_a, curve_b):
         """
-        Geodesic distance between two parameterized manifolds
-        sampled at points_a and points_b.
+        Geodesic distance between two discretized curves.
         """
-        assert point_a.shape == point_b.shape
-        shape_a = point_a.shape
-        n_coord = shape_a[-1]
-        n_sampling_points = shape_a[-2]
-        shape_cumprod = np.cumprod(point_a.shape)
-        point_a = point_a.reshape(shape_cumprod[-2], n_coord)
-        point_b = point_b.reshape(shape_cumprod[-2], n_coord)
+        assert curve_a.shape == curve_b.shape
+        curve_a = gs.to_ndarray(curve_a, to_ndim=3)
+        curve_b = gs.to_ndarray(curve_b, to_ndim=3)
 
-        dist = self.embedding_metric.dist(point_a, point_b)
-        dist = dist.reshape(shape_a[:-1])
+        n_curves = curve_a.shape[0]
+        n_sampling_points = curve_a.shape[1]
+        n_coords = curve_a.shape[2]
+
+        curve_a = curve_a.reshape(n_curves * n_sampling_points, n_coords)
+        curve_b = curve_b.reshape(n_curves * n_sampling_points, n_coords)
+
+        dist = self.embedding_metric.dist(curve_a, curve_b)
+        dist = dist.reshape(n_curves, n_sampling_points)
         dist = gs.sqrt(gs.sum(dist ** 2, -1) / n_sampling_points)
 
         return dist
 
-    def exp(self, tangent_vec, base_point):
+    def exp(self, tangent_vec, base_curve):
         """
-        Riemannian exponential of a tangent vector wrt to a base point.
+        Riemannian exponential of a tangent vector wrt to a base curve.
         """
-        base_point = base_point.squeeze()
-        mult_tangent_vecs = 0
-        if tangent_vec.ndim > base_point.ndim:
-            mult_tangent_vecs = 1
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
 
-        point_shape = base_point.shape
-        tangent_vec_shape = tangent_vec.shape
+        n_curves = base_curve.shape[0]
+        n_sampling_points = base_curve.shape[1]
+        n_coords = base_curve.shape[2]
+        n_tangent_vecs = tangent_vec.shape[0]
 
-        n_coord = point_shape[-1]
-        shape_cumprod = np.cumprod(point_shape)
+        new_dim = n_curves * n_sampling_points
+        new_base_curve = base_curve.reshape(new_dim, n_coords)
+        new_tangent_vec = tangent_vec.reshape(new_dim, n_coords)
 
-        new_base_point = base_point.reshape(shape_cumprod[-2], n_coord)
-        if mult_tangent_vecs == 1:
-            new_shape = [tangent_vec.shape[0], shape_cumprod[-2], n_coord]
-        else:
-            new_shape = [shape_cumprod[-2], n_coord]
-        new_tangent_vec = tangent_vec.reshape(new_shape)
-
-        exp = self.embedding_metric.exp(new_tangent_vec, new_base_point)
-        exp = exp.reshape(tangent_vec_shape)
+        exp = self.embedding_metric.exp(new_tangent_vec, new_base_curve)
+        exp = exp.reshape(n_tangent_vecs, n_sampling_points, n_coords)
+        exp = exp.squeeze()
 
         return exp
 
-    def log(self, point, base_point):
+    def log(self, curve, base_curve):
         """
-        Riemannian logarithm of a point wrt a base point.
+        Riemannian logarithm of a curve wrt a base curve.
         """
-        assert point.shape == base_point.shape
-        point_shape = point.shape
-        n_coord = point_shape[-1]
-        shape_cumprod = np.cumprod(point_shape)
+        assert curve.shape == base_curve.shape
+        curve = gs.to_ndarray(curve, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
 
-        point = point.reshape(shape_cumprod[-2], n_coord)
-        base_point = base_point.reshape(shape_cumprod[-2], n_coord)
-        log = self.embedding_metric.log(point, base_point)
-        log = log.reshape(point_shape)
+        n_curves = curve.shape[0]
+        n_sampling_points = curve.shape[1]
+        n_coords = curve.shape[2]
+
+        curve = curve.reshape(n_curves * n_sampling_points, n_coords)
+        base_curve = base_curve.reshape(n_curves * n_sampling_points, n_coords)
+        log = self.embedding_metric.log(curve, base_curve)
+        log = log.reshape(n_curves, n_sampling_points, n_coords)
+        log = log.squeeze()
 
         return log
 
-    def geodesic(self, initial_point,
-                 end_point=None, initial_tangent_vec=None):
+    def geodesic(self, initial_curve,
+                 end_curve=None, initial_tangent_vec=None):
         """
         Geodesic specified either by an initial point and an end point,
         either by an initial point and an initial tangent vector.
         """
         #TODO(alice): vectorize
-        point_ndim = 2
-        point_shape = initial_point.shape
-        initial_point = gs.to_ndarray(initial_point,
-                                      to_ndim=point_ndim+1)
+        curve_ndim = 2
+        initial_curve = gs.to_ndarray(initial_curve,
+                                      to_ndim=curve_ndim+1)
+        n_sampling_points = initial_curve.shape[1]
+        n_coords = initial_curve.shape[2]
 
-        if end_point is None and initial_tangent_vec is None:
-            raise ValueError('Specify an end point or an initial tangent '
+        if end_curve is None and initial_tangent_vec is None:
+            raise ValueError('Specify an end curve or an initial tangent '
                              'vector to define the geodesic.')
-        if end_point is not None:
-            end_point = gs.to_ndarray(end_point,
-                                      to_ndim=point_ndim+1)
-            shooting_tangent_vec = self.log(point=end_point,
-                                            base_point=initial_point)
+        if end_curve is not None:
+            end_curve = gs.to_ndarray(end_curve,
+                                      to_ndim=curve_ndim+1)
+            shooting_tangent_vec = self.log(curve=end_curve,
+                                            base_curve=initial_curve)
             if initial_tangent_vec is not None:
                 assert gs.allclose(shooting_tangent_vec, initial_tangent_vec)
             initial_tangent_vec = shooting_tangent_vec
         initial_tangent_vec = gs.array(initial_tangent_vec)
         initial_tangent_vec = gs.to_ndarray(initial_tangent_vec,
-                                            to_ndim=point_ndim+1)
+                                            to_ndim=curve_ndim+1)
 
-        def point_on_geodesic(t):
+        def curve_on_geodesic(t):
             t = gs.cast(t, gs.float32)
             t = gs.to_ndarray(t, to_ndim=1)
             t = gs.to_ndarray(t, to_ndim=2, axis=1)
-            new_initial_point = gs.to_ndarray(
-                                          initial_point,
-                                          to_ndim=point_ndim+1)
-            new_initial_tangent_vec = gs.to_ndarray(
-                                          initial_tangent_vec,
-                                          to_ndim=point_ndim+1)
+            new_initial_curve = gs.to_ndarray(initial_curve,
+                                              to_ndim=curve_ndim+1)
+            new_initial_tangent_vec = gs.to_ndarray(initial_tangent_vec,
+                                                    to_ndim=curve_ndim+1)
 
             tangent_vecs = gs.einsum('il,nkm->ikm', t, new_initial_tangent_vec)
 
-            shape_point_at_time_t = gs.hstack([len(t), point_shape])
-            point_at_time_t = gs.zeros(shape_point_at_time_t)
+            curve_shape_at_time_t = gs.hstack(
+                    [len(t), n_sampling_points, n_coords])
+            curve_at_time_t = gs.zeros(curve_shape_at_time_t)
             for k in range(len(t)):
-                point_at_time_t[k, :] = self.exp(
+                curve_at_time_t[k, :] = self.exp(
                         tangent_vec=tangent_vecs[k, :],
-                        base_point=new_initial_point)
-            return point_at_time_t
+                        base_curve=new_initial_curve)
+            return curve_at_time_t
 
-        return point_on_geodesic
+        return curve_on_geodesic
 
 
 class SRVMetric(RiemannianMetric):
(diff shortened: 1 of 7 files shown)


AliceLeBrigant:
Thanks for all the comments ! 

It's not really easy to determine what should the ```dimension``` variable be in ```ParameterizedManifold```. Rigorously, the value should be infinite since the space of curves or surfaces is infinite-dimensional. If we consider sampled curves or surfaces defined by a number ```n```of sample points, then it should be ```n * base_manifold.dimension```, where ```base_manifold``` is the space where the sample points live (a Euclidean space or a Riemannian manifold such as the sphere). But then, there should probably be another variable that distinguishes between curve and surface. For now, ```dimension``` is rather the dimension of the ```ParameterizedManifold``` seen as a _submanifold_ of the ambient space : ```dimension=1``` for a curve and ```dimension=2``` for a surface. What do you think ?

geomstats/parameterized_manifold.py
```diff
@@ -0,0 +1,158 @@
...
+        assert base_manifold.belongs(points).all()
+        super(ParameterizedManifold, self).__init__(
+                dimension=dimension)
+        self.points = points
+        self.base_manifold = base_manifold
+
+
+class L2Metric(RiemannianMetric):
+    """
+    L2 Riemannian metric on the space of parameterized curves or surfaces.
```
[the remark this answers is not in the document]
AliceLeBrigant: I am not sure what you mean by field. In the end, we should end up with one file with
- a class ```ParameterizedManifold```
- a class ```ParameterizedManifoldSpace```
- a class ```L2Metric``` (and other metric classes)
is that correct ?

geomstats/parameterized_manifold.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    def __init__(self, base_manifold):
+        self.base_manifold = base_manifold
```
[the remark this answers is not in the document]
AliceLeBrigant: I don't really get what is the meaning of the ```signature``` attribute of ```Riemannian Metric``` ?

ninamiolane:
Here is what I understood, let me know if I got it right or wrong?

We have a manifold of manifolds, so we have to implement in fact two classes of manifolds:

- the class `ParameterizedManifold`, already implemented: since `points` is given to the constructor, an object instantiated from this class is *one* curve (made of `points`) or *one* surface (made of `points`). The `dimension` of this manifold is 1 for curve, or 2 for surface, it is the dimension of one parameterized manifold: i.e. the `dimension` parameter is easy in this case.

One object of the above class (e.g. a curve) becomes one point in the following manifold:
- the class, not implemented yet, of all curves / all surfaces, i.e. the manifold of the parameterized manifolds. We could call this class `ParameterizedManifoldsSpace` or something like this. Rigorously, `dimension` of this manifold should be infinite indeed. How about we put `dimension = math.inf` for now?

ninamiolane:
I'm also thinking: we called this class "ParameterizedManifold", but we don't really have a parameterization so far.

geomstats/parameterized_manifold.py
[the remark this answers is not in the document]
ninamiolane: By field, I mean "attribute", in the sense of Object Oriented Programming. "Object-oriented programming (OOP) is a programming paradigm based on the concept of "objects", which may contain data, in the form of fields, often known as attributes; and code, in the form of procedures, often known as methods. "

ninamiolane:
Yes I agree. In the meantime, since we don't have a parameterization per se, how about moving the methods of the current class ` ` to the class `EmbeddedManifold`?

Yes! :D

A potential problem though, is that it would change the default shape of vectors: which is (n_points, vec_dim) everywhere by convention now. We would need to check if this doesn't impact the code.

Another potential problem with doing [n_tangent_vecs, n_points, dim] is that it might only works when the points on the manifold are represented as vectors, for example a point on the Euclidean space is represented as a vector. If points are represented as matrices, for example a point on a matrix space is represented as a matrix, it might induce confusion because if the exponential output something of shape (10, 3, 3), we don't know if it's a vector case with n_tangent_vecs = 10, n_points=3 and dim=3 or a matrix case with n_tangent_vec = 10 and a matrix of size 3x3.

But we might be able to go around! Could you open a github issue with this? Do you need it soon? Thanks!

AliceLeBrigant:
I have rebased to update my branch with the master, deleted ```parameterized_manifold.py``` and created ```discretized_curves_space.py```. 

I haven't forgotten about the ``ìnner_product_matrix``` method of ```HypersphereMetric``` that you asked me to remove, however without it I can't use the ```inner_product``` method, which I need for the same method of the ```L2_metric``` class when the embedding manifold is the sphere. I could call the ``ìnner_product``` method of the embedding manifold instead (R3 in this case), however this only works for embedded manifolds and not Euclidean spaces for example. Do you see a way around this ?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,424 @@
...
+        srv_shape[-2] -= 1
+
+        curve = curve.reshape(n_points_in_total, n_coords)
+        velocity = n_sampling_points * self.embedding_metric.log(
+                point=curve[1:, :], base_point=curve[:-1, :])
+        velocity_norm = self.embedding_metric.norm(velocity, curve[:-1, :])
+        assert gs.all(velocity_norm != 0)
+        srv = velocity / gs.sqrt(velocity_norm)
+        index = gs.arange(n_points_in_total - 1)
+        index_select = index[(index + 1) % n_sampling_points != 0]
```
ninamiolane: what is this index selection?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,424 @@
...
+                                          base_point=base_point[:, 0, :])
+        log_0 = gs.transpose(np.tile(log_0, (1, 1, 1)), (1, 0, 2))
+
+        log_cumsum = gs.hstack([gs.zeros((n_points, 1, 3)),
+                                np.cumsum(log_derivative, -2)])
+        log = log_0 + 1 / n_sampling_points * log_cumsum
+
+        return log
+
+    def geodesic(self, initial_point,
```
ninamiolane: Same: can this function call the parent's function `geodesic` to avoid copy-pasting code?

geomstats/embedded_manifold.py
```diff
@@ -11,7 +13,8 @@ class EmbeddedManifold(Manifold):
     """
 
     def __init__(self, dimension, embedding_manifold):
-        assert isinstance(dimension, int) and dimension > 0
+        assert isinstance(dimension, int) or dimension == math.inf
```
ninamiolane: Nice!

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
+"""
+Unit tests for parameterized manifolds.
+"""
+
+import unittest
+
+import geomstats.backend as gs
+
+from geomstats.hypersphere import Hypersphere
```
ninamiolane: Alphabetic order

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+POINT_A = CURVE_A(SAMPLING_TIMES)
+POINT_B = CURVE_B(SAMPLING_TIMES)
+POINT_C = CURVE_C(SAMPLING_TIMES)
+
+N_POINTS = 5
+TIMES = gs.linspace(0, 1, N_POINTS)
+ATOL = 1e-8
+
+
+class TestDiscretizedCurvesSpaceMethods(unittest.TestCase):
```
ninamiolane: Nice, thanks for the tests!

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+CURVE_A = S2.metric.geodesic(initial_point=INITIAL_POINT,
+                             initial_tangent_vec=INITIAL_TANGENT_VEC_A)
+CURVE_B = S2.metric.geodesic(initial_point=INITIAL_POINT,
+                             initial_tangent_vec=INITIAL_TANGENT_VEC_B)
+CURVE_C = S2.metric.geodesic(initial_point=INITIAL_POINT,
+                             initial_tangent_vec=INITIAL_TANGENT_VEC_C)
+
+N_SAMPLING_POINTS = 10
+SAMPLING_TIMES = gs.linspace(0, 1, N_SAMPLING_POINTS)
+POINT_A = CURVE_A(SAMPLING_TIMES)
```
ninamiolane: Maybe `DISCRETIZED_CURVE_A` and same for the others?

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+ATOL = 1e-8
+
+
+class TestDiscretizedCurvesSpaceMethods(unittest.TestCase):
+    _multiprocess_can_split_ = True
+
+    def setUp(self):
+        gs.random.seed(1234)
+        self.space_curves_R3 = DiscretizedCurvesSpace(embedding_manifold=R3)
+        self.space_curves_S2 = DiscretizedCurvesSpace(embedding_manifold=S2)
```
ninamiolane: Nice!

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+        self.point_a = POINT_A
+        self.point_b = POINT_B
+        self.point_c = POINT_C
+        points_ab = self.space_curves_S2.L2_metric.geodesic(
```
ninamiolane: Maybe defined:
`s2_l2_metric = self.space_curves_s2.l2_metric` since you're using it everywhere?

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+TIMES = gs.linspace(0, 1, N_POINTS)
+ATOL = 1e-8
+
+
+class TestDiscretizedCurvesSpaceMethods(unittest.TestCase):
+    _multiprocess_can_split_ = True
+
+    def setUp(self):
+        gs.random.seed(1234)
+        self.space_curves_R3 = DiscretizedCurvesSpace(embedding_manifold=R3)
```
ninamiolane: No uppercase letters in variable names. Maybe
`space_curves_in_euclidean_3d` and `space_curves_in_sphere_2d`?

tests/test_discretized_curves_space.py
ninamiolane: space_curves_sphere_2d

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+
+N_POINTS = 5
+TIMES = gs.linspace(0, 1, N_POINTS)
+ATOL = 1e-8
+
+
+class TestDiscretizedCurvesSpaceMethods(unittest.TestCase):
+    _multiprocess_can_split_ = True
+
+    def setUp(self):
```
ninamiolane: Actually, this should be `setUpClass(cls)`, otherwise it is run at the beginning and end of each test:
https://stackoverflow.com/questions/23667610/what-is-the-difference-between-setup-and-setupclass-in-python-unittest/23670844

We should correct it elsewhere in geomstats too :)

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+CURVE_C = S2.metric.geodesic(initial_point=INITIAL_POINT,
+                             initial_tangent_vec=INITIAL_TANGENT_VEC_C)
+
+N_SAMPLING_POINTS = 10
+SAMPLING_TIMES = gs.linspace(0, 1, N_SAMPLING_POINTS)
+POINT_A = CURVE_A(SAMPLING_TIMES)
+POINT_B = CURVE_B(SAMPLING_TIMES)
+POINT_C = CURVE_C(SAMPLING_TIMES)
+
+N_POINTS = 5
```
ninamiolane: This is `N_DISCRETIZED_CURVES`?

tests/test_discretized_curves_space.py
```diff
@@ -0,0 +1,250 @@
...
+        result = self.space_curves_R3.SRV_metric.exp(tangent_vec=log,
+                                                     base_point=self.points_ab)
+        expected = self.points_bc
+
+        gs.testing.assert_allclose(result.squeeze(), expected, atol=ATOL)
+
+    def test_srv_metric_geodesic(self):
+        """
+        Test that the geodesic between two curves in a Euclidean space
+        for the srv metric is the L2 geodesic betweeen the curves srvs.
```
ninamiolane: Nice!

AliceLeBrigant:
Thanks for the review ! While I'm doing the corrections, here are some answers to your general comments:

- Yes you're right, I can use ```to_ndarray``` and it makes the code more readable. Thanks !

- I have indeed re-used the code of the ```geodesic``` method of ```RiemannianMetric```, because I couldn't call it directly. This is because of the issue I told you about in a previous message : it's not possible to compute the exponential map of several vector fields at several base_points for the ```HypersphereMetric``` (whereas it is possible for a ```EuclideanMetric```). If we managed to fix this, I could simply call the method of the parent class.

- Do you mean change ```point``` to ```curve```, ```n_sampling_points``` to ```n_points```, and ```n_points``` to ```n_curves```, in all the methods of the ```L2Metric``` and ```SRVMetric``` classes ?

ninamiolane:
- Cool!

- I see, this makes sense.

- Maybe `n_points` to `n_curves` and `point` to `curve`, but keeping the `n_sampling_points`. What do you think? I'm thinking in terms on how easy it would be for us to read the code when we'll come back to it later.

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,424 @@
...
+                signature=(embedding_manifold.dimension, 0, 0))
+        self.embedding_metric = embedding_manifold.metric
+
+    def inner_product(self, tangent_vec_a, tangent_vec_b, base_point):
+        """
+        Inner product between two tangent vectors at a base point.
+        """
+        assert tangent_vec_a.shape == tangent_vec_b.shape
+        assert tangent_vec_a.shape == base_point.shape
+        n_coord = base_point.shape[-1]
```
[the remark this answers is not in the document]
AliceLeBrigant: In the case of a Euclidean space yes, but in the case of the Hypersphere the coordinates used to define a point are usually exterior, i.e. ```n_coord = embedding_manifold.dimension + 1```. What would be right is ```n_coord = embedding_manifold.embedding_manifold.dimension```. 

After using ```to_ndarray```, the simplest is probably to write:
```
n_curves = tangent_vec_a.shape[0]
n_points = tangent_vec_a.shape[1]
n_coords = tangent_vec_a.shape[2]
```

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,424 @@
...
+        return dist
+
+    def exp(self, tangent_vec, base_point):
+        """
+        Riemannian exponential of a tangent vector wrt to a base point.
+        """
+        base_point = base_point.squeeze()
+        mult_tangent_vecs = 0
+        if tangent_vec.ndim > base_point.ndim:
+            mult_tangent_vecs = 1
```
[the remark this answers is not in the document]
AliceLeBrigant: Indeed it becomes unnecessary when using ```to_ndarray```. It was useful before to handle the case when only one base point is given but several tangent vectors, which is not covered I think by the log method (one base points but several points).

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,424 @@
...
+    (see Srivastava et al. 2011).
+    """
+    def __init__(self, embedding_manifold):
+        super(SRVMetric, self).__init__(
+                dimension=embedding_manifold.dimension,
+                signature=(embedding_manifold.dimension, 0, 0))
+        self.embedding_metric = embedding_manifold.metric
+        self.L2_metric = L2Metric(embedding_manifold=embedding_manifold)
+
+    def pointwise_inner_product(self, tangent_vec_a, tangent_vec_b,
```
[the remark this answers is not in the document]
AliceLeBrigant: It does when tangent vec has shape ```[n_sampling_points, n_coords]``` but not when it has shape ```[n_curves, n_sampling_points, n_coords]```, which I need for the vectorization of my methods.

geomstats/discretized_curves_space.py
AliceLeBrigant: This is again linked to vectorization. When ```curve``` has shape ```[n_curves, n_sampling_points, n_coords]```, it has to be reshaped into ```[n_curves * n_sampling_points, n_coords]``` before computing the velocity using
```
velocity = n_sampling_points * self.embedding_metric.log(
                point=curve[1:, :], base_point=curve[:-1, :])
```

At regular intervals of length ```n_sampling_points```, this computes a log that is meaningless, namely the log between the end point of ```curve[k, :, :]``` and the starting point of the next curve ```curve[k + 1, :, :]```. This index selection gets rid of these vectors.

AliceLeBrigant:
Yes, keeping ```n_sampling_points``` makes sense !

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        assert tangent_vec_a.shape == tangent_vec_b.shape
+        assert tangent_vec_a.shape == base_curve.shape
+        tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=3)
+        tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
+
+        n_curves = tangent_vec_a.shape[0]
+        n_sampling_points = tangent_vec_a.shape[1]
+        n_coords = tangent_vec_a.shape[2]
```
ninamiolane: You can write:
```
n_curves, n_sampling_points, n_coords = tangent_vec_a.shape
```
directly.

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def dist(self, curve_a, curve_b):
+        """
+        Geodesic distance between two discretized curves.
+        """
+        assert curve_a.shape == curve_b.shape
+        curve_a = gs.to_ndarray(curve_a, to_ndim=3)
+        curve_b = gs.to_ndarray(curve_b, to_ndim=3)
+
+        n_curves = curve_a.shape[0]
```
ninamiolane: Same as above.

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        return dist
+
+    def exp(self, tangent_vec, base_curve):
+        """
+        Riemannian exponential of a tangent vector wrt to a base curve.
+        """
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
+
+        n_curves = base_curve.shape[0]
```
ninamiolane: Same/

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def log(self, curve, base_curve):
+        """
+        Riemannian logarithm of a curve wrt a base curve.
+        """
+        assert curve.shape == base_curve.shape
+        curve = gs.to_ndarray(curve, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
+
+        n_curves = curve.shape[0]
```
ninamiolane: Same.

geomstats/discretized_curves_space.py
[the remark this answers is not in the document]
ninamiolane: I see!

geomstats/discretized_curves_space.py
ninamiolane: I see, thanks for explaining! Could you add a comment to explain (very rapidly) this for future collaborators? Thanks!

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+                                 'implemented for dicretized curves embedded '
+                                 'in a Euclidean space.')
+        if srv.ndim != starting_point.ndim:
+            starting_point = gs.transpose(
+                    np.tile(starting_point, (1, 1, 1)), (1, 0, 2)
+                    )
+        srv_shape = srv.shape
+        srv = gs.to_ndarray(srv, to_ndim=3)
+        n_curves = srv.shape[0]
+        n_sampling_points = srv.shape[1] + 1
```
ninamiolane: why +1?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+            raise AssertionError('The square root velocity inverse is only '
+                                 'implemented for dicretized curves embedded '
+                                 'in a Euclidean space.')
+        if srv.ndim != starting_point.ndim:
+            starting_point = gs.transpose(
+                    np.tile(starting_point, (1, 1, 1)), (1, 0, 2)
+                    )
+        srv_shape = srv.shape
+        srv = gs.to_ndarray(srv, to_ndim=3)
+        n_curves = srv.shape[0]
```
ninamiolane: maybe use `srv_shape` since you've defined it above? Or get rid of line 268?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        n_sampling_points = srv.shape[1] + 1
+        n_coords = srv.shape[2]
+
+        srv = srv.reshape(n_curves * (n_sampling_points - 1), n_coords)
```
ninamiolane: why -1?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+                base_curve[:, 1:, :] - base_curve[:, :-1, :])
+        base_curve_velocity_norm = self.pointwise_norm(base_curve_velocity,
+                                                       base_curve[:, :-1, :])
+
+        coef_1 = 1 / gs.sqrt(base_curve_velocity_norm)
+        coef_2 = - 1 / (2 * base_curve_velocity_norm ** (5/2)) * \
+            self.pointwise_inner_product(tangent_vec_derivative,
+                                         base_curve_velocity,
+                                         base_curve[:, :-1, :])
+        srv_initial_derivative = gs.transpose(coef_1) * \
```
ninamiolane: This line is a bit long. Any chance you could decompose the computation, giving meaningful variable names to the auxiliary blocks, and then compute srv_initial_derivative? Thank you in advance!

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+            gs.transpose(tangent_vec_derivative, (2, 1, 0)) + \
+            gs.transpose(coef_2) * gs.transpose(base_curve_velocity, (2, 1, 0))
+        srv_initial_derivative = gs.transpose(srv_initial_derivative,
```
ninamiolane: There seems to be several transposes: aren't cancelling them out, and could we write it directly without tranposing?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3)
+        n_sampling_points = base_curve.shape[1]
+
+        base_curve_srv = self.square_root_velocity(base_curve)
+
+        tangent_vec_derivative = n_sampling_points * (
+                tangent_vec[:, 1:, :] - tangent_vec[:, :-1, :])
+        base_curve_velocity = n_sampling_points * (
+                base_curve[:, 1:, :] - base_curve[:, :-1, :])
+        base_curve_velocity_norm = self.pointwise_norm(base_curve_velocity,
```
ninamiolane: Indentation looks a bit weird here?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        Riemannian logarithm of a curve wrt a base curve.
+        """
+        if not isinstance(self.embedding_metric, EuclideanMetric):
+            raise AssertionError('The logarithm map is only implemented '
+                                 'for dicretized curves embedded in a '
+                                 'Euclidean space.')
+        curve = gs.to_ndarray(curve, to_ndim=3)
+        base_curve = gs.to_ndarray(base_curve, to_ndim=3)
+        n_curves = curve.shape[0]
```
ninamiolane: Same:
```
n_curves, n_sampling_points, n_coords = curve.shape
```

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+
+        coef_1 = gs.sqrt(base_curve_velocity_norm)
+        coef_2 = 1 / base_curve_velocity_norm ** (3/2) * \
+            self.pointwise_inner_product(curve_srv - base_curve_srv,
+                                         base_curve_velocity,
+                                         base_curve[:, :-1, :])
+        log_derivative = gs.transpose(coef_1) * \
+            gs.transpose(curve_srv - base_curve_srv, (2, 1, 0)) + \
+            gs.transpose(coef_2) * gs.transpose(base_curve_velocity, (2, 1, 0))
+        log_derivative = gs.transpose(log_derivative, (2, 1, 0))
```
ninamiolane: Same: several transposes in a row, can we do it without tranposing? Maybe using `gs.einsum` function?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+        log_0 = self.embedding_metric.log(point=curve[:, 0, :],
```
ninamiolane: maybe `_initial_sampling_points` instead of `_0` would make the code more readable for later?

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+                                          base_point=base_curve[:, 0, :])
+        log_0 = gs.transpose(np.tile(log_0, (1, 1, 1)), (1, 0, 2))
+
+        log_cumsum = gs.hstack([gs.zeros((n_curves, 1, n_coords)),
+                                np.cumsum(log_derivative, -2)])
```
ninamiolane: Use `gs.cumsum` and define `cumsum` in the numpy backend:
https://github.com/geomstats/geomstats/blob/master/geomstats/backend/numpy.py

geomstats/discretized_curves_space.py
```diff
@@ -0,0 +1,437 @@
...
+            self.pointwise_inner_product(curve_srv - base_curve_srv,
+                                         base_curve_velocity,
+                                         base_curve[:, :-1, :])
+        log_derivative = gs.transpose(coef_1) * \
+            gs.transpose(curve_srv - base_curve_srv, (2, 1, 0)) + \
+            gs.transpose(coef_2) * gs.transpose(base_curve_velocity, (2, 1, 0))
+        log_derivative = gs.transpose(log_derivative, (2, 1, 0))
+        log_0 = self.embedding_metric.log(point=curve[:, 0, :],
+                                          base_point=base_curve[:, 0, :])
+        log_0 = gs.transpose(np.tile(log_0, (1, 1, 1)), (1, 0, 2))
```
ninamiolane: why do we need this?

geomstats/hypersphere.py
```diff
@@ -178,6 +178,13 @@ def __init__(self, dimension):
         self.signature = (dimension, 0, 0)
         self.embedding_metric = EuclideanMetric(dimension + 1)
 
+    def inner_product_matrix(self, base_point=None):
+        """
+        Inner product matrix.
+        """
+        mat = self.embedding_metric.inner_product_matrix(base_point)
+        return mat
+
```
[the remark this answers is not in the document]
ninamiolane: Can we call this method `embedding_inner_product_matrix`?

geomstats/discretized_curves_space.py
AliceLeBrigant: To be consistent with the names of the other methods : ```n_sampling_points``` is the number of sampling points of the curve, and so when we deal with its derivative or srv (computed with finite differences) we 'loose' a point. I suggest to rename it ```n_sampling_points_minus_one``` ?

This made me realize I should be multiplying by ```n_sampling_points - 1``` rather than ```n_sampling_points``` everywhere where I compute velocities (the curves are supposed to be parameterized between 0 and 1, at regular intervals of 1/(n-1) if there are n points).

geomstats/discretized_curves_space.py
AliceLeBrigant: ```srv.shape``` doesn't yield the same result as ```srv_shape``` because of the ```to_ndarray``` in between. I am defining ```srv_shape``` to keep the information of the initial shape of ```srv``` so as to give the output in the same shape (e.g. I want different outputs for input shapes ```[1, n_sampling_points, n_coords]``` and ```[n_sampling_points, n_coords]```.

geomstats/discretized_curves_space.py
AliceLeBrigant: Same as before.

geomstats/discretized_curves_space.py
AliceLeBrigant: Yes you're right, it can be done with the ```einsum``` function ! I think I finally understand how it works.

geomstats/discretized_curves_space.py
AliceLeBrigant: How about ```_initial_starting_points``` to be coherent with the input variable name of ```square_root_velocity_inverse``` ?

geomstats/discretized_curves_space.py
AliceLeBrigant: to go from shape ```[n_curves, n_coords]``` to shape ```[n_curves, 1, n_coords]``` to do the the ```cumsum``` afterwards.

AliceLeBrigant:
Thanks for the review !

If I change the name of the ```inner_product``` method of Hypersphere to ```embedding_inner_product_matrix``` then when I call the ```inner_product``` method of the class ```L2Metric``` when the embedding manifold is a sphere (which itself calls the ``ìnner_product``` method of ```Hypersphere```) I get the error 'The computation of the inner product matrix is not implemented'.

ninamiolane:
Ah, I see! 

Then, let's implement the method `inner_product`, using the inner product of the embedding metric, instead of the `inner_product_matrix`. This way, the `inner_product` is the correct method (by definition of the pullback metric) .

(The inner_product_matrix method was incorrect: the dimensionality of the inner_product_matrix on the sphere was not the same as the dimensionality of inner_product_matrix in the Euclidean space.)

Do you agree?

Thanks!

AliceLeBrigant:
Do you mean replacing the ```inner_product_matrix``` method of ```Hypersphere``` by an ```inner_product``` method which calls ```self.embedding_metric.inner_product``` ?

ninamiolane:
Yes! Would this work?

AliceLeBrigant:
Yes it works !

Issue #155, opened 2018-10-17
geomstats tf backend result != numpy backend result on loss function


farrell236:
I have managed to run the geodesic loss function using geomstats with tensorflow backend after a few fixes, however the result is not the same as the numpy backend.

[L330](https://github.com/geomstats/geomstats/blob/a4e753f110618f718099758f96f38aa02e3a41b4/geomstats/special_euclidean_group.py#L330) in jacobian_translation() in special_euclidean_group.py. ```jacobian.ndim == 3``` gives ```AttributeError: 'Tensor' object has no attribute 'ndim'```. This can be fixed with a backend function e.g. get_tensor_rank() so that it can be attributed to numpy/tensorflow/pytorch.

With this commented out (as a temporary solution), the loss function would only run for N=1. N<1 will cause the einsum problem again (#153) in multiple areas.

The scripts I'm running are as follows:

Numpy:
```
import os
os.environ['GEOMSTATS_BACKEND'] = 'numpy'  # NOQA

import numpy as np
import geomstats.lie_group as lie_group
from geomstats.special_euclidean_group import SpecialEuclideanGroup

# Reproducable results
np.random.seed(0)

n_vecs = 1

SE3_GROUP = SpecialEuclideanGroup(3)
metric=SE3_GROUP.left_canonical_metric

y_pred = np.random.rand(n_vecs, 6)
y_true = np.random.rand(n_vecs, 6)

loss = lie_group.loss(y_pred, y_true, SE3_GROUP, metric)

print('y_pred:',y_pred)
print('y_true:',y_true)
print('loss:', loss)

'''
Output:
Using numpy backend
y_pred: [[0.5488135  0.71518937 0.60276338 0.54488318 0.4236548  0.64589411]]
y_true: [[0.43758721 0.891773   0.96366276 0.38344152 0.79172504 0.52889492]]
loss: [[0.33912556]]
'''
```

Tensorflow:
```
import os
os.environ['GEOMSTATS_BACKEND'] = 'tensorflow'  # NOQA

import tensorflow as tf

import numpy as np
import geomstats.lie_group as lie_group
from geomstats.special_euclidean_group import SpecialEuclideanGroup

# Reproducable results
np.random.seed(0)

n_vecs = 1

SE3_GROUP = SpecialEuclideanGroup(3)
metric=SE3_GROUP.left_canonical_metric

y_pred = tf.placeholder(tf.float32, shape=[n_vecs, 6])
[...]

PR #159, opened 2018-11-06
tf backend fixes 3

farrell236:
- geomstats/invariant_metric.py:
	inner_product_at_identity():
		- ensures that tangent_vec_a and tangent_vec_b have the same cardinality
		- stack self.inner_product_mat_at_identity N times w.r.t. tangent_vec
	inner_product_matrix():
		- stack self.inner_product_mat_at_identity N times w.r.t. base_point

- geomstats/lie_group.py:
	group_log():
		- ensure point and base_point cardinality match for tf.einsum()

- geomstats/special_euclidean_group.py:
	jacobian_translation():
		- fix assert gs.ndim(jacobian) == 3

- geomstats/special_orthogonal_group.py:
	regularize_tangent_vec():
		- ensure jacobian has the same cardinality as tangent_vec
		- N.B. TODO: check, this fix assumes there will always be one base_point to many tangent_vec
	compose():
		- ensures that point_1 and point_2 have the same cardinality
		- stack point_1 and point_2 N times for tf.einsum() fix

- geomstats/backend/tensorflow_linalg.py
	svd(x)
		- np.linalg.svd() returns U, S, V
		- tf.svd() returns S, U, V_transposed

tests/test_special_orthogonal_group.py
	test_group_log_vectorization()
		- gs.allclose() is not required in tuple formation

### DIFF
--- a/examples/loss_and_gradient_se3.py
+++ b/examples/loss_and_gradient_se3.py
@@ -44,9 +44,9 @@ def grad(y_pred, y_true,
 
     :return: tangent vector at point y_pred.
     """
-    if y_pred.ndim == 1:
+    if gs.ndim(y_pred) == 1:
         y_pred = gs.expand_dims(y_pred, axis=0)
-    if y_true.ndim == 1:
+    if gs.ndim(y_true) == 1:
         y_true = gs.expand_dims(y_true, axis=0)
 
     if representation == 'vector':
@@ -60,10 +60,6 @@ def grad(y_pred, y_true,
         y_true_pose = gs.hstack([y_true_rot_vec, y_true[:, 4:]])
         grad = lie_group.grad(y_pred_pose, y_true_pose, SE3, metric)
 
-        differential = gs.zeros((1, 6, 7))
-
-        upper_left_block = gs.zeros((1, 3, 4))
-        lower_right_block = gs.zeros((1, 3, 3))
         quat_scalar = y_pred[:, :1]
         quat_vec = y_pred[:, 1:4]
 
@@ -73,19 +69,24 @@ def grad(y_pred, y_true,
 
         quat_arctan2 = gs.arctan2(quat_vec_norm, quat_scalar)
         differential_scalar = - 2 * quat_vec / (quat_sq_norm)
-        differential_vec = (2 * (quat_scalar / quat_sq_norm
-                                 - 2 * quat_arctan2 / quat_vec_norm)
-                            * gs.outer(quat_vec, quat_vec) / quat_vec_norm ** 2
+        differential_vec = (2 * (quat_scalar / quat_sq_norm - 2 * quat_arctan2 / quat_vec_norm)
+                            * gs.einsum('ni,nj->nij',quat_vec, quat_vec) / quat_vec_norm * quat_vec_norm
                             + 2 * quat_arctan2 / quat_vec_norm * gs.eye(3))
 
-        upper_left_block[0, :, :1] = differential_scalar.transpose()
-        upper_left_block[0, :, 1:] = differential_vec
-        lower_right_block[0, :, :] = gs.eye(3)
+        differential_scalar_t = gs.transpose(differential_scalar,axes=(1,0))
 
-        differential[0, :3, :4] = upper_left_block
-        differential[0, 3:, 4:] = lower_right_block
+        upper_left_block = gs.hstack((differential_scalar_t, differential_vec[0]))
+        upper_right_block = gs.zeros((3, 3))
+        lower_right_block = gs.eye(3)
+        lower_left_block = gs.zeros((3, 4))
 
-        grad = gs.matmul(grad, differential)
+        top = gs.hstack((upper_left_block, upper_right_block))
+        bottom = gs.hstack((lower_left_block, lower_right_block))
+
+        differential = gs.vstack((top, bottom))
+        differential = gs.expand_dims(differential, axis=0)
+
+        grad = gs.einsum('ni,nij->ni', grad, differential)
 
     grad = gs.squeeze(grad, axis=0)
     return grad
@@ -110,7 +111,7 @@ def main():
     scalar = gs.array(cos)
     vec = sin * u
     translation = gs.array([5., 6., 7.])
-    y_pred_quaternion = gs.hstack([scalar, vec, translation])
+    y_pred_quaternion = gs.concatenate([[scalar], vec, translation], axis=0)
 
     angle = gs.pi / 7
     cos = gs.cos(angle / 2)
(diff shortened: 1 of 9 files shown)


geomstats/invariant_metric.py
```diff
@@ -57,9 +57,26 @@ def inner_product_at_identity(self, tangent_vec_a, tangent_vec_b):
             tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=2)
             tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=2)
 
+            n_tangent_vec_a = tangent_vec_a.shape[0]
+            n_tangent_vec_b = tangent_vec_b.shape[0]
+
+            assert (tangent_vec_a.shape == tangent_vec_a.shape
```
ninamiolane: `tangent_vec_b.shape` on the right?

geomstats/invariant_metric.py
```diff
@@ -57,9 +57,26 @@ def inner_product_at_identity(self, tangent_vec_a, tangent_vec_b):
...
+                    or n_tangent_vec_a == 1
+                    or n_tangent_vec_b == 1)
```
ninamiolane: Does this assert work in TF without the eager execution mode?

geomstats/lie_group.py
```diff
@@ -146,10 +146,26 @@ def group_log(self, point, base_point=None, point_type=None):
             point_type = self.default_point_type
 
         identity = self.get_identity(point_type=point_type)
-        identity = self.regularize(identity, point_type=point_type)
+        #identity = self.regularize(identity, point_type=point_type)
```
ninamiolane: left-over comment?

geomstats/lie_group.py
```diff
@@ -146,10 +146,26 @@ def group_log(self, point, base_point=None, point_type=None):
...
         if base_point is None:
             base_point = identity
+
+        point = self.regularize(point, point_type=point_type)
         base_point = self.regularize(base_point, point_type=point_type)
+
+        n_points = point.shape[0]
+        n_base_points = base_point.shape[0]
+
+        assert (point.shape == base_point.shape
```
ninamiolane: Same question: does this work with TF?

geomstats/special_euclidean_group.py
```diff
@@ -323,7 +323,7 @@ def jacobian_translation(
             jacobian = gs.concatenate(
                 [jacobian_block_line_1, jacobian_block_line_2], axis=1)
 
-            assert jacobian.ndim == 3
+            assert gs.ndim(jacobian) == 3
```
ninamiolane: thanks!

geomstats/backend/tensorflow_linalg.py
```diff
@@ -16,8 +16,8 @@ def eig(x):
 
 
 def svd(x):
-    s, u, v = tf.linalg.svd(x)
-    return u, s, v
+    s, u, v_t = tf.svd(x, full_matrices=True)
+    return u, s, tf.transpose(v_t, perm=(0, 2, 1))
```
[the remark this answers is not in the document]
farrell236: glad to find this was the issue causing the mismatch :D

geomstats/invariant_metric.py
farrell236: Yes you're right! good catch

geomstats/lie_group.py
farrell236: I thought it was weird to regularise an identity point... commented it out for the time being, but yes it should be removed.

geomstats/invariant_metric.py
farrell236: Yes, I ran the tf script [here](https://github.com/geomstats/geomstats/issues/155#issue-371004876) and set a breakpoint at the location. It passes this location and provide correct assert values:

```
/vol/medic01/users/bh1511/_venv/geomstats-py3/bin/python3.6 /Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py --cmd-line --multiproc --qt-support=auto --client 127.0.0.1 --port 50116 --file /vol/medic01/users/bh1511/_build/geomstats-farrell/tests/bh_test_tf.py
pydev debugger: process 11617 is connecting
Connected to pydev debugger (build 182.4505.26)
/vol/medic01/users/bh1511/_venv/geomstats-py3/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: compiletime version 3.5 of module 'tensorflow.python.framework.fast_tensor_util' does not match runtime version 3.6
  return f(*args, **kwds)
Using tensorflow backend
import sys; print('Python %s on %s' % (sys.version, sys.platform))
Python 3.6.3 |Intel Corporation| (default, May  3 2018, 23:25:54) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.0.1 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 7.0.1
Python 3.6.3 |Intel Corporation| (default, May  3 2018, 23:25:54) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
In[2]: tangent_vec_a.shape == tangent_vec_b.shape
Out[2]: True
In[3]: n_tangent_vec_a
Out[3]: Dimension(4)
In[4]: n_tangent_vec_b
Out[4]: Dimension(4)

```

geomstats/lie_group.py
farrell236: yes, detailed answered above

farrell236:
I'm currently running test_special_orthogonal_group.py and test_special_euclidean_group.py with the tf backend and eager execution enabled.

```
import os
os.environ['GEOMSTATS_BACKEND'] = 'tensorflow'  # NOQA

import tensorflow as tf
tf.enable_eager_execution()
```

It's failing a lot of them, but I think the test should be agnostic to the backend used and the result should be the same regardless of which. The fix is a long process, but its getting there :)

examples/loss_and_gradient_se3.py
```diff
@@ -110,7 +111,7 @@ def main():
     scalar = gs.array(cos)
     vec = sin * u
     translation = gs.array([5., 6., 7.])
-    y_pred_quaternion = gs.hstack([scalar, vec, translation])
+    y_pred_quaternion = gs.concatenate([[scalar], vec, translation], axis=0)
```
ninamiolane: why the `[]` around scalar?

farrell236: Scalar tensors does not have a shape (returns ```shape=()``` or ```TensorShape([])```), so it's not possible to concatenate with tensor vectors. The square bracket assigns it with ```shape=(1,)``` or ```TensorShape([Dimension(1)])```

farrell236:
I think the syntax should be very similar despite the backend, e.g. like Keras. The same code written with Keras works with Tensorflow, Theano and CNTK, which was why I thought test files should be the same irrespective of which geomstats backend used. This unfortunately will be difficult to implement. 

> How about sending a subset of running TF tests (fully running, without eager mode) and then move one unit test at a time?

I think this is right call, shall we prioritise on the most frequently used functions/tests and get them working first? Then move onto subsequent ones.

PR #163, opened 2018-11-17
Nina warnings

ninamiolane:
This PR:
- adds the inner_product function missing in hyperbolic space
- addresses most of the warnings during unit testing


geomstats/hyperbolic_space.py
```diff
@@ -154,6 +153,14 @@ def __init__(self, dimension):
         self.signature = (dimension, 0, 0)
         self.embedding_metric = MinkowskiMetric(dimension + 1)
 
+    def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
+        """
+        Inner product.
```
johmathe: One line should be good enough

PR #188, opened 2018-12-12
Nina tf post refactor

ninamiolane:
TF unit tests for euclidean space
Delete tf.session for the assertAllClose


tests/test_euclidean_space.py
```diff
@@ -21,239 +21,219 @@ def setUp(self):
         self.space = EuclideanSpace(self.dimension)
         self.metric = self.space.metric
 
-        self.n_samples = 10
+        self.n_samples = 3
+
+        self.one_point_a = gs.array([0., 1.])
```
johmathe: actually not a huge fan of factorizing this kind of stuff in the setup. It's roughly the same amount of letters to write down gs.array([0., 1.]) and it's way more clear to keep the whole thing.

PR #189, opened 2018-12-12
Nina tf2

ninamiolane:
Enable all unit tests for minkowski with TF


tests/test_minkowski_space.py
```diff
@@ -23,39 +23,35 @@ def setUp(self):
...
 
-    @geomstats.tests.np_only
     def test_random_uniform_and_belongs(self):
         point = self.space.random_uniform()
-        with self.session():
-            self.assertTrue(gs.eval(self.space.belongs(point)))
+        result = self.space.belongs(point)
+        expected = gs.array([[True]])
+
+        self.assertAllClose(result, expected)
```
johmathe: no session needed here?

PR #194, opened 2018-12-26
Unit tests for stiefel in tensorflow + General unit test cleaning

ninamiolane:


geomstats/backend/tensorflow_linalg.py
```diff
@@ -40,5 +56,16 @@ def eigvalsh(x):
...
+def qr(*args, mode='reduced'):
+    def qr_aux(x, mode):
+        if mode == 'complete':
+            aux = tf.linalg.qr(x, full_matrices=True)
+        else:
+            aux = tf.linalg.qr(x)
+
+        return (aux.q, aux.r)
+
+    return tf.map_fn(
```
johmathe: can you do this in 2 times?

johmathe: assign and return

PR #195, opened 2018-12-26
Unit tests in TF for stiefel

ninamiolane:


geomstats/backend/numpy.py
```diff
@@ -269,8 +269,18 @@ def clip(*args, **kwargs):
...
+def diag(x):
+    x = to_ndarray(x, to_ndim=2)
+    _, n = shape(x)
+    aux = np.vectorize(
+        np.diagflat,
+        signature='(m,n)->(k,k)')(x)
+    k, k = shape(aux)
+    m = int(k / n)
+    result = [
+        aux[i*n:(i+1)*n, i*n:(i+1)*n]
```
johmathe: rule for list comprehension is that if it's too hard to read make it back in a for loop. Here I think it would be better to have this in a for loop.

geomstats/stiefel.py
```diff
@@ -165,7 +161,7 @@ def exp(self, tangent_vec, base_point):
 
         return exp
 
-    def log(self, point, base_point, max_iter=100, tol=1e-6):
+    def log(self, point, base_point, max_iter=30):
```
johmathe: did you change this because it was taking too long?

ninamiolane: Yes, because I deleted the early stopping condition because of the lack of easy if statement in tf.

PR #203, opened 2018-12-28
Unit tests in TF for angles close to pi in SO(3) and SE(3)

ninamiolane:


tests/test_special_orthogonal_group.py
```diff
@@ -3168,40 +3132,20 @@ def test_exp_then_log_with_angles_close_to_pi(self):
...
-                             metric_type,
-                             angle_type,
-                             tangent_vec, reg_tangent_vec,
-                             angle_type_base,
-                             base_point, group.regularize(base_point),
-                             result, regularized_result,
-                             expected, regularized_expected))
+                    with self.session():
+                        self.assertTrue(
+                            gs.eval(gs.allclose(
```
johmathe: so we need the evals here?

ninamiolane: yes!

PR #213, opened 2019-06-14
Pytorch backend

ninamiolane:


geomstats/hypersphere.py
```diff
@@ -236,14 +236,6 @@ def log(self, point, base_point):
         mask_0_float = gs.cast(mask_0, gs.float32)
         mask_else_float = gs.cast(mask_else, gs.float32)
 
-        angle_0 = gs.boolean_mask(angle, mask_0)
```
johmathe: why remove this?

tests/test_hypersphere.py
```diff
@@ -124,6 +124,8 @@ def test_log_and_exp_general_case(self):
         expected = point
         expected = helper.to_vector(expected)
 
+        print(result)
```
johmathe: leftover

tests/test_hypersphere.py
```diff
@@ -315,6 +317,8 @@ def test_norm_and_dist(self):
         expected = self.metric.dist(point_a, point_b)
         expected = helper.to_scalar(expected)
 
+        #print('normresult', result)
```
johmathe: leftover

geomstats/hypersphere.py
ninamiolane: it was not used. left over from a previous version probably

PR #216, opened 2019-07-01
Pytorch backend for GLn and SPD matrices

ninamiolane:


geomstats/backend/pytorch.py
```diff
@@ -10,6 +10,20 @@
 int32 = 'torch.LongTensor'
 int8 = 'torch.ByteTensor'
 
+CUDA = torch.cuda.is_available()
+DEVICE = torch.device("cuda" if CUDA else "cpu")
```
johmathe: single quotes

geomstats/backend/pytorch.py
```diff
@@ -317,6 +331,7 @@ def where(*args, **kwargs):
 
 def tile(x, y):
     # TODO(johmathe): Native tile implementation
+    y = tuple([int(one_y) for one_y in y])
```
johmathe: - why need the tuple
- why need of the cast to an int?
- can't you do the cast to an int with something more backend friendly?


```
y.int() 
```

ninamiolane: no need of the tuple -> corrected

need to cast to an int bcs gs.maximum is sometimes used as a index for tile and outputs a torch tensor (which is what we want it to do)
cannot do .int() bcs integers are also passed as indices

geomstats/backend/pytorch.py
```diff
@@ -354,13 +373,6 @@ def diagonal(*args, **kwargs):
 
 def exp(*args, **kwargs):
     return torch.exp(*args, **kwargs)
```
johmathe: why do we have 2 returns line by line?

PR #220, opened 2019-08-25
Debug pytorch tests hypersphere

AliceLeBrigant:
Debug of pytorch unit tests of hypersphere. The general precision of the results in the unit tests is usually a little worse in pytorch than in numpy, and so I have increases the tolerances.

Still the last one ```test_optimat_quantization.py``` is not working well, because the ```gap``` variable in the ```optimal_quantization``` method of ```riemannian_metric.py```, computed as the distance between ```center_to_update``` and ```new_center```, is sometimes zero with pytorch whereas it is non zero (but close) with numpy. I have traced this to the ```inner_product``` method of the embedding Euclidean metric, which seems to be less precise in pytorch than in numpy, but I don't understand why.

### DIFF
--- a/geomstats/backend/pytorch.py
+++ b/geomstats/backend/pytorch.py
@@ -11,6 +11,33 @@ int32 = 'torch.LongTensor'
 int8 = 'torch.ByteTensor'
 
 
+def indexing(x):
+    return x.type(torch.LongTensor)
+
+
+def float_to_double(x):
+    return x.double()
+
+
+def byte_to_float(x):
+    return x.type(torch.FloatTensor)
+
+
+def while_loop(cond, body, loop_vars, maximum_iterations):
+    iteration = 0
+    while cond(*loop_vars):
+        loop_vars = body(*loop_vars)
+        iteration += 1
+        if iteration >= maximum_iterations:
+            break
+    return loop_vars
+
+
+def logical_or(x, y):
+    bool_result = x or y
+    return bool_result
+
+
 def cond(pred, true_fn, false_fn):
     if pred:
         return true_fn()
@@ -115,7 +142,7 @@ def empty_like(*args, **kwargs):
 def all(x, axis=None):
     if axis is None:
         return x.byte().all()
-    return torch.from_numpy(np.all(x, axis=axis).astype(int))
+    return torch.from_numpy(np.all(np.array(x), axis=axis).astype(int))
 
 
 def allclose(a, b, **kwargs):
@@ -296,11 +323,11 @@ def equal(a, b, **kwargs):
         a = cast(a, torch.uint8).float()
     if b.dtype == torch.ByteTensor:
         b = cast(b, torch.uint8).float()
-    return torch.equal(a, b, **kwargs)
+    return torch.eq(a, b, **kwargs)
 
 
 def floor(*args, **kwargs):
-    return torch.floor(*args, **kwargs)
+    return np.floor(*args, **kwargs)
 
 
 def cross(x, y):
@@ -352,8 +379,8 @@ def diagonal(*args, **kwargs):
     return torch.diagonal(*args, **kwargs)
 
 
-def exp(*args, **kwargs):
-    return torch.exp(*args, **kwargs)
+def exp(input):
+    return torch.exp(torch.tensor(input))
 
 
 def log(*args, **kwargs):
@@ -401,3 +428,7 @@ def mean(x, axis=None):
         return torch.mean(x)
     else:
         return np.mean(x, axis)
+
+
+def argmin(*args, **kwargs):
+    return torch.argmin(*args, **kwargs)
--- a/geomstats/backend/pytorch.py
+++ b/geomstats/backend/pytorch.py
@@ -11,17 +11,6 @@ int32 = 'torch.LongTensor'
 int8 = 'torch.ByteTensor'
 
 
-def indexing(x):
-    return x.type(torch.LongTensor)
-
-
-def float_to_double(x):
-    return x.double()
-
-
-def byte_to_float(x):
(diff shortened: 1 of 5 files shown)


tests/test_hypersphere.py
```diff
@@ -302,7 +301,7 @@ def test_squared_dist_vectorization(self):
         result = self.metric.squared_dist(n_points_a, n_points_b)
         self.assertAllClose(gs.shape(result), (n_samples, 1))
 
-    @geomstats.tests.np_and_tf_only
+    #@geomstats.tests.np_and_tf_only
```
ninamiolane: Leftover comment

tests/test_hypersphere.py
```diff
@@ -418,35 +417,32 @@ def test_inner_product(self):
...
         self.assertAllClose(expected, result)
 
-    @geomstats.tests.np_and_tf_only
     def test_variance(self):
         point = gs.array([0., 0., 0., 0., 1.])
-        points = gs.array([
-            point,
-            point])
+        points = gs.zeros((2, point.shape[0]))
+        points[0, :] = point
```
ninamiolane: These assignments make the tests fail in TF. Is it possible to rewrite without assignment in ``points"?

tests/test_hypersphere.py
```diff
@@ -418,35 +417,32 @@ def test_inner_product(self):
...
         self.assertAllClose(expected, result)
 
-    @geomstats.tests.np_and_tf_only
     def test_mean(self):
         point = gs.array([0., 0., 0., 0., 1.])
-        points = gs.array([
-            point,
-            point])
+        points = gs.zeros((2, point.shape[0]))
+        points[0, :] = point
```
ninamiolane: Same.

tests/test_hypersphere.py
```diff
@@ -418,35 +417,32 @@ def test_inner_product(self):
...
 
-    @geomstats.tests.np_and_tf_only
     def test_mean_and_belongs(self):
         point_a = gs.array([1., 0., 0., 0., 0.])
         point_b = gs.array([0., 1., 0., 0., 0.])
-        points = gs.array([
-            point_a,
-            point_b])
+        points = gs.zeros((2, point_a.shape[0]))
+        points[0, :] = point_a
```
ninamiolane: Same.

geomstats/backend/pytorch.py
```diff
@@ -296,11 +323,11 @@ def equal(a, b, **kwargs):
         a = cast(a, torch.uint8).float()
     if b.dtype == torch.ByteTensor:
         b = cast(b, torch.uint8).float()
-    return torch.equal(a, b, **kwargs)
+    return torch.eq(a, b, **kwargs)
 
 
 def floor(*args, **kwargs):
-    return torch.floor(*args, **kwargs)
+    return np.floor(*args, **kwargs)
```
ninamiolane: Why this np.?

geomstats/riemannian_metric.py
```diff
@@ -393,7 +393,7 @@ def optimal_quantization(self, points, n_centers=N_CENTERS,
 
         random_indices = gs.random.randint(low=0, high=n_points,
                                            size=(n_centers,))
-        centers = points[gs.ix_(random_indices, gs.arange(dimension))]
+        centers = points[gs.indexing(random_indices), :]
```
ninamiolane: Could we use the function's name gs.ix_ and match numpy's API?

geomstats/backend/pytorch.py
```diff
@@ -11,6 +11,33 @@
 int8 = 'torch.ByteTensor'
 
 
+def indexing(x):
+    return x.type(torch.LongTensor)
+
+
+def float_to_double(x):
```
ninamiolane: Can we use the function "gs.cast" to cast into a specific type? The types can also be referred to as gs.float32, gs.float64, etc. (see backends)

For example: ```gs.cast(x, gs.float64)```

geomstats/backend/pytorch.py
```diff
@@ -352,8 +379,8 @@ def diagonal(*args, **kwargs):
     return torch.diagonal(*args, **kwargs)
 
 
-def exp(*args, **kwargs):
-    return torch.exp(*args, **kwargs)
+def exp(input):
+    return torch.exp(torch.tensor(input))
```
ninamiolane: Let's put the conversion to tensor outside of this function (maybe in the core code). As we do not perform such conversions for the other functions for now, it would keep the code consistent.

tests/test_hypersphere.py
AliceLeBrigant: Removed.

tests/test_hypersphere.py
AliceLeBrigant: Works now only in numpy and pytorch.

tests/test_hypersphere.py
AliceLeBrigant: Same.

tests/test_hypersphere.py
AliceLeBrigant: Same.

geomstats/backend/pytorch.py
AliceLeBrigant: Removed.

geomstats/riemannian_metric.py
AliceLeBrigant: Replaced by cast.

geomstats/backend/pytorch.py
AliceLeBrigant: Done.

geomstats/backend/pytorch.py
AliceLeBrigant: Done.

PR #235, opened 2019-09-02
Modif powers

ythanwerdas:


geomstats/geometry/spd_matrices_space.py
```diff
@@ -119,7 +119,7 @@ def random_tangent_vec_uniform(self, n_samples=1, base_point=None):
 
 class SPDMetric(RiemannianMetric):
 
-    def __init__(self, n):
+    def __init__(self, n, power_p=1, power_q=-1):
```
ninamiolane: Are you sure power_p is the best variable name?

PR #259, opened 2020-01-13
Update explained variance plot

cshewmake2:
Change visualization of fraction of variance explained by the tangent PCA vectors to a barchart instead of a standard plot.


ninamiolane:
Good idea, thanks!

PR #260, opened 2020-01-13
Set GEOMSTATS_BACKEND if not given

nkoep:
Define the 'GEOMSTATS_BACKEND' environment variable if it's not
explicitly given when geomstats is imported.

This change also
- makes sure we error out if we're given an invalid backend name
- removes the 'common' module of the 'backend' package' as it only
  re-exports 'numpy.pi'

Signed-off-by: Niklas Koep <niklas.koep@gmail.com>

### DIFF
--- a/geomstats/backend/common.py
+++ b/geomstats/backend/common.py
@@ -1,3 +1,0 @@
-import numpy as np
-
-pi = np.pi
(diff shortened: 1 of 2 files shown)


geomstats/backend/common.py
```diff
@@ -1,3 +0,0 @@
-import numpy as np
-
-pi = np.pi
```
ninamiolane: Don't we need this in common to be able to call `gs.pi`?

ninamiolane:
Great, thanks!

PR #288, opened 2020-01-15
Yanntwd spd power affine

ythanwerdas:
Power affine inner product


geomstats/geometry/spd_matrices_space.py
```diff
@@ -246,19 +246,23 @@ def __init__(self, n, power_affine=1):
...
         tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=3)
         n_tangent_vecs_a, _, _ = tangent_vec_a.shape
         tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=3)
         n_tangent_vecs_b, _, _ = tangent_vec_b.shape
 
         base_point = gs.to_ndarray(base_point, to_ndim=3)
-        n_base_points, _, _ = base_point.shape
+        n_base_points, dim, _ = base_point.shape
+
+        spd_space = SPDMatricesSpace(dim)
```
ninamiolane: You don't need this, you can call `self`.

geomstats/geometry/spd_matrices_space.py
```diff
@@ -285,10 +289,28 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point):
...
-        inner_product = gs.trace(gs.matmul(aux_a, aux_b), axis1=1, axis2=2)
-        inner_product = gs.to_ndarray(inner_product, to_ndim=2, axis=1)
+        if power_affine == 1:
+            aux_a = gs.matmul(inv_base_point, tangent_vec_a)
+            aux_b = gs.matmul(inv_base_point, tangent_vec_b)
+            inner_product = gs.trace(gs.matmul(aux_a, aux_b), axis1=1, axis2=2)
+            inner_product = gs.to_ndarray(inner_product, to_ndim=2, axis=1)
+        else:
+            modified_tangent_vec_a =\
+                spd_space.differential_power(power_affine, tangent_vec_a,
```
ninamiolane: `self.differential_power`

tests/test_spd_matrices_space.py
```diff
@@ -125,6 +124,20 @@ def test_procrustes_inner_product(self):
...
 
+    @geomstats.tests.np_and_tf_only
+    def test_power_affine_inner_product(self):
+        base_point = gs.array([[1., 0., 0.],
+                               [0., 2.5, 1.5],
+                               [0., 1.5, 2.5]])
+        tangent_vec = gs.array([[2., 1., 1.],
+                               [1., .5, .5],
+                               [1., .5, .5]])
+        metric = SPDMetric(3, power_affine=.5)
```
ninamiolane: Let's put this line in the setup, so that we can then call self.metric in the tests.

ythanwerdas: Line 135? But we use different metrics in different tests... I don't get your point.

geomstats/geometry/spd_matrices_space.py
ninamiolane: Aaah right! Sorry, I read too fast. The question then is: should we have `space` as an attribute of each SPDMetric?

geomstats/geometry/spd_matrices_space.py
ninamiolane: same, sorry

tests/test_spd_matrices_space.py
ninamiolane: then self.affine_metric?

PR #296, opened 2020-01-15
Restrict backend tests

nkoep:
This change makes sure that we don't switch the backend unnecessarily when testing a specific backend. This effectively suppresses curious messages like "Using numpy backend" when running tests for tensorflow. Similiarly, running the test suite on the numpy backend now no longer produces tensorflow deprecation warnings as we don't import tensorflow in the first place.

### DIFF
--- a/tests/test_backend_numpy.py
+++ b/tests/test_backend_numpy.py
@@ -13,8 +13,6 @@ from geomstats.geometry.special_orthogonal_group import SpecialOrthogonalGroup
 
 @geomstats.tests.np_only
 class TestBackendNumpy(geomstats.tests.TestCase):
-    _multiprocess_can_split_ = True
-
     def setUp(self):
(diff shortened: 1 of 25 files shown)


tests/test_backend_numpy.py
```diff
@@ -2,29 +2,19 @@
 Unit tests for numpy backend.
 """
 
-import importlib
 import os
 import unittest
 import warnings
 
+import geomstats.tests
```
ninamiolane: import order b < t

tests/test_backend_tensorflow.py
```diff
@@ -2,29 +2,18 @@
 Unit tests for tensorflow backend.
 """
 
-import importlib
 import os
-import tensorflow as tf
 
+import geomstats.tests
```
ninamiolane: import order t > b

PR #298, opened 2020-01-15
exponential map on the grassmannian

opeltre:
add GrassmannianCanonicalMetric.exp and its tests


geomstats/geometry/grassmannian.py
```diff
@@ -4,9 +4,11 @@
 where p <= n
 """
 
+import geomstats.backend as gs
+
 from geomstats.geometry.embedded_manifold import EmbeddedManifold
 from geomstats.geometry.euclidean_space import EuclideanMetric
-from geomstats.geometry.matrices_space import MatricesSpace
+from geomstats.geometry.matrices_space import MatricesSpace as Mat
```
ninamiolane: No renaming of the classes, otherwise it's very difficult for other developers to understand what is referred to.

geomstats/geometry/grassmannian.py
```diff
@@ -55,3 +59,22 @@ def __init__(self, n, p):
                 dimension=dimension,
                 signature=(dimension, 0, 0))
         self.embedding_metric = EuclideanMetric(n * p)
+
+    def exp(self, v, p):
```
ninamiolane: v -> tangent_vector
p -> base_point

As a rule of thumb, we avoid single letter as variable names, in order to be able to use automatic find and replace

tests/test_grassmannian.py
```diff
@@ -6,12 +6,48 @@
 import geomstats.tests
 
 from geomstats.geometry.grassmannian import Grassmannian
+from geomstats.geometry.grassmannian import GrassmannianCanonicalMetric
+
+Pxy = gs.diag([1., 1., 0.])[0]
```
ninamiolane: no uppercase letters in variable names, it is a convention chosen so that developers only have to remember variable names and not which letter is uppercase or not

tests/test_grassmannian.py
```diff
@@ -6,12 +6,48 @@
...
+    [1., 0., 0.],
+    [0., 0., 0.]])
+pi_2 = gs.pi/2
 
 
 class TestGrassmannianMethods(geomstats.tests.TestCase):
     def setUp(self):
         gs.random.seed(1234)
 
+<<<<<<< HEAD
```
ninamiolane: Left-over from a merge?

tests/test_grassmannian.py
```diff
@@ -6,12 +6,48 @@
...
+    def test_exp(self): 
+        result = self.metric.exp(pi_2 * Ry, [Pxy, Pyz])
+        expected = gs.array([Pyz, Pxy])
+        self.assertAllClose(result, expected)
+
+        result = self.metric.exp(
+            pi_2 * gs.array([Ry, Rz]), 
+            gs.array([Pxy, Pyz]))
+        expected = gs.array([Pyz, Pxz])
+        self.assertAllClose(result, expected)
```
ninamiolane: Nice!

tests/test_grassmannian.py
```diff
@@ -6,12 +6,48 @@
...
+
+if __name__ == '__main__':
```
ninamiolane: These three lines should be removed as per a recent PR from Niklas

ninamiolane:
Some small flake8 problems on the imports, all good otherwise!

PR #300, opened 2020-01-15
Added a gradient descent method for the mean

xpennec:
Several algorithms for computing the mean can be implemented. It would be nice to have a method to select which one we want to use.


tests/test_hypersphere.py
```diff
@@ -436,6 +436,24 @@ def test_mean(self):
...
+        expected = gs.zeros(n_tests)
+
+        for i in range(n_tests):
+            # take 2 random points, compute their mean, and verify that log of each at the mean is opposite
+            points = self.space.random_uniform(n_samples=2)
+            mean = self.metric.adaptive_gradientdescent_mean(points)
+            logs = self.metric.log(point=points, base_point=mean)
+            result[i] = gs.linalg.norm( logs[1, :] + logs[0, :] )
+            print(result[i])
+        # print(max(result))
```
ninamiolane: Left-over comment

PR #307, opened 2020-01-16
parallel transport and pole ladder

nguigs:
- reformat in connection file
- exact parallel transport on sphere
- approximation with pole ladder


geomstats/geometry/hypersphere.py
```diff
@@ -408,3 +408,13 @@ def dist(self, point_a, point_b):
         dist = gs.arccos(cos_angle)
 
         return dist
+
+    def parallel_transport(self, tan_a, tan_b, point):
```
ninamiolane: Add docstring

use naming API: tangent_vector and base_point

tests/test_connection.py
```diff
@@ -50,3 +51,22 @@ def test_christoffel_symbols(self):
         expected = gs.zeros((1,) + (self.dimension, ) * 3)
 
         gs.testing.assert_allclose(result, expected)
+
+    def test_parallel_transport(self):
```
ninamiolane: Really cool!

PR #313, opened 2020-01-16
matrix transpose

opeltre:
add vectorised matrix transpose. 

N.B:  this is only a first push to expose a vectorized transpose map with **correct shape signature**. 
Many other methods resort to matrix transposition usually via an inline implementation and resorting to `gs.to_ndim` for vectorisation. They might have inconsistent vectorization behaviour, next step is to check this, point to `Mat.transpose` while checking `spd_matrices` stays ok. 

For instance: 
+ `is_symmetric`
+ `make_symmetric`
+ `inverse` on orthogonal matrices!
+ ... 

Look at the implementation to see how array shapes should be checked, and pass the correct `axes` argument to `gs.transpose`. 

```
tr = MatricesSpace.transpose
a = gs.array([
    [1, 0],
    [0, 1]])
b = tr(a) 

tr([a, b]) 
>>> [b, a]
```

### DIFF
--- a/geomstats/geometry/matrices_space.py
+++ b/geomstats/geometry/matrices_space.py
@@ -65,7 +65,7 @@ class MatricesSpace(EuclideanSpace):
     @staticmethod
     def transpose(mat):
         """
-        Return the transpose of matrices. 
+        Return the transpose of matrices.
 
         Parameters
         ----------
--- a/geomstats/geometry/matrices_space.py
+++ b/geomstats/geometry/matrices_space.py
@@ -63,6 +63,23 @@ class MatricesSpace(EuclideanSpace):
         return gs.matmul(a, b) - gs.matmul(b, a)
 
     @staticmethod
+    def transpose(mat):
+        """
+        Return the transpose of matrices. 
+
+        Parameters
+        ----------
+        mat : array-like, shape=[n_samples, dim, dim]
+
+        Returns
+        -------
+        transpose : array-like, shape=[n_samples, dim, dim]
+        """
+        is_vec = (gs.ndim(gs.array(mat)) == 3)
+        axes = (0, 2, 1) if is_vec else (1, 0)
+        return gs.transpose(mat, axes)
+
+    @staticmethod
     def vector_from_matrix(matrix):
         """
         Conversion function from (_, m, n) to (_, mn).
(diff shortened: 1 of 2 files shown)


geomstats/geometry/matrices_space.py
```diff
@@ -62,6 +62,23 @@ def commutator(a, b):
...
+
+        Parameters
+        ----------
+        mat : array-like, shape=[n_samples, dim, dim]
+
+        Returns
+        -------
+        transpose : array-like, shape=[n_samples, dim, dim]
+        """
+        is_vec = (gs.ndim(gs.array(mat)) == 3)
```
ninamiolane: NIT: call this `is_vectorized`, in order not to be confused with `is_vector`?

tests/test_matrices_space.py
```diff
@@ -59,6 +59,15 @@ def test_commutator(self):
         result = self.space.commutator(x, [x, y, z])
         expected = gs.array([gs.zeros((3,3)), z, -y])
         self.assertAllClose(result, expected)
+    
+    @geomstats.tests.np_only
+    def test_transpose(self):
+        tr = self.space.transpose
```
ninamiolane: let's not rename functions: it might confuse other developers who are used to see transpose and array

opeltre: I actually think it's more readable:
+ it's only renamed inside the body function, for the next couple lines, and not globally
+ the following lines would actually be unreadable with the full names
+ this is exactly like declaring a local variable `tr_a = MatricesSpace.transpose(a)`, except that by letting `tr = MatricesSpace.transpose` I can reuse `tr` and apply it to `b`

geomstats/geometry/matrices_space.py
opeltre: ok fair enough... 
Je vais changer le nom dans mon master, mais laisse-moi le push dans une future PR stp! ... j'ai encore fait évoluer ma branche avec grassmann.log et c'est suffisament mineur pour que je ne me rebatte pas avec `git cherry-pick` ^^

tests/test_matrices_space.py
ninamiolane: @johmathe any opinion on this renaming?

ninamiolane:
@johmathe I'm merging this in the meantime so that @opeltre can move on, but let's chat on this soon?

tests/test_matrices_space.py
johmathe: This violates the style guide originally agreed and can lead to very bad things because of confusions between variable names and function names. Please revert this PR ASAP. As mentioned in the google style guide the most important thing is consistency across the code base.

PR #318, opened 2020-01-16
Add spherical coord and christoffel symbols in spherical coord to sphere

AliceLeBrigant:


geomstats/geometry/hypersphere.py
```diff
@@ -423,3 +491,36 @@ def parallel_transport(self, tangent_vec_a, tangent_vec_b, base_point):
             + gs.einsum('n,ni->ni', gs.cos(theta) * pb, normalized_b)\
             + p_orth
         return transported
+
+    def christoffel_spherical(self, point):
```
ninamiolane: Should this be `christoffel_symbol` to be consistent with the connection API, and then have a parameter `point_type='spherical'` to say that the christoffel symbols will be expressed in spherical coords?

AliceLeBrigant:
1. Changed 'christoffel_symbols' into 'christoffel' everywhere 
2. Deleted online_kmeans method of Hypersphere and corresponding test in test_hypersphere

PR #323, opened 2020-01-16
Change implementation of random uniform sampling on the hypersphere

nkoep:


geomstats/backend/numpy_random.py
```diff
@@ -1,19 +1,3 @@
...
-    return np.random.randint(*args, **kwargs)
-
-
-def seed(*args, **kwargs):
-    return np.random.seed(*args, **kwargs)
-
-
-def normal(*args, **kwargs):
-    return np.random.normal(*args, **kwargs)
+from numpy.random import normal, rand, randint, randn, seed  # NOQA
```
ninamiolane: If we do this change, shouldn't we do it in all the backends, otherwise we introduce style inconsistency. What do you think?

geomstats/geometry/hypersphere.py
```diff
@@ -167,45 +169,39 @@ def extrinsic_to_intrinsic_coords(self, point_extrinsic):
...
+        size = (n_samples, self.dimension + 1)
+
+        samples = gs.random.normal(size=size)
+        while True:
+            norms = gs.linalg.norm(samples, axis=1)
+            # TODO(nkoep): Should the 'bad sample tolerance' be configurable
+            #              via a function argument? Is TOLERANCE a good
+            #              default?
+            indcs = norms < TOLERANCE
+            num_bad_samples = gs.sum(indcs)
```
ninamiolane: gs.sum(gs.isclose(norms, 0.))

or RANDOM_TOL

?

geomstats/backend/numpy_random.py
nkoep: Jepp, see https://github.com/geomstats/geomstats/pull/326.

geomstats/geometry/hypersphere.py
nkoep: Jepp, that works :+1:

PR #326, opened 2020-01-16
Forward-export numpy functions

nkoep:
The majority of functions defined in the numpy backend were straightforward forward-calls to the bare numpy functions. This change simply forward-exports these functions rather than adding wrapper functions which simply pass along their `*args, **kwargs` arguments.

### DIFF
--- a/geomstats/backend/numpy.py
+++ b/geomstats/backend/numpy.py
@@ -1,6 +1,81 @@
 """Numpy based computation backend."""
 
 import numpy as np
+from numpy import (  # NOQA
+    abs,
+    all,
+    allclose,
+    amax,
+    any,
+    arange,
+    arccos,
+    arccosh,
+    arcsin,
+    arctan2,
+    argmax,
+    argmin,
+    array,
+    asarray,
+    average,
+    clip,
+    concatenate,
+    cos,
+    cosh,
+    cov,
+    cross,
+    diagonal,
+    divide,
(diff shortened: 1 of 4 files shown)


geomstats/backend/numpy.py
```diff
@@ -386,36 +181,7 @@ def ndim(x):
...
-def normal(*args, **kwargs):
-    return np.random.normal(*args, **kwargs)
-
-
 def cumprod(x, axis=0):
     if axis is None:
         raise NotImplementedError('cumprod is not defined where axis is None')
-    else:
-        return np.cumprod(x, axis=axis)
+    return np.cumprod(x, axis=axis)
```
ninamiolane: Thanks!

geomstats/backend/numpy.py
```diff
@@ -1,6 +1,81 @@
 """Numpy based computation backend."""
 
 import numpy as np
+from numpy import (  # NOQA
```
ninamiolane: And in alphabetic order, thank you so much!

geomstats/backend/numpy_linalg.py
```diff
@@ -32,8 +42,7 @@ def expm(x):
 
     if ndim == 2:
         return result[0]
-    else:
-        return result
+    return result
```
ninamiolane: 👍

PR #327, opened 2020-01-16
grassmannian logarithm

opeltre:
An implementation of the logarithm will stay on this branch.
tests: run on numpy 
issues: `logm` non-existent on torch, not properly vectorised on numpy

I honestly today don't have the courage to change the code into obfuscated one, just to prevent from assigning functions to local variable names. As mentioned, @johmathe, I don't see the point of this diktat, so I'll leave it like this for now, pull if you want else not, I'll work on different branches in the meantime. 

It seems to me there are more important topics, like having a consistent `expm` and `logm` methods in the backend, to be used accross all matrix classes.

### DIFF
--- a/geomstats/geometry/grassmannian.py
+++ b/geomstats/geometry/grassmannian.py
@@ -10,7 +10,6 @@ from geomstats.geometry.matrices_space import MatricesSpace
 from geomstats.geometry.matrices_space import MatricesMetric
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
-
 TOLERANCE = 1e-5
 EPSILON = 1e-6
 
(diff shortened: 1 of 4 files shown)


geomstats/geometry/grassmannian.py
```diff
@@ -78,3 +84,34 @@ def exp(self, vector, point):
         expm = gs.linalg.expm
         mul = MatricesSpace.mul
         return mul(mul(expm(vector), point), expm(-vector))
+
+    def log(self, point, base_point):
+        """
```
ninamiolane: Do you have references / formulas for the expressions of the log and the exp that we could put in the docstring?

geomstats/geometry/grassmannian.py
```diff
@@ -78,3 +84,34 @@ def exp(self, vector, point):
...
+        Returns:
+        -------
+        log : array-like, shape=[n_samples, n, n]
+        """
+        svd = gs.linalg.svd
+        mul = gs.matmul
+        tr = MatricesSpace.transpose
+        logm = gs.linalg.logm
+
+        def closest(rot):
```
ninamiolane: Closest to rot, in the grassmanian, for its own metric? maybe add a docstring so that it's easier for us to understand quickly what is being computed.

geomstats/geometry/grassmannian.py
opeltre: Will do for the exp soon!

geomstats/geometry/grassmannian.py
opeltre: Hmm... as the name suggests `closest(rot)` is supposed to yield the closest representant of `rot` in SO(n) up to SO(k) x SO(n-k) 
But I realized this was actually a hack and although it does yield an interpolating curve in Gr(k, n) this curve is not ensured to be geodesic.

opeltre:
fix log

PR #335, opened 2020-01-17
Support python 3.7 builds in travis

nkoep:
This is another attempt at adding python 3.7 support in the integration tests after the recent version bump of the pytorch dependency.


ninamiolane:
Wooooohooo!

PR #341, opened 2020-01-17
New try on better numerics for exp and log in hypersphere

ninamiolane:
Follows #340


geomstats/geometry/hypersphere.py
```diff
@@ -357,14 +357,41 @@ def exp(self, tangent_vec, base_point):
...
+
+        mask_0 = gs.isclose(norm_tangent_vec, 0.)
+        mask_non0 = ~mask_0
+
+        coef_1 = gs.zeros((n_exps, 1))
+        coef_2 = gs.zeros((n_exps, 1))
+        norm2 = norm_tangent_vec[mask_0]**2
+        norm4 = norm2**2
+        norm6 = norm2**3
+        coef_1[mask_0] = 1. - norm2/2. + norm4/24. - norm6/720.
```
nkoep: Sorry to revive this after the PR was already merged, but for clarity these constants should probably be kept in a look-up table like https://github.com/geomstats/geomstats/blob/c95984472e9e91eec3eed455a95d1d62132d02a1/geomstats/geometry/hypersphere.py#L18 In fact, `COS_TAYLOR_COEFFS` is nowhere used in the module at all atm. I'm surprised flake8 doesn't catch this...

ninamiolane: Creating an issue now

PR #342, opened 2020-01-17
Pull version info from *requirements.txt files inside setup.py

nkoep:
This change makes the *requirements.txt files the 'single source of
truth' regarding version dependencies. This information was previously
duplicated in `__about__.py`, which had been out-of-sync for a while.

Signed-off-by: Niklas Koep <niklas.koep@gmail.com>


ninamiolane:
Awesome, thank you!

PR #360, opened 2020-01-17
Docs geom 2.py

cshewmake2:
Format and add further documentation to files.

Note that hyperbolic space is not complete in this respect as uncertainties exist about certain functions internal.


geomstats/geometry/general_linear_group.py
```diff
@@ -34,9 +39,16 @@ def get_identity(self, point_type=None):
                 ' implemented for a point_type that is not \'matrix\'.')
     identity = property(get_identity)
 
-    def belongs(self, mat):
-        """
-        Check if mat belongs to GL(n).
+    def belongs(self, mat):  # TODO Should this stay generic? (point)
```
ninamiolane: Good question, I could not decide on this. A TODO makes sense here

cshewmake2: What's the best way to annotate these things? I don't want them to get exposed to the users unnecessarily, but I also want contributors to see them. Hmm.

Issue #338, opened 2020-01-17
Vectorization: Make behaviour clear and work the same everywhere


stefanheyder:
Currently, vectorization is handled differently by different functions (and backends); the behaviour should be uniform across all of geomstats. Suppose we have a function
```
foo(bar, baz)
```
where `foo` is vectorized over both `bar` and `baz`, `bar` being arrays of the same shape s and `foo` returns an array of shape r. We should be able to handle the following cases:

### bar and baz are of shape s (e.g. only one object passed for each argument)
`foo(bar, baz)` is of shape r
### bar is of shape (n, s) and baz is of shape s
`foo(bar, baz)` should be of shape (n,r) where each of the r-arrays is the result of applying `foo` pointwise to bar.
### bar is of shape s and baz is of shape (n,s)
Same as above
### bar is of shape (n,s) and baz is of shape (n,s)
`foo(bar, baz)` should be of shape (n,r), applying `foo` across the first dimension: the first element of the output is the result of `foo(bar[0], baz[0])`, the second `foo(bar[1], baz[1])` and so on.

@cshewmake2 suggested adding a decorator function doing the following:
Remember the shape of the input arrays, cast them using `to_ndarray` to have shape (n,s) (even if n = 1), calculate the function result and cast back to the desired shape.

This would allow the developers to focus on vectorized functions, not having to worry about casting between arrays of different shape. It might also be useful to have a look at numpy, as their functions implement exactly this behaviour.

ninamiolane:
Some documentation on how NumPy does it: https://docs.scipy.org/doc/numpy/reference/internals.code-explanations.html#broadcasting

And C code:
https://github.com/numpy/numpy/blob/f31ce546045ccb765ef0992b9c841d4682eee775/numpy/core/src/multiarray/iterators.c#L1162

ninamiolane:
Duplicate of issue #48, I am closing #48

ninamiolane:
Update w.r.t. #128, closing #128

ninamiolane:
The two main problems of this issue are:
- **a.** Functions output unintuitive shapes: 
    - e.g. `gs.inner_product([1, 2, 3], [1, 1, 1]) == gs.array([[6]])` instead of `gs.array(6)`,
- **b.** Many reshapings happen in the codebase, with `gs.to_ndarray`, which:
    - **b1:** slows it down,
    - **b2:** make the code less readable.

I started to address these, with the following strategy:
[...]

PR #361, opened 2020-01-17
Geodesics with no closed form

nguigs:


geomstats/backend/numpy/__init__.py
```diff
@@ -1,7 +1,7 @@
 """Numpy based computation backend."""
 
-import numpy as _np
-from numpy import (  # NOQA
+import autograd.numpy as _np
```
ninamiolane: 🎉

geomstats/backend/numpy_linalg.py
```diff
@@ -0,0 +1,62 @@
+"""Numpy based linear algebra backend."""
```
ninamiolane: There is a rebase problem, this file shouldn't be there. The organization of the backends has changed, it is in different folders now

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
...
-        raise NotImplementedError(
-                'The affine connection exponential is not implemented.')
 
-    def log(self, point, base_point):
+        n_steps: int
+        """
+        tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=2)
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
+
+        def geodesic_equation(x, v):
```
ninamiolane: Avoid single letters to denote variables, otherwise it becomes impossible to use automatic find and replace later

ninamiolane: Why did you choose to have geodesic_equation as an inner function, and not a method of the class connection?

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
...
+            v = gs.to_ndarray(v, to_ndim=2)
+            gamma = self.christoffels(x)
+            return - gs.einsum('lkij,li,lj->lk', gamma, v, v)
+
+        initial_state = (base_point, tangent_vec)
+        flow, _ = self.integrate(geodesic_equation, initial_state,
+                                 n_steps=n_steps)
+        return flow[-1]
+
+    def log(self, point, base_point, n_steps=N_STEPS):
```
ninamiolane: 🎉 woooo

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
         raise NotImplementedError(
                 'connection is not implemented.')
 
-    def exp(self, tangent_vec, base_point):
+    def exp(self, tangent_vec, base_point, n_steps=N_STEPS):
```
ninamiolane: What is n_steps?

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
...
+
+        n_steps: int
         """
-        raise NotImplementedError(
-                'The affine connection logarithm is not implemented.')
+
+        point = gs.to_ndarray(point, to_ndim=2)
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
+
+        def objective(vel):
```
ninamiolane: what is vel, velocity? should we call it tangent_vec to be consistent with the naming?

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
...
+            vel = vel.reshape(base_point.shape)
+            delta = self.exp(vel, base_point, n_steps) - point
+            loss = 1 / self.dimension * gs.sum(delta ** 2, axis=1)
+            return 1 / n_samples * gs.sum(loss)
```
ninamiolane: don't you get an error here, where is n_samples defined?

geomstats/geometry/connection.py
```diff
@@ -41,39 +49,85 @@ def connection(self, tangent_vector_a, tangent_vector_b, base_point):
...
+
+        objective_grad = autograd.elementwise_grad(objective)
+
+        def objective_with_grad(vel):
+            return objective(vel), objective_grad(vel)
+
+        n_samples = len(base_point)
```
ninamiolane: define n_samples above?

geomstats/geometry/connection.py
```diff
@@ -93,6 +147,13 @@ def pole_ladder_step(self, base_point, next_point, base_shoot):
 
         end_point: array-like, shape=[n_samples, dimension]
                                                 or shape=[1, dimension]
+
+        References
```
ninamiolane: Thanks for the reference!

geomstats/geometry/connection.py
```diff
@@ -198,6 +322,46 @@ def torsion(self, base_point):
         raise NotImplementedError(
                 'The torsion tensor is not implemented.')
 
+    @staticmethod
+    def _symplectic_euler_step(y, force, dt):
+        x, v = y
```
ninamiolane: avoid single letter variables

geomstats/geometry/connection.py
```diff
@@ -198,6 +322,46 @@ def torsion(self, base_point):
...
+
+        Returns
+        ----------
+        a tuple of sequences of solutions every end_time / n_steps
+        """
+        dt = end_time / n_steps
+        positions = [initial_state[0]]
+        velocities = [initial_state[1]]
+        current_state = (positions[0], velocities[0])
+        for i in range(n_steps):
```
ninamiolane: for _ in range(n_steps), as i is not used.

geomstats/geometry/hypersphere.py
```diff
@@ -532,14 +549,14 @@ def christoffels(self, point, point_type='spherical'):
...
-                                self.dimension,
-                                self.dimension,
-                                self.dimension))
-        christoffel[:, 0, 1, 1] = - gs.sin(point[:, 0]) * gs.cos(point[:, 0])
-        christoffel[:, 1, 0, 1] = gs.cos(point[:, 0]) / gs.sin(point[:, 0])
-        christoffel[:, 1, 1, 0] = gs.cos(point[:, 0]) / gs.sin(point[:, 0])
-
-        return christoffel
+        christoffel = []
+        for p in point:
```
ninamiolane: avoid one letter variable.

tests/test_connection.py
```diff
@@ -64,8 +67,35 @@ def test_parallel_transport(self):
             tan_vec_a, tan_vec_b, base_point)
         result = connection.pole_ladder_parallel_transport(
             tan_vec_a, tan_vec_b, base_point)
+
         self.assertAllClose(result, expected, rtol=1e-7, atol=1e-5)
 
+    @geomstats.tests.np_only
+    def test_exp(self):
+        p = gs.array([[gs.pi / 2, 0], [gs.pi / 6, gs.pi / 4]])
```
ninamiolane: avoid one letter variable, p -> point

tests/test_connection.py
```diff
@@ -64,8 +67,35 @@ def test_parallel_transport(self):
...
+        self.connection.christoffels = self.hypersphere.metric.christoffels
+        expected = self.hypersphere.metric.exp(vector_ext, point_ext)
+        result_spherical = self.connection.exp(vec, p, n_steps=1000)
+        result = self.hypersphere.spherical_to_extrinsic(result_spherical)
+
+        self.assertAllClose(result, expected, rtol=1e-3)
+
+    @geomstats.tests.np_only
+    def test_log(self):
+        p = gs.array([[gs.pi / 3, gs.pi / 4], [gs.pi / 2, gs.pi / 4]])
```
ninamiolane: same

geomstats/geometry/connection.py
nguigs: the number of steps to discretize the time interval in the integration of the geodesic equation

geomstats/geometry/connection.py
nguigs: It wasn't used anywhere else. In fact it outputs the acceleration, it can be renamed if keeping geodesic_equation as a method is important, but I don't think it was useful: how do you output an ODE?

geomstats/geometry/connection.py
nguigs: The number of steps to discretize the time interval in the integration scheme

geomstats/geometry/connection.py
nguigs: I called it tangent_vec outside this subfunction

tests/test_connection.py
nguigs: point is already a variable!

geomstats/geometry/connection.py
ninamiolane: It can be useful during inheritance, if we want to have a geodesic_equation method for LeviCivita connection (if we keep this class), or RiemannianMetric (if it inherits from Connection, like in the code refactorization that we have discussed with @opeltre )

opeltre: yep we'll definitely have to see how to put some order in all of this! 
In the meantime it's good to have it working at some place so that we only have to re-reference or copy-paste when it's figured out :)

We worked on some other kind of cool ODE methods with nguigs yesterday using autograd! it may be good to have all these in a separate file and import them as needed, so that we don't mix up autograd-related ODE integration techniques with the code defining the Connection class interface...

geomstats/geometry/connection.py
opeltre: @nguigs  :carrot:  exactly what I told you in the kitchen! 
we should use '...' einsum ellipsis instead of 'to_ndarray' to vectorize consistently and return a 1D-array if given one.
+ delete line `v = gs.to_ndarray(v, to_ndim=2)`
+ replace `gs.einsum('...kij,...i,...j->...k', gamma, v, v)`
(I even think the last `->...k` can be omitted and figured out by einsum)

geomstats/geometry/connection.py
ninamiolane: Agreed! @AliceLeBrigant and I were also talking about creating an integration_utils.py somewhere.

ninamiolane: Agreed! @AliceLeBrigant and I were also talking about creating an integration_utils.py somewhere.

geomstats/geometry/connection.py
ninamiolane: We had the einsum ellipsis before, but had to replace all of them because pytorch and/or tensorflow do not support them. This might have changed as we use newer versions of pytorch and tensorflow now, but it's worth testing before converting everything.

ninamiolane: We had the einsum ellipsis before, but had to replace all of them because pytorch and/or tensorflow do not support them. This might have changed as we use newer versions of pytorch and tensorflow now, but it's worth testing before converting everything.

nguigs: I started changing the pytorch einsum in the backend, I will do the tests and try to get the exp and log work with this backend as well

nguigs: We need to solve #338 first and then reformat this. As it is, I think it won't work if tangent vectors are matrices.

Issue #282, opened 2020-01-15
Make dependencies work with newer python versions


stefanheyder:
See also #278.

ninamiolane:
Now working with python 3.5, 3.6, 3.7 and 3.8

Issue #280, opened 2020-01-15
Make dependecy on tensorflow and pytorch optional for development


stefanheyder:
These cause issues with several python version (see e.g. #278).

ninamiolane:
This does not cause issues with newer python versions anymore.

If developers do not want torch or tensorflow, they are free to comment the corresponding lines in the requirements.txt.

Issue #278, opened 2020-01-15
Write Installation guide


stefanheyder:
Currently there are some problems installing geomstats under different Python versions, in particular Python 3.7 and 3.8 cause problems with tensorflow and pytorch.

To make installation more consistent this, a working installation method should be documented, including the required programs (pyenv / pipenv (?)) and operations required to install all dependecies without causing any error.

ninamiolane:
Now `requirements.txt` contains a working installation method, for python versions 3.5, 3.6, 3.7 and 3.8.

PR #368, opened 2020-01-19
Add python 3.8 in travis

ninamiolane:
Since tensorflow 2.0 does not support python 3.8, we avoid testing in tf 2.0 in the 3.8 version.


ninamiolane:
This PR, together with #362, #363, #367, #369, finishes to address issues #282, #280 and #278.
- Python 3.5, 3.6, 3.7 and 3.8 are now tested with np, tf and torch.
- With the exception that tf 2.0 and torch are not available for Python 3.8, `pip3 install -r requirements` with python 3.8 will disregard tf and torch installations.

PR #379, opened 2020-01-20
Add decorator for readability in vectorization operations

ninamiolane:
Start to address #338, following @cshewmake2 's recommendation about decorators.

This needs unit tests.


ninamiolane:
This needs to be adapted when we want to decorate methods and not functions.

PR #359, opened 2020-01-17
Yanntwd spd logeucl metric

ythanwerdas:

PR #383, opened 2020-01-20
Run flake8 only on files differing from master

stefanheyder:
This should fix the problem @ythanwerdas mentioned in https://github.com/geomstats/geomstats/pull/359#issuecomment-576285413, by running the flake8 check only on the files that differ in the submitted PR from master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
     - pip install -q -r dev-requirements.txt
     - pip install -q -r requirements.txt --only-binary=numpy,scipy
 script:
-    - flake8 examples geomstats tests
+    - flake8 $(git diff --name-only master HEAD geomstats examples tests)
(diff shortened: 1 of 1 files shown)


ninamiolane:
I think @ythanwerdas' problem was elsewhere, but do you still think we should still do this? I like that flake8 tests everything every time, but it's probably also slower, although not much slower compared to nose2.

stefanheyder:
Yes, I also mistunderstood the problem. Still I agree this is a good thing to have, as it will make CI faster and formatting violations can only arise in files that are touched (whereas tests may fail in files completely unrelated to the current changes, thus all tests should always be run).

ninamiolane:
Yes! Let's just be careful when we integrate flake8-docstring to travis, because right now most of the current code is not flake8-docstring compliant.

PR #386, opened 2020-01-21
Add simple example using geomstats in pymanopt

nkoep:
This implements a simple toy example from pymanopt on top of geomstats. I've intentionally not included the example in `tests/test_examples.py`, as the example obviously requires pymanopt to run, which we probably shouldn't add as a (hard) dev dependency. The example merely demonstrates how to adapt geomstats to pymanopt's manifold interface.

### DIFF
--- a/examples/geomstats_in_pymanopt.py
+++ b/examples/geomstats_in_pymanopt.py
@@ -68,9 +68,9 @@ def estimate_dominant_eigenvector(matrix):
     the Rayleigh quotient -x' * A * x / (x' * x).
     """
     num_rows, num_columns = gs.shape(matrix)
-    assert num_rows == num_columns, "matrix must be square"
+    assert num_rows == num_columns, 'matrix must be square'
     assert gs.allclose(gs.sum(matrix - gs.transpose(matrix)), 0.0), \
-        "matrix must be symmetric"
+        'matrix must be symmetric'
 
     def cost(vector):
         return -gs.dot(vector, gs.dot(matrix, vector))
@@ -84,10 +84,10 @@ def estimate_dominant_eigenvector(matrix):
     return solver.solve(problem)
 
 
-if __name__ == "__main__":
-    if os.environ.get("GEOMSTATS_BACKEND") != "numpy":
+if __name__ == '__main__':
+    if os.environ.get('GEOMSTATS_BACKEND') != 'numpy':
         raise SystemExit(
-            "This example currently only supports the numpy backend")
+            'This example currently only supports the numpy backend')
 
     ambient_dimension = 128
     matrix = gs.random.normal(size=(ambient_dimension, ambient_dimension))
@@ -101,11 +101,11 @@ if __name__ == "__main__":
             gs.sign(dominant_eigenvector_estimate[0])):
         dominant_eigenvector_estimate = -dominant_eigenvector_estimate
 
-    print("l2-norm of dominant eigenvector:",
+    print('l2-norm of dominant eigenvector:',
           gs.linalg.norm(dominant_eigenvector))
-    print("l2-norm of dominant eigenvector estimate:",
+    print('l2-norm of dominant eigenvector estimate:',
           gs.linalg.norm(dominant_eigenvector_estimate))
     error_norm = gs.linalg.norm(
         dominant_eigenvector - dominant_eigenvector_estimate)
-    print("l2-norm of difference vector:", error_norm)
-    print("solution found: %s" % gs.isclose(error_norm, 0.0, atol=1e-3))
+    print('l2-norm of difference vector:', error_norm)
+    print('solution found: %s' % gs.isclose(error_norm, 0.0, atol=1e-3))
--- a/examples/geomstats_in_pymanopt.py
+++ b/examples/geomstats_in_pymanopt.py
@@ -9,76 +9,79 @@ The example currently requires installing the pymanopt HEAD from git:
 
 import functools
 import os
-os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"  # NOQA
 
-import numpy as np
-from numpy import linalg as la, random as rnd
 from pymanopt import Problem
 from pymanopt.manifolds.manifold import Manifold
 from pymanopt.solvers import SteepestDescent
 
+import geomstats.backend as gs
 from geomstats.geometry.hypersphere import Hypersphere
 
 
-def squeeze_ndarray(f):
+def squeeze_output(function):
     """Decorator to remove singleton dimensions from a numpy.ndarray returned
     by the decorated function.
     """
-    @functools.wraps(f)
+    @functools.wraps(function)
     def inner(*args, **kwargs):
-        return np.squeeze(f(*args, **kwargs))
+        return gs.squeeze(function(*args, **kwargs))
     return inner
 
 
 class GeomstatsSphere(Manifold):
     """A simple adapter class which proxies calls by pymanopt's solvers to
-    manifolds to the underlying geomstats Hypersphere class.
+    `Manifold` subclasses to the underlying geomstats `Hypersphere` class.
     """
 
-    def __init__(self, n):
-        self._sphere = Hypersphere(n-1)
+    def __init__(self, ambient_dimension):
+        self._sphere = Hypersphere(ambient_dimension-1)
 
-    def norm(self, x, g):
-        return self._sphere.metric.norm(g, base_point=x)
+    def norm(self, base_vector, tangent_vector):
+        return self._sphere.metric.norm(tangent_vector, base_point=base_vector)
 
-    @squeeze_ndarray
-    def proj(self, x, g):
-        return self._sphere.projection_to_tangent_space(g, base_point=x)
+    def inner(self, base_vector, tangent_vector_a, tangent_vector_b):
+        return self._sphere.metric.inner_product(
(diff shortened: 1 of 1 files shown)


examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+steepest descent solver.
+
+The example currently requires installing the pymanopt HEAD from git:
+
+    pip install git+ssh://git@github.com/pymanopt/pymanopt.git
+"""
+
+import functools
+import os
+os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"  # NOQA
```
ninamiolane: Do we need this? This is related to tensorflow, which is not imported?

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+import numpy as np
+from numpy import linalg as la, random as rnd
+from pymanopt import Problem
+from pymanopt.manifolds.manifold import Manifold
+from pymanopt.solvers import SteepestDescent
+
+from geomstats.geometry.hypersphere import Hypersphere
+
+
+def squeeze_ndarray(f):
```
ninamiolane: Very nice, and way shorter than the one that I started to implement in PR #379 . (which implements a decorator's decorator in order to have parameters in the decorator) Could we refactorize this code somehow?

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+    def norm(self, x, g):
+        return self._sphere.metric.norm(g, base_point=x)
+
+    @squeeze_ndarray
+    def proj(self, x, g):
+        return self._sphere.projection_to_tangent_space(g, base_point=x)
+
+    @squeeze_ndarray
+    def retr(self, x, g):
+        # geomstats's hypersphere implementation doesn't provide a retraction
```
ninamiolane: Maybe a docstring instead of a comment? It could be parsed by the automatic documentation generator later this way (if we put the examples there).

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+
+class GeomstatsSphere(Manifold):
+    """A simple adapter class which proxies calls by pymanopt's solvers to
+    manifolds to the underlying geomstats Hypersphere class.
+    """
+
+    def __init__(self, n):
+        self._sphere = Hypersphere(n-1)
+
+    def norm(self, x, g):
```
ninamiolane: `x, g` is pymanopt's API, right?

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+
+    pip install git+ssh://git@github.com/pymanopt/pymanopt.git
+"""
+
+import functools
+import os
+os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"  # NOQA
+
+import numpy as np
+from numpy import linalg as la, random as rnd
```
ninamiolane: No renaming of packages, for consistency of code, faster reading through the codebase, and ease of automatic replacement tools

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+from pymanopt.solvers import SteepestDescent
+
+from geomstats.geometry.hypersphere import Hypersphere
+
+
+def squeeze_ndarray(f):
+    """Decorator to remove singleton dimensions from a numpy.ndarray returned
+    by the decorated function.
+    """
+    @functools.wraps(f)
```
ninamiolane: NIT: Avoid single-letter variables as a general rule, to be able to do automatic replacement and grep, bcs grep 'f' would return the entire codebase.

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+
+    sphere = GeomstatsSphere(n)
+    problem = Problem(manifold=sphere, cost=cost, egrad=egrad)
+    solver = SteepestDescent()
+    xopt = solver.solve(problem)
+    return xopt
+
+
+if __name__ == "__main__":
+    if os.environ.get("GEOMSTATS_BACKEND") != "numpy":
```
ninamiolane: Even if only numpy is supported, does it make sense to use `gs.` syntax if we try to generalize it later on?

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+    by the decorated function.
+    """
+    @functools.wraps(f)
+    def inner(*args, **kwargs):
+        return np.squeeze(f(*args, **kwargs))
+    return inner
+
+
+class GeomstatsSphere(Manifold):
+    """A simple adapter class which proxies calls by pymanopt's solvers to
```
ninamiolane: 🎉

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+    """Decorator to remove singleton dimensions from a numpy.ndarray returned
+    by the decorated function.
+    """
+    @functools.wraps(f)
+    def inner(*args, **kwargs):
+        return np.squeeze(f(*args, **kwargs))
+    return inner
+
+
+class GeomstatsSphere(Manifold):
```
stefanheyder: Could this wrapper be extended to a Wrapper for any GeomstatsManifold? Quickly glancing at the interface in pymanopt, most of the interface is available, except for `egrad2rgrad`, `ehess2rhess`, `retr` and some information about the structure of the manifold e.g. `typicaldist`.

examples/geomstats_in_pymanopt.py
```diff
@@ -0,0 +1,113 @@
...
+    xopt = solver.solve(problem)
+    return xopt
+
+
+if __name__ == "__main__":
+    if os.environ.get("GEOMSTATS_BACKEND") != "numpy":
+        raise SystemExit(
+            "This example currently only supports the numpy backend")
+
+    # Generate random problem data.
```
ninamiolane: I've heard it is good practice to avoid comments (they are difficult to maintain), and favor explanatory variable names.

examples/geomstats_in_pymanopt.py
stefanheyder: What is this line necessary for?

stefanheyder:
I checked out the PR and ran it locally, on my machine it works :) Thank you for the example!

examples/geomstats_in_pymanopt.py
nkoep: It's not needed, just silences an ugly warning that get's triggered since pymanopt imports tensorflow. I can remove it if it's preferable. Might otherwise give the impression it's needed for something.

examples/geomstats_in_pymanopt.py
nkoep: Yes, though the naming convention fairly inconsistently mirrors the one used in manopt. I've been meaning to unify and break with the manopt conventions in a future release anyway. I'll change this to match closer what you do in geomstats.

examples/geomstats_in_pymanopt.py
nkoep: Fair enough. Guess I really ought to read up on the contributing guidelines :sunglasses:

examples/geomstats_in_pymanopt.py
nkoep: Will do!

examples/geomstats_in_pymanopt.py
nkoep: Jepp, makes sense.

examples/geomstats_in_pymanopt.py
nkoep: This was definitely the idea for future examples as it's fairly straightforward to wrap the API. The question is where such an adapter class should live as I'd think it odd if a pymanopt-specific class popped up somewhere in the geomstats codebase since pymanopt is not (yet^^) a hard dependency.

examples/geomstats_in_pymanopt.py
nkoep: Just silences an ugly TF warning, see above. I'll remove the line for readability's sake.

examples/geomstats_in_pymanopt.py
nkoep: I guess this is a philosophical question. Of course code should ideally be self-documenting, but a comment here or there goes a long way IMO, but maybe an example should be self-explanatory. Your call. I guess this example is simple enough to do away with excessive comments.

ninamiolane: Agreed, totally philosophical and maybe also personal pref. A "no comment style rule" also forces the user to write code's specificities in the docstrings, which are parsed automatically in the documentation. And knowing that it will be in the online documentation might also encourage writing something understandable. I have a tiny preference for no-comment (as you can see x)), but this is an example and not the core codebase so both are fine afaic.

examples/geomstats_in_pymanopt.py
nkoep: I've had a look at doing this via `geomstats.vectorization.decorator`, and was bitten by the fact that the decorator currently doesn't support instance methods (see https://github.com/geomstats/geomstats/pull/379#issuecomment-576250069). Isn't it overkill to use the decorator in this case anway? I feel like we're muddying terminology a bit at this point. The `squeeze_array` decorator here isn't concerned with vectorization, just with the removal of singleton dimensions.

PR #388, opened 2020-01-21
Merge matrices.py

opeltre:
This merge updates the `Matrices` class to provide with convenience methods discussed: equal, transpose, is_symmetric etc. 

I spent some time recabling the tests as well, while doing so I noticed many of them had really inconsistent shape signatures and this fixes them. 

This merge is an important step towards unifying vectorization behaviour across matrices-related files. Inline implementations are bug-prone and sometimes really hard to find.

I should also mention the convoluted way SPD(n) resorted to symmetry checks: 
```python
SPD.__init__:
     SPDn.embedding_manifold = GLn
SPD.belongs:
     #gs.to_ndarray cooking
     return self.embedding_manifold.is_symmetric(...)

GL.is_symmetric: 
      # inherited from MatricesSpace
```

### DIFF
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -1,8 +1,4 @@
-"""
-This module exposes the `GeneralLinearGroup` class. 
-
-Note: This class is temporarily kept for compatibility purposes
-but will soon be renamed `GeneralLinear`.
+"""Module exposing the GeneralLinear group class.
 """
 
 import geomstats.backend as gs
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -1,4 +1,9 @@
-"""The General Linear Group, i.e. the matrix group GL(n)."""
+"""
+This module exposes the `GeneralLinearGroup` class. 
+
+Note: This class is temporarily kept for compatibility purposes
+but will soon be renamed `GeneralLinear`.
+"""
 
 import geomstats.backend as gs
 from geomstats.geometry.lie_group import LieGroup
--- a/geomstats/geometry/grassmannian.py
+++ b/geomstats/geometry/grassmannian.py
@@ -1,8 +1,4 @@
-"""Grassmannian manifold Gr(n, p).
-
-The Grassmannian manifold is the set of all p-dimensional
-subspaces in n-dimensional space, where p <= n.
-"""
+"""This module exposes the `Grassmannian` and `GrassmannianMetric` class."""
 
 import geomstats.backend as gs
 from geomstats.geometry.embedded_manifold import EmbeddedManifold
@@ -87,4 +83,4 @@ class GrassmannianCanonicalMetric(RiemannianMetric):
         """
         expm = gs.linalg.expm
         mul = Matrices.mul
-        return mul(mul(expm(vector), point), expm(-vector))
+        return mul(expm(vector), point, expm(-vector))
--- a/geomstats/geometry/matrices.py
+++ b/geomstats/geometry/matrices.py
@@ -1,7 +1,8 @@
-"""This module exposes the `Matrices` and `MatricesMetric` class."""
+"""Module exposing the `Matrices` and `MatricesMetric` class."""
+
+from functools import reduce
 
 import geomstats.backend as gs
-from functools import reduce
 from geomstats.geometry.euclidean_space import EuclideanSpace
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
@@ -25,7 +26,7 @@ class Matrices(Euclidean):
         point = gs.to_ndarray(point, to_ndim=3)
         _, mat_dim_1, mat_dim_2 = point.shape
         return mat_dim_1 == self.m & mat_dim_2 == self.n
-    
+
     @staticmethod
     def equal(a, b, atol=TOLERANCE):
         """
@@ -40,8 +41,8 @@ class Matrices(Euclidean):
         -------
         eq : array-like boolean, shape=[n_samples]
         """
-        is_vectorized = (gs.ndim(gs.array(a)) == 3)\
-                or (gs.ndim(gs.array(b)) == 3)
+        is_vectorized = \
+            (gs.ndim(gs.array(a)) == 3) or (gs.ndim(gs.array(b)) == 3)
         axes = (1, 2) if is_vectorized else (0, 1)
         return gs.all(gs.isclose(a, b, atol=atol), axes)
 
--- a/geomstats/geometry/matrices.py
+++ b/geomstats/geometry/matrices.py
@@ -67,8 +67,7 @@ class Matrices(Euclidean):
     @classmethod
     def bracket(cls, mat_a, mat_b):
         """
-        Return the commutator of a and b,
-        i.e. `[a, b] = ab - ba`.
+        Return the commutator of a and b, i.e. `[a, b] = ab - ba`.
 
         Parameters
         ----------
@@ -101,7 +100,7 @@ class Matrices(Euclidean):
     def is_symmetric(cls, mat, atol=TOLERANCE):
         """
         Check if a matrix is symmetric.
-        
+
         Parameters
         ----------
         mat : array-like, shape=[n_samples, n, n]
@@ -121,7 +120,7 @@ class Matrices(Euclidean):
         Parameters
         ----------
         mat : array-like, shape=[n_samples, n, n]
-        
+
         Returns
         -------
         sym : array-like, shape=[n_samples, n, n]
--- a/geomstats/geometry/matrices.py
+++ b/geomstats/geometry/matrices.py
@@ -1,7 +1,8 @@
-"""The space of matrices (m, n), which is the Euclidean space R^{mn}."""
+"""This module exposes the `Matrices` and `MatricesMetric` class."""
 
 import geomstats.backend as gs
-from geomstats.geometry.euclidean import Euclidean
+from functools import reduce
+from geomstats.geometry.euclidean_space import EuclideanSpace
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
 
@@ -24,10 +25,11 @@ class Matrices(Euclidean):
         point = gs.to_ndarray(point, to_ndim=3)
         _, mat_dim_1, mat_dim_2 = point.shape
         return mat_dim_1 == self.m & mat_dim_2 == self.n
-
+    
     @staticmethod
-    def mul(a, b):
-        """Return the product of matrices a and b.
+    def equal(a, b, atol=TOLERANCE):
+        """
+        Test if matrices a and b are close.
 
         Parameters
         ----------
@@ -36,13 +38,36 @@ class Matrices(Euclidean):
 
         Returns
         -------
-        c : array-like, shape=[n_samples, dim1, dim3]
+        eq : array-like boolean, shape=[n_samples]
         """
-        return gs.matmul(a, b)
+        is_vectorized = (gs.ndim(gs.array(a)) == 3)\
+                or (gs.ndim(gs.array(b)) == 3)
+        axes = (1, 2) if is_vectorized else (0, 1)
+        return gs.all(gs.isclose(a, b, atol=atol), axes)
 
     @staticmethod
-    def commutator(a, b):
-        """Return the commutator of a and b, i.e. :math: `[a, b] = ab - ba`.
+    def mul(*args):
+        """
+        Return the product of matrices a1, ..., an.
+
+        Parameters
+        ----------
+        a1 : array-like, shape=[n_samples, dim_1, dim_2]
+        a2 : array-like, shape=[n_samples, dim_2, dim_3]
+        ...
+        an : array-like, shape=[n_samples, dim_n-1, dim_n]
+
+        Returns
+        -------
+        mul : array-like, shape=[n_samples, dim_1, dim_n]
+        """
+        return reduce(gs.matmul, args)
+
+    @classmethod
+    def commutator(cls, a, b):
+        """
+        Return the commutator of a and b,
+        i.e. `[a, b] = ab - ba`.
 
         Parameters
         ----------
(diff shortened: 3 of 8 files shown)


geomstats/geometry/matrices.py
```diff
@@ -36,13 +39,36 @@ def mul(a, b):
...
         Returns
         -------
-        c : array-like, shape=[n_samples, dim1, dim3]
+        eq : array-like boolean, shape=[n_samples]
         """
-        return gs.matmul(a, b)
+        is_vectorized = \
+            (gs.ndim(gs.array(a)) == 3) or (gs.ndim(gs.array(b)) == 3)
+        axes = (1, 2) if is_vectorized else (0, 1)
+        return gs.all(gs.isclose(a, b, atol=atol), axes)
```
ninamiolane: does `gs.allclose` work? 

https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html

geomstats/geometry/general_linear.py
```diff
@@ -1,4 +1,9 @@
-"""The General Linear Group, i.e. the matrix group GL(n)."""
+"""
+This module exposes the `GeneralLinearGroup` class. 
+
+Note: This class is temporarily kept for compatibility purposes
+but will soon be renamed `GeneralLinear`.
```
ninamiolane: It is already renamed GeneralLinear, so we can delete this note.

opeltre:
@ythanwerdas you may want to look at the tiny changes and tell us if you'd like some more in this direction

Note I haven't fixed the `Matrices.belongs` method yet because it can get quite tedious to correct the tests so right now we still have `belongs : matrix -> [ [[bool]] ]`

This is fixed in SPD so that:
```python
belongs: matrix -> bool 
# or
belongs : [matrix] -> [bool]
```

geomstats/geometry/grassmannian.py
```diff
@@ -87,4 +83,4 @@ def exp(self, vector, point):
         """
         expm = gs.linalg.expm
```
ninamiolane: Do you want to make this `Matrices.expm` at some point? I have no strong opinion on this, it is also good not to re-implement numpy x)

geomstats/geometry/spd_matrices.py
```diff
@@ -21,19 +22,12 @@ def __init__(self, n):
...
-        belongs = mask_is_symmetric & mask_is_invertible
-        belongs = gs.to_ndarray(belongs, to_ndim=1)
-        belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
-        return belongs
+    def belongs(self, mat, atol=TOLERANCE):
+        """
+        Check if a matrix belongs to the manifold of
+        symmetric positive definite matrices.
+        """
+        return Matrices.is_symmetric(mat)
```
ninamiolane: yayy, this simplifies a lot indeed 🙏 !

tests/test_spd_matrices.py
```diff
@@ -33,19 +33,18 @@ def setUp(self):
...
-        expected = gs.array([[True]])
+        expected = gs.array(True)
         self.assertAllClose(result, expected)
 
     @geomstats.tests.np_and_tf_only
     def test_random_uniform_and_belongs_vectorization(self):
         """
-        Test that the random uniform method samples
-        on the hypersphere space.
+        same: fix belongs.
```
ninamiolane: Use `# FIXME` syntax, it will be easier to `grep` when we have "Fix it days"

geomstats/geometry/matrices.py
```diff
@@ -36,13 +39,36 @@ def mul(a, b):
...
+        an : array-like, shape=[n_samples, dim_n-1, dim_n]
+
+        Returns
+        -------
+        mul : array-like, shape=[n_samples, dim_1, dim_n]
+        """
+        return reduce(gs.matmul, args)
+
+    @classmethod
+    def commutator(cls, a, b):
```
ninamiolane: Can we use `mat_a`, `mat_b`? It makes automatic parsing and refactoring easier not to have single-letter variables.

Otherwise the use of `git grep -l 'a'` will return half of the words of the files

geomstats/geometry/matrices.py
```diff
@@ -67,40 +93,49 @@ def transpose(mat):
...
 
     @staticmethod
-    def make_symmetric(matrix):
-        """Make a matrix fully symmetric to avoid numerical issues."""
+    def vector_from_matrix(matrix):
+        """
+        Conversion function from (_, m, n) to (_, mn).
+        """
         matrix = gs.to_ndarray(matrix, to_ndim=3)
         n_mats, m, n = matrix.shape
```
ninamiolane: I know it's not your code, but maybe `dim_1`, `dim_2` instead of `m, n` will be more consistent with the naming of this file + avoid single letter variable

geomstats/geometry/matrices.py
opeltre: Not exactly, that's something @stefanheyder walked me through on the first day :)

`gs.all` would also test along the `n_samples` dimension and behave like: `[ mats ] -> True`

opeltre: BTW it's good to have a `[mats] -> [bool]` method so that we'll be able to add methods like: 
+ `keep_points = filter(belongs)` to filter out those outside 
+ `are_points = gs.all(belongs)` to end up with the same test as `gs.all([mats])`

geomstats/geometry/general_linear.py
opeltre: oops I got really mixed up in the rebase^^

geomstats/geometry/matrices.py
stefanheyder: Yes, if the input is vectorized, the output should also be a vector of booleans. `np.allclose` checks whether the arrays as a whole are equal, it is not vectorized.

tests/test_spd_matrices.py
opeltre: right! and belongs is already fixed...

geomstats/geometry/matrices.py
```diff
@@ -67,40 +93,49 @@ def transpose(mat):
         -------
         transpose : array-like, shape=[n_samples, dim, dim]
         """
-        is_vec = (gs.ndim(gs.array(mat)) == 3)
-        axes = (0, 2, 1) if is_vec else (1, 0)
+        is_vectorized = (gs.ndim(gs.array(mat)) == 3)
```
stefanheyder: this is the second time I see this in the PR, maybe consider refactoring into a method?

geomstats/geometry/spd_matrices.py
stefanheyder: Shouldn't this method also check whether mat is invertible?

geomstats/geometry/matrices.py
```diff
@@ -26,8 +28,9 @@ def belongs(self, point):
         return mat_dim_1 == self.m & mat_dim_2 == self.n
 
     @staticmethod
-    def mul(a, b):
-        """Return the product of matrices a and b.
+    def equal(a, b, atol=TOLERANCE):
```
stefanheyder: Agree with @ninamiolane, for consistency these should be called `mat_a` and `mat_b`

geomstats/geometry/matrices.py
```diff
@@ -36,13 +39,36 @@ def mul(a, b):
...
+        a1 : array-like, shape=[n_samples, dim_1, dim_2]
+        a2 : array-like, shape=[n_samples, dim_2, dim_3]
+        ...
+        an : array-like, shape=[n_samples, dim_n-1, dim_n]
+
+        Returns
+        -------
+        mul : array-like, shape=[n_samples, dim_1, dim_n]
+        """
+        return reduce(gs.matmul, args)
```
stefanheyder: Just a comment: I think this can be made much faster using `gs.einsum`, because with such an implementation you don't have to incrementally multiply all the matrices together. On the other hand I don't know you would do this to support any number of factors ...

geomstats/geometry/grassmannian.py
opeltre: Yes! 
In the end I think this ought to point to `GL.exp` -- or even to `SO.exp` if we figure out what numpy does and does not do :)

(in methods like `linalg.svd` for instance, I had to remove the `hermitian=True` flag because unsupported in numpy 1.16.  It's not clear what numpy `expm` does as 1) we had to fix vectorization and 2) we don't know what algorithm it uses when given a symmetric matrix, as `expm` does not accept such flags as arguments, same for `logm`. This is related to @ythanwerdas change of the algorithm for SPD and it would be nice to have some benchmarks! in the meantime it will be good to have all the `expm` calls point to one common place so we can update these more easily)

opeltre: BTW I'll come with updates for the `grassmannian.py` PR but after GL and SO

geomstats/geometry/matrices.py
opeltre: I also renamed the method to 'bracket' it will be more uniform with stefan's

geomstats/geometry/matrices.py
opeltre: Ok I can change these, but as per @stefan's remarks I could as well delete all these vector to matrix methods

opeltre: I do agree it's not the right place for them, and if ever needed we could start putting those in separate mappings files

geomstats/geometry/matrices.py
opeltre: what do you mean? an `is_vectorized` method?

opeltre: ... this might be a great idea, especially for unification of vectorization accross classes!  :smiley:

geomstats/geometry/matrices.py
opeltre: Hmm I guess you mean something like `gs.einsum('...ij, ...jk, ...kl->...il', a, b, c)`

opeltre: I think it's good to have it that way for now because we know `gs.matmul` vectorizes correctly

stefanheyder: For three matrices this works, but your current implementation even supports n matrices :) I am fine with this, if we need to go fast we can always add a special case for three or four matrices later on.

opeltre: The 'any number of factors' is just to make explicit the associativity of `mul` 
It gets more convenient and readable when you do matrix conjugation for instance: 
```
point2 = Mat.mul(Mat.exp(vec), point, Mat.exp(-vec))
```

stefanheyder: Yes I totally agree with you :) I just wanted to make the point that `gs.einsum` might be faster in multiplying for example four matrices (because you can skip the intermediate results), but it is not clear (at least to me) how one would implement it for an arbitrary number of factors.

Thus I prefer the solution you commited right now.

geomstats/geometry/matrices.py
ninamiolane: Yes! To be put in the `vectorization.py` in this case.

ninamiolane:
The travis error comes from the pip upgrade that happens ~2h ago, should be fix after PR #390 if you rebase on master @opeltre

geomstats/geometry/matrices.py
ninamiolane: Got it, yes the [mats] -> [bools] format would be better!

geomstats/geometry/grassmannian.py
ninamiolane: Numpy does not have expm, you're referring to scipy, right? Yes to have everything at a commong place for sure.

ninamiolane: Cool for GL, SO and the Grassmanians, but let's all agree on the inheritance structure first? cf #382

geomstats/geometry/matrices.py
ninamiolane: yes, `vectorization.py` 's goal is to factorize these. #379

geomstats/geometry/matrices.py
opeltre: Thanks for the idea stefan I really think it's good! 
I'll keep it like as it is for now but let's keep on thinking about it in #338

opeltre: This `class.base_shape` attribute would be especially usefull for handling higher dimensional vectorization, for instance when computing 4-D arrays of geodesics, e.g. returning `[n_samples, time, dim1, dim2]` shape

geomstats/geometry/grassmannian.py
opeltre: Hmm...  it's still not so clear to me. I'll write more on that in 382 soon, see also the GL PR to come.

geomstats/geometry/matrices.py
ninamiolane: Yes! Currently, it is provided by an object's attribute called `default_point_type` or `point_type`. I agree that having it as a class attribute would probably be better if default computations are always in a certain format for the whole class. Happy to rename it `base_shape` or `default_shape` (probably `base_ndim` though?) if it's more intuitive, but let's rename it everywhere in this case to avoid namings' complexity explosion 💥

geomstats/geometry/spd_matrices.py
opeltre: Yes! had noticed as well 
This will be fixed by calling `GL.belongs` from the next PR:
```
GL.belongs(mat) = gs.where(gs.linalg.det(mat) != 0, True, false) 
```

tests/test_spd_matrices.py
```diff
@@ -29,14 +29,14 @@ def setUp(self):
         self.metric_logeuclidean = SPDMetricLogEuclidean(n=self.n)
         self.n_samples = 4
 
-    @geomstats.tests.np_and_tf_only
+    @geomstats.tests.np_only
```
ninamiolane: Why doesn't it pass tf tests?

opeltre: shape mismatch? otherwise no idea, I think it should be fixed from the backend

opeltre: I'll do some tests of `gs.where`, `gs.linalg.det` etc and add a "#FIXME" as soon as it's figured out

ninamiolane: Yes, it relates to the vectorization issue... Would it be possible to fix it now? I know it's not fun, but otherwise, it's a step back in the code base as @ythanwerdas 's code was working before, and we should not take the habit of breaking existing unit tests if not really required?

ninamiolane:
Merging this as per our discussion, but future PRs should not break any existing unit tests. (I will fix this one in the meantime).

PR #390, opened 2020-01-21
Hot fix following pip upgrade

ninamiolane:
Recent pip upgrade breaks geomstats' travis. This is a hot fix, thanks to @nkoep .

https://github.com/pypa/pip/issues/7620


stefanheyder:
LGTM!

PR #391, opened 2020-01-21
Merge general linear

opeltre:
The `GeneralLinear` class to provide base methods for invertible matrices such as `exp`, `log`, `inv`, `orbit`... This has little to no effect on the existing code base as very few classes rely on it. 

Most collateral work was to fix tests from the `test_general_linear.py` file, which actually called nothing but `SO(3)` methods (sic).

This commit also fixes `SPD.belongs` which now calls `GL.belongs` to check invertibility. 

__Note__: @ninamiolane @stefanheyder @ythanwerdas 
+ right now `SPD` inherits from [EmbeddedManifold](https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/embedded_manifold.py) which does not provide anything but `raise "NotImplementedErrors"`
+ its `embedding_manifold` attribute is set to `GeneralLinear(n)`, and calls: 
    
    (i) `self.embedding_manifold.is_symmetric` and `self.embedding_manifold.make_symmetric` in the [vector_from_symmetric_matrix](https://github.com/geomstats/geomstats/blob/f29d609f21f64669751e16518b70d70e3c8fca8a/geomstats/geometry/spd_matrices.py#L38) method, which should be deleted or moved.
    
    (ii) `self.embedding_manifold.exp` in its [random_uniform](https://github.com/geomstats/geomstats/blob/f29d609f21f64669751e16518b70d70e3c8fca8a/geomstats/geometry/spd_matrices.py#L75) method, which actually (1) samples uniformly in `Matrices(n)` (2) implements inline an unconsistently vectorized form `Matrices.make_symmetric`, and (3) exponentiates the obtained symmetric matrices to SPD via its `embedding_manifold` 

Point (i) shows how the present implementation is convoluted: `is_symmetric` and `make_symmetric` are actually not implemented in `embedding_manifold = GL(n)` but inherited from `Mat(n,n)`!  (it was also the case _before_ my matrix PR)

Point (ii) illustrates well the kind of problems we wish to tackle by this refactoring. Methods like `random_uniform` are bug prone because they provide an inline implementation of `a => ( a + t(a))/2` or something similar. They shall one day yield broadcast errors that will be very tedious to track, and in the long run unnecessary code will be impossible to maintain.


geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
+            return logm(point)
+        else:
+            return logm(cls.mul(point, cls.inv(base_point)))
 
-    def exp_not_from_identity(
-            self, tangent_vec, base_point, point_type=None):
-        """Compute group exponential from a base point.
+    @classmethod
+    def orbit(cls, point, base_point=None):
+        r"""
```
ninamiolane: r""" ?

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
-        point_near_identity = self.compose(
-            self.inverse(base_point), point)
-
-        log_from_identity = self.log_from_identity(
-            point_near_identity)
+        vector = cls.log(point, base_point)
 
-        group_log = self.compose(
-            base_point, log_from_identity)
+        def path(time):
```
ninamiolane: maybe `point_on_orbit` to micmic the geodesic's API? https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/riemannian_metric.py#L261

opeltre:
Also, I do not think it is good that the `EmbeddedManifold` attribute

geomstats/geometry/general_linear.py
ninamiolane: (or change geodesic's API to `path`, whatever is more understandable)

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
+            vecs = gs.einsum('t,...ij->...tij', time, vector)
```
ninamiolane: We need to check if ellipsis work in tf and pytorch (later PR probably)

geomstats/geometry/matrices.py
```diff
@@ -26,34 +28,57 @@ def belongs(self, point):
...
         ----------
-        a : array-like, shape=[n_samples, dim1, dim2]
-        b : array-like, shape=[n_samples, dim2, dim3]
+        mat_a : array-like, shape=[n_samples, dim1, dim2]
+        mat_b : array-like, shape=[n_samples, dim2, dim3]
 
         Returns
         -------
-        c : array-like, shape=[n_samples, dim1, dim3]
+        eq : array-like boolean, shape=[n_samples]
```
ninamiolane: Rebase on master, so that we do not see the changes of the other PR.

tests/test_general_linear.py
```diff
@@ -9,105 +9,56 @@
...
-        # 1. Composition by identity, on the right
-        # Expect the original transformation
-        rot_vec = gs.array([0.2, -0.1, 0.1])
-        mat = self.so3_group.matrix_from_rotation_vector(rot_vec)
-
-        result = self.group.compose(mat, self.group.identity)
-        expected = mat
-        expected = helper.to_matrix(mat)
-
+    def test_compose_and_identity(self):
```
ninamiolane: this is `test_compose`

tests/test_general_linear.py
```diff
@@ -9,105 +9,56 @@
...
+        result = self.group.inv(mat)
         expected = 1. / 3. * gs.array([
             [-2., -4., 3.],
             [-2., 11., -6.],
             [3., -6., 3.]])
-        expected = helper.to_matrix(expected)
-
-        self.assertAllClose(result, expected)
-
-    def test_compose_and_inverse(self):
```
ninamiolane: You removed this in order to have unit tests that are really "unit"?

geomstats/geometry/matrices.py
```diff
@@ -67,40 +92,40 @@ def transpose(mat):
...
         transpose : array-like, shape=[n_samples, dim, dim]
         """
-        is_vec = (gs.ndim(gs.array(mat)) == 3)
-        axes = (0, 2, 1) if is_vec else (1, 0)
+        is_vectorized = (gs.ndim(gs.array(mat)) == 3)
+        axes = (0, 2, 1) if is_vectorized else (1, 0)
         return gs.transpose(mat, axes)
 
-    @staticmethod
-    def vector_from_matrix(matrix):
```
ninamiolane: I agree that this was not the most useful method, but is there a specific reason why you deleted it?

tests/test_matrices.py
```diff
@@ -74,14 +74,14 @@ def test_is_symmetric(self):
         sym_mat = gs.array([[1., 2.],
                             [2., 1.]])
         result = self.space.is_symmetric(sym_mat)
-        expected = gs.array([[True]])
+        expected = gs.array(True)
```
ninamiolane: These weird shapes were enforced in order to have vectorized functions. Since you are changing them, could you add 1 or 2 tests to see if your functions are vectorized? (they are probably, as they rely on numpy). Thanks!

ninamiolane:
About your PR description: 
- the `raise NotImplementedError` is the syntax used for abstract classes (https://stackoverflow.com/questions/1151212/equivalent-of-notimplementederror-for-fields-in-python) although apparently this is the python 2.X syntax, and not the 3.X syntax, so we should indeed change it.
- SPD(n) is indeed embedded in GL(n): do you think it is a problem that its attribute `embedding_space` is GL(n) instead of M(n)? What would be the pros/cons of having M(n) instead of GL(n)?

Thanks for noticing all the other problems! Could you create corresponding issues? (one per problem) This way they will not get forgotten and it will create incentive for us to tackle them asap.

geomstats/geometry/general_linear.py
opeltre: flake8 told me to
(for the tex code and \ I think!)

geomstats/geometry/general_linear.py
opeltre: Hm yes I've seen this, a bit obscure don't you think? 
I would expect `point_on_orbit` to test whether point is in the orbit

geomstats/geometry/general_linear.py
opeltre: on this subject, I looked for the pytorch einsum docs which seem ok:
 https://pytorch.org/docs/stable/torch.html#torch.einsum
there's definitely some testing to do!

geomstats/geometry/matrices.py
opeltre: think I managed :)

tests/test_general_linear.py
opeltre: yes there were many tests to fix so I removed those that seemed redundant!

opeltre: I might have been a bit quick doing so, so let me know if you think something's missing...
I should ask one for `orbit` BTW

tests/test_matrices.py
opeltre: done :)

geomstats/geometry/matrices.py
opeltre: Yes, this was guided by what we discussed with stefan about manifold classes handling only one kind of representation for points and tangent vectors...

I think it will be less misleading for users as maintainers, so that we never have to wonder what kind of shape a specific method expects

opeltre: There's plenty of similar code in other matrix classes, for instance in SPD

opeltre: there's even a `symmetric_matrix_from_vector` using for loops

opeltre: I also did a 'search:vector from matrix' accross the repo and such methods clearly don't seem to be used anywhere, apart from some tests. Personnally I would delete all of them so that we feel lighter :)

opeltre: ok, so after a check such conversion methods are all absent in 'matrices.py' and 'general_linear.py'. I suggest we deal with SPD in another PR later, this guy is bigger ^^

opeltre: and if we ever want those back it'll be easy to reimplement them properly vectorized, by calling `gs.reshape` with the correct `axes` argument passed

geomstats/geometry/general_linear.py
opeltre: just renamed this everywhere :)
btw doing so I noticed that `SPD.geodesic` does not behave as expected and has prototype:
```
SPD.geodesic(point, vector) => path
```
I'll put this in separate issues as you suggested, just to mention. Do we agree that this should be either:
+ `geodesic(point2, base_point) -> path`
+ `geodesic(vector, base_point) -> path` ?

opeltre: Clearly not the time,  but we might want to think about having two names, like `interpolate` vs `shoot` someday...
If you have any idea ^^

opeltre:
> SPD(n) is indeed embedded in GL(n): do you think it is a problem that its attribute embedding_space is GL(n) instead of M(n)? What would be the pros/cons of having M(n) instead of GL(n)?

I'll create those issues so that we can talk more on this, thanks to all your challenging comments I've done some thinking :)

For me we still have to figure out what this `EmbeddedManifold` class would be supposed to do.

In short, keeping any code considerations away for now, I think that "embedded manifold" from differential geometry textbooks is often used in first chapters to define "submanifolds of R^n", and this notion is useful because it allows the student to carry out calculations in R^n before introducing atlases, charts, etc.

Then an "embedding" of manifolds clearly means something more general, i.e. a map whose tangent map is everywhere injective, but this is just terminology to characterise some smooth maps in C(M,N) but it cannot be generically very useful...

So I think that if we do figure out a way to make this parent class useful, it would be by leveraging on the linear structure of an ambient vector space (as we already do by manipulating np arrays).

opeltre:
this breaks SO3 tests :sob:

opeltre:
good to go now, but note that SO3 ran 96 tests in 20s getting my fan crazy!

tests/test_spd_matrices.py
```diff
@@ -78,7 +87,7 @@ def vector_and_symmetric_matrix_vectorization(self):
 
         self.assertTrue(gs.allclose(result, expected))
 
-    @geomstats.tests.np_and_tf_only
+    @geomstats.tests.np_and_pytorch_only
```
ninamiolane: why the change to pytorch there?

tests/test_spd_matrices.py
```diff
@@ -250,7 +259,7 @@ def test_log_euclidean_inner_product(self):
 
         self.assertAllClose(result, expected)
 
-    @geomstats.tests.np_and_tf_only
+    @geomstats.tests.np_only
```
ninamiolane: Let's not break tests in other backends, what was the issue?

geomstats/geometry/special_orthogonal.py
```diff
@@ -615,7 +615,7 @@ def rotation_vector_from_matrix(self, rot_mat):
 
             rot_vec *= (1. + fact)
         else:
-            skew_mat = self.embedding_manifold.log_from_identity(rot_mat)
+            skew_mat = self.embedding_manifold.log(rot_mat)
```
ninamiolane: 🎉 shorter is nice!

geomstats/geometry/spd_matrices.py
```diff
@@ -929,20 +931,17 @@ def log(self, point, base_point):
...
-        geodesic : callable
+        path : callable
+            The time parameterized geodesic.
         """
-        def point_on_geodesic(t):
+        def path(t):
             return self.exp(t*initial_tangent_vec, initial_point)
 
-        return point_on_geodesic
+        return path
```
ninamiolane: Yes, `path` is more intuitive!

geomstats/geometry/spd_matrices.py
```diff
@@ -22,12 +22,14 @@ def __init__(self, n):
...
-    def belongs(self, mat, atol=TOLERANCE):
-        """Check if a matrix belongs to the manifold.
-
-        Check if a matrix belongs to the manifold of symmetric positive
-        definite matrices.
-        """
+    @staticmethod
+    def belongs(mat, atol=TOLERANCE):
+        """Check if a matrix is symmetric and invertible."""
+        # FIXME: check positivity, implying invertibility (Yann)
```
ninamiolane: If you want to attribute the FIXME to someone, use the syntax: 
```
# TODO(ythanwerdas): ...
```
Although, it's better practice to leave TODOs to yourself rather than to others.

TODOs and FIXMEs are automatically parsed by pycharm

geomstats/geometry/spd_matrices.py
```diff
@@ -22,12 +22,14 @@ def __init__(self, n):
...
+        #
+        # note : vectorized "and" on numpy works with:
+        #       [bool] * [bool] -> bool
+        # but does not on tf.
```
ninamiolane: What do you recommend there? We'll tackle this with decorators?

tests/test_general_linear.py
```diff
@@ -125,31 +86,43 @@ def test_group_exp_vectorization(self):
...
+            [gs.exp(2.), 0.],
+            [0., gs.exp(1.)]])
+        idty = GeneralLinear(2).identity()
+
+        path = GeneralLinear(2).orbit(point)
+        time = gs.linspace(0., 1., 3)
+
+        result = path(time)
+        expected = gs.array([idty, sqrt, point])
+        self.assertAllClose(result, expected)
```
ninamiolane: Nice! And it works in tf too 🎉

tests/test_spd_matrices.py
```diff
@@ -96,7 +105,7 @@ def test_differential_power(self):
                               [1/3, .125, .125]]])
         self.assertAllClose(result, expected)
 
-    @geomstats.tests.np_and_tf_only
+    @geomstats.tests.np_and_pytorch_only
```
ninamiolane: same

geomstats/geometry/matrices.py
ninamiolane: Yes, agree, we can delete it, especially if it's not used, and this would be a good opportunity to rewrite it better later.

I think manifolds should have different representations, but with one unique default representation so that the users do not worry indeed.

ninamiolane:
And yes, cleaning SO3 tests would be great for our personal laptops to love us back

geomstats/geometry/spd_matrices.py
opeltre: ok thanks :)

tests/test_general_linear.py
opeltre: yes! :smiley:

geomstats/geometry/spd_matrices.py
opeltre: We should probably provide a `gs.and` and `gs.or` in the backend for these kind of tests...
Actually I was surprised not to find those in numpy until I realized that `+` and `*` did the job!  

Only thing is to figure how tf and torch do this: if they really don't like booleans, we'll probably have to return 1. and 0. floats...

geomstats/geometry/spd_matrices.py
opeltre: does `# TODO(ythanwerdas, opeltre): ...` work ?

tests/test_spd_matrices.py
opeltre: Theses tests yied 'shape mismatch' I can't figure out...
I'm really surprised as these functions shouldn't be affected by the commit

opeltre:
tf tests removed:
+ differential_power
+ inverse_differential_power
+ log_and_exp_affine_invariant

opeltre:
Apparently moving to tensorflow 2.0 fixed it, though I really have no idea why 1.15 had so much trouble running the tests

geomstats/geometry/connection.py
```diff
@@ -301,7 +300,7 @@ def geodesic(self, initial_point,
         initial_tangent_vec = gs.to_ndarray(initial_tangent_vec,
                                             to_ndim=point_ndim+1)
 
-        def point_on_geodesic(t):
+        def path(t):
```
stefanheyder: Better naming :tada:

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
+    def log(cls, point, base_point=None):
         """
-        mat_a = gs.to_ndarray(mat_a, to_ndim=3)
-        mat_b = gs.to_ndarray(mat_b, to_ndim=3)
-        composition = gs.einsum('nij,njk->nik', mat_a, mat_b)
-        return composition
-
-    def inverse(self, mat):
-        """Compute matrix inverse.
+        Search for a left-invariant vector field bringing base_point to point.
```
stefanheyder: Maybe use "calculate" instead of "search"?

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
-        mat
+        point :         array-like, shape=[..., n, n]
+        base_point :    array-like, shape=[..., n, n]
+            Defaults to identity.
 
         Returns
         -------
-        inverse
+        vector :        array-like, shape=[..., n, n]
+            A matrix such that `exp(vector, base_point) = point`.
```
stefanheyder: `tangent_vec` ?

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
+        Denoting `point` by :math: `g` and `base_point` by :math: `h`,
+        the orbit :math: `\gamma` satisfies:
 
-        exp_from_identity = self.exp_from_identity(
-                tangent_vec_at_identity)
+        .. math::
 
-        group_exp = self.compose(
-            base_point, exp_from_identity)
+            \gamma(t) = {\mathrm e}^{t X} \cdot h \\
```
stefanheyder: :tada:

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
-    """Class for the General Linear Group, i.e. the matrix group GL(n).
-
-    Note: The default representation for elements of GL(n)
-    are matrices. For now, SO(n) and SE(n) elements are represented
-    by a vector by default.
-    """
+class GeneralLinear(Matrices):
+    """Class for the general linear group GL(n)."""
 
     def __init__(self, n):
```
ninamiolane: You do not instantiate from the parent class here?

ninamiolane: why did you delete the assert?

geomstats/geometry/general_linear.py
```diff
@@ -1,192 +1,125 @@
...
-        Returns
-        -------
-        identity : TODO
+        self.n = n
+
+    @staticmethod
+    def belongs(point):
+        """Test if a matrix is invertible."""
+        det = gs.linalg.det(point)
+        return gs.where(det != 0., True, False)
```
ninamiolane: I wonder if we should use something like ` not gs.isclose(det, 0.)`, in order to be flexible with respect to numerical imprecisions (if the determinant is 1e-5, should we consider it is 0.?)

geomstats/geometry/general_linear.py
opeltre: Good question, I'll restore `Matrices.__init__(self, n, n)` so that the metric is set to the Frobenius metric 
`assert` will then be redundant as matrices already check n > 0, but this not proper error catching either btw... nothing is printed!

opeltre: I checked the `lie_group.py` file but it does not seem a good idea to call its `__init__` method: this would override the metric to `InvariantMetric(...)` which is not defined

geomstats/geometry/general_linear.py
opeltre: I guess we could take the benefit of avoiding where!
Matrices should be invertible as soon `det != 0.` though.

opeltre: We don't have a vectorized working `not`  :( 
We have to add logical operations to the backend...

ninamiolane:
I fixed the conflicts, but flake8 is failing.

ninamiolane:
Ready to merge! Just need to solve the remaining conflicts + flake8 problem.

ninamiolane:
I don't understand why there are still conflicts here? 🤔 "This branch cannot be rebased due to conflicts"

PR #393, opened 2020-01-22
Add flake8 docstrings for geomstats/geometry

ninamiolane:
This PR runs flake8 with docstrings on geomstats/geometry that has been cleaned by @cshewmake2 .

It is a step towards PR #319 where we will run flake8 with docstrings on the whole codebase.


ninamiolane:
@stefanheyder I'd like to use flake8 on docstrings in `geomstats/geometry`, which has been cleaned by Christian - but not on the other folders yet. 

When `git diff --name-only master HEAD geomstats/geometry` outputs nothing, the whole codebase is being run through flake8, which led me to this modification of .travis.yml.

PR #396, opened 2020-01-23
Gen api ref

cshewmake2:
Check the commit message.
Essentially a rewrite of the api-reference.rst and supporting pages and several error corrections.

Next we need to fix up and complete all of the method docstrings which require further details.


cshewmake2:
**NOTE** If you try to use sphinx-build from inside the docs directory, a FileNotFound error occurs for ~ bch_coefficients.dat. This is because of the way that we currently read in the ~ bch_coefficients.dat file inside one of the geometry modules (Lie Algebra/Group?). 

Essentially this is related to how sphinx sets the path for its work. We can either change the way we import the .dat file, or you can modify how you run sphinx (of course, other ways may exist as well)

A quick change in how we call Sphinx avoids the issue:
cd into the **root directory** for the project and run
`sphinx-build -b html docs/ docs/html`

You can open and sample the generated files using
`open docs/html/index.html`

docs/geometry.rst
```diff
@@ -0,0 +1,132 @@
...
+----------------
+.. automodule:: geomstats.geometry.stiefel
+   :members:
+
+Lie Groups
+----------
+.. automodule:: geomstats.geometry.lie_group
+   :members:
+
+Lie Algebra
```
nguigs: Does it make sense to put Lie Algebra (and Matrices space, and Skew Symmetric Matrices) in Manifolds?

geomstats/geometry/stiefel.py
```diff
@@ -94,8 +94,8 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point):
 
         References
         ----------
-        .. [1] Formula from:
-        http://noodle.med.yale.edu/hdtag/notes/steifel_notes.pdf
+        .. [sn] Formula from:
+          http://noodle.med.yale.edu/hdtag/notes/steifel_notes.pdf
```
nkoep: Web links like this are generally discouraged in docstrings. Maybe we can find a proper reference.

ninamiolane: Right. For formulas (like here), we can write the formula in latex.
For references, we can add the full reference, with the link to the .pdf which is less likely to get deprecated than just a weblink.

docs/conf.py
```diff
@@ -14,9 +14,18 @@
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
     'sphinx.ext.githubpages',
+    'sphinx.ext.napoleon', # enable numpy or google docstrings
```
ninamiolane: Nice!

docs/geometry.rst
ninamiolane: Good question. Since the doc is generated automatically, did it put it there because of our inheritance structure?

If we want to change the API doc, we probably need to change the code first.

docs/conf.py
```diff
@@ -14,9 +14,18 @@
...
 ]
 
-templates_path = ['templates']
+# # Choose numpy docstring
```
ninamiolane: One # is maybe enough?

ninamiolane:
Only NITs: good to go as soon as it passes flake8 on travis. Also, could you add a TODO so that we replace the pdf weblink in stiefel with the right formula? (TODO(ninamiolane) works).

Thank you for this 🎉 ! We will finally have a clean doc!

geomstats/geometry/stiefel.py
nkoep: [This](https://epubs.siam.org/doi/abs/10.1137/16M1074485?journalCode=sjmael) one ought to do for instance (cf. the equation right after (2.1) in the [preprint](https://arxiv.org/pdf/1604.05054.pdf)).

docs/geometry.rst
cshewmake2: I did that. I wasn't sure what the best categorization scheme was for the geometry section. Given that they all inherit Manifold, I put them in that section for now; please update as you see fit.

I wish those were generated automatically; perhaps in the future they will be.

geomstats/geometry/stiefel.py
cshewmake2: Yes I agree. This will be one of our big todo's in the docstring cleanup.

docs/conf.py
cshewmake2: Thanks for catching that.

geomstats/geometry/stiefel.py
cshewmake2: Is this better?
image

ninamiolane: 🎉 yay!

PR #411, opened 2020-01-29
Add Beta distributions

nguigs:
- Add the manifold of beta distributions (upper-right quadrant of the plane) with the Fisher information metric
- The LeviCivitaConnection are removed, and methods therein are now in RiemannianMetric which inherits from Connection. This allows to use the Christoffel symbols computed from the metric matrix to compute geodesics.
- Integrators are put in a separate file and called by `connection.exp`. I added rk4 as integration scheme.


geomstats/integrator.py
```diff
@@ -0,0 +1,62 @@
+def _symplectic_euler_step(state, force, dt):
```
ninamiolane: Great!! Maybe add a doctring as a title of this file?

geomstats/integrator.py
```diff
@@ -0,0 +1,62 @@
...
+    point_new
+    vector_new
+    """
+    point, vector = state
+    point_new = point + vector * dt
+    vector_new = vector + force(point, vector) * dt
+    return point_new, vector_new
+
+
+def rk4_step(state, force, dt, k1=None):
```
ninamiolane: what is rk4?

tests/test_beta.py
```diff
@@ -0,0 +1,64 @@
...
+        n_samples = self.n_samples
+        gs.random.seed(123)
+        base_point = self.beta.random_uniform(n_samples=n_samples, bound=5)
+        point = self.beta.random_uniform(n_samples=n_samples, bound=5)
+        log = self.metric.log(point, base_point, n_steps=500)
+        expected = point
+        result = self.metric.exp(tangent_vec=log, base_point=base_point)
+        self.assertAllClose(result, expected, rtol=1e-2)
+
+    def test_christoffels_vectorization(self):
```
ninamiolane: It would be nice to also test the function (not only the shape output when used vectorized).

ninamiolane:
test_beta.py -> test_beta_distributions.py
or
beta_distributions.py -> beta.py

so that the file and its tests have the same name

geomstats/geometry/beta_distributions.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class BetaMetric(RiemannianMetric):
+    """Class for the Fisher information metric on beta distributions."""
+
+    def __init__(self):
+        super(RiemannianMetric, self).__init__(dimension=2)
+
+    @staticmethod
+    def detg(param_a, param_b):
```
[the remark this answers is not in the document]
nguigs: This might be misleading as it is not the general linalg.det function, maybe metric_det?

geomstats/geometry/beta_distributions.py
```diff
@@ -0,0 +1,292 @@
...
+        Returns
+        -------
+        base_point : array-like, shape=[n_samples, 2, 2]
+        """
+        assert ~ (base_point is None), 'The metric depends on the base point'
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
+        matrices = []
+        for point in base_point:
+            param_a, param_b = point
+            g0 = gs.array(
```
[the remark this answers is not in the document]
nguigs: for future autograd differentiation, we cannot use completion with indexes, so there is no intuition behind these other that g0 is the first line and g1 the second...

tests/test_beta.py
nguigs: How? I can't really compute them by hand...

nguigs: I added a test for the exponential (we know that the first diagonal is a geodesic). This is a higher-level method that uses `christoffels` but we (@AliceLeBrigant and I) don't have any other idea to test whether the Christoffel symbols themselves are right.

PR #420, opened 2020-02-05
Remove redundant code of tangent pca in the RiemannianMetric class

ninamiolane:
Tangent PCA algorithm was present twice: 
- as a method of the class RiemannianMetric
- as an algorithm in the folder geomstats/learning

This PR removes the first one and cleans docstrings.

### DIFF
--- a/geomstats/learning/pca.py
+++ b/geomstats/learning/pca.py
@@ -107,56 +107,81 @@ class TangentPCA(_BasePCA):
         self.iterated_power = iterated_power
         self.random_state = random_state
 
-    def fit(self, X, point_type='vector', y=None):
+    def fit(self, X, y=None, base_point=None, point_type='vector'):
         """Fit the model with X.
 
         Parameters
         ----------
-        X : array-like, shape (n_samples, n_features)
+        X : array-like, shape=[n_samples, n_features]
             Training data, where n_samples is the number of samples
             and n_features is the number of features.
-        point_type :
-
-        y : Ignored
+        y : Ignored (Compliance with scikit-learn interface)
+        base_point : array-like, shape=[n_samples, n_features]
+            Point at which to perform the tangent PCA
+            Optional, default to Frechet mean if None
+        point_type : str, {'vector', 'matrix'}
+            Optional
 
         Returns
         -------
         self : object
             Returns the instance itself.
         """
-        self._fit(X, point_type)
+        self._fit(X, base_point=base_point, point_type=point_type)
         return self
 
-    def fit_transform(self, X, point_type='vector', y=None):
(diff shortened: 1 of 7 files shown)


geomstats/learning/pca.py
```diff
@@ -108,15 +107,15 @@ def __init__(self, metric, n_components=None, copy=True,
         self.iterated_power = iterated_power
         self.random_state = random_state
 
-    def fit(self, X,
```
stefanheyder: You removed the `base_point` argument (which defaults to the Fréchet mean), I assume this is because tPCA doesn't really make sense if you do not look at the Fréchet mean? But maybe its reasonable to keep it as an argument so that the user can supply the (already precalculated) Fréchet mean. Or is the computation time for the mean very small compared to the whole tPCA?

geomstats/learning/pca.py
```diff
@@ -108,15 +107,15 @@ def __init__(self, metric, n_components=None, copy=True,
...
-            base_point=None, point_type='vector', y=None):
+    def fit(self, X, point_type='vector', y=None):
         """Fit the model with X.
 
         Parameters
         ----------
         X : array-like, shape (n_samples, n_features)
             Training data, where n_samples is the number of samples
             and n_features is the number of features.
+        point_type :
```
stefanheyder: Documentation :tada:

geomstats/learning/pca.py
```diff
@@ -108,15 +107,15 @@ def __init__(self, metric, n_components=None, copy=True,
         self.iterated_power = iterated_power
         self.random_state = random_state
 
-    def fit(self, X,
-            base_point=None, point_type='vector', y=None):
```
stefanheyder: Just to be sure: The `y` argument ist to comply with the scikit-learn interface to PCA?

nguigs: Actually, if we want this to be compatible with sickit-learn (and the Pipeline module in particular, for which `pipeline.fit(X, y)` propagates through all the steps of the pipeline without specifying `y=y`), I think `y=None` needs to be the second argument of `fit`.

geomstats/learning/pca.py
ninamiolane: You're right, the Frechet mean computation being an iterative algorithm that can take time, it is good to conserve this argument. Putting it back.

geomstats/learning/pca.py
ninamiolane: Good eye, thanks. I have added it.

geomstats/learning/pca.py
ninamiolane: Yes, it is for compliance. Thanks @nguigs for the arguments order, I have put y as second parameter now.

PR #425, opened 2020-02-07
Documentation for embedded_manifold, hypersphere, hyperbolic, product_*, stiefel

ninamiolane:
Clean documentation in files
- embedded_manifold.py
- hypersphere.py
- hyperbolic.py
- product_manifold.py
- product_riemannian_metric.py
- stiefel.py

in order to address issue #398 @cshewmake2

Issue #398, opened 2020-01-23
Resolve outstanding docstring issues


cshewmake2:
Christian Shewmake
January 23, 2020

**Hello geomstats wranglers!**

Thanks for your patience while I was putting together v1 of the API documentation. I think the GeomStats API reference page is going to be totally awesome using these new tools/standards. However, there’s quite a lot of work remaining before we can take it live.

**First, let’s restrict our attention to the geometry section.**
There are a couple of design decisions to make which should be discussed:
1. _Succinct vs Verbose Docstrings_
    1. How do we want to handle small, simple functions?
		-> It’s painful to annotate Descriptions, Parameters, and Returns for these functions.

2. _Singular or Plural Input Language_
    1. How do we want to handle docstrings for vectorized operations?
		-> I vote singular language.
		-> How do we annotate whether or not something supports vectorization?

**Next, there are outstanding issues in each of the files related to:**

3. _Absent or Unclear descriptions and parameter types_
    1. Some summaries and descriptions aren’t very useful / descriptive.
    2. Many ‘Parameters’ and ‘Returns’ entries lack type annotations

4. _Inconsistent Referencing_
    1. Some docstrings have weblinks which could potentially expire over the lifetime of the project
    2. Some docstrings don’t use a references section
		-> We need citations in the format you’d use in a paper (accompanied by links to paper pdfs would be nice)

5. _Inconsistent Math Formatting_
    1. Some docstrings use LaTeX inline or in sections via “:math:`eqn`” or “.. math:: eqn”, respectively.
    2. Some docstrings don’t use these options at all
    3. Some docstrings use the ‘math’ directives but have syntax errors (from rst or LaTeX or both) and don’t render properly.
--------
**Given that there are so many files to address, I think we should divide them up, resolve the issues, then merge them in piece-by-piece.** 
I’ve gone through some of these fixes myself, so I am happy to answer any and all questions related to syntax and our internal ‘style guide’. We can also discuss items for which you’re unsure once you raise a PR; I would be happy to review and give prompt feedback on these.

opeltre:
Could you edit the post with a link to the documentation web page? :)

ninamiolane:
From @nguigs : 
"We should divide files to each look at. I will take care of
connection
riemannian_metric
landmarks_space
"

stefanheyder:
I will take care of 

```
geometry/lie_algebra.py
geometry/lie_group.py
geometry/special_euclidean.py
geometry/speical_orthogonal.py
[...]

PR #428, opened 2020-02-07
Doc/more documentation

stefanheyder:
Some more documentation for `{euclidean,matrices,grassmannian}.py`.


stefanheyder:
See also #398.

geomstats/geometry/matrices.py
```diff
@@ -22,15 +22,24 @@ def __init__(self, m, n):
         self.metric = MatricesMetric(m, n)
 
     def belongs(self, point):
-        """Check if point belongs to the Matrix space."""
+        """Check if point belongs to the Matrix space.
```
ninamiolane: "Evaluate if a point ..." to be consistent with other docstrings? :/ Or should we put "Check" everywhere? It's also a detail.

geomstats/geometry/matrices.py
```diff
@@ -22,15 +22,24 @@ def __init__(self, m, n):
...
+
+
+        Parameters
+        ----------
+        point : array-like, shape=[n_samples, (m,n)]
```
ninamiolane: Descriptions?

geomstats/geometry/matrices.py
```diff
@@ -48,8 +57,7 @@ def equal(mat_a, mat_b, atol=TOLERANCE):
 
     @staticmethod
     def mul(*args):
-        """
-        Return the product of matrices a1, ..., an.
+        """Calculate the product of matrices a1, ..., an.
```
ninamiolane: Should we use "Calculate"/ "Compute" to describe the functions? I had deleted them all in my PR, to save space and also bcs I thought it was not essential, but happy to put them back everywhere.

geomstats/geometry/matrices.py
```diff
@@ -128,7 +133,20 @@ def make_symmetric(cls, mat):
...
-        """Generate n samples from a uniform distribution."""
+        """Generate n samples from a uniform distribution.
+
+        Each entry in the matrix is drawn from a uniform distribution with
+        support [0,1]
+
+        Parameters
+        ----------
+        n_samples : int
+            amount of samples to generate
```
ninamiolane: Should the first letter be upper-case or lower-case? I had used upper case, but happy to change my PR!

geomstats/geometry/matrices.py
```diff
@@ -143,7 +161,19 @@ def __init__(self, m, n):
...
+        Parameters
+        ----------
+        tangent_vec_a : array-like, shape=[n_samples, m, n]
+        tangent_vec_b : array-like, shape=[n_samples, m, n]
+        base_point : array-like, shape=[n_samples, m, n], optional
+
+
+        Returns
+        -------
+        inner_prod : the Frobenius inner product of a and b
```
ninamiolane: array-like, shape=...

geomstats/geometry/matrices.py
```diff
@@ -143,7 +161,19 @@ def __init__(self, m, n):
...
 
     def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
-        """Compute Frobenius inner product of two tan vecs at `base_point`."""
+        """Compute Frobenius inner product of two tan vecs at `base_point`.
+
+        Parameters
+        ----------
+        tangent_vec_a : array-like, shape=[n_samples, m, n]
+        tangent_vec_b : array-like, shape=[n_samples, m, n]
+        base_point : array-like, shape=[n_samples, m, n], optional
```
ninamiolane: Descriptions? You can copy paste from PR #425 , I have docstrings for inner products there.

ninamiolane:
Flake8 needs to pass

geomstats/geometry/matrices.py
ninamiolane: I think `Test if a point...` would be best, is has a boolean feel to it.

ninamiolane:
According to https://numpydoc.readthedocs.io/en/latest/format.html (thanks @cshewmake2 ) and follow-up conversation on slack:

Parameters' description starts with uppercase, ends with a .
optional is on the same line as the parameters' type
spaces around :

ninamiolane:
@stefanheyder any update on this?

PR #444, opened 2020-02-13
Fix docstrings

Hzaatiti:
Fixed some docstrings for geometry (mobius_add, math latex render) and learning (k_means)


geomstats/learning/kmeans.py
```diff
@@ -8,10 +9,18 @@
...
+    -------
+    Available example on the Poincaré Ball and Hypersphere manifolds
+    :mod:`examples.plot_kmeans_manifolds`
+
+    """
 
     def __init__(self, riemannian_metric, n_clusters=8, init='random',
                  tol=1e-2, mean_method='default', verbose=0):
-        """ K-Means algorithm using Riemannian manifolds
+        """k-means algorithm using Riemannian manifolds.
```
ninamiolane: Actually, there should be no docstring in the constructors:
see https://numpydoc.readthedocs.io/en/latest/format.html for example.

Could you fix this also, since we're on it?

geomstats/learning/kmeans.py
```diff
@@ -1,3 +1,4 @@
+"""k-means clustering."""
```
ninamiolane: """K-means clustering.""" ?

geomstats/learning/kmeans.py
ninamiolane: """K-means

PR #455, opened 2020-02-25
Return trajectories in Pole Ladder

nguigs:
Add options to return the geodesic parallelograms of the pole ladder construction. Example to visualize them will come in a separate PR.
Also fix the docstrings in connection.


examples/plot_pole_ladder_S2.py
```diff
@@ -0,0 +1,56 @@
+"""Plot the pole ladder scheme for parallel transport on S2.
+
+Sample a point on S2 and two tangent vectors to transport one along the
+other.
+
```
ninamiolane: Delete blank line?

examples/plot_pole_ladder_S2.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import matplotlib.pyplot as plt
+
+import geomstats.backend as gs
+import geomstats.visualization as visualization
+from geomstats.geometry.hypersphere import Hypersphere
+
```
ninamiolane: Nice! Could you use the `def main` syntax for the example:
```
if __name__ == "__main__":
    main()
```

to be consistent with other examples?

examples/plot_pole_ladder_S2.py
```diff
@@ -0,0 +1,56 @@
...
+t_main = gs.linspace(0, 1, n_points * 4)
+for points in trajectory:
+    main_geodesic, diagonal, final_geodesic = points
+    sphere_visu.draw_points(ax, main_geodesic(t_main), marker='o', c='b', s=2)
+    sphere_visu.draw_points(ax, diagonal(-t), marker='o', c='r', s=2)
+    sphere_visu.draw_points(ax, diagonal(t), marker='o', c='r', s=2)
+    sphere_visu.draw_points(ax, final_geodesic(-t), marker='o', c='g', s=2)
+    sphere_visu.draw_points(ax, final_geodesic(t), marker='o', c='g', s=2)
+
+V = gs.concatenate([tan_b / n_steps, tan_a, pole_ladder])
```
ninamiolane: Is this uppercase needed? maybe write a lower case more explanatory name?

geomstats/geometry/connection.py
```diff
@@ -136,27 +156,39 @@ def objective_with_grad(velocity):
...
-        end_point : array-like
-            shape=[n_samples, dimension] or shape=[1, dimension]
+        transported_tangent_vector : array-like, shape=[n_samples, dimension]
+            Tangent vector at end point.
+        end_point : array-like, shape=[n_samples, dimension]
+            Point on the manifold, closes the geodesic parallelogram of the
+            construction.
+        trajectories : list of callable, len=3 (only if
+            `return_geodesics=True`)
+            The 3 three geodesics of the construction.
```
ninamiolane: Nice, thanks!

geomstats/geometry/connection.py
```diff
@@ -188,34 +220,51 @@ def pole_ladder_step(self, base_point, next_point, base_shoot):
...
+                initial_point=base_point,
+                end_point=next_point)
+            diagonal = self.geodesic(
+                initial_point=mid_point,
+                end_point=base_shoot)
+            final_geodesic = self.geodesic(
+                initial_point=next_point,
+                end_point=end_shoot)
+            trajectories = [main_geodesic, diagonal, final_geodesic]
+        return transported_tangent_vector, end_point, trajectories
```
ninamiolane: Do we want to return a blank list when return_geodesic is false? Or rather have a completely different return: `return transported_tangent_vector, end_point`?

geomstats/geometry/connection.py
```diff
@@ -230,29 +279,32 @@ def pole_ladder_parallel_transport(
...
-                              tangent_vec=transported_tangent_vector)
+                              tangent_vec=transported_tangent_vec)
+        trajectory = []
         for i_point in range(0, n_steps):
             frac_tangent_vector_b = (i_point + 1) / n_steps * tangent_vec_b
             next_point = self.exp(
                 base_point=base_point,
                 tangent_vec=frac_tangent_vector_b)
-            transported_tangent_vector, base_shoot = self.pole_ladder_step(
+            transported_tangent_vec, base_shoot, pts = self._pole_ladder_step(
```
ninamiolane: `points`? we haven't abbreviated points much in the codebase so far.

geomstats/geometry/connection.py
```diff
@@ -230,29 +279,32 @@ def pole_ladder_parallel_transport(
...
                 base_point=current_point,
                 next_point=next_point,
-                base_shoot=base_shoot)
+                base_shoot=base_shoot,
+                **single_step_kwargs)
             current_point = next_point
+            trajectory.append(pts)
 
-        return transported_tangent_vector
+        return transported_tangent_vec, trajectory
```
ninamiolane: Same question here

geomstats/geometry/connection.py
nguigs: That's what I was wandering, the problem is that if you write that you have to adapt the way you then unpack the output l291 depending on the value of `return_geodesics` which will make the code longer and with repetitions...
I don't know what is the best practice here...

geomstats/geometry/connection.py
nguigs: Yes, same probleme:
```
pole_ladder, trajectory = metric.ladder_parallel_transport(
    tan_a, tan_b, point, n_steps=n_steps, return_geodesics=True)
```
would run, but
```
pole_ladder, trajectory = metric.ladder_parallel_transport(
    tan_a, tan_b, point, n_steps=n_steps, return_geodesics=False)
```
would crash, which I think is more annoying than always returning a tuple even if one may be an empty list

geomstats/geometry/connection.py
ninamiolane: Typo: 3 three

Issue #429, opened 2020-02-07
Uniform handling of logging / warnings


stefanheyder:
Currently we are using different mechanisms to notify the user of logging information (such as convergence warnings, which backend is in use etc.).

1. `print` https://github.com/geomstats/geomstats/blob/c4e6d959db7b1bcc99b00b535b8aa5d832b62e28/geomstats/backend/__init__.py#L19
2. `logging` https://github.com/geomstats/geomstats/blob/c4e6d959db7b1bcc99b00b535b8aa5d832b62e28/geomstats/geometry/hyperbolic.py#L129
3. `warning`https://github.com/geomstats/geomstats/blob/c4e6d959db7b1bcc99b00b535b8aa5d832b62e28/geomstats/geometry/riemannian_metric.py#L511

I think we should unify this, and after a quick research the accepted way seems to be to use the `logging` module everywhere (of course, `print` is bad coding style). After we decide which of the above options to use, we should also append best practices to our contributing file.

ninamiolane:
Agreed to uniformize!  @est271 do you have any ideas/comments on this?

est271:
@stefanheyder , @ninamiolane 

Yes, using the logging module is the best option. 

I used the warnings module in one of my commits because I wanted a warning to print only once during testing and using logging would have printed the warning multiple times. 

But this issue has since been fixed so changing `warnings.warn()` to `logging.warning()` will work just fine. We would just have to keep an eye out if a similar issue appears in another part of the code.

est271:
Quick question regarding the appearance of statements before I go any further.

Using the default root logging setup, the default message format will display the levelname, the word root which is the default logger, and finally the message that is passed by using `logging.info()` or `logging.warning()`. A sample of how it will appear on the console is shown below (some info omitted for clarity).

![default](https://user-images.githubusercontent.com/56150991/74997960-43ad6800-540c-11ea-87c1-ff36104de30f.png)

I think removing the root part and just having the levelname followed by the message will be sufficient but I could leave it as is. [Other attributes can be added or taken out as well](https://docs.python.org/3/library/logging.html#logrecord-attributes) . Any suggestions about the format?

ninamiolane:
Agreed, let's just keep the levelname and not the word root! I looked at the other attributes on your weblink, and I don't see a need for printing others. But happy to know if you have suggestions / if you have seen other attributes printed in other libraries and the advantages of it!

Thanks for looking at this!

nkoep:
[...]

PR #456, opened 2020-02-28
Uniformize logging

est271:
Fixes part of issue #429 

This PR addresses the uniform handling of print / logging / warnings. Work is in progress as some warning statements still need to be changed.

**Note:**
Because nose2 introduces its own logging configuration, the logging **format** will not be displayed as implemented in `geomstats/backend/__init__py` during testing. I tried to find a way around this problem but everything I tried just introduced more issues. 
Nevertheless, the correct logging format will be displayed when running examples or other task that do not involve nose2 testing.

### DIFF
--- a/examples/empirical_frechet_mean_uncertainty_sn.py
+++ b/examples/empirical_frechet_mean_uncertainty_sn.py
@@ -8,6 +8,8 @@ for isotropic distributions on hyper-spheres of radius 0 < theta < Pi in
 the sphere S_dim (called here a bubble).
 """
 
+import logging
+
 import matplotlib.pyplot as plt
 
 import geomstats.backend as gs
@@ -136,7 +138,8 @@ def plot_modulation_factor(n_samples, dim, n_expectation=1000, n_theta=20):
             n_samples, theta_i, dim, n_expectation=n_expectation)
         measured_modulation_factor.append(var)
         error.append(std_var)
-        print(n_samples, theta_i, var, std_var, '\n')
+        logging.info(
+            '{} {} {} {}\n'.format(n_samples, theta_i, var, std_var))
         small_var_modulation_factor.append(
             1.0 + 2.0 / 3.0 * theta_i ** 2
             * (1.0 - 1.0 / dim) * (1.0 - 1.0 / n_samples))
@@ -197,7 +200,8 @@ def multi_plot_modulation_factor(dim, n_expectation=1000, n_theta=20):
             (var, std_var) = modulation_factor(
                 n_samples, theta_i, dim, n_expectation=n_expectation)
             measured_modulation_factor.append(var)
-            print(n_samples, theta_i, var, std_var, '\n')
+            logging.info(
+                '{} {} {} {}\n'.format(n_samples, theta_i, var, std_var))
         plt.plot(theta, measured_modulation_factor,
                  color=color[n_samples], label="N={0}".format(n_samples))
     plt.xlabel(r'Standard deviation $\theta$')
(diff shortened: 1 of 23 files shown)


examples/empirical_frechet_mean_uncertainty_sn.py
```diff
@@ -136,7 +138,8 @@ def plot_modulation_factor(n_samples, dim, n_expectation=1000, n_theta=20):
             n_samples, theta_i, dim, n_expectation=n_expectation)
         measured_modulation_factor.append(var)
         error.append(std_var)
-        print(n_samples, theta_i, var, std_var, '\n')
+        logging.info(
```
ninamiolane: Is \n needed? Same questions for next `logging.info`

examples/tangent_pca_s2.py
```diff
@@ -38,10 +40,10 @@ def main():
     geodesic_points_0 = geodesic_0(t)
     geodesic_points_1 = geodesic_1(t)
 
-    print(
+    logging.info(
         'Coordinates of the Log of the first 5 data points at the mean, '
         'projected on the principal components:')
-    print(tangent_projected_data[:5])
+    logging.info('\n{}'.format(tangent_projected_data[:5]))
```
ninamiolane: Is \n needed here?

examples/tangent_pca_so3.py
```diff
@@ -29,10 +31,10 @@ def main():
     tpca = TangentPCA(metric=METRIC, n_components=N_COMPONENTS)
     tpca = tpca.fit(data, base_point=mean_estimate)
     tangent_projected_data = tpca.transform(data)
-    print(
+    logging.info(
         'Coordinates of the Log of the first 5 data points at the mean, '
         'projected on the principal components:')
-    print(tangent_projected_data[:5])
+    logging.info('\n{}'.format(tangent_projected_data[:5]))
```
ninamiolane: same question

geomstats/backend/__init__.py
```diff
@@ -1,8 +1,11 @@
+import logging
 import os
-import sys
 
 from numpy import pi  # NOQA
 
+logging.basicConfig(format='%(levelname)s: %(message)s')
+logging.getLogger().setLevel(logging.INFO)
```
ninamiolane: Nice!

geomstats/backend/__init__.py
```diff
@@ -16,4 +19,8 @@
     from geomstats.backend.tensorflow import *  # NOQA
 else:
     raise RuntimeError('Unknown backend \'{:s}\''.format(_BACKEND))
-print('Using {:s} backend'.format(_BACKEND), file=sys.stderr)
+logging.info('Using {:s} backend'.format(_BACKEND))
+
+loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
+if loggers and loggers[0].name.startswith('nose2'):
+    logging.getLogger().setLevel(logging.WARNING)
```
ninamiolane: Nice, thanks!

examples/empirical_frechet_mean_uncertainty_sn.py
est271: For this case I was trying to keep the spacing of the console messages the same way it was spaced before I made my changes as shown below.
q1
the \n is not needed and I can remove it if you’d like. If I remove the \n more messages can fit on the screen, so it has that advantage. What do you think?

examples/tangent_pca_s2.py
est271: For this case I was looking to align all 5 data points since the first entry is shifted to the right due to the addition of the message logging level. I did this just for aesthetic reasons and is not required. Without the newline, the message will appear as shown below .
q2

examples/tangent_pca_so3.py
est271: My reasoning for this case is the same as `examples/tangent_pca_s2.py `

examples/tangent_pca_s2.py
ninamiolane: Oh I get it, great!

examples/tangent_pca_so3.py
ninamiolane: Ok!

examples/empirical_frechet_mean_uncertainty_sn.py
ninamiolane: Got it, yes it's easier to read this way. Thanks for clarifying!

PR #465, opened 2020-03-30
Remove binary 'out.mp4' file re-introduced in fc8baac

nkoep:


nguigs:
Well spotted, thanks!

Issue #245, opened 2019-11-15
invariant metric with point_type=matrix


nguigs:
The log seems to expect vectors even when using matrices to represent points
```
SO3_GROUP = SpecialOrthogonalGroup(n=3, point_type='matrix')
METRIC = SO3_GROUP.bi_invariant_metric

N_SAMPLES = 10

data = SO3_GROUP.random_uniform(n_samples=N_SAMPLES)
mean = METRIC.mean(data)
```
This raises a ValueError, due to
https://github.com/geomstats/geomstats/blob/e786bd606532cce887acf5d4d7adf5630f8b241f/geomstats/geometry/invariant_metric.py#L291

which can be removed as n_points is not used later, but then
https://github.com/geomstats/geomstats/blob/e786bd606532cce887acf5d4d7adf5630f8b241f/geomstats/geometry/invariant_metric.py#L260

Is it necessary?

nguigs:
In fact rotation_vector_from_matrix should be used as the log in that case

stefanheyder:
So if I remember correctly, the Riemannian `exp` and `log` for left/right invariant metrics should coincide with the Lie group `exp` and `log` (from the left/right). Shouldn't we then just call `group.log(....)` in the `invariant_metric.left_exp_from_identity` (and related functions)?

stefanheyder:
Also should we add `point_type` as an argument to the methods in the class?

nguigs:
> So if I remember correctly, the Riemannian `exp` and `log` for left/right invariant metrics should coincide with the Lie group `exp` and `log`
That's not true if the metric is not bi-invariant I think, e.g. on SE(n): group `exp` is the matrix `exp` while the left inv. Riemannian `exp` is: `exp(\Omega), tv`. See https://en.wikipedia.org/wiki/Exponential_map_(Riemannian_geometry)#Relationships_to_exponential_maps_in_Lie_theory

nguigs:
This is outdated, the following now works:

```python
from geomstats.geometry.special_orthogonal import SpecialOrthogonal
from geomstats.learning.frechet_mean import FrechetMean
SO3_GROUP = SpecialOrthogonal(n=3, point_type='matrix')
METRIC = SO3_GROUP.bi_invariant_metric

N_SAMPLES = 10

data = SO3_GROUP.random_uniform(n_samples=N_SAMPLES)
estimator = FrechetMean(METRIC).fit(data)
mean = estimator.estimate_
```

PR #467, opened 2020-03-30
Fix general_linear and plot_bch_so3

### LINKED ISSUE #463: Use left-invariant vector-fields in general linear exp
Following discussions with @opeltre
https://github.com/geomstats/geomstats/blob/3ea3e47934de399eaccecd8aa438ff3e113ec1f6/geomstats/geometry/general_linear.py#L54

- the name `tangent_vec` is inappropriate as `expm` can only be used on an element of the lie algebra, so a tangent vector at `base_point` has to be transported to identity first.
- right multiplication by `base_point` is used here, while the convention is usually to use left multiplication

pchauchat:
Fixes plot_bch_so3 so that it works with numpy backend (still not with the other two)

Fixes #463 :
- order of product changed in both exp and log
- tangent_vec renamed to algebra_mat, as it represents a member of the Lie algebra, in matrix form.

### DIFF
--- a/0.22.1
+++ b/0.22.1
@@ -1,0 +1,3 @@
+Requirement already satisfied: scikit-learn in d:\apps\anaconda3\lib\site-packages (0.20.3)
+Requirement already satisfied: numpy>=1.8.2 in d:\apps\anaconda3\lib\site-packages (from scikit-learn) (1.16.2)
+Requirement already satisfied: scipy>=0.13.3 in d:\apps\anaconda3\lib\site-packages (from scikit-learn) (1.2.1)
--- a/0.22.1
+++ b/0.22.1
@@ -1,3 +1,0 @@
-Requirement already satisfied: scikit-learn in d:\apps\anaconda3\lib\site-packages (0.20.3)
-Requirement already satisfied: numpy>=1.8.2 in d:\apps\anaconda3\lib\site-packages (from scikit-learn) (1.16.2)
-Requirement already satisfied: scipy>=0.13.3 in d:\apps\anaconda3\lib\site-packages (from scikit-learn) (1.2.1)
--- a/1.18.1
+++ b/1.18.1
@@ -1,0 +1,1 @@
+Requirement already satisfied: numpy in d:\apps\anaconda3\lib\site-packages (1.16.2)
--- a/1.18.1
+++ b/1.18.1
@@ -1,1 +1,0 @@
-Requirement already satisfied: numpy in d:\apps\anaconda3\lib\site-packages (1.16.2)
--- a/examples/plot_bch_so3.py
+++ b/examples/plot_bch_so3.py
@@ -1,4 +1,5 @@
-"""
+"""Visualizes convergenve of the BCH formula approximation on so(n).
+
 Visualize the first, second and third order approximation of the Baker Campbell
 Hausdorf formula on so(n). To this end, sample 2 random elements a,b of so(n)
 and compute both the BCH approximations of different orders as well as
--- a/examples/plot_bch_so3.py
+++ b/examples/plot_bch_so3.py
@@ -1,4 +1,4 @@
-"""Visualizes convergenve of the BCH formula approximation on so(n).
+"""Visualize convergenve of the BCH formula approximation on so(n).
 
 Visualize the first, second and third order approximation of the Baker Campbell
 Hausdorf formula on so(n). To this end, sample 2 random elements a,b of so(n)
--- a/examples/plot_bch_so3.py
+++ b/examples/plot_bch_so3.py
@@ -32,15 +32,19 @@ ALGEBRA = SkewSymmetricMatrices(n=N)
 
 
 def main():
-    norm_rv_1 = gs.normal(size=DIM)
+    # norm_rv_1 = gs.normal(size=DIM)
+    norm_rv_1 = gs.random.normal(size=DIM)
     tan_rv_1 = ALGEBRA.matrix_representation(
-        norm_rv_1 / gs.norm(norm_rv_1, axis=0) / 2
+        # norm_rv_1 / gs.norm(norm_rv_1, axis=0) / 2
+        norm_rv_1 / gs.linalg.norm(norm_rv_1, axis=0) * 2
     )
     exp_1 = gs.linalg.expm(tan_rv_1)
 
-    norm_rv_2 = gs.normal(size=DIM)
+    # norm_rv_2 = gs.normal(size=DIM)
+    norm_rv_2 = gs.random.normal(size=DIM)
     tan_rv_2 = ALGEBRA.matrix_representation(
-        norm_rv_2 / gs.norm(norm_rv_2, axis=0) / 2
+        # norm_rv_2 / gs.norm(norm_rv_2, axis=0) / 2
+        norm_rv_2 / gs.linalg.norm(norm_rv_2, axis=0) * 2
     )
     exp_2 = gs.linalg.expm(tan_rv_2)
 
@@ -86,7 +90,7 @@ def main():
     plt.grid()
 
     plt.show()
-    plt.close()
+    # plt.close()
 
 
 if __name__ == '__main__':
--- a/examples/plot_bch_so3.py
+++ b/examples/plot_bch_so3.py
@@ -86,7 +86,6 @@ def main():
     plt.grid()
 
     plt.show()
-    # plt.close()
 
 
 if __name__ == '__main__':
--- a/examples/plot_bch_so3.py
+++ b/examples/plot_bch_so3.py
@@ -32,19 +32,15 @@ ALGEBRA = SkewSymmetricMatrices(n=N)
 
 
 def main():
-    # norm_rv_1 = gs.normal(size=DIM)
     norm_rv_1 = gs.random.normal(size=DIM)
     tan_rv_1 = ALGEBRA.matrix_representation(
-        # norm_rv_1 / gs.norm(norm_rv_1, axis=0) / 2
-        norm_rv_1 / gs.linalg.norm(norm_rv_1, axis=0) * 2
+        norm_rv_1 / gs.linalg.norm(norm_rv_1, axis=0) / 2
     )
     exp_1 = gs.linalg.expm(tan_rv_1)
 
-    # norm_rv_2 = gs.normal(size=DIM)
     norm_rv_2 = gs.random.normal(size=DIM)
     tan_rv_2 = ALGEBRA.matrix_representation(
-        # norm_rv_2 / gs.norm(norm_rv_2, axis=0) / 2
-        norm_rv_2 / gs.linalg.norm(norm_rv_2, axis=0) * 2
+        norm_rv_2 / gs.linalg.norm(norm_rv_2, axis=0) / 2
     )
     exp_2 = gs.linalg.expm(tan_rv_2)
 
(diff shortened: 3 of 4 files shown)


0.22.1
```diff
@@ -0,0 +1,3 @@
+Requirement already satisfied: scikit-learn in d:\apps\anaconda3\lib\site-packages (0.20.3)
```
ninamiolane: This file should not be committed with the PR.

1.18.1
```diff
@@ -0,0 +1 @@
+Requirement already satisfied: numpy in d:\apps\anaconda3\lib\site-packages (1.16.2)
```
ninamiolane: This file should not be committed with the PR.

examples/plot_bch_so3.py
```diff
@@ -1,4 +1,5 @@
-"""
+"""Visualizes convergenve of the BCH formula approximation on so(n).
```
ninamiolane: NIT: """Visualize.
(no s)

Typo: convergenve -> convergence

geomstats/geometry/general_linear.py
```diff
@@ -114,9 +114,9 @@ def orbit(cls, point, base_point=None):
 
         Will work when expm gets properly 4-D vectorized.
```
ninamiolane: "Will work when..."
Sorry, I had missed this comment in a previous code review.

Could you update this comment into a TODO(ninamiolane)? Thanks!

geomstats/geometry/general_linear.py
```diff
@@ -31,27 +31,27 @@ def inv(point):
         return gs.linalg.inv(point)
 
     @classmethod
-    def exp(cls, tangent_vec, base_point=None):
+    def exp(cls, algebra_mat, base_point=None):
```
ninamiolane: I saw issue #463, but I am not sure that I understand. Shouldn't we keep `tangent_vec`, transport it to the lie algebra via `base_point`, then use `expm`, and then again the group multiplication by `base_point`?

This way, this group `exp` follows the same logic, and the same API, at the Riemannian exponentials.

@nguigs @pchauchat

examples/plot_bch_so3.py
```diff
@@ -32,15 +33,15 @@
 
 
 def main():
-    norm_rv_1 = gs.normal(size=DIM)
+    norm_rv_1 = gs.random.normal(size=DIM)
```
ninamiolane: Great, thanks!

geomstats/geometry/general_linear.py
nguigs: Yes, it still makes sense as the flow of left-invariant vector fields with initial condition base_point, tangent_vec. What do you think @opeltre ?

nguigs: See remark 3 p.612 of https://www.seas.upenn.edu/~jean/diffgeom-spr-I.pdf

opeltre: Yes @nguigs the formula `x(t) = exp(t v) . x` is the right one for left-invariant orbits, i.e. the initial velocity `dx/dt (0) = v` and this coincides with the logic of riemaniann exp

opeltre: You were right, this is the left invariant one in the sense that we always have `x(t) . dx/dt = x(0) . v` !

opeltre: Where `V` is in the Lie algebra, but `v = dx/dt(0) = x(0) . V` is in T_x GL
So as @ninamiolane points out we should keep this clear in the docstrings. The renaming to `algebra_mat` is probably confusing as this function does not identify T_x GL with gl = T_1 GL. It's probably better to stick with the riemannian convention as one can always compute the exponential from `V` in the Lie algebra via: 
```python
GL.mul(x, GL.exp(V)) = GL.exp(v, x)      #when v = x . V
```

opeltre: Sorry I got confused again :sweat_smile:  
Letting `x(t) = x(0) . exp(t V)` has left-invariant velocity `dx/dt(t) = x(t) . V` but of course`dx/dt(0) = x(0) . V`.
Hence @ninamiolane is right and to stay consistent with riemaniann exp we should have `tangent_vec = dx/dt(0)` as argument... 
```python
GL.exp(v, x) = GL.mul(x, GL.exp(GL.mul(GL.inv(x), v)))  //  x . exp(x^{-1} . v)   
// one can still compute from V in the Lie algebra with v = x . V via:
GL.mul(GL.exp(V), x) = GL.exp(v, x)
```

nguigs: I almost agree with you @opeltre, but if  V is in the Lie algebra with `v = x . V` then:
```GL.mul(x, GL.exp(V)) = GL.exp(v, x)```

1.18.1
pchauchat: Sorry about that, I was still getting used to the whole thing.

geomstats/geometry/general_linear.py
pchauchat: Just to be sure I understand: to be consistent with what is done in the Riemannian case, the first argument of exp should be, with your notations, `v (= x . V)`, while until now it was `V` the associated Lie algebra vector ?

I get the consistency issue, but from a computational point of view, won't that put some overhead ? For instance in the `orbit` function, `log` outputs directly a member of the Lie algebra, which should thus be multiplied by `base_point` to enter `exp`, to finally get brought back to the algebra. Or did I miss something ?

opeltre: > I almost agree with you @opeltre, but if V is in the Lie algebra with `v = x . V` then:
> `GL.mul(x, GL.exp(V)) = GL.exp(v, x)`
Haha yes exactly I keep getting mixed up with left and right ^^

opeltre: > Just to be sure I understand: to be consistent with what is done in the Riemannian case, the first argument of exp should be, with your notations, v (= x . V), while until now it was V the associated Lie algebra vector ?

Yes that's it, it's my bad I hadn't been careful with this... 

> I get the consistency issue, but from a computational point of view, won't that put some overhead ? 

No, actually it wouldn't because if you want to work from a Lie algebra vector `V` you can exponentiate from the identity with `GL.exp(V)` before **left**-multiplying the orbit by `x`. 

This should also be fixed accordingly in `log` if we want to be consistent, and we should probably insist on that in the docstring, recalling that `GL.mul(x, GL.exp(V)) = GL.exp(v, x)` in equations...

pchauchat: So `log` should also output `v` instead of `V` ?

Just one quick question, though, from a "Lie grouper" point of view: people often insist on the fact that the riemannian and Lie group exponentials usually differ. Is it good to give them the same treatment here ? Being even more stubborn, one could want to explicitly rename it lie_exp and lie_log.
I don't really have a fixed opinion on this, what do you think ?

nguigs: - Yes to the first question :), this is useful for e.g. exponential barycenters
- To the second, this as come up in #245, I think they could be renamed group_exp/log or left as is, and it coincides with the Riemannian exp iff the metric is bi-invariant. This could be mentionned in the docstring?

ninamiolane: - Also yes to the first question :) Even more given the remark by @xpennec at the Visio today: the group Exp/Log are affine connection Exp/Log of the Cartan-Shouten connection, and therefore should have the same API than Exp/Log of affine connections, which have the same API than Exp/Log of Riemannian metrics.

- Yes indeed they differ, and we should not mix them up! I don't think we should rename them though, because the object's type does it for us.
    - From a user perspective, you either write `my_gln_group.exp` or `my_metric.exp`; therefore it is already clear which `exp` is being used. Having `my_gln_group.group_exp` would be redundant IMO.
    - Even if the `exp` are different, they are still "exp-methods" taking a tangent_vec and a base_point and returning a point, therefore a consistent API will be easier for the users to remember.

ninamiolane: Down to mention the links between group exp and metric exp in docstrings!

pchauchat: Ok, I'll do this !

I have a newcomer question then: if someone were to use another metric on general_linear, then there would be no mix up because they would call the metric's exponential rather than that of the group ?

ninamiolane: Yes, exactly! This is why we distinguish the Manifold objects and the RiemannianMetric objects.

They would have to instantiate an object my_metric from the class of the metric that they need, and use my_metric.exp indeed!

pchauchat: I think I get it now.

At some point, it could be interesting, especially for you and the other main contributors, to look at the structure of the whole package, to be able to distinguish between objects. It might be easier for people who want to dig in differential geometry to have clear separations between metrics, connections, spaces, etc.

ninamiolane: You mean, for example, having two folders:
- `geomstats/geometry/manifolds/`
- `geomstats/geometry/metrics/`
?

I completely agree that some more structure would be helpful for newcomers.

pchauchat: Something like that, yes. Basically, a structure that makes people say "Ok, now I'm calling this class from this particular folder, so this is the kind of object that I am using".

opeltre: Aaah this is a tough question we ran into at the last hackathon :) 

To my understanding @pchauchat for now GL is one of the only places where `exp` and `log` are defined as attributes of the manifold, i.e. group, and not of a metric object. Going further in matrix Lie groups such as GA, SO, SE might help figure out what the structure looks should look like... 

The other place I think of is connection where `exp` and `log` are numerically integrated using autograd, assuming you supply with a way to compute christoffel symbols or the metric tensor (@nguigs ?) hence the mechanism is quite different with the rest of the package, providing closed formulas

pchauchat: Since the group exp and log come from the Cartan-Shouten connection, a solution could be to create such a class, place exp and log there containing the closed forms, and make it the connection by default for Lie groups. Although that could frighten people (for instance roboticians which are used to working with SE(3) but have no idea what in the world a connection is)

opeltre: > Although that could frighten people (for instance roboticians which are used to working with SE(3) but have no idea what in the world a connection is)

Yes, I agree with you. 

For me there is no reason to rename `group.exp` as in the first and strict sense the exponential is the canonical map `exp: g -> G` from a Lie algebra to its Lie group. (In GeneralLinear, this canonical map is obtained by supplying no `base_point` argument: `GL.exp(V)` ).  Its definition does not resort to a connection... 

The "riemannian exponential" is a by-product of this exponential, and a slight abuse of terminology. It actually stands for: the group exponential map `exp: g -> G` where `g = X(TM)` is the algebra of vector fields on `TM` and `G = Diff(TM)` is the group of diffeomorphisms, **applied** to the vector field `V` of `X(TM)` defined by the geodesic equations of motion... 

We also had a discussion about this [here](https://github.com/geomstats/geomstats/pull/382#discussion_r368866950)

1.18.1
pchauchat: Is there a way to remove them from the PR ? Or are will you do it yourself when merging ?

nguigs: I just deleted them and commited the change from the github interface!

PR #468, opened 2020-03-31
Add options for code coverage

ninamiolane:
This commit adds a .codecov.yml that allows to set options relative to
code coverage with codecov. Specifically, the option to ignore data/ and
imgs/ folder from the examples/ is added.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,18 +1,12 @@
 # Team Yaml
 coverage:
-  round: down
-  precision: 5
+  round: up
+  precision: 2
 
 # Repository Yaml
 coverage:
   round: up
-  range: 0..10
-
-# Used in Codecov after updating
-coverage:
-  round: up
-  range: 0..10
-  precision: 5
+  range: 50..100
 
 # Files to ignore
 ignore:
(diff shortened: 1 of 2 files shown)


.codecov.yml
```diff
@@ -0,0 +1,20 @@
...
+  round: down
+  precision: 5
+
+# Repository Yaml
+coverage:
+  round: up
+  range: 0..10
+
+# Used in Codecov after updating
+coverage:
```
nkoep: It seems that this block overrides all the previous settings so I think we can remove them. At least that's how I interpret the validation output:
```shell
[archibald ⚡ /tmp] » curl --data-binary @codecov.yml https://codecov.io/validate
Valid!

{
  "ignore": [
    "^examples/data.*", 
    "^examples/imgs.*"
  ], 
  "coverage": {
    "range": [
      0.0, 
      10.0
    ], 
    "round": "up", 
    "precision": 5
  }
}
```
Also, the `range` parameter seems to define the region between which coverage status color transitions from red to green, see https://docs.codecov.io/docs/coverage-configuration#section-range. So `range: 0..10` is probably a little premature :joy: Maybe we can change the range to `50..100` or `50..90`. I don't have a strong opinion regarding the precision.

PR #470, opened 2020-03-31
add projection from St(n, p) to Gr(n, p)

opeltre:
Given an orthonormal p-frame `x` in Mat(n, p) returns the projector onto its image `p = x . x*`

### DIFF
--- a/geomstats/geometry/stiefel.py
+++ b/geomstats/geometry/stiefel.py
@@ -78,6 +78,24 @@ class Stiefel(EmbeddedManifold):
         belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
         return belongs
 
+    def to_grassmannian(self, point):
+        r"""Project a point of St(n, p) to Gr(n, p).
+
+        If :math:`U \in St(n, p)` is an orthonormal frame,
+        return the orthogonal projector :math:`P = U U^T`
+        onto the subspace of :math:`\mathbb{R}^n` spanned by
+        :math:`U`.
+
+        Parameters
+        ----------
+        point : array-like, shape=[n_samples, n, p]
+
+        Returns
+        -------
+        projector : array-like, shape=[n_samples, n, n]
+        """
+        return Matrices.mul(point, Matrices.transpose(point))
+
     def random_uniform(self, n_samples=1):
         r"""Sample on St(n,p) from the uniform distribution.
 
--- a/geomstats/geometry/stiefel.py
+++ b/geomstats/geometry/stiefel.py
@@ -272,7 +272,6 @@ class StiefelCanonicalMetric(RiemannianMetric):
         matrix_n : array-like
         """
         matrix_k = point - gs.matmul(base_point, matrix_m)
-
         matrix_q, matrix_n = gs.linalg.qr(matrix_k)
         return matrix_q, matrix_n
 
--- a/geomstats/geometry/stiefel.py
+++ b/geomstats/geometry/stiefel.py
@@ -78,7 +78,8 @@ class Stiefel(EmbeddedManifold):
         belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
         return belongs
 
-    def to_grassmannian(self, point):
+    @staticmethod
+    def to_grassmannian(point):
         r"""Project a point of St(n, p) to Gr(n, p).
 
         If :math:`U \in St(n, p)` is an orthonormal frame,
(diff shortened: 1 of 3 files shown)


nguigs:
Cool! Does Grassmann has a belong method? It could be nice to add a test that checks that `Grassmann.belongs(to_grassmann(point)=True` with `point \in Stiefel`

geomstats/geometry/stiefel.py
```diff
@@ -78,6 +78,24 @@ def belongs(self, point, tolerance=TOLERANCE):
         belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
         return belongs
 
+    def to_grassmannian(self, point):
```
ninamiolane: This could be a static method: it does not depend on the object (self) nor the class (cls).

Also, should we have it in stiefel.py, or maybe in matrices.py that would gather all the conversions from matrices types? (I have no strong opinion on it, happy to go with whatever you recommend).

geomstats/geometry/stiefel.py
```diff
@@ -78,6 +78,24 @@ def belongs(self, point, tolerance=TOLERANCE):
...
+        r"""Project a point of St(n, p) to Gr(n, p).
```
ninamiolane: This method needs a unit-test.

opeltre:
> Cool! Does Grassmann has a belong method? It could be nice to add a test that checks that `Grassmann.belongs(to_grassmann(point)=True` with `point \in Stiefel`

Not yet but we should be working on that with Pablo: the problem had encountered was no consistent way to perform logical operations on boolean arrays accross backends, we need a `gs.logical_and` in the backend files...

geomstats/geometry/stiefel.py
opeltre: I put it here for starters but I'm also not sure it's such a very good place... 
We could end up defining the same method in `grassmannian.py` as: 
```
def from_stiefel(self, point):
    return Stiefel(self.n, self.p).to_grassmann(point)
```

opeltre: I was thinking about 2 options: 

1. Chose to implement only methods of the form `mfdA.to_mfdB` in the `mfdA` class by convention, at least for the code, and optionally later link `mfdB.from_mfdA` to this method

2. Define functions `mfdA_to_mfdB` in a separate `mappings.py` file, imported by the mfdA and the mfdB class files

I don't really know which is better... 1. is simple and allows to mfdA methods seamlessly but 2. has the advantage of grouping all mappings together so that it's easier to keep track of which have been implemented or not...

opeltre: Maybe we could keep track of these mappings in a "feature request" issue anyway

ninamiolane: @opeltre I like both options, I don't really have a strong opinion for one vs the other! Maybe let's define a new issue and ask for other opinions to choose between the two?

Still, a related issue is: how do we name methods that change coordinates representations from one type to another, for example: poincare ball coordinates to extrinsic coordinates in the hyperbolic space? (@tgeral68). It would be good to have a consistent logic with these as well.

ninamiolane:
> > Cool! Does Grassmann has a belong method? It could be nice to add a test that checks that `Grassmann.belongs(to_grassmann(point)=True` with `point \in Stiefel`
> 
> Not yet but we should be working on that with Pablo: the problem had encountered was no consistent way to perform logical operations on boolean arrays accross backends, we need a `gs.logical_and` in the backend files...

Yes, there is a:
```
def logical_or(x, y):
    return x or y
```
but no logical_and so far.

ninamiolane:
Good to know, thanks! (Just waiting for Travis to pass).

ninamiolane:
@opeltre travis fails because of flake8 and import orders in test_stiefel. Ready to merge when this is corrected!

ninamiolane:
@opeltre it fails in tf and pytorch:
https://travis-ci.org/github/geomstats/geomstats/builds/669746781?utm_source=github_status&utm_medium=notification

Maybe add the decorator only_ to the unit test?

opeltre:
Do you understand why `gs.sqrt(2)` errors in tf??

ninamiolane:
@opeltre you solved gs.sqrt(2)? I do not see it.

Flake8 fails though.

nguigs:
> Do you understand why `gs.sqrt(2)` errors in tf??

from [tf documentation](https://www.tensorflow.org/api_docs/python/tf/math/sqrt), the argument must be one of following types: `bfloat16, half, float32, float64, complex64, complex128`.

So `gs.sqrt(2.)` will work.

PR #471, opened 2020-03-31
Product Metric for vector representation

nguigs:
- Wrap the iteration over manifolds/metrics for all methods, and parallelize them with joblib
- Vectorize the code for the matrix representation
- Add a few tests
This doesn't work for pytorch because of inconsistent behavior of tensor.split and array.split.
This only works for 1d points/vectors


geomstats/geometry/product_riemannian_metric.py
```diff
@@ -3,14 +3,14 @@
 Define the metric of a product manifold endowed with a product metric.
 """
 
+from joblib import delayed, Parallel
```
ninamiolane: Avoid relative imports

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -52,21 +55,52 @@ def inner_product_matrix(self, base_point=None):
...
             b = dim_current + dim_next
-            matrix_next = self.metrics.inner_product_matrix(base_point[i + 1])
+            matrix_next = self.metrics[i].inner_product_matrix(
+                base_point[i + 1])
             matrix[a:b, a:b] = matrix_next
 
         return matrix
 
-    def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
+    def _detect_intrinsic_extrinsic(self, point):
```
ninamiolane: Ditto: `_is_intrinsic`? Although it could be nice to have this method coded only once, in product_manifold.py: how about **using** the method from product_manifold.py in this file?

geomstats/geometry/product_manifold.py
```diff
@@ -117,3 +151,34 @@ def regularize(self, point, point_type=None):
         elif point_type == 'matrix':
             regularized_point = gs.vstack(regularized_point)
         return gs.all(regularized_point)
+
+    def random_uniform(self, n_samples, point_type=None):
+        """Sample in the the product space from the uniform distribution.
```
ninamiolane: Typo: the the

geomstats/geometry/product_riemannian_metric.py
nkoep: Relative imports refer to things like `from . import sibling` though. But maybe a possible rule-of-thumb could be that if one uses more than, say, 2 objects from a module, one should import the toplevel module itself rather than individual objects.

PR #474, opened 2020-04-01
Special Euclidian matrix representation (focus on Numpy)

pchauchat:
Not that much of a "small" PR, sorry about that !
Redid commits from a branch I couldn't create a PR from (sorry about that also)

Mostly implementations of matrix form for SE(n), and associated tests. Focused on numpy implementation, some tricks are needed for Tensorflow and Pytorch (for which, in particular, .any() and .reshape() do not work, or differently)

Some changes use GeneralLinear, some others come from nguigs (especially jacobian and barycenter).

inverse function in matrix form to be more efficient (i.e., avoid matrix inversion)

Created matrix form of test points in a separated variable, to ease the associated tests.

Changed point_type check for identity (was only using the default one).

Created get_dimension which takes representation into account.

Renamed random_transfo to random_point in random_uniform.

Some flake8 corrections and unused variables deletions.

### DIFF
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -131,6 +131,15 @@ def flatten(x):
     return x.flatten()
 
 
+def logical_and(x, y):
+    bool_result = x and y
+    return bool_result
+
+
+def flatten(x):
+    return x.flatten()
+
+
 def get_mask_i_float(i, n):
     range_n = arange(n)
     i_float = cast(array([i]), int32)[0]
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -222,7 +222,7 @@ def eval(x):
 
 
 def ndim(x):
-    return _np.array(x).ndim
+    return x.ndim
 
 
 def cumprod(x, axis=0):
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -127,6 +127,15 @@ def logical_or(x, y):
     return bool_result
 
 
+def logical_and(x, y):
+    bool_result = x and y
+    return bool_result
+
+
+def flatten(x):
+    return x.flatten()
+
+
 def get_mask_i_float(i, n):
     range_n = arange(n)
     i_float = cast(array([i]), int32)[0]
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -7,7 +7,6 @@ from autograd.numpy import (  # NOQA
     allclose,
     amax,
     amin,
-    any,
     append,
     arange,
     arccos,
@@ -112,6 +111,11 @@ def byte_to_float(x):
     return x
 
 
+def any(x, axis=0):
+    print('there')
+    return _np.any(x, axis)
+
+
 def while_loop(cond, body, loop_vars, maximum_iterations):
     iteration = 0
     while cond(*loop_vars):
@@ -142,6 +146,23 @@ def get_mask_i_float(i, n):
     mask_i = equal(range_n, i_float)
     mask_i_float = cast(mask_i, float32)
     return mask_i_float
+
+
+def assignment(x, values, indices, axis=0):
+    x_new = copy(x)
+    single_index = ndim(indices) == 0 or (ndim(indices) <= 1 and ndim(x) > 1)
+    if single_index:
+        indices = [indices]
+    if ndim(values) == 0:
+        values = [values] * len(indices)
+    for (nb_index, index) in enumerate(indices):
+        if len(indices[0]) < len(shape(x)):
+            for n_axis in range(shape(x)[axis]):
+                extended_index = index[:axis] + (n_axis,) + index[axis:]
+                x_new[extended_index] = values[nb_index]
+        else:
+            x_new[index] = values[nb_index]
+    return x_new
 
 
 def gather(x, indices, axis=0):
@@ -201,7 +222,7 @@ def eval(x):
 
 
 def ndim(x):
-    return x.ndim
+    return _np.array(x).ndim
 
 
 def cumprod(x, axis=0):
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -52,6 +52,8 @@ from autograd.numpy import (  # NOQA
     less_equal,
     linspace,
     log,
+    logical_and,
+    logical_or,
     matmul,
     maximum,
     mean,
@@ -112,7 +114,6 @@ def byte_to_float(x):
 
 
 def any(x, axis=0):
-    print('there')
     return _np.any(x, axis)
 
 
@@ -124,16 +125,6 @@ def while_loop(cond, body, loop_vars, maximum_iterations):
         if iteration >= maximum_iterations:
             break
     return loop_vars
-
-
-def logical_or(x, y):
-    bool_result = x or y
-    return bool_result
-
-
-def logical_and(x, y):
-    bool_result = x and y
-    return bool_result
 
 
 def flatten(x):
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -141,11 +141,10 @@ def get_mask_i_float(i, n):
 
 def assignment(x, values, indices, axis=0):
     x_new = copy(x)
-    single_index = ndim(array(indices)) == 0 or (
-                ndim((array(indices))) <= 1 and ndim(x) > 1)
+    single_index = not isinstance(indices, list)
     if single_index:
         indices = [indices]
-    if ndim((array(values))) == 0:
+    if not isinstance(values, list):
         values = [values] * len(indices)
     for (nb_index, index) in enumerate(indices):
         if len(indices[0]) < len(shape(x)):
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -150,10 +150,10 @@ def get_mask_i_float(i, n):
 
 def assignment(x, values, indices, axis=0):
     x_new = copy(x)
-    single_index = ndim(indices) == 0 or (ndim(indices) <= 1 and ndim(x) > 1)
+    single_index = ndim(array(indices)) == 0 or (ndim((array(indices))) <= 1 and ndim(x) > 1)
     if single_index:
         indices = [indices]
-    if ndim(values) == 0:
+    if ndim((array(values))) == 0:
         values = [values] * len(indices)
     for (nb_index, index) in enumerate(indices):
         if len(indices[0]) < len(shape(x)):
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -131,15 +131,6 @@ def flatten(x):
     return x.flatten()
 
 
-def logical_and(x, y):
-    bool_result = x and y
-    return bool_result
-
-
-def flatten(x):
-    return x.flatten()
-
-
 def get_mask_i_float(i, n):
     range_n = arange(n)
     i_float = cast(array([i]), int32)[0]
@@ -150,7 +141,8 @@ def get_mask_i_float(i, n):
 
 def assignment(x, values, indices, axis=0):
     x_new = copy(x)
-    single_index = ndim(array(indices)) == 0 or (ndim((array(indices))) <= 1 and ndim(x) > 1)
+    single_index = ndim(array(indices)) == 0 or (
+                ndim((array(indices))) <= 1 and ndim(x) > 1)
     if single_index:
         indices = [indices]
     if ndim((array(values))) == 0:
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -393,10 +393,10 @@ def get_mask_i_float(i, n):
 
 def assignment(x, values, indices, axis=0):
     x_new = copy(x)
-    single_index = ndim(indices) == 0 or (ndim(indices) <= 1 and ndim(x) > 1)
+    single_index = ndim(array(indices)) == 0 or (ndim(array(indices)) <= 1 and ndim(x) > 1)
     if single_index:
         indices = [indices]
-    if ndim(values) == 0:
+    if ndim(array(values)) == 0:
         values = [values] * len(indices)
     for (nb_index, index) in enumerate(indices):
         if len(indices[0]) < len(shape(x)):
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -29,12 +29,10 @@ def logical_or(x, y):
 
 
 def logical_and(x, y):
-    print('there')
     return x and y
 
 
 def any(x, axis=0):
-    print('there')
     return torch.any(x.type(torch.bool), axis)
 
 def cond(pred, true_fn, false_fn):
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -142,7 +142,8 @@ def empty_like(*args, **kwargs):
 def all(x, axis=None):
     if axis is None:
         return x.byte().all()
-    return torch.from_numpy(_np.array(_np.all(_np.array(x), axis=axis).astype(int)))
+    numpy_result = _np.array(_np.all(_np.array(x), axis=axis).astype(int))
+    return torch.from_numpy(numpy_result)
 
 
 def allclose(a, b, **kwargs):
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -33,7 +33,10 @@ def logical_and(x, y):
 
 
 def any(x, axis=0):
-    return torch.any(x.type(torch.bool), axis)
+    numpy_result = _np.array(_np.any(_np.array(x), axis=axis))
+    return torch.from_numpy(numpy_result)
+    # return x.type(torch.bool).any(axis)
+
 
 def cond(pred, true_fn, false_fn):
     if pred:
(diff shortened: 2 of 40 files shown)


pchauchat:
I closed it as it was still failing some tests. I'll try reopen it now that everything (should) work

pchauchat:
Also had to add some functions to the backend : logical_and, flatten and reshape.

All in all, that was slightly chaotic, but I think I reached a somewhat stable base now.

geomstats/backend/numpy/__init__.py
```diff
@@ -126,6 +126,15 @@ def logical_or(x, y):
     return bool_result
 
 
+def logical_and(x, y):
+    bool_result = x and y
+    return bool_result
```
ninamiolane, line 131: Great, thanks for this!

geomstats/geometry/special_euclidean.py
```diff
@@ -88,11 +91,15 @@ def get_identity(self, point_type=None):
             point_type = self.default_point_type
 
         identity = gs.zeros(self.dimension)
-        if self.default_point_type == 'matrix':
-            identity = gs.eye(self.n)
+        if point_type == 'matrix':
+            identity = gs.eye(self.n + 1)
```
ninamiolane, line 95: Good catch!

geomstats/geometry/special_euclidean.py
```diff
@@ -88,11 +91,15 @@ def get_identity(self, point_type=None):
...
         return identity
     identity = property(get_identity)
 
+    def get_dimension(self):
+        """Get the dimension of each object given the default_point_style."""
```
ninamiolane, line 100: default_point_style -> default_point_type, or point_type?

geomstats/geometry/special_euclidean.py
```diff
@@ -115,12 +122,32 @@ def belongs(self, point, point_type=None):
             point = gs.to_ndarray(point, to_ndim=2)
             n_points, point_dim = point.shape
             belongs = point_dim == self.dimension
-            belongs = gs.to_ndarray(belongs, to_ndim=1)
-            belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
-            belongs = gs.tile(belongs, (n_points, 1))
+            belongs = gs.logical_and(belongs,
+                                     self.rotations.belongs(point[:, :self.n]))
+            # TODO: Should to_ndarray do this if to_ndim = ndim(x) - 1 ?
```
ninamiolane, line 127: TODO syntax needs a github account name, eg: TODO(pchauchat).

This way, you can search for all your TODOs in the codebase easily.

ninamiolane: Also, NIT: TODO items should be actionable, eg:
TODO(pchauchat): Decide if to_ndarray.... .

geomstats/geometry/special_euclidean.py
```diff
@@ -115,12 +122,32 @@ def belongs(self, point, point_type=None):
...
+            belongs = gs.flatten(belongs)
         elif point_type == 'matrix':
             point = gs.to_ndarray(point, to_ndim=3)
-            raise NotImplementedError()
+            n_points, point_dim1, point_dim2 = point.shape
+            # Check dimensions
```
ninamiolane, line 132: No comments! Here are the reasons:
- Different developers have very different comment styles and the codebase can become a giant mess if everybody comments,
- Comments are difficult to maintain, they are not unit-tested for example, and you might have comments remaining that have nothing to do with the next code line in a few weeks from now,
- Prohibiting comments forces the collaborators to have a cleaner code! If you feel that your code is not understandable, maybe: find better variable names? add an auxiliary function, with a good function name, that has a docstring that explains what is going on? etc.

This no-comment guideline is a guideline used in software engineering at Google and many other startups that I know of in SF.

geomstats/geometry/special_euclidean.py
```diff
@@ -115,12 +122,32 @@ def belongs(self, point, point_type=None):
...
+            # Check rotation part
+            rotation = point[:, :self.n, :self.n]
+            rot_belongs = self.rotations.belongs(rotation,
+                                                 point_type=point_type)
+            belongs = gs.logical_and(belongs, rot_belongs)
+            # Check that last line is [0, ..., 0, 1]
+            last_line_except_last_term = point[:, self.n:, :-1]
+            all_but_last_zeros = ~ last_line_except_last_term.any(axis=(1, 2))
+            # Preferred to gs.all(last_line == 0, axis=1)
+            # TODO Fails with Pytorch, and probably Tensorflow too
```
ninamiolane, line 143: Follow TODO guidelines if you really need a TODO.

geomstats/geometry/special_euclidean.py
```diff
@@ -238,6 +263,34 @@ def regularize_tangent_vec(
 
         return regularized_vec
 
+    def matrix_from_vector(self, vec):  # TODO Fails with Tensorflow for now
```
ninamiolane, line 266: It can probably be clear from the tests that this fails with TF: the corresponding test will have the decorator @...only_np_and_pytorch

so no need for this comment.

geomstats/geometry/special_euclidean.py
```diff
@@ -238,6 +263,34 @@ def regularize_tangent_vec(
...
+        rot_vec = vec[:, :self.rotations.dimension]
+        trans_vec = vec[:, self.rotations.dimension:]
+
+        rot_mat = self.rotations.matrix_from_rotation_vector(rot_vec)
+        # Concatenate rotation and translation
+        trans_vec = trans_vec.reshape(n_vecs, self.n, 1)
+        mat = gs.concatenate((rot_mat, trans_vec), axis=2)
+        # Add the last line for homogeneous coordinates
+        last_lines = gs.zeros((n_vecs, 1, self.n + 1))
+        last_lines[:, :, -1:] = 1
```
ninamiolane, line 289: Probably the assignment fails with TF?

geomstats/geometry/special_euclidean.py
```diff
@@ -363,6 +427,7 @@ def jacobian_translation(
         Compute the jacobian matrix of the differential
         of the left/right translations from the identity to point in SE(n).
         Currently only implemented for point_type == 'vector'.
+        # TODO(nguigs): implement for matrix, formula p157 in Xavier's thesis
```
ninamiolane, line 430: 🎉

geomstats/geometry/special_euclidean.py
```diff
@@ -617,33 +683,37 @@ def random_uniform(self, n_samples=1, point_type=None):
...
+        point_type: str, {'vector', 'matrix'}, optional
             default: self.default_point_type
 
-
         Returns
         -------
-        random_transfo: array-like,
+        random_point: array-like,
             shape=[n_samples, {dimension, [n + 1, n + 1]}]
             an array of random elements in SE(n) having the given point_type
```
ninamiolane, line 693: I know this is not your line, but could you update this parameter description so that it follows the docstring guidelines? (see Contributing file).

    An array of random elements in SE(n) having the given point_type.

tests/test_special_euclidean.py
```diff
@@ -93,6 +93,12 @@ def setUp(self):
                 'point_1': point_1,
                 'point_2': point_2}
 
+        else:  # matrix_from_vector fails with tensorflow
```
ninamiolane, line 96: Remove this omment

tests/test_special_euclidean.py
```diff
@@ -140,23 +146,61 @@ def setUp(self):
...
+    def test_random_and_belongs_matrix_form(self):
+        """
+        Test that the random uniform method samples
+        on the special euclidean group.
+        """
+        local_point_type = 'matrix'
+        base_point_1 = self.group.random_uniform(1, local_point_type)
+        base_point_2 = gs.copy(base_point_1)
+        base_point_3 = gs.copy(base_point_1)
+        # Violates SE(n) structure on the last line
```
ninamiolane, line 177: However, comments are usually fine in unit tests, where we are a bit less strict on the coding style :)

geomstats/geometry/special_euclidean.py
ninamiolane: Also, this should be renamed, as it does not really return the `dimension` of the Lie group: we reserve this name for the actual dimension of SE(n), which is n(n + 1)/2.

Maybe `get_point_shape`? (It's not great, happy to choose something else if you have a better idea!)

tests/test_special_euclidean.py
```diff
@@ -735,8 +881,8 @@ def test_log_from_identity_left(self):
     @geomstats.tests.np_only
     def test_exp_then_log_from_identity_left(self):
         """
-        Test that the riemannian left exponential from the identity
-        and the riemannian left logarithm from the identity
+        Test that the Riemannian left exponential from the identity
```
ninamiolane, line 884: Thanks!

geomstats/backend/tensorflow/__init__.py
```diff
@@ -23,6 +23,10 @@ def logical_or(x, y):
     return tf.logical_or(x, y)
 
 
+def logical_and(x, y):
```
ninamiolane, line 26: For the functions that have exactly the same syntax, I wonder if we could the same code as here:
https://github.com/geomstats/geomstats/blob/446f0f0a754b4d0ff7660fb240f4822e8eb673d3/geomstats/backend/numpy/__init__.py#L4

Eg:

from tensorflow import (
    logical_and,
...)

ninamiolane:
@pchauchat travis fails because of a flake8 error:
https://travis-ci.org/github/geomstats/geomstats/jobs/669720087?utm_medium=notification&utm_source=github_status

geomstats/geometry/special_euclidean.py
pchauchat: @nguigs is the one to thank for this !

geomstats/geometry/special_euclidean.py
pchauchat: - That's a good question. With the way things are coded now, most functions do not specify point_type, therefore relying on the default_point_type.
- Peut-être `get_representation_size`, pour bien appuyer sur le fait que c'est lié à la représentation ?

geomstats/geometry/special_euclidean.py
pchauchat: I didn't know who to assign this to. Since I'm new here, and it involves a quite important backend function, I mainly wanted to raise the issue.

geomstats/geometry/special_euclidean.py
pchauchat: Challenge accepted !

Is this rule also true with "uglier" languages (say, C++) ?

tests/test_special_euclidean.py
pchauchat: Is there a more "official" way to pinpoint things that we know need to be improved or fixed ?

tests/test_special_euclidean.py
pchauchat: Someone else is to be thanked for that, don't know who though.

pchauchat:
Indeed, the failure comes from an assignment issue. Is there a preferred way to solve that ?

I will correct the TODOs and other comment issues. (And the flake8, but normally it was only trailing spaces so it should be quick)

geomstats/geometry/special_euclidean.py
ninamiolane: Good question... I don't actually know, but I could very well be different!

tests/test_special_euclidean.py
ninamiolane: Yes!
- TODO(pchauchat) with an actionable item,
- FIXME
- issues, that can refer lines in the code, or even the corresponding TODOs.

Having a standardized syntax for TODOs and FIXMEs allows to use numerous online tools, for example:
https://github.com/naholyr/github-todos

ninamiolane:
Yes, solving for assignment issues should be done as follows:

- Use the hack we discussed during the Visio today: 
    1) initialize a my_tensor = gs.zeros() tensor, 
    2) create a mask that is [False, ...., False, True, False,... False], selecting the indices where you need to assign,
    3) my_tensor += mask * value_to_be_assigned.

Elements that help: 
- You can look at these lines from SO(3):
https://github.com/geomstats/geomstats/blob/886c1eca4b84501304c609ea07299761a47e0ea0/geomstats/geometry/special_orthogonal.py#L174
and note that the mask needs to be converted to a gs.float because tf does not know (at least didn't know in 1.X versions...) how to multiply bools and floats.
- The backend function `gs.get_mask_i_float` automatically creates a mask with True only at the i-th index.

- Ideally, this hack should be factorized *in* the backends (which is not what is done now, and consequently, the code in SO(3) is very ugly...:( ). So maybe create a `gs.assignment`function in the backends, that can account for the assignment cases where (i) one index is given, (ii) several indices are given, (iii) a mask is given.

geomstats/geometry/special_euclidean.py
pchauchat: Seems like Skyping in French and writing in English do not go well along. I had thought of `get_representation_size`to emphasize that it was linked to the chosen representation. Why not be even more direct and say `get_point_type_shape` with `point_type` as an optional argument ?

pchauchat:
I'm currently thinking about it and trying various ways to achieve that.

One question though, @ninamiolane , @nguigs, how bad would it be to create the mask (or even directly the dx such that x <- x + dx) with numpy and call convert_to_tensor on it ?

geomstats/geometry/special_euclidean.py
ninamiolane: - I would suggest self.point_type (no need for the default_*), since - as you said - point_type is more often used as the object's attributes than as a method's parameter. Then, methods should have `point_type=None` as an optional parameter, and within the method's body: `if point_type is None: point_type = self.point_type`.
- `size` is tricky, because in numpy, `np.size(my_array)` returns the total number of elements in the array, so we should prefer `shape` there. I like `get_point_type_shape`! I think it's better to be over-explicit for now indeed.

geomstats/geometry/special_euclidean.py
ninamiolane: Makes sense! In this case:
- either raise a issue and reference the line (go in the code, click on the number of line, then `Copy permalink` and paste this into the issue post), - preferred.
- or assign me x) TODO(ninamiolane) -> use my github login.

ninamiolane:
@pchauchat Hm, you can try to create the mask in np, yes. However, since tf tensors are not evaluated until a tf.session is created, you might not be able to get the _values_ of the mask, and therefore might not be able to create a np.array at all. But it's worth a try!

geomstats/geometry/special_euclidean.py
pchauchat: In the way the things are right now, there is no `point_type` attribute, only the default one. So I'll stick to this one, with `point_type` as optional argument, with None as base value.

geomstats/backend/numpy/__init__.py
```diff
@@ -191,7 +212,7 @@ def eval(x):
 
 
 def ndim(x):
-    return x.ndim
+    return _np.array(x).ndim
```
ninamiolane, line 215: I understand that you added this to prevent the error `int object has no attributed 'ndim'`. However, 1) this changes the behavior of numpy, 2) now geomstats does not follow the behavior of numpy.

On 1): now if someone writes gs.ndim(3) in numpy backend, there will be no error: this is confusing for numpy users.
On 2): we decided to follow numpy's API everywhere, to have a global logic.

Therefore, I would suggest dealing with this error __outside of the backend__, so that numpy stays consistent with itself and geomstats' stays consistent with numpy's logic. Happy to talk more about it though!

geomstats/backend/numpy/__init__.py
```diff
@@ -134,6 +138,23 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
+def assignment(x, values, indices, axis=0):
```
ninamiolane, line 141: Nice, thanks!

geomstats/backend/pytorch/__init__.py
```diff
@@ -412,6 +419,8 @@ def eval(x):
 
 
 def ndim(x):
+    if not torch.is_tensor(x):
```
ninamiolane, line 422: Same remark as before.

geomstats/backend/tensorflow/__init__.py
```diff
@@ -31,6 +40,103 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
+def get_mask_float(indices, mask_shape):
```
ninamiolane, line 43: Yessss 🎉

geomstats/geometry/special_euclidean.py
ninamiolane: @pchauchat update on this?

geomstats/backend/tensorflow/__init__.py
ninamiolane: update on this?

geomstats/geometry/special_euclidean.py
ninamiolane: update?

tests/test_special_euclidean.py
```diff
@@ -92,6 +92,15 @@ def setUp(self):
             elements = {
                 'point_1': point_1,
                 'point_2': point_2}
+            elements_matrices_all = {
+                key: group.matrix_from_vector(elements_all[key]) for key in
+                elements_all}
+
+        else:
+            elements_matrices_all = {
```
ninamiolane, line 100: This code seems to be duplicate: could we put it outside of the if/else?

pchauchat:
I basically took the day trying to sort the assignment thing, and deal with pytorch on some issues also. I wanted to push them so that other people working or wondering about assignments with Tensorflow could look at it.

I didn't have time to check the other TODOs and comments. Will do asap !

pchauchat:
I put back `ndim` to work as in numpy. I added a `dtype` input to `array`, as tensorflow is really picky on this.`t

I avoided the question of reducing the dimension of arrays, i.e. the question of whether `to_ndarray` should also be able to do that or not.

I needed to force a higher `atol` on the `belongs` of  SO(n), in matrix form and with `n=3`, because with Tensorflow the difference gets slightly above `1e-8`. I don't know if that only comes from the fact that it does 32-bits computations.

I will add some documentation to the assignment function (in a new PR ?)

PR #475, opened 2020-04-01
add grassmannian logarithm

opeltre:
Given two projectors `p1` and `p2`, a rotation from `p1` to `p2` is given by the square root of their composed symmetries (see Batzies et al in docstring) 
```
U = (2. p2 - 1) . (2. p1 - 1)
```
And the log can be computed as `Gr.metric.log(p2, p1) = (1/2) logm(U)`

Tests pass on numpy but we still lack a matrix logarithm on pytorch, would it be ok to use `scipy.linalg.logm` as a workaround @ninamiolane ? (on a separate backend PR!)

### DIFF
--- a/geomstats/geometry/grassmannian.py
+++ b/geomstats/geometry/grassmannian.py
@@ -31,7 +31,6 @@ class Grassmannian(EmbeddedManifold):
             dimension=dimension,
             embedding_manifold=Matrices(n, n))
 
-
     def belongs(self, point, tolerance=TOLERANCE):
         """Check if the point belongs to the manifold.
 
--- a/geomstats/geometry/grassmannian.py
+++ b/geomstats/geometry/grassmannian.py
@@ -88,7 +88,7 @@ class GrassmannianCanonicalMetric(RiemannianMetric):
         return mul(expm(tangent_vec), base_point, expm(-tangent_vec))
 
     def log(self, point, base_point):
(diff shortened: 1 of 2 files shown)


geomstats/geometry/grassmannian.py
```diff
@@ -83,4 +85,39 @@ def exp(self, vector, point):
         """
         expm = gs.linalg.expm
         mul = Matrices.mul
-        return mul(expm(vector), point, expm(-vector))
+        return mul(expm(tangent_vec), base_point, expm(-tangent_vec))
+
+    def log(self, point, base_point):
+        r"""Compute the riemannian logarithm of point w.r.t. base_point.
```
ninamiolane, line 91: Nit: Riemannian.

ninamiolane:
Flake8 error in Travis:
tests/test_grassmannian.py:65:1: W293 blank line contains whitespace

https://travis-ci.org/github/geomstats/geomstats/jobs/669700420?utm_medium=notification&utm_source=github_status

Could you fix it, and the riemannian -> Riemannian? Thanks!

geomstats/geometry/grassmannian.py
opeltre: Really? as adjective I wouldn't capitalize it would you?

opeltre: Hmm my bad it seems the usage is Riemannian in english indeed

PR #476, opened 2020-04-01
Symmetric matrices

ythanwerdas:
New class SymmetricMatrices with corresponding tests.


geomstats/geometry/symmetric_matrices.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, n):
+        assert isinstance(n, int) and n > 0
+        super(SymmetricMatrices, self).__init__(
+            dimension=int(n * (n + 1) / 2),
+            embedding_manifold=Matrices(n, n))
+        self.n = n
+
+    def belongs(self, mat, atol=TOLERANCE):
+        """Check if mat belongs to the manifold of symmetric matrices."""
```
nguigs, line 21: Although both are true, maybe consistently use vector space or manifold to describe the space

nguigs: Shouldn't it inherit from Matrices?

ythanwerdas: That is all the question... If a class inherits from Matrices, it has to use its constructor which asks for two parameters m and n (size of the matrix). Then your class is considered as a manifold of dimension m*n since Matrices constructor defines dimension=m*n.

ythanwerdas: For now, all matrix subgroups don't inherit from GeneralLinear so I consistently did not make the subvector space SymmetricMatrices inherit from Matrices. But I agree this should be modified.

geomstats/geometry/spd_matrices.py
```diff
@@ -111,7 +111,7 @@ def aux_differential_power(self, power, tangent_vec, base_point):
 
         Parameters
         ----------
-        power : int
+        power : float
```
ninamiolane, line 114: Next PR, add a description to this parameter? ;)

geomstats/geometry/symmetric_matrices.py
```diff
@@ -0,0 +1,22 @@
+"""The vector space of symmetric matrices."""
+
+from geomstats.geometry.embedded_manifold import EmbeddedManifold
+from geomstats.geometry.matrices import Matrices
+
+EPSILON = 1e-6
+TOLERANCE = 1e-12
+
+
+class SymmetricMatrices(EmbeddedManifold):
```
ninamiolane, line 10: Great, thanks!

tests/test_symmetric_matrices.py
```diff
@@ -0,0 +1,32 @@
...
+        """Set up the test."""
+        warnings.simplefilter('ignore', category=ImportWarning)
+
+        gs.random.seed(1234)
+
+        self.n = 3
+
+    def test_belongs(self):
+        """Test of belongs method."""
+        Sym_n = SymmetricMatrices(self.n)
```
ninamiolane, line 23: Ok since the coding style in unit tests is less strict, but remember than python (pep8) convention is to have all variable names in lower-cases: so this should be `sym_n`. It should only have an uppercase if you were using the whole class, eg:
Sym = SymmetricMatrices.

ninamiolane:
Actually, travis fails on the `belongs` method with pytorch:
https://travis-ci.org/github/geomstats/geomstats/builds/669706122?utm_medium=notification&utm_source=github_status

Could you address this and also my two NIT comments? Thanks!

ythanwerdas:
Yes! I finally found the problem! It comes from torch.all, the following line bugs:
`    return torch.from_numpy(_np.all(_np.array(x), axis=axis).astype(int))`
It seems that from_numpy is not able to convert an integer but an np.ndarray. So I added a `_np.array` before applying the `from_numpy`. Now it looks like:
`    return torch.from_numpy(_np.array(_np.all(_np.array(x), axis=axis).astype(int)))`
Any other suggestion instead?

PR #478, opened 2020-04-01
Add the KNN classifier.

YannCabanes:
Add the KNN classifier and its test.

### DIFF
--- a/geomstats/learning/knn.py
+++ b/geomstats/learning/knn.py
@@ -2,11 +2,6 @@
 
 from sklearn.neighbors import KNeighborsClassifier
 
-from geomstats.geometry.euclidean import Euclidean
-
-EUCLIDEAN = Euclidean(dimension=1)
-EUCLIDEAN_METRIC = EUCLIDEAN.metric
-
 
 class KNearestNeighborsClassifier(KNeighborsClassifier):
     """Classifier implementing the k-nearest neighbors vote on manifolds.
@@ -25,23 +20,21 @@ class KNearestNeighborsClassifier(KNeighborsClassifier):
         - [callable] : a user-defined function which accepts an
           array of distances, and returns an array of the same shape
           containing the weights.
-    metric : callable or string, default is Euclidean metric in dimension 1.
-        The distance metric to use.
-        The 'minkowski' string metric with p=2 is equivalent to the standard
-        Euclidean metric.
-        See the documentation of the DistanceMetric class in the scikit-learn
-        library for a list of available metrics.
-        If metric is "precomputed", X is assumed to be a distance matrix and
-        must be square during fit.
-    dimension : int, optional (default = 1)
-        Dimension of the manifold used if the metric parameter is callable.
-    metric_params : dict, optional (default = None)
-        Additional keyword arguments for the metric function.
     p : integer, optional (default = 2)
         Power parameter for the 'minkowski' string metric.
         When p = 1, this is equivalent to using manhattan_distance (l1),
         and euclidean_distance (l2) for p = 2.
         For arbitrary p, minkowski_distance (l_p) is used.
+    metric : string or callable, default 'minkowski'
+        The distance metric to use.
+        The default metric is minkowski, and with p=2 is equivalent to the
+        standard Euclidean metric.
+        See the documentation of the DistanceMetric class in the scikit-learn
+        library for a list of available metrics.
+        If metric is "precomputed", X is assumed to be a distance matrix and
+        must be square during fit.
+    metric_params : dict, optional (default = None)
+        Additional keyword arguments for the metric function.
     n_jobs : int or None, optional (default=None)
         The number of parallel jobs to run for neighbors search.
         ``None`` means 1; ``-1`` means using all processors.
@@ -72,9 +65,9 @@ class KNearestNeighborsClassifier(KNeighborsClassifier):
 
     def __init__(self, n_neighbors=5,
                  weights='uniform',
-                 metric=EUCLIDEAN_METRIC,
-                 metric_params=None,
                  p=2,
+                 metric='minkowski',
+                 metric_params=None,
                  n_jobs=None,
                  **kwargs):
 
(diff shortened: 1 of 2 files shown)


geomstats/learning/knn.py
```diff
@@ -0,0 +1,93 @@
...
+        weight function used in prediction.  Possible values:
+        - 'uniform' : uniform weights.  All points in each neighborhood
+          are weighted equally.
+        - 'distance' : weight points by the inverse of their distance.
+          in this case, closer neighbors of a query point will have a
+          greater influence than neighbors which are further away.
+        - [callable] : a user-defined function which accepts an
+          array of distances, and returns an array of the same shape
+          containing the weights.
+    metric : callable or string, default EuclideanMetric
```
ninamiolane, line 25: See my next comment a few lines later: this argument should be an object of the class RiemannianMetric, for consistency with KMeans and TangentPCA.

geomstats/learning/knn.py
```diff
@@ -0,0 +1,93 @@
...
+        if isinstance(metric, str):
+            distance = metric
+        else:
+            manifold_metric = metric(dimension=dimension)
+            distance = manifold_metric.dist
+
+        super().__init__(
+            n_neighbors=n_neighbors,
+            weights=weights,
+            algorithm='brute',
```
ninamiolane, line 88: Nice!

tests/test_knn.py
```diff
@@ -0,0 +1,46 @@
...
+
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.euclidean import EuclideanMetric
+from geomstats.learning.knn import KNearestNeighborsClassifier
+
+TOLERANCE = 1e-4
+
+
+class TestOnlineKmeansMethods(geomstats.tests.TestCase):
```
ninamiolane, line 11: Need to rename this class ;)

tests/test_knn.py
```diff
@@ -0,0 +1,46 @@
...
+    """Class defining the Poincare polydisk tests."""
+
+    def setUp(self):
+        """Define the parameters to test."""
+        gs.random.seed(1234)
+        self.n_neighbors = 3
+        self.metric = EuclideanMetric
```
ninamiolane, line 18: Should be an object, see other unit test files for example

ninamiolane: TODO (maybe next PR): test if the method works with a non-Euclidean metric?

geomstats/learning/knn.py
```diff
@@ -0,0 +1,93 @@
...
+        When p = 1, this is equivalent to using manhattan_distance (l1),
+        and euclidean_distance (l2) for p = 2.
+        For arbitrary p, minkowski_distance (l_p) is used.
+    n_jobs : int or None, optional (default=None)
+        The number of parallel jobs to run for neighbors search.
+        ``None`` means 1; ``-1`` means using all processors.
+
+    Attributes
+    ----------
+    classes_ : array of shape (n_classes,)
```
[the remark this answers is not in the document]
YannCabanes: Thanks, I modified for shape=[n_classes,] 
(It really is the number of classes, not the number of samples.)

tests/test_knn.py
YannCabanes: For the example of the next PR, I plot the result of a classification on the sphere, so it works for other metrics.

geomstats/learning/knn.py
ninamiolane: The docstring needs to be updated.

But now I see why you wanted the class and not the object: in order to be flexible with respect to the dimension?

In this case, I would recommend:
- `metric=None` in the `__init__` signature, as a default value,
- `if metric is None:` condition in the `__init__`, that can access the data's dimension (from metric' params? or something else?) and create the Euclidean metric of the corresponding dimension?

geomstats/learning/knn.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    def __init__(self, n_neighbors=5,
+                 weights='uniform',
+                 metric=EUCLIDEAN_DISTANCE,
+                 metric_params=None,
+                 p=2,
+                 n_jobs=None,
+                 **kwargs):
+
```
ninamiolane, line 80: Here you could have `if metric is None:`

geomstats/learning/knn.py
ninamiolane: What version do you prefer? Your current code, or this proposal in my comment above?

ninamiolane:
@YannCabanes Travis fails.

YannCabanes:
C'est bon  pour Travis maintenant.
Qu'est-ce que tu en penses ?
Le codecov ne passe pas, est-ce que c'est un problème ?

Yann

Le mer. 1 avr. 2020 à 21:59, Nina Miolane <notifications@github.com> a
écrit :

YannCabanes:
Ok, I like the idea. 
What do you mean by the good dimension ? 
We don't know the dimension before seeing the data. How would you like to do ?
If I let the default parameters from the scikit-learn library (metric='minkowski', p=2),
we have the Euclidean metric and it will adapt to the dimension of the data.
If its okay for you, I can let these default parameters.
What do you think ?
The users will see how to use other metrics with the example on the sphere.

ninamiolane:
Ah, you're right! We don't see the dimension of the data (= the "right" dimension) when we initialize the KNN object. Could we add the dimension of the data to come using `metric_params`?

Otherwise I like your alternative, `(metric='minkowski', p=2)`, but maybe `(metric='euclidean', p=2)` instead? Since the Euclidean is more known than the Minkowski.

YannCabanes:
I prefer minkowski to stick to the scikit-learn code. Otherwise I will have to write in the __init__ of the class, before __super__: if metric is 'euclidean', then metric = 'minkowski'.
If I use 'minkowski', everything is direct.
The link between the Minkowski metric and the Euclidean metric is detailed in the docstring when we present the parameter metric.
And in the docstring of the attribute 'effective_metric_' it is said that its value is automatically replaced by 'euclidean' if the `metric` parameter is set to 'minkowski' and the `p` parameter is set to 2.
What do you think ?

ninamiolane:
Oh ok, I hadn't realized that scikit-learn was using minkowski. Good with this then, thanks for explaining!

YannCabanes:
Some checks did not pass. Is it still fine ?

PR #481, opened 2020-04-01
Update decorator that performs automatic vectorization of functions/methods

ninamiolane:
This is the first of a series of PRs to tackle issue #338.

The vectorization decorator allows to tackle cases where vectorized functions have led to ambiguities, for example not being able to distinguish between:
- n input vectors of dimension n,
- one matrix of shape n x n.

This PR:
- simplifies the signature of the decorator, removing the `param_names` parameter,
- allows the decorator to work on methods,
    - by adding the flag `'else'` that refers to a method's input that is not expected to vectorized/not-vectorized: for example `self`,
- allows the decorator to work on methods with optional parameters,
with the corresponding unit tests.

This PR also updates docstrings in the vectorization.py file.

### DIFF
--- a/geomstats/vectorization.py
+++ b/geomstats/vectorization.py
@@ -13,7 +13,7 @@ POINT_TYPES_TO_NDIMS = {
 
 
 def squeeze_output_dim_0(initial_ndims, point_types):
-    """Determine if the output needs to squeeze a singular dimension 0.
+    """Determine if the output needs to be squeezed on dim 0.
 
     The dimension 0 is squeezed iff all input parameters:
     - contain one sample,
--- a/geomstats/vectorization.py
+++ b/geomstats/vectorization.py
@@ -52,11 +52,11 @@ def is_scalar(vect_array):
 
 
 def squeeze_output_dim_1(result, initial_shapes, point_types):
-    """Determine if the output needs to squeeze a singular dimension 1.
+    """Determine if the output needs to be squeezed on dim 1.
 
     This happens if the user represents scalars as array of shapes:
     [n_samples,] instead of [n_samples, 1]
-    Dimension 1 is squeezed by default if the return point type is a scalar.
+    Dimension 1 is squeezed by default if point_type is 'scalar'.
     Dimension 1 is not squeezed if the user inputs at least one scalar with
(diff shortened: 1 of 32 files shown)


geomstats/vectorization.py
```diff
@@ -21,78 +20,136 @@ def squeeze_output_dim_0(initial_ndims, point_types):
...
 def squeeze_output_dim_1(result, initial_shapes, point_types):
     """Determine if the output needs to squeeze a singular dimension 1.
 
     This happens if the user represents scalars as array of shapes:
     [n_samples,] instead of [n_samples, 1]
-
     Dimension 1 is squeezed by default if the return point type is a scalar.
     Dimension 1 is not squeezed if the user inputs at least one scalar with
     a singleton in dimension 1.
+
```
nguigs, line 62: The above docstring is not very clear:
- use passive mode for the title: output needs to be squeezed
- why 'return' in return point type?

geomstats/vectorization.py
```diff
@@ -21,78 +20,136 @@ def squeeze_output_dim_0(initial_ndims, point_types):
...
+    """Vectorize geomstats functions.
+
+    This decorator assumes that its function:
+    - works with fully-vectorized inputs,
+    - returns fully-vectorized outputs,
+
+    where "fully-vectorized" means that:
+    - one scalar has shape [1, 1],
+    - n scalars have shape [n, 1],
+    - on d-D vector has shape [1, d],
```
nguigs, line 100: typo on -> one

geomstats/vectorization.py
```diff
@@ -21,78 +20,136 @@ def squeeze_output_dim_0(initial_ndims, point_types):
...
+
+            for i_arg, arg in enumerate(args):
+                point_type = point_types[i_arg]
 
                 if point_type == 'scalar':
-                    vect_param = gs.to_ndarray(param, to_ndim=1)
-                    vect_param = gs.to_ndarray(vect_param, to_ndim=2, axis=1)
+                    arg = gs.array(arg)
+
+                if point_type == 'else' or arg is None:
```
nguigs, line 128: Nice!

PR #483, opened 2020-04-01
Adapt einsum functions in tf and pytorch backends

ninamiolane:
This PR adapts the einsum function is tf and pytorch backends:
- Pytorch was using numpy's einsum function the cover, now pytorch's einsum is used. This has led to a few modifications in the code and the unit tests.
- Tf's and pytorch's einsum are less flexible on ellipsis than numpy. For example, einsum('...i,...i->...', a, b) with a of shape [1, d] and b of shape [n, d] does not work with tf and pytorch. This PR adapts the einsum's code in both tf and pytorch backends to address this.

The unit tests are added correspondingly.

### DIFF
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -285,7 +285,52 @@ def sum(x, axis=None, keepdims=None, **kwargs):
 
 
 def einsum(*args, **kwargs):
-    return torch.from_numpy(_np.einsum(*args, **kwargs)).float()
+    einsum_str = args[0]
+    input_tensors_list = args[1:]
+
+    einsum_list = einsum_str.split('->')
+    input_str = einsum_list[0]
+    output_str = einsum_list[1]
+
+    input_str_list = input_str.split(',')
+
+    is_ellipsis = [input_str[:3] == '...' for input_str in input_str_list]
+    at_least_one_ellipsis = bool(_np.sum(is_ellipsis))
+
+    if at_least_one_ellipsis:
+        if len(input_str_list) > 2:
+            raise NotImplementedError(
+                'Ellipsis support not implemented for >2 input tensors')
+        tensor_a = input_tensors_list[0]
+        tensor_b = input_tensors_list[1]
+        n_tensor_a = tensor_a.shape[0]
+        n_tensor_b = tensor_b.shape[0]
+
+        if n_tensor_a != n_tensor_b:
+            if n_tensor_a == 1:
+                tensor_a = squeeze(tensor_a, axis=0)
+                input_prefix_list = ['', 'n']
+                output_prefix = 'n'
+            elif n_tensor_b == 1:
+                tensor_b = squeeze(tensor_b, axis=0)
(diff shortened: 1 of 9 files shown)


tests/test_backends.py
```diff
@@ -237,3 +237,34 @@ def test_cumsum(self):
...
+
+        np_result = _np.einsum('...i,...i->...', np_array_1, np_array_2)
+        gs_result = gs.einsum('...i,...i->...', array_1, array_2)
+
+        self.assertAllCloseToNp(gs_result, np_result)
+
+        np_array_1 = _np.array([[1, 1], [1, 1]])
+        np_array_2 = _np.array([[1, 1]])
+        array_1 = gs.array([[1, 1], [1, 1]])
+        array_2 = gs.array([[1, 1]])
```
ythanwerdas, line 255: The test could be even more faithful if [1, 1] were not used everywhere. For example, if in einsum code you invert the rows or the columns of array_1 before performing einsum, there is no way to see this error in the test.

geomstats/backend/tensorflow/__init__.py
```diff
@@ -252,9 +252,9 @@ def einsum(equation, *inputs, **kwargs):
     input_str_list = input_str.split(',')
 
     is_ellipsis = [input_str[:3] == '...' for input_str in input_str_list]
-    at_least_one_ellipsis = bool(_np.sum(is_ellipsis))
+    all_ellipsis = bool(_np.prod(is_ellipsis))
 
-    if at_least_one_ellipsis:
+    if all_ellipsis:
```
ythanwerdas: Why is it different between pytorch (at_least_one_ellipsis) and tf (all_ellipsis)?

geomstats/backend/pytorch/__init__.py
```diff
@@ -285,7 +285,52 @@ def sum(x, axis=None, keepdims=None, **kwargs):
...
+                'Ellipsis support not implemented for >2 input tensors')
+        tensor_a = input_tensors_list[0]
+        tensor_b = input_tensors_list[1]
+        n_tensor_a = tensor_a.shape[0]
+        n_tensor_b = tensor_b.shape[0]
+
+        if n_tensor_a != n_tensor_b:
+            if n_tensor_a == 1:
+                tensor_a = squeeze(tensor_a, axis=0)
+                input_prefix_list = ['', 'n']
```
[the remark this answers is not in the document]
ninamiolane: Aaaah good point! Thanks.

geomstats/backend/tensorflow/__init__.py
ninamiolane: Good catch! I had tried something different and forgot to change back.

tests/test_backends.py
ninamiolane: Good point! I changed.

ninamiolane:
Yes, I added decorators because the new version of pytorch's einsum made these tests fail. This is because the previous version of pytorch's einsum was actually np's einsum. So these test were passing in pytorch, but in fact... they weren't!

Issue #423, opened 2020-02-07
Rename dimension into dim everywhere


ninamiolane:
Currently, `dimension` and `dim` are used. We can use `dim` everywhere, as it is self-explanatory.

ninamiolane:
Duplicate.

PR #484, opened 2020-04-02
add SOn, move SO3 

opeltre:
SpecialOrthogonal for general n is in `special_orthogonal.py`, and for n = 3 is in `so3.py`: I only changed the import statements accordingly in the tests and examples to not break anything.

Still need to add unit tests for SOn :)

@pchauchat @nguigs


examples/plot_bch_so3.py
```diff
@@ -19,7 +19,7 @@
 
 import geomstats.backend as gs
 from geomstats.geometry.skew_symmetric_matrices import SkewSymmetricMatrices
-from geomstats.geometry.special_orthogonal import SpecialOrthogonal
+from geomstats.geometry.so3 import SpecialOrthogonal
```
ninamiolane, line 22: Naming inconsistency in the .py files:
- special_orthogonal.py
versus
- so3.py

We should:
- either have special_orthogonal and special_orthogonal_3,
- or have son and so3.

What do you prefer? If we move to `SO`, then the objects should probably also be called `SO` and `SO3`, and then we should probably rename GeneralLinear to GL, etc etc.

I would slightly prefer staying very explanatory at first (keep special_orthogonal.py, create special_orthogonal_3.py) - so that non-mathematician users have clear keywords to google if they don't know these names - but happy to hear others opinions. @opeltre @nguigs @pchauchat ?

geomstats/geometry/grassmannian.py
```diff
@@ -50,6 +50,16 @@ def belongs(self, point, tolerance=TOLERANCE):
             'The Grassmann `belongs` is not implemented.'
             'It shall test whether p*=p, p^2 = p and rank(p) = k.')
 
+    def is_tangent(self, tangent_vec, point):
```
ninamiolane, line 53: Missing docstrings in these 2 methods.

ninamiolane: `point` should be called `base_point` and (maybe?) default to `None` if there is an intuitive default case?

geomstats/geometry/so3.py
```diff
@@ -0,0 +1,1664 @@
...
+                         - 1. / 480.]
+
+
+class SpecialOrthogonal(LieGroup, EmbeddedManifold):
+    """Class for the special orthogonal group SO(n).
+
+    i.e. the Lie group of rotations.
+    """
+
+    def __init__(self, n, point_type=None, epsilon=0.):
```
ninamiolane, line 34: This is SO3, why is there a 'n' parameter?

geomstats/geometry/special_orthogonal.py
```diff
@@ -1,1664 +1,34 @@
-"""The special orthogonal group SO(n).
+"""Exposes the `SpecialOrthogonal` group class."""
```
ninamiolane, line 1: We should agree on a convention for docstrings at the top of the files. But right now, the implicit convention is not to be as concise as possible. Thus: `SpecialOrthogonal class.` would work. 

Being concise also allows avoiding inconsistency for example between:
"""Exposes the `SpecialOrthogonal` group class."""
and
"""Module exposing `Grassmannian` and `GrassmannianMetric` classes."""

Inconsistencies may not seem like a big deal, but on the automatic documentation website where all the .py files are listed together, it looks weird.

geomstats/geometry/so3.py
```diff
@@ -0,0 +1,1664 @@
...
+            the dimension of the euclidean space that SO(n) acts upon
+        point_type : str, {'vector', 'matrix'}, optional
+            if None is given, point_type is set to 'vector for dimension 3
+            and matrix otherwise
+        epsilon : float, optional
+            precision to use for calculations involving potential divison by in
+            rotations
+            default: 0
+        """
+        assert isinstance(n, int) and n > 1
```
ninamiolane, line 49: This whole file is in fact the copy-paste of son, right? We should adapt and change n to 3, otherwise, users will be very confused.

ninamiolane: A of doing it would be:
- rename `SpecialOrthogonal` to `SpecialOrthogonal3` and have it inherit from `SpecialOrthogonal`
- Remove `n` parameter in __init__
- set `self.n = 3`
- Delete methods that do not explicitly require the 3D, so that the parent's methods are called.

geomstats/geometry/special_orthogonal.py
```diff
@@ -1,1664 +1,34 @@
...
-            assert gs.ndim(exp_bar) == 2, gs.ndim(exp_bar)
-
-        elif point_type == 'matrix':
-            points = self.rotation_vector_from_matrix(points)
-            exp_bar = self.exponential_barycenter(
-                points, weights, point_type='vector')
-            exp_bar = self.matrix_from_rotation_vector(exp_bar)
-
-        return exp_bar
+        self.dim = int((n * (n - 1)) / 2)
```
ninamiolane, line 11: Yes to rename dimension into `dim`, see issue #423.

But let's keep this change to a PR that does only this change of naming and does it everywhere? Unless the code is already inconsistent?

geomstats/geometry/special_orthogonal.py
```diff
@@ -1,1664 +1,34 @@
...
+
+    @classmethod
+    def is_tangent(cls, vector):
+        """Check whether vector is a skew-symmetric matrix."""
+        return cls.equal(
+            cls.transpose(vector),
+            - vector)
+
+    @classmethod
+    def to_tangent(cls, vector):
```
ninamiolane, line 32: These docstrings need parameter descriptions.

ninamiolane:
Also, not separating all the way SOn and SO3 breaks a lot of unit tests.

examples/plot_bch_so3.py
pchauchat: Non-mathematician people will most probably look for "3D rotations", which they would find by googling either name I think. The only thing is that, if at some point you decide to create files for the Lie algebras, `so3` will be confusing, but that's a little far-fetched.

ninamiolane:
@opeltre any update on this? (see Travis' errors).

nguigs:
This was done in #582 and #613

PR #485, opened 2020-04-02
Add an example of the KNN classifier on the sphere.

YannCabanes:
Add an example of the KNN classifier on the sphere and its corresponding test.

### DIFF
--- a/examples/plot_knn_s2.py
+++ b/examples/plot_knn_s2.py
@@ -27,8 +27,8 @@ def main():
         kappa=10,
         n_samples=n_samples_per_dataset)
     training_dataset = gs.concatenate((dataset_1, dataset_2), axis=0)
-    labels_dataset_1 = gs.zeros([n_samples_per_dataset], dtype=int)
-    labels_dataset_2 = gs.ones([n_samples_per_dataset], dtype=int)
+    labels_dataset_1 = gs.zeros([n_samples_per_dataset], dtype=gs.int64)
+    labels_dataset_2 = gs.ones([n_samples_per_dataset], dtype=gs.int64)
     labels = gs.concatenate((labels_dataset_1, labels_dataset_2))
     target = sphere.random_uniform(n_samples=n_targets)
 
--- a/examples/plot_knn_s2.py
+++ b/examples/plot_knn_s2.py
@@ -32,13 +32,14 @@ def main():
     labels = gs.concatenate((labels_dataset_1, labels_dataset_2))
     target = sphere.random_uniform(n_samples=n_targets)
 
-    neigh = KNearestNeighborsClassifier(n_neighbors=2,
-                                        metric=sphere_distance)
+    neigh = KNearestNeighborsClassifier(
+        n_neighbors=2,
+        metric=sphere_distance)
     neigh.fit(training_dataset, labels)
     target_labels = neigh.predict(target)
 
     plt.figure(0)
-    ax = plt.subplot(111, projection="3d")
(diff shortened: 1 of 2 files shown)


examples/plot_knn_s2.py
```diff
@@ -0,0 +1,68 @@
...
+    n_targets = 200
+
+    dataset_1 = sphere.random_von_mises_fisher(
+        kappa=10,
+        n_samples=n_samples_per_dataset)
+    dataset_2 = - sphere.random_von_mises_fisher(
+        kappa=10,
+        n_samples=n_samples_per_dataset)
+    training_dataset = gs.concatenate((dataset_1, dataset_2), axis=0)
+    labels_dataset_1 = gs.zeros([n_samples_per_dataset], dtype=int)
```
ninamiolane, line 30: Use gs.int instead of int: it will make the example more portable to pytorch and tensorflow later ;)

examples/plot_knn_s2.py
```diff
@@ -0,0 +1,68 @@
...
+    labels_dataset_2 = gs.ones([n_samples_per_dataset], dtype=int)
+    labels = gs.concatenate((labels_dataset_1, labels_dataset_2))
+    target = sphere.random_uniform(n_samples=n_targets)
+
+    neigh = KNearestNeighborsClassifier(n_neighbors=2,
```
ninamiolane, line 35: NIT: for indentation, I usually prefer the syntax:
```
neigh = KNearestNeighborsClassifier(
    n_neighbors=2, metric=sphere_distance)
```

because if you change the variable name `neigh` to a name with a different number of characters, you have to re-align all your code, whereas the above syntax does not need it!

examples/plot_knn_s2.py
```diff
@@ -0,0 +1,68 @@
...
+                                        metric=sphere_distance)
+    neigh.fit(training_dataset, labels)
+    target_labels = neigh.predict(target)
+
+    plt.figure(0)
+    ax = plt.subplot(111, projection="3d")
```
ninamiolane, line 41: Single quote instead of double quote is the convention in geomstats

examples/plot_knn_s2.py
```diff
@@ -0,0 +1,68 @@
...
+    plt.title('Training set')
+    sphere_plot = visualization.Sphere()
+    sphere_plot.draw(ax=ax)
+    for i_label in range(n_labels):
+        points_label_i = training_dataset[labels == i_label, ...]
+        sphere_plot.draw_points(ax=ax, points=points_label_i)
+
+    plt.figure(1)
+    ax = plt.subplot(111, projection="3d")
```
ninamiolane, line 50: Ditto

examples/plot_knn_s2.py
YannCabanes: I have the following error:
AttributeError: module 'geomstats.backend' has no attribute 'int'

YannCabanes: The nose2 test does not pass if I change int for gs.int.
Is it a recent change ?

YannCabanes: Do I add int in the backend ?

YannCabanes:
I did not change int for gs.int 
If I do, I have the following error:
AttributeError: module 'geomstats.backend' has no attribute 'int'
The nose2 test does not pass if I change int for gs.int.
Do I add "int" in the backend ?

YannCabanes:
That's fine now, I chose to change int for gs.int64

PR #486, opened 2020-04-02
lie algebra and skew-sym

nguigs:
Add is_skew_sym method and corresponding tests

### DIFF
--- a/geomstats/backend/tensorflow/__init__.py
+++ b/geomstats/backend/tensorflow/__init__.py
@@ -136,6 +136,10 @@ def assignment(x, values, indices, axis=0):
         return x
 
 
+def get_from_sparse(indices, data, shape):
+    return tf.sparse.to_dense(tf.SparseTensor(indices, data, shape))
+
+
 def gather(*args, **kwargs):
     return tf.gather(*args, **kwargs)
 
@@ -172,13 +176,6 @@ def cond(*args, **kwargs):
 
 def reshape(*args, **kwargs):
     return tf.reshape(*args, **kwargs)
-
-
-def flatten(x):
-    """Collapses the tensor into 1-D.
-
-    Following https://www.tensorflow.org/api_docs/python/tf/reshape"""
-    return tf.reshape(x, [-1])
 
 
 def arange(*args, **kwargs):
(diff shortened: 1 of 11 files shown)


geomstats/backend/tensorflow/__init__.py
```diff
@@ -31,6 +31,10 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
+def get_from_sparse(indices, data, shape):
```
ninamiolane, line 34: is `get_from_sparse` the name of the numpy function?

If not, does numpy have a similar function and could we use its name and syntax?

geomstats/geometry/lie_algebra.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, dimension, n):
         dimension: int
             The dimension of the Lie algebra as a real vector space
         n: int
-            The amount of rows and columns in the matrx representation of the
+            The amount of rows and columns in the matrix representation of the
             Lie algebra
```
ninamiolane, line 25: "." missing at the end of the parameter descriptions.

(Sorry, I know it's not your code, thanks for cleaning it already!)

geomstats/geometry/skew_symmetric_matrices.py
```diff
@@ -25,11 +30,24 @@ def __init__(self, n):
...
+
+        else:
+            for row in gs.arange(n - 1):
+                for col in gs.arange(row + 1, n):
+                    self.basis[loop_index, row, col] = 1
+                    self.basis[loop_index, col, row] = -1
+                    loop_index += 1
+
+    def belongs(self, mat, atol=TOLERANCE):
+        """Check if mat belongs to the vector space of symmetric matrices."""
```
ninamiolane, line 49: Parameter descriptions?

tests/test_lie_group.py
```diff
@@ -1,6 +1,4 @@
-"""
-Unit tests for Lie groups.
-"""
+"""Unit tests for Lie groups."""
```
ninamiolane, line 1: Thanks!

geomstats/geometry/skew_symmetric_matrices.py
```diff
@@ -25,11 +30,24 @@ def __init__(self, n):
         self.basis = gs.zeros((dimension, n, n))
         loop_index = 0
 
-        for row in gs.arange(n - 1):
-            for col in gs.arange(row + 1, n):
-                self.basis[loop_index, row, col] = 1
-                self.basis[loop_index, col, row] = -1
-                loop_index += 1
+        if tests.tf_backend():
```
ninamiolane, line 33: - Maybe create a method: `canonical_basis`?

- We should also avoid testing for backend in the code (`tf_backend()`): can't we abstract this by calling only backend functions?

geomstats/backend/tensorflow/__init__.py
nguigs: - `get_from_sparse` is not the numpy of a numpy function
- I don't know any similar function per se, there a different ways of doing it in numpy e.g. fromfunction. My idea here would be to add this function to all backends using each specific tools, just like `to_ndarray` does

geomstats/geometry/skew_symmetric_matrices.py
nguigs: in the long term yes, maybe once we have a `get_from_sparse` for all backends, or stg similar with assignements

geomstats/backend/tensorflow/__init__.py
ninamiolane: OK! Maybe call it `sparse_array()`? Sorry I'm nit-picking, but get_from_sparse is not super intuitive?

nguigs: Yes but it's not a sparse array in the usual sense, but a dense one. Sparse refers to scipy.sparse or SparseTensor which are different data types that don't store all the zeros. The proposed method converts a sparse array to a dense (usual) one.

nguigs: Maybe `array_from_sparse`?

ninamiolane: Got it! `array_from_sparse` would be great: the naming is also closer to our conversion functions.

PR #494, opened 2020-04-02
Yt symmetric matrices basic functions

ythanwerdas:
Function from_vector_to_diagonal_matrix in the 3 backends.
Function expm in symmetric_matrices (using the previous function).
Corresponding test.


geomstats/backend/numpy/__init__.py
```diff
@@ -203,3 +203,10 @@ def cumprod(x, axis=0):
 
 def copy(x):
     return x.copy()
+
+
+def from_vector_to_diagonal_matrix(x):
```
ninamiolane, line 208: Awesome! This vectorizes the numpy function `diag`, right?

I was just wondering if we should name it `diag`, but actually probably not: otherwise it might be confusing for users since it's not working the exact same way as the numpy function does.

geomstats/geometry/symmetric_matrices.py
```diff
@@ -20,3 +21,13 @@ def __init__(self, n):
     def belongs(self, mat, atol=TOLERANCE):
         """Check if mat belongs to the vector space of symmetric matrices."""
         return Matrices(self.n, self.n).is_symmetric(mat=mat, atol=atol)
+
+    def expm(x):
```
ninamiolane, line 25: Wooohooo 🎉 !

geomstats/geometry/symmetric_matrices.py
```diff
@@ -20,3 +21,13 @@ def __init__(self, n):
...
+        """Compute the matrix exponential."""
```
ninamiolane, line 26: Missing description of parameters in docstring

ninamiolane: No `self` nor `cls` in input parameters? Then maybe make it a static method: @staticmethod decorator.

geomstats/backend/numpy/__init__.py
ythanwerdas: Exactly! But the name could be shortened anyway

geomstats/geometry/symmetric_matrices.py
ythanwerdas: Oh I forgot `self`. For now, I prefer it to be a standard method.

ythanwerdas: At some point, it could be useful to make all the classes consistent with each other for standard/static/class methods. It seems to me that, for now, the choice is a bit random. Or if there is a rule, it is not clear to me.

PR #498, opened 2020-04-02
Remove unnecessary list comprehension

nguigs:
Adresses DeepSource performance issues


geomstats/geometry/product_riemannian_metric.py
```diff
@@ -31,9 +31,9 @@ def __init__(self, metrics, default_point_type='vector', n_jobs=1):
         self.default_point_type = default_point_type
         self.n_jobs = n_jobs
 
-        sig_0 = sum([sig[0] for sig in signatures])
-        sig_1 = sum([sig[1] for sig in signatures])
-        sig_2 = sum([sig[2] for sig in signatures])
+        sig_0 = sum(sig[0] for sig in signatures)
+        sig_1 = sum(sig[1] for sig in signatures)
+        sig_2 = sum(sig[2] for sig in signatures)
```
ninamiolane, line 36: Nice!

geomstats/geometry/spd_matrices.py
```diff
@@ -367,7 +367,8 @@ def __init__(self, n, power_affine=1):
         self.space = SPDMatrices(n)
         self.power_affine = power_affine
 
-    def _aux_inner_product(self, tangent_vec_a, tangent_vec_b, inv_base_point):
+    @staticmethod
+    def _aux_inner_product(tangent_vec_a, tangent_vec_b, inv_base_point):
```
ninamiolane, line 371: What are these? Why inv_base_point?

geomstats/learning/frechet_mean.py
```diff
@@ -253,9 +253,7 @@ def _adaptive_gradient_descent(points,
 
     sum_weights = gs.sum(weights)
 
-    n_init = len(init_points)
-
-    if n_init == 0:
+    if init_points is None:
```
ninamiolane, line 256: Will this also works when `init_points` is given as `[]` by the user?

nguigs: According to DeepSource setting [] as default value is not recommended. There is no other reason a user would pass []?

geomstats/geometry/spd_matrices.py
nguigs: This was already here, I just added the @staticmethod decorator according to DeepSource

nguigs: inv stands for inverse.

geomstats/learning/frechet_mean.py
nguigs: In fact even if different init_points are passed, only the first one is used

PR #502, opened 2020-04-02
Add the Agglomerative Hierarchical Clustering (AHC) on manifolds.

YannCabanes:
Add the Agglomerative Hierarchical Clustering (AHC) on manifolds and its test.


geomstats/learning/ahc.py
```diff
@@ -0,0 +1,118 @@
+"""The Agglomerative Hierarchical Clustering (AHC) on manifolds."""
+
+from sklearn.cluster import AgglomerativeClustering
+
+import geomstats.backend as gs
+
+
+class AgglomerativeHierarchicalClustering(AgglomerativeClustering):
```
ninamiolane, line 8: Did you modify this class, w.r.t. to scikit-learn? Or are we "only" using it?

tests/test_ahc.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TestAgglomerativeHierarchicalClustering(geomstats.tests.TestCase):
+    """Class defining the Poincare polydisk tests."""
+
+    def setUp(self):
+        """Define the parameters to test."""
+        gs.random.seed(1234)
+        self.n_clusters = 2
+        self.dimension = 2
+        self.space = Euclidean(dimension=self.dimension)
```
ninamiolane, line 17: Let's test this with a non-Euclidean metric, to see if it works with geomstats' distances? It'll be good to have it in the test, not only in the examples.

YannCabanes: I will show that it is working well with other metrics in my next PR with an example on the sphere.
Is this fine for you ?

geomstats/learning/ahc.py
YannCabanes: I made it easier to use thanks to the distance parameter instead of the affinity parameter which in scikit-learn, when callable, should be a function that takes the data as input and return a distance matrix.
In my code, we just need to specify the distance and this function is automatically defined.
Plus I changed the default parameter of the linkage parameter for it to be compatible with non-euclidean metrics.

tests/test_ahc.py
YannCabanes: Ok, I will add other metrics in the test too.

tests/test_ahc.py
```diff
@@ -0,0 +1,57 @@
+"""Unit tests for the KNN classifier."""
```
nguigs, line 1: Wrong docstring!

tests/test_ahc.py
```diff
@@ -0,0 +1,57 @@
...
+
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.euclidean import Euclidean
+from geomstats.geometry.hypersphere import Hypersphere
+from geomstats.learning.ahc import AgglomerativeHierarchicalClustering
+
+
+class TestAgglomerativeHierarchicalClustering(geomstats.tests.TestCase):
+    """Class defining the Poincare polydisk tests."""
```
nguigs, line 11: Wrong docstring

tests/test_ahc.py
```diff
@@ -0,0 +1,57 @@
...
+        """Define the parameters to test."""
+        gs.random.seed(1234)
+
+    @geomstats.tests.np_only
+    def test_fit_euclidean_distance(self):
+        """Test the 'fit' class method using the Euclidean distance."""
+        n_clusters = 2
+        dimension = 2
+        space = Euclidean(dimension=dimension)
+        distance = space.metric.dist
```
nguigs, line 23: Nice! Could you duplicate this test with a string distance argument so that the condition `if isinstance(distance, str)` is met and actually tested? We would reach 100% covering then!

geomstats/learning/ahc.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Recursively merges the pair of clusters that minimally increases
+    a given linkage distance.
+
+    Parameters
+    ----------
+    n_clusters : int or None, default=2
+        The number of clusters to find. It must be ``None`` if
+        ``distance_threshold`` is not ``None``.
+    distance : str or callable, default='euclidean'
```
nguigs, line 19: Cool! Could you add an exemple section in the docstring so we know how to pass a geomstats geodesic distance? As in https://github.com/geomstats/geomstats/blob/6cf2aba8f917718028e445207eb3705fb56c98ab/geomstats/learning/online_kmeans.py#L142 for example:
```
    Example
    -------
    >>> from geomstats.geometry.hypersphere import Hypersphere
    >>> from geomstats.learning.ahc import AgglomerativeHierarchicalClustering
    >>> sphere = Hypersphere(dimension=2)
    >>> X = sphere.random_uniform(10)
    >>> distance = sphere.metric.dist
    >>> AgglomerativeHierarchicalClustering(distance=distance)
    >>> clustering.fit(X)
```

PR #509, opened 2020-04-03
Assignment debug, documentation, and unit test

pchauchat:
Debugged assignment in non-trivial cases (e.g., vectorizing along axis!=0), and allowed assigning even when the former element was non-zero. Added unit tests for both cases.

Note that assignment_by_sum seems a bit redundant, especially in the case of numpy and pytorch.

Added corresponding docstrings.

I also thought of a way of documenting evolving shapes along a provided axis :
 new_x: array_like, shape=[dimension[:axis], n_samples, dimension[axis:]]
What do you think @ninamiolane @nguigs @nkoep  ?

(Also corrected a typo in plot_bch_so3)

### DIFF
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -148,8 +148,9 @@ def assignment(x, values, indices, axis=0):
         Initial array.
     values: {float, list(float)}
         Value or list of values to be assigned.
-    indices: {tuple, list(tuple)}
-        Single tuple, or list of tuples of indices where values are assigned.
+    indices: {int, tuple, list(int), list(tuple)}
+        Single int or tuple, or list of ints or tuples of indices where value
+        is assigned.
         If the length of the tuples is shorter than ndim(x), values are
         assigned to each copy along axis.
     axis: int, optional
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -173,7 +173,9 @@ def assignment(x, values, indices, axis=0):
     if not isinstance(values, list):
         values = [values] * len(indices)
     for (nb_index, index) in enumerate(indices):
-        if len(indices[0]) < len(shape(x)):
+        if not isinstance(index, tuple):
+            index = index,
+        if len(index) < len(shape(x)):
             for n_axis in range(shape(x)[axis]):
                 extended_index = index[:axis] + (n_axis,) + index[axis:]
                 x_new[extended_index] = values[nb_index]
--- a/geomstats/backend/numpy/__init__.py
+++ b/geomstats/backend/numpy/__init__.py
@@ -140,7 +140,7 @@ def get_mask_i_float(i, n):
 
 
 def assignment(x, values, indices, axis=0):
-    """Create a vectorized binary mask.
+    """Assign values at given indices of an array.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


geomstats/backend/tensorflow/__init__.py
```diff
@@ -40,25 +40,33 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
-def get_mask_float(indices, mask_shape):
+def get_mask_float(indices, mask_shape, dtype=float32):
```
ninamiolane, line 43: Ah, good catch! 🎉 

Actually, this could just be called `index_mask` or `indices_mask` or `mask_from_indices` and we have the `dtype` as parameter. Any chance you could update this?

geomstats/backend/tensorflow/__init__.py
```diff
@@ -70,57 +78,211 @@ def get_mask_float(indices, mask_shape):
...
     axis: int
         Axis along which the mask is vectorized.
+    dtype: dtype
+        Type of the returned array.
 
     Returns
     -------
-    tf_mask : array
-    TODO(pchauchat): give shape of the output according to guideline
+    tf_mask : array, shape=[mask_shape[:axis], n_samples, mask_shape[axis:]]
```
ninamiolane, line 126: Yes for this syntax: could you update the contributing file as well?

tests/test_special_euclidean.py
```diff
@@ -159,7 +159,6 @@ def test_random_and_belongs(self):
         base_point = self.group.random_uniform()
         result = self.group.belongs(base_point)
         expected = gs.array([True])
-        print(result, expected)
         self.assertAllClose(result, expected)
 
     # @geomstats.tests.np_and_tf_only
```
ninamiolane, line 164: Left-over comment (I know it's not your code, but could you still delete it?)

geomstats/backend/numpy/__init__.py
```diff
@@ -140,14 +140,42 @@ def get_mask_i_float(i, n):
 
 
 def assignment(x, values, indices, axis=0):
```
ninamiolane, line 142: Could you explain why assignment is complicated in numpy? I'm just worried that we will loose a lot in performances if this part is not optimized.

For example, `x` shouldn't be copied, it should be modified, right?

geomstats/backend/tensorflow/__init__.py
```diff
@@ -40,25 +40,33 @@ def get_mask_i_float(i, n):
...
-    tf_mask : array
+    tf_mask : array, shape=[mask_shape]
     """
     np_mask = _np.zeros(mask_shape)
-    if ndim(array(indices)) <= 1 and ndim(np_mask) == 1:
-        indices = list(indices)
+    single_index = not isinstance(indices, list)
+
+    if single_index:
+        indices = [indices]
```
ninamiolane, line 63: The edge case single_index is not tested in the unit tests.

geomstats/backend/tensorflow/__init__.py
```diff
@@ -40,25 +40,33 @@ def get_mask_i_float(i, n):
...
 
-    if ndim(array(indices)) == 1 and ndim(np_mask) > 1:
+    for (nb_index, index) in enumerate(indices):
+        if not isinstance(index, tuple):
+            indices[nb_index] = (index,)
+
+    if single_index:
         if len(indices) != len(mask_shape):
```
ninamiolane, line 70: Ditto

geomstats/backend/tensorflow/__init__.py
```diff
@@ -70,57 +78,211 @@ def get_mask_float(indices, mask_shape):
...
+
+    Notes
+    -----
+    If a single value is provided, it is assigned at all the indices.
+    If a list is given, it must have the same length as indices.
+    """
+    if not isinstance(values, list):
+        return assignment_single_value_by_sum(x, values, indices, axis)
+
+    else:
```
ninamiolane, line 203: This else is not tested.

geomstats/backend/numpy/__init__.py
pchauchat: The problem is that, since Tensorflow does not allow modifying (or at least, I don't know how), if we want behaviors to be as similar as possible, this is needed. In some cases (weird, but not that uncommon), having backends which modify an object or create a new one might lead to completely different outcomes, for instance in lists.
However, there's also the possibility of specifying in the doc that behaviors differ, asking them to be cautious.

geomstats/backend/tensorflow/__init__.py
pchauchat: True, the way it is coded, single_index is tested in `assignment_single_value`, which always call `get_mask_float` (or the vectorized version) with a list, even of length one. I'll correct it.

pchauchat: Or should it be kept in case get_mask_float is directly called ? But with an additional test ?

geomstats/backend/tensorflow/__init__.py
pchauchat: I had added additonal tests on an awaiting PR. Do I bring them on this one ?

geomstats/backend/tensorflow/__init__.py
pchauchat: Just to be sure I understand correctly:
- You want to rename the function, or one of its parameters ?
- In the first case, there would only be two parameters, `indices` and `dtype` ?

tests/test_special_euclidean.py
pchauchat: It might be mine, as I was looking at which tests now worked with Pytorch and Tensorflow. I just forgot to remove them when doing the PR.
Same as above, in the next one they're gone.

geomstats/backend/tensorflow/__init__.py
ninamiolane: If `get_mask_float` is always called with a list, then maybe we can just remove this `if`?

geomstats/backend/tensorflow/__init__.py
ninamiolane: Let's wait for the other PR then!

tests/test_special_euclidean.py
ninamiolane: OK! Can't wait for this next one :p

pchauchat:
About the renaming and changes to reduce Deepsource errors, since they all apply to the backends on which you are, I think, currently working @nkoep, wouldn't it be better to wait until these modifications are merged and rebase on them ?

There's already a commit with renaming, but I can revert it.

Issue #246, opened 2019-11-18
tangent PCA bug


nbara:
Hello,

I'm trying to apply tangent PCA on SPD matrices, something along the lines of: https://github.com/geomstats/geomstats/blob/master/examples/tangent_pca_so3.py

The following code doesn't work:

```python
manifold = SPDMatricesSpace(10)
X = manifold.random_uniform(n_samples=140)
# X = manifold.vector_from_symmetric_matrix(X)   # this doesn't work either
mean = manifold.metric.mean(X[None, :])
tpca = TangentPCA(metric=manifold.metric)
tpca = tpca.fit(X)
tangent_projected_data = tpca.transform(X)
```

Passing the SPD matrices as vectors, I get the following error : 

<details>

```bash
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/learning/pca.py", line 128, in fit
    self._fit(X, base_point, point_type)
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/learning/pca.py", line 164, in _fit
    tangent_vecs = self.metric.log(X, base_point=base_point)
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/spd_matrices_space.py", line 233, in log
    sqrt_base_point = gs.linalg.sqrtm(base_point)
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/backend/numpy_linalg.py", line 19, in sqrtm
    scipy.linalg.sqrtm, signature='(n,m)->(n,m)')(x)
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/scipy/linalg/_matfuncs_sqrtm.py", line 170, in sqrtm
    T, Z = schur(A)
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/scipy/linalg/decomp_schur.py", line 126, in schur
    raise ValueError('expected square matrix')
ValueError: expected square matrix
```

</details>

When passing a matrix, I get a different error in `riemannian_metric`:

<details>

```bash
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/riemannian_metric.py", line 389, in <lambda>
    lambda i, m, v, sq: while_loop_body(i, m, v, sq),
  File "/Users/nicolas/anaconda3/lib/python3.7/site-packages/geomstats/riemannian_metric.py", line 340, in while_loop_body
[...]

PR #510, opened 2020-04-03
Tangent PCA

nguigs:
- fix symmetric_matrix_from_vector and its inverse and add these methods to symmetric_matrices. They could be removed from the SPDMatrices class when it will inherit from SymmetricMatrices. Only works for numpy so far.
- implement tangent PCA for `point_type='matrix'`, thus addressing issue #246 
- add inverse_tranfrom method (to override the inherited one from sklearn). It outputs points on the manifold and not tangent vectors
- add unit tests for PCA methods


geomstats/geometry/spd_matrices.py
```diff
@@ -38,36 +38,20 @@ def vector_from_symmetric_matrix(self, mat):
...
-                if i == j:
-                    vec[idx] = mat[j, j]
-                else:
-                    vec[idx] = mat[j, i]
-                idx += 1
-
-        return vec
+        _, dim, _ = mat.shape
+        i, j = gs.tril_indices(dim)
+        return mat[:, i, j]
```
ninamiolane, line 43: 🎉

geomstats/geometry/spd_matrices.py
```diff
@@ -38,36 +38,20 @@ def vector_from_symmetric_matrix(self, mat):
...
-        diag_indices = gs.diag_indices(mat_dim)
-
-        mat[lower_triangle_indices] = 2 * vec
-        mat[diag_indices] = vec
-
-        mat = self.embedding_manifold.make_symmetric(mat)
-        return mat
+        mask = gs.tril(gs.ones((mat_dim, mat_dim))) != 0
+        sym = gs.zeros((n_samples, mat_dim, mat_dim))
+        sym[..., mask != 0] = vec
```
ninamiolane, line 52: This will not work with TF... Use assignment function from the backend?

geomstats/learning/pca.py
```diff
@@ -99,39 +101,38 @@ class TangentPCA(_BasePCA):
...
+        self.point_type = point_type
 
-    def fit(self, X, y=None, base_point=None, point_type='vector'):
+    def fit(self, X, y=None, base_point=None):
         """Fit the model with X.
 
         Parameters
         ----------
-        X : array-like, shape=[n_samples, n_features]
+        X : array-like, shape=[n_samples, n_features], optional
```
ninamiolane, line 119: X is not optional.

geomstats/geometry/spd_matrices.py
nguigs: Yes I updated the decorator, but it was not tested at all before because the test's name did not start by test...

nguigs:
> @nguigs there seems to be a github problem :/

No I think I messed up the rebase

PR #512, opened 2020-04-03
Fix docstring error in the test_knn.py file.

YannCabanes:
Fix docstring error in the test_knn.py file: the class docstring refered to the Poincare polydisk instead of the KNN.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,14 @@
+# Team Yaml
+coverage:
+  round: up
+  precision: 2
+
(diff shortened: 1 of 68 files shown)


ninamiolane:
@YannCabanes there seems to be an error: no file is attached to this PR.

YannCabanes:
Oops, I pushed from the branch master instead of the branch ycabanes_knn.
You can delete my PR, I did another one from the right branch.

PR #522, opened 2020-04-03
Tangent PCA - picking-up

nguigs:
- This PR picks up #510 which was corrupted by an abusive use of `git rebase`
- fix symmetric_matrix_from_vector and its inverse and add these methods to symmetric_matrices. They could be removed from the SPDMatrices class when it will inherit from SymmetricMatrices. Only works for numpy so far. They don't require assignement but may not be memory/time efficient
- implement tangent PCA for point_type='matrix', thus addressing issue #246
- add inverse_tranfrom method (to override the inherited one from sklearn). It outputs points on the manifold and not tangent vectors
- add unit tests for PCA methods


ninamiolane:
@nguigs Can you try and solve DeepSource issues? https://deepsource.io/gh/geomstats/geomstats/run/a8d3fb41-a470-4681-b6df-dd2ac88c69e0/python/

It is configured to trigger only on the lines that have been changed for a commit, maybe we should try to go ahead and use it?

nguigs:
> @nguigs Can you try and solve DeepSource issues? https://deepsource.io/gh/geomstats/geomstats/run/a8d3fb41-a470-4681-b6df-dd2ac88c69e0/python/
> 
> It is configured to trigger only on the lines that have been changed for a commit, maybe we should try to go ahead and use it?

I think it triggers on the the files touched rather than the lines. I cannot address the remaining DeepSource errors:
- one in numpy/__init__, stg I did not touch, and would break consistency with tensorflow syntax
- one in pytorch/__init__, same as above
- unused `y=None` argument in pca: compliance with sklearn

ninamiolane:
That’s weird though, because it is explicitly configured to trigger on the lines only in the settings. Anyway, yes! Some errors do not necessarily make sense for us and I am des-activating them one at a time so that we can rely more and more on deep source in the future. Thank you for these!

PR #530, opened 2020-04-05
Inverse and random_uniform work in matrix form with all backends

pchauchat:
Only failing tests are Pytorch and Tensorflow giving wrong output in some log and exp tests (especially Pytorch), for which the decorator `@...np_only` was kept.
I removed the other decorators around the tests which all backends succeed, except `dist_is_symmetric` which takes ages. Better to slow down even the tests even more, or coverage completeness ? (Tensorflow is already tested with less testcases)

`logical_and` and `array` were modified in Pytorch backend to have behavior more similar to numpy's: the former can be applied to a bool and a Tensor for instance; the latter gives an nd_array with the same dimension.

### DIFF
--- a/geomstats/backend/pytorch/__init__.py
+++ b/geomstats/backend/pytorch/__init__.py
@@ -29,6 +29,10 @@ def logical_or(x, y):
 
 
 def logical_and(x, y):
+    if torch.is_tensor(x):
+        return x.eq(y)
+    elif torch.is_tensor(y):
+        return y.eq(x)
     return x and y
 
 
@@ -109,7 +113,7 @@ def array(val):
         if not isinstance(val[0], torch.Tensor):
             val = _np.copy(_np.array(val))
         else:
-            val = concatenate(val)
+            val = stack(val)
 
     if isinstance(val, bool):
         val = _np.array(val)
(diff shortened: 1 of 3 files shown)


geomstats/backend/pytorch/__init__.py
```diff
@@ -29,6 +29,10 @@ def logical_or(x, y):
 
 
 def logical_and(x, y):
+    if torch.is_tensor(x):
+        return x.eq(y)
+    elif torch.is_tensor(y):
```
ninamiolane, line 34: No need for elif since there is return

geomstats/geometry/special_euclidean.py
```diff
@@ -699,12 +698,8 @@ def random_uniform(self, n_samples=1, point_type=None):
         elif point_type == 'matrix':
```
ninamiolane, line 698: Same

pchauchat:
This came from a non-numpy-like behavior of Pytorch's `array` which creates tensors of `ndim = 1` when given ints and floats. I changed that, I tried to look around for things I potentially broke, ran `nose2 tests` and it looks fine. Let's see what Travis has to say.

PR #533, opened 2020-04-06
Vectorization: belongs, random_uniform in Euclidean and Minkowski

ninamiolane:
This PR partly addresses #338, and specifically update the methods `random_uniform` and `belongs` from Euclidean and Minkowski, while modifying methods using them.

Now, `random_uniform`:
- returns samples of size (dimension, ) when `n_samples` == 1, or (n_samples, dimension) when n_samples != 1,

And `belongs`:
- returns True if called on one sample,
- returns [True] * n_samples if called on several samples.


geomstats/geometry/lie_group.py
```diff
@@ -397,9 +397,9 @@ def log(self, point, base_point=None, point_type=None):
 
         return result
 
-    def exponential_barycenter(
-        self, points, weights=None, point_type=None
-    ):
+    def exponential_barycenter(self,
+                               points, weights=None,
```
nguigs, line 401: I thought we had the other indentation convention?

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -61,7 +61,10 @@ def inner_product_matrix(self, base_point=None, point_type=None):
         """
         if point_type is None:
             point_type = self.default_point_type
+
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
```
nguigs, line 65: why is this necessary? if base_point is only 1d it means there is only one manifold, and this case could be handled in the __init__?

geomstats/geometry/special_euclidean.py
```diff
@@ -311,44 +312,45 @@ def compose(self, point_1, point_2, point_type=None):
...
-                    or n_points_2 == 1)
+            if not (point_a.shape == point_b.shape
+                    or n_points_a == 1
+                    or n_points_b == 1):
+                raise ValueError()
 
-            if n_points_1 == 1:
-                point_1 = gs.stack([point_1[0]] * n_points_2)
+            if n_points_a == 1:
+                point_a = gs.stack([point_a[0]] * n_points_b)
```
nguigs, line 328: I don't think this is necessary: `gs.matmul(point_a, point_b)` will distribute the multiplication of `point_b` to all points of `point_a` and conversely

geomstats/geometry/special_euclidean.py
```diff
@@ -311,44 +312,45 @@ def compose(self, point_1, point_2, point_type=None):
...
 
-            if n_points_2 == 1:
-                point_2 = gs.stack([point_2[0]] * n_points_1)
+            if n_points_b == 1:
+                point_b = gs.stack([point_b[0]] * n_points_a)
```
nguigs, line 331: idem

geomstats/geometry/special_euclidean.py
```diff
@@ -796,18 +795,19 @@ def exponential_barycenter(
 
         if point_type == 'vector':
             n_points = points.shape[0]
-            assert n_points > 0
+            if n_points <= 0:
```
nguigs, line 798: btw this method is not unit-tested!

geomstats/geometry/special_euclidean.py
```diff
@@ -765,7 +765,6 @@ def exponential_matrix(self, rot_vec):
             term_2[i] = gs.matmul(skew_rot_vec[i], skew_rot_vec[i]) * coef_2[i]
```
nguigs, line 765: Good! By the way this method is not unit tested. It is a bit surprising to find it here though, shouldn't it be in SpecialOrthogonal?

geomstats/geometry/special_orthogonal.py
```diff
@@ -105,12 +106,12 @@ def belongs(self, point, point_type=None):
...
 
         if point_type == 'vector':
-            point = gs.to_ndarray(point, to_ndim=2)
-            n_points, vec_dim = point.shape
+            vec_dim = point.shape[-1]
             belongs = vec_dim == self.dimension
+            if point.ndim == 2:
+                belongs = gs.tile([belongs], (point.shape[0],))
             belongs = gs.to_ndarray(belongs, to_ndim=1)
             belongs = gs.to_ndarray(belongs, to_ndim=2, axis=1)
```
nguigs, line 114: Shouldn't this one be removed to output a 1d array?

geomstats/geometry/minkowski.py
```diff
@@ -25,12 +25,10 @@ def belongs(self, point):
         -------
         belongs : array-like, shape=[n_samples, 1]
```
nguigs, line 26: shouldn't this be [n_samples,] now?

geomstats/geometry/special_orthogonal.py
```diff
@@ -1365,27 +1361,23 @@ def compose(self, point_1, point_2, point_type=None):
...
 
-        assert (point_1.shape == point_2.shape
-                or n_points_1 == 1
-                or n_points_2 == 1)
+        n_points_a = point_a.shape[0]
+        n_points_b = point_b.shape[0]
 
-        if n_points_1 == 1:
-            point_1 = gs.stack([point_1[0]] * n_points_2)
+        if n_points_a == 1:
```
nguigs, line 1374: same as in SpecialEuclideand, this is not necessary. However there is an issue with einsum: numpy einsum does this well
```
point_a = rotations.random_uniform(4)
point_b = rotations.random_uniform(1)[0]
rot_mat_a = rotations.matrix_from_rotation_vector(point_a)
rot_mat_b = rotations.matrix_from_rotation_vector(point_b)[0] #2d-array
composition_rot_mat = np.einsum('...ij,...jk->...ik', rot_mat_a, rot_mat_b)
```
while gs.einsum throws `ValueError: Shape mismatch for einsum.`

tests/test_connection.py
```diff
@@ -23,15 +23,15 @@ def test_metric_matrix(self):
         base_point = gs.array([0., 1., 0., 0.])
 
         result = self.euc_metric.inner_product_matrix(base_point)
-        expected = gs.array([gs.eye(self.dimension)])
+        expected = gs.eye(self.dimension)
```
nguigs, line 26: great!

geomstats/geometry/lie_group.py
ninamiolane: Yes, good catch! I thought flake8 was throwing me an error there, but it must have been for something else. It is changed now, thanks!

geomstats/geometry/product_riemannian_metric.py
ninamiolane: nose2 tests.test_product_manifold is failing without it: this is because gs.to_ndarray cannot convert a 1D to a 3D directly, it needs to go through a 2D...

Maybe it would be worth thinking about modifying gs.to_ndarray, however the vectorization PRs might remove a lot of them. Thus, maybe let's postpone this discussion for later?

geomstats/geometry/special_euclidean.py
ninamiolane: Indeed, thank you!

geomstats/geometry/special_euclidean.py
ninamiolane: 🎉

geomstats/geometry/special_euclidean.py
ninamiolane: It is in this file because it is an auxiliary function for the exponential barycenter of SE(n). I have converted it to a private method, to make this more clear. If it gets used elsewhere, I agree that we could think about moving it.

geomstats/geometry/special_orthogonal.py
ninamiolane: Yes! But so far I have only modified belongs and random_uniform for Euclidean and Minkowski (and fixed the issues that were appearing in the other files). I will move on and modify belongs and random on the other files now.

geomstats/geometry/minkowski.py
ninamiolane: Yes, changed!

geomstats/geometry/special_orthogonal.py
ninamiolane: Great, thanks! Yes, this ValueError is one I have implemented because the new versions of einsum for tf and pytorch are not done yet. Working on it!

PR #536, opened 2020-04-07
Debugging SO(3) in vector form for all backends

pchauchat:
A number of definitions and changes in functions to make Pytorch and Tensorflow cope with SO(3) in vector form.

- Another way of computing the rotation vector from a rotation matrix, based on the axis-angle representation for the case `theta ~ pi` (https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation#Log_map_from_SO(3)_to_%7F'%22%60UNIQ--postMath-0000000D-QINU%60%22'%7F(3)). The other one made Tensorflow and Pytorch output wrong results (and I did not have the courage to dig into it). Many tests are now successful with Pytorch and Tensorflow, to the price of raising the threshold a bit. The limit cases are indeed troubling for them, might come from the fact that many computations are done in `float32`. In particular, it fails some tests that numpy passes (e.g. `gs.allclose(base_point, self.group.identity)`), which induces more computations and increases numerical errors.

- `set_diag` and `get_slice` were defined to mimic numpy's behavior (and, of course, because of Tensorflow).


geomstats/_backend/pytorch/__init__.py
```diff
@@ -51,14 +50,12 @@ def _raise_not_implemented_error(*args, **kwargs):
...
 
-flip = _raise_not_implemented_error
-hsplit = _raise_not_implemented_error
 searchsorted = _raise_not_implemented_error
 vectorize = _raise_not_implemented_error
 
 
-def empty(shape, dtype=float64):
-    return torch.empty(*shape, dtype=dtype)
+def empty(tensor_shape, dtype=float64):
```
nkoep, line 57: I know this masks the module-level `shape` function otherwise, but the name was chosen to match the numpy interface. I think we need to decide in general how we are gonna handle argument names so it's possible to use kwargs consistently. A lot of the functions we simply re-export or alias have different parameter names, which prevents this from working properly.

geomstats/_backend/pytorch/__init__.py
```diff
@@ -127,6 +124,14 @@ def repeat(a, repeats, axis=None):
     return torch.repeat_interleave(a, repeats, axis)
 
 
+def flip(x, axis):
```
nkoep, line 127: Remember to remove the previous `flip = _raise_not_implemented_error` at the top of the file.

geomstats/_backend/pytorch/__init__.py
pchauchat: I did that to try and reduce the number of Deepsources flags, but I agree this feels weird.

Also, Deepsource says `torch.tensor` is not supposed to be callable, any idea why ?

geomstats/_backend/pytorch/__init__.py
```diff
@@ -151,8 +156,13 @@ def vstack(seq):
 
 def array(val):
     if isinstance(val, list):
-        if not isinstance(val[0], torch.Tensor):
+        if not any([isinstance(t, torch.Tensor) for t in val]):
             val = _np.copy(_np.array(val))
+        elif any([not isinstance(t, torch.Tensor) for t in val]):
+            for (index, t) in enumerate(val):
```
nkoep, line 162: That's a style question, but the parentheses around `index, t` aren't needed.

geomstats/_backend/pytorch/__init__.py
pchauchat: Yep, done in the next commit (normally) !

geomstats/_backend/pytorch/__init__.py
```diff
@@ -394,6 +408,21 @@ def outer(*args, **kwargs):
     return torch.ger(*args, **kwargs)
 
 
+def hsplit(x, indices_or_section):
```
nkoep, line 411: See comment regarding the `flip` function.

geomstats/_backend/pytorch/random.py
```diff
@@ -16,6 +16,8 @@ def seed(*args, **kwargs):
 
 
 def normal(loc=0.0, scale=1.0, size=(1,)):
+    if isinstance(size, int):
```
nkoep, line 19: I think it's better to test whether `size` is not a sequence by doing `if not hasattr(size, '__iter__')`. Keep in mind that a numpy integer like `np.int32(1)` does not inherit from python's builtin `int` type.

Edit: I also realize that the same logic should apply to `uniform` below, which should have `size=(1,)` as default, and then check whether the `size` parameter is iterable.

tests/test_special_orthogonal.py
```diff
@@ -303,15 +299,14 @@ def test_regularize(self):
                 self.assertAllClose(result, expected)
 
             else:
-                angle = 0.345
+                angle = gs.array(0.345)
```
nkoep, line 302: Are all these conversions below necessary? Shouldn't the backends be able to handle number literals the same way primitives like `np.sin` do?

geomstats/_backend/pytorch/__init__.py
nkoep: Best guess would be that it's a false alarm.

tests/test_special_orthogonal.py
pchauchat: If they do, we need to redefine all of these functions for Pytorch and Tensorfow.

Since this is (for now) only for tests, I wonder whether this is worth the trouble ?

geomstats/_backend/pytorch/random.py
pchauchat: Smart ! I'll change that.

geomstats/_backend/numpy/__init__.py
```diff
@@ -246,6 +246,11 @@ def diag(x):
     return result
 
 
+def set_diag(x, new_diag):
```
ninamiolane, line 249: Is there a difference with this numpy function, and if not, should we call it fill_diagonal?
https://docs.scipy.org/doc/numpy/reference/generated/numpy.fill_diagonal.html

geomstats/_backend/pytorch/__init__.py
ninamiolane: Yes, DeepSource is not perfectly configured yet. Let me know which errors don't make sense and I can deactivate them for a whole file.

geomstats/_backend/numpy/__init__.py
```diff
@@ -201,7 +201,7 @@ def assignment_by_sum(x, values, indices, axis=0):
     return x_new
 
 
-def gather(x, indices, axis=0):
+def get_slice(x, indices):
```
ninamiolane, line 204: `gather` was referring to:
https://www.tensorflow.org/api_docs/python/tf/gather
(which does not exist in np),

but I agree that get_slice maybe more clear!

geomstats/_backend/pytorch/__init__.py
pchauchat: For now I rollbacked on the change of `shape` to `tensor_shape`. Indeed, a decision must be made concerning these names. Here it's the first argument, so maybe it would be less problematic ?

geomstats/geometry/special_orthogonal.py
```diff
@@ -542,127 +543,36 @@ def rotation_vector_from_matrix(self, rot_mat):
...
-            # This avois dividing by 0.
-            mask_0 = gs.isclose(angle, 0.)
-            mask_0_float = gs.cast(mask_0, gs.float32) + self.epsilon
-
-            rot_vec *= (1. + mask_0_float * (.5 - (trace - 3.) / 12. - 1.))
-
-            # This avois dividing by 0.
-            mask_pi = gs.isclose(angle, gs.pi)
-            mask_pi_float = gs.cast(mask_pi, gs.float32) + self.epsilon
+            rot_vec_not_pi = self.vector_from_skew_matrix(
```
ninamiolane, line 549: Great, thanks!! 🎉

tests/test_special_euclidean.py
```diff
@@ -1100,8 +1100,8 @@ def test_log_then_exp_left_with_angles_close_to_pi(self):
                     inv_expected = helper.to_vector(inv_expected)
 
                     self.assertTrue(
-                        gs.allclose(result, expected)
-                        or gs.allclose(result, inv_expected))
+                        gs.allclose(result, expected, atol=1e-6)
```
ninamiolane, line 1103: Define constant variable ATOL=1e-6 at the top of the file? It could be easier to change it later.

tests/test_special_orthogonal.py
```diff
@@ -158,7 +158,7 @@ def setUp(self):
         self.angles_close_to_pi_all = angles_close_to_pi_all
         self.n_samples = 4
 
-    @geomstats.tests.np_only
+    @geomstats.tests.np_and_tf_only
```
ninamiolane, line 161: Great!

tests/test_special_orthogonal.py
ninamiolane: Right, torch.Tensor(3.4) gives:
```
TypeError: sin(): argument 'input' (position 1) must be Tensor, not float
```

We could create an issue to flag this for later?

geomstats/_backend/numpy/__init__.py
pchauchat: Yes, there is a difference, because `fill_diagonal` fills indices `[i, ..., i]`, that is, the diagonal across all dimensions, while this function only wants to set the diagonal of the last two dimensions.

geomstats/_backend/numpy/__init__.py
pchauchat: I also changed the name because of this: the Tensorflow version relies on `gather_nd` and not `gather` (maybe there was a way to achieve that with it, but I couldn't find any easy one)

tests/test_special_euclidean.py
pchauchat: For some tests, I needed to increase it more than that (up to `5e-3` for angles close to pi, with Tensorflow and Pytorch). Leaving it like that might allow later improvements of the functions ?
We might be able to get away with the rtol though I'll check that.

ninamiolane:
@pchauchat I've relaxed DeepSource for the backend files. Could you see if it passes now?

Also, maybe using torch.Tensor instead of torch.tensor could help with the not callable error?

nkoep:
> @pchauchat I've relaxed DeepSource for the backend files. Could you see if it passes now?
> 
> Also, maybe using torch.Tensor instead of torch.tensor could help with the not callable error?

`torch.Tensor` is the PyTorch equivalent of `np.ndarray`, i.e., the constructor takes an integer (or a shape tuple) and creates an empty tensor much like `torch.empty`, whereas `torch.tensor` converts its arguments to a `torch.Tensor` instance.

tests/test_special_orthogonal.py
nkoep: I guess you mean `torch.sin(1.)`, but yes, that throws an exception, which I find a bit surprising. Still, if we want to mimic the numpy behavior, we should be able to handle such cases without manually having to wrap scalars in the backend's array representation.

geomstats/_backend/pytorch/__init__.py
nkoep: Yes, that's what I was thinking as well just now actually, see also #540.

tests/test_special_orthogonal.py
pchauchat: I am not convinced it is that surprising. torch is made to work with tensors only, to be able to better use GPUs. `numpy.sin(1.)` outputs a float for instance. Since the original point of torch was to do neural networks, the gradient needs to be computed at each step, which requires (or is easier) with the tensor structure I think. And from what a friend told me, Pytorch is also used in other contexts for lazy parallelisation, which is also easier done with tensors (I guess ?)

ninamiolane:
@nkoep any idea why we get torch.tensor is not callable? I was wondering if it was related to this: https://github.com/pytorch/pytorch/issues/24807.

@pchauchat there should be only a few deepsource issues remaining, let us know if you can address them? Thanks.
https://deepsource.io/gh/geomstats/geomstats/run/34abee0a-8d51-4126-9e15-87a51d8fca39/python

pchauchat:
Damn it. I tried things on special_orthognal, reverted them, which made Deepsource go and have a look at it. I'll get rid of the asserts.
Last question: what about the `shape` renaming ? Is it ok or not in the end ?

ninamiolane:
I thought that I got rid in the asserts in `special_orthogonal` in my last PR, maybe try rebase on master first?

ninamiolane:
For the `shape` argument, I say that we can keep it to match numpy's API, and I'll deactivate this error on Deep Source once this is merged. (I can deactivate some types of DeepSource errors on whole files, when they appear at the repository level - not during the PR). But happy to hear your thoughts!

pchauchat:
Good catch, I thought I had rebased, but I didn't.

About the `shape` argument, I think it is better to keep it as is, especially if it is the same in numpy.

nkoep:
> @nkoep any idea why we get torch.tensor is not callable? I was wondering if it was related to this: [pytorch/pytorch#24807](https://github.com/pytorch/pytorch/issues/24807).
> 
> @pchauchat there should be only a few deepsource issues remaining, let us know if you can address them? Thanks.
> https://deepsource.io/gh/geomstats/geomstats/run/34abee0a-8d51-4126-9e15-87a51d8fca39/python

Ah yes, that's very likely it. So DS gets confused because the torch package contains a 'tensor.py' file, so it confuses 'torch.tensor' as trying to call a module. Good catch! I think we can ignore this for now then until it changes upstream.

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -354,6 +356,10 @@ def array_from_sparse(indices, data, target_shape):
         tf.SparseTensor(indices, data, target_shape)))
 
 
+def get_slice(x, indices):
+    return gather_nd(x, list(zip(*indices)))
```
nkoep, line 360: Ah, I think it's better to have `tf.gather_nd`... here instead of importing the function explicitly. I prefer for the import list at the beginning of the file to contain only functions which are exported or aliased without change.

pchauchat: That sounds sensible, I'm on it.

pchauchat:
I think the question of the unused `path` argument had already been asked, and this was a matter of compatibility/consistency. Can you confirm or infirm @nkoep ?

nguigs:
> * [...]. In particular, it fails some tests that numpy passes (e.g. `gs.allclose(base_point, self.group.identity)`), which induces more computations and increases numerical errors.

I've been facing this problem as well (I added this line), which comes from shape mismatches I think: did you find a solution?

pchauchat:
When I looked at it, the test was really failing because of the values, (~1e-7, while the threshold is 1e-8), at least in the case of a single point. For some reason, Pytorch (or the implementation of `allclose`) allows comparing `x` and `[x]`.

pchauchat:
I found the discussion on the `path` parameter, in the conversation about the `_\backend\__init__.py` file: https://github.com/geomstats/geomstats/pull/515/files#r403679894
Is there a way to trigger Travis on its own ?

ninamiolane:
@pchauchat so weird that Travis does not trigger. Maybe it's because there are many GitHub conflicts remaining: This branch has conflicts that must be resolved. Could you try to solve this? I'll try to trigger travis independently now as well.

pchauchat:
I'll close this one to redo it in #557

Issue #531, opened 2020-04-06
Exponential Barycenter method in LieGroups


nguigs:
https://github.com/geomstats/geomstats/blob/9c3dd5095d4cefba3a34f53249e68b33f9020af6/geomstats/geometry/lie_group.py#L400
Do we want to keep this method here? In spirit it is very close to `FrechetMean` estimator, which is now in `learning`, with the difference that it uses canonical group operations (`exp, log`) rather than the Riemannian ones.

The method is implemented here for instance
https://github.com/geomstats/geomstats/blob/9c3dd5095d4cefba3a34f53249e68b33f9020af6/geomstats/geometry/special_euclidean.py#L767
But it can be factorized to the above to be available to all groups. I fixed it a while ago because I needed it on `SEn` but haven't pushed it. Should it remain a `LieGroup` method or should it be moved to the `learning` module?

ninamiolane:
Well spotted! I'd say we put it in the `learning` module, just as FrechetMean is, as you suggested.

PR #537, opened 2020-04-07
Factorize exponential Barycenter in learning module

nguigs:
This PR addresses #531
- add estimator `ExponentialBarycenter`
- remove corresponding method from all groups, and move corresponding tests
- allow invariant_metric exp to work with matrices.
- add `default_point_type` attribute to Lie Groups
- it converges with other backends but somehow `assertAllclose` throws exceptions because of shape issues


geomstats/geometry/lie_group.py
```diff
@@ -65,7 +65,7 @@ class LieGroup(Manifold):
     and columns of the matrix.
     """
 
-    def __init__(self, dimension):
+    def __init__(self, dimension, point_type='matrix'):
```
ninamiolane, line 68: Nice! 🙏

geomstats/geometry/lie_group.py
```diff
@@ -205,11 +205,12 @@ def exp_not_from_identity(self, tangent_vec, base_point, point_type):
         exp : array-like, shape=[n_samples, {dimension,[n,n]}]
             the computed exponential
         """
-        jacobian = self.jacobian_translation(
-            point=base_point, left_or_right="left", point_type=point_type
-        )
+        if point_type is None:
+            point_type = self.default_point_type
 
         if point_type == "vector":
```
ninamiolane, line 211: single quotes, if possible.

geomstats/geometry/lie_group.py
```diff
@@ -229,9 +230,13 @@ def exp_not_from_identity(self, tangent_vec, base_point, point_type):
...
         if point_type == "matrix":
             tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3)
-            raise NotImplementedError()
+            lie_vec = self.compose(
+                self.inverse(base_point), tangent_vec)
+            return self.compose(
+                base_point, self.exp_from_identity(lie_vec, point_type))
 
-        raise ValueError('point_type should be \'vector\' or \'matrix\'.')
+        raise ValueError('Invalid point_type, expected \'vector\' or '
```
ninamiolane, line 238: Nice! I'm thinking that, at some point, we should uniformize the error messages. Right now we have various syntaxes across the codebase. I'm writing an issue.

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
+"""Frechet mean."""
```
ninamiolane, line 1: Change docstring

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
...
+
+from sklearn.base import BaseEstimator
+
+import geomstats.backend as gs
+from geomstats.geometry.euclidean import Euclidean
+
+EPSILON = 1e-6
+
+
+def linear_mean(points, weights=None):
```
ninamiolane, line 13: This is the same code as in frechet_mean.py, right? Maybe call that function instead of rewriting it?

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
...
+    elif weights is None:
+        weights = gs.ones((n_points, ))
+
+    weighted_points = gs.einsum('...,...j->...j', weights, points)
+    mean = (gs.sum(weighted_points, axis=0) / gs.sum(weights))
+    mean = gs.to_ndarray(mean, to_ndim=2)
+    return mean
+
+
+def _default_gradient_descent(
```
ninamiolane, line 49: Should we create a file with all the gradient descents, that would be used across learning algorithms, so that we do not have to copy-paste (this is copy-pasted from frechet_mean.py right?)

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
...
+            'Maximum number of iterations {} reached. '
+            'The mean may be inaccurate'.format(max_iter))
+
+    if verbose:
+        logging.info('n_iter: {}, final norm: {}'.format(last_iteration, norm))
+
+    return mean[0]
+
+
+class ExponentialBarycenter(BaseEstimator):
```
ninamiolane, line 127: 🎉

tests/test_exponential_barycenter.py
```diff
@@ -0,0 +1,90 @@
...
+        point = self.so.random_uniform(self.n_samples)
+        estimator = ExponentialBarycenter(self.so, max_iter=2)
+        estimator.fit(point)
+        barexp = estimator.estimate_
+        result = self.so.belongs(barexp)
+        expected = True
+        self.assertAllClose(result, expected)
+
+    @geomstats.tests.np_only
+    def test_coincides_with_frechet_so(self):
```
ninamiolane, line 77: Nice, thanks!

geomstats/geometry/special_euclidean.py
```diff
@@ -767,80 +767,3 @@ def exponential_matrix(self, rot_vec):
         assert exponential_mat.ndim == 3
```
ninamiolane, line 767: You didn't move the private function `exponential_matrix`, aren't you using it?

tests/test_exponential_barycenter.py
```diff
@@ -0,0 +1,90 @@
...
+from geomstats.geometry.special_euclidean import SpecialEuclidean
+from geomstats.geometry.special_orthogonal import SpecialOrthogonal
+from geomstats.learning.exponential_barycenter import ExponentialBarycenter
+from geomstats.learning.frechet_mean import FrechetMean
+
+
+class TestExponentialBarycenter(geomstats.tests.TestCase):
+
+    def setUp(self):
+        self.se = SpecialEuclidean(n=3, point_type='matrix')
```
ninamiolane, line 14: Does the ExponentialBarycenter work with point_type='vector'?

The code in special_orthogonal and special_euclidean had the 'vector' implementation too, it will be good not to lose it, as it is probably faster than going to 'matrix', using ExponentialBarycenter, and converting back to 'vector'.

ninamiolane:
Also, Travis and DeepSource need to pass.

tests/test_exponential_barycenter.py
nguigs: It called FrechetMean in `special_orthogonal`, which is valid for the bi-invariant metric. For the `special_euclidean` case it called FrechetMean on the rotation part and the linear mean on the translation part (which I am not sure is valid). `point_type='vector'` is such a limited case of groups that I thought it was not relevant... But I can put it back if you insist

nguigs: And it was probably slower because FrechetMean calls `metric.squared_dist` and `metric.norm` for the stopping criterion, which is much slower than `gs.linalg.norm`used here. This is actually almost the only difference with FrechetMean

nguigs:
> Also, Travis and DeepSource need to pass.

Thanks for the review! Very weird for Travis, all the tests worked on my machine last night...
DeepSource won't pass because of unused variables in the `fit` method, `lambda` in the `while_loop_iterator`, and unused variables in the iterator, which I cannot do anything about as this is the required syntax as I understood it.

geomstats/learning/exponential_barycenter.py
nguigs: Not exactly copy-paste but very close:
- different stopping criterion (which could be used in FrechetMean to make it faster maybe)
- factorizes `log(points, current)` to `log_from_id(compose(inverse(current), points))`

ninamiolane: - Maybe add a flag about the stopping criterion, and then factorize this _default_gradient_descent?
- Does this modification make the code faster? If so, we should probably compute all Lie group logs this way, and factorize _default_gradient_descent.

geomstats/learning/exponential_barycenter.py
ninamiolane: how about this one?

tests/test_exponential_barycenter.py
ninamiolane: I am not sure that the exponential barycenter formula for SE(3) is valid, especially since it seems that you didn't use the function `exponential_matrix`, which is needed to link the rotation part with the translation part (see original implementation).

ninamiolane: I agree that `point_type='vector'` is a limited case for Lie groups when you take the point of view of the mathematician and consider all Lie groups which equal weights.

However, from an applied perspective, `point_type='vector'` applies to SO(3) and SE(3) which are probably the most widely used Lie groups. The vector representation might:
- speed-up the computations,
- decrease the memory space needed,
- or simply be the favorite representation for users computing with SO(3) and SE(3) - and we might have many users in this case -
which I think are 3 good reasons to keep them. What do you think?

geomstats/learning/exponential_barycenter.py
nguigs: This is already the way lie group logs are computed, and also in `invariant_metric` although I'm not sure it's right

geomstats/geometry/special_euclidean.py
nguigs: I'm not using it, unless `self.exp` calls it. Was I supposed to move it?

geomstats/learning/exponential_barycenter.py
ninamiolane: In that case, no need to change the code of `default_gradient_descent` and we can factorize it?

geomstats/geometry/special_euclidean.py
ninamiolane: The exponential barycenter in SE(3) needs it, see other comment.

tests/test_exponential_barycenter.py
nguigs: > I am not sure that the exponential barycenter formula for SE(3) is valid, especially since it seems that you didn't use the function `exponential_matrix`, which is needed to link the rotation part with the translation part (see original implementation).

In matrix representation this is automatically done by the matrix exponential of the squared matrix of size n+1

ninamiolane: Got it!

nguigs: > I agree that `point_type='vector'` is a limited case for Lie groups when you take the point of view of the mathematician and consider all Lie groups which equal weights.
> 
> However, from an applied perspective, `point_type='vector'` applies to SO(3) and SE(3) which are probably the most widely used Lie groups. The vector representation might:
> 
>     * speed-up the computations,
> 
>     * decrease the memory space needed,
> 
>     * or simply be the favorite representation for users computing with SO(3) and SE(3) - and we might have many users in this case -
>       which I think are 3 good reasons to keep them. What do you think?

Ok I agree, but we need to separate the vector and matrix implementations of these groups because they become confusing being so long and not dealing with the matrix type all the time (there were many e.g. `assert logs.ndim == 2` which crash when using matrices, etc).

ninamiolane: Definitely agreed to this! It also goes back to @opeltre remark that the special_orthogonal was full with computations for the n=3 case, see PR #484.

geomstats/geometry/lie_group.py
```diff
@@ -143,8 +144,7 @@ def inverse(self, point, point_type=None):
         raise NotImplementedError("The Lie group inverse is not implemented.")
 
     def jacobian_translation(
-        self, point, left_or_right="left", point_type=None
-    ):
+            self, point, left_or_right="left", point_type=None):
```
ninamiolane, line 147: single quotes

geomstats/geometry/lie_group.py
```diff
@@ -205,11 +205,12 @@ def exp_not_from_identity(self, tangent_vec, base_point, point_type):
...
         """
-        jacobian = self.jacobian_translation(
-            point=base_point, left_or_right="left", point_type=point_type
-        )
+        if point_type is None:
+            point_type = self.default_point_type
 
         if point_type == "vector":
+            jacobian = self.jacobian_translation(
+                point=base_point, left_or_right="left", point_type=point_type)
```
ninamiolane, line 213: single quotes

geomstats/geometry/lie_group.py
```diff
@@ -321,24 +326,31 @@ def log_not_from_identity(self, point, base_point, point_type):
...
+        if point_type == 'vector':
+            jacobian = self.jacobian_translation(
+                point=base_point, left_or_right="left", point_type=point_type)
+            point_near_id = self.compose(
+                self.inverse(base_point), point, point_type=point_type)
+            log_from_id = self.log_from_identity(
+                point=point_near_id, point_type=point_type)
+
+            log = gs.einsum(
+                "ni,nij->nj",
```
ninamiolane, line 341: single quotes

geomstats/geometry/lie_group.py
```diff
@@ -321,24 +326,31 @@ def log_not_from_identity(self, point, base_point, point_type):
...
+            point_type = self.default_point_type
 
-        log = gs.einsum(
-            "ni,nij->nj",
-            log_from_id,
-            gs.transpose(jacobian, axes=(0, 2, 1)),
-        )
+        if point_type == 'vector':
+            jacobian = self.jacobian_translation(
+                point=base_point, left_or_right="left", point_type=point_type)
```
ninamiolane, line 334: single quotes

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
...
+    -------
+    exp_bar : array-like, shape=[n,n]
+        The exponential_barycenter of the input points.
+    """
+    n_points = points.shape[0]
+    if weights is None:
+        weights = gs.ones((n_points,))
+    sum_weights = gs.sum(weights)
+
+    def while_loop_cond(iter_index, current, grad_norm):
```
ninamiolane, line 83: `current` is not used, remove it?

geomstats/learning/exponential_barycenter.py
```diff
@@ -0,0 +1,192 @@
...
+        result = grad_norm > epsilon
+        return result or iter_index == 0
+
+    def while_loop_body(iter_index, current, grad_norm):
```
ninamiolane, line 87: `grad_norm` is not used (it is assigned later with a new value), remove it?

geomstats/learning/exponential_barycenter.py
nguigs: It was not possible to remove it with the way `while_loop` in the backend worked, but after #544 i won't use it anymore

PR #540, opened 2020-04-08
Alias argument names of backend functions

nkoep:
This is just a quick PoC to demonstrate how we might alias argument names for backend functions which offer the same behavior as their Numpy counterpart but use different argument names. Without special handling, this currently prevents the use of keyword arguments for functions where argument names diverge from Numpy's interface.

The PR demonstrates this for PyTorch's `clamp` function (aliased `clip`), which accepts arguments `input`, `min`, `max` instead of `a`, `a_min`, `a_max` as in Numpy's case. To remap arguments, we could wrap the function using the provided `alias_argument_names` decorator (factory):
```python
clip = alias_argument_names(a="input", a_min="min", a_max="max")(clip)
```
It's questionable though whether this is really more convenient than simply doing
```python
def clip(a, a_min, a_max):
    return torch.clamp(input=a, min=a_min, max=a_max)
```

The alternative is what we pretty much already do, albeit somewhat coincidentally. We might simply make it a convention that keyword arguments are not allowed unless for arguments with default values.

I'd be interested to hear your views on this.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -45,16 +45,13 @@ from torch import (  # NOQA
 
 from . import linalg  # NOQA
 from . import random  # NOQA
+from geomstats._backend.utils import mark_not_supported
 
 
-def _raise_not_implemented_error(*args, **kwargs):
-    raise NotImplementedError
-
-
-flip = _raise_not_implemented_error
-hsplit = _raise_not_implemented_error
-searchsorted = _raise_not_implemented_error
-vectorize = _raise_not_implemented_error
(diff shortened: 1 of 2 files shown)


ninamiolane:
I like the one-liner! However, its current form might actually be less explicit than the syntax that we already have? Specifically: from the current syntax, it was not 100% intuitive to me to know which were the original arguments and which are the new ones. But it does save us space!

Why "We might simply make it a convention that keyword arguments are not allowed unless for arguments with default values."?

ninamiolane:
@nkoep @nguigs: down to move forward with this, if Travis passes!

PR #542, opened 2020-04-08
Fix vectorization in Poincare Ball

tgeral68:


tgeral68:
There are some missing functionalities with the Pytorch backend for broadcasting in einsum, particularly `gs.einsum('j...,j...->j...',a, b)`  do not works having `a.shape=(1,d)` and `b.shape=n,1` in pytorch only (and here, it is a possible case).  

But I still can change the function to use einsum but keeping some repeat.

ninamiolane:
How about: `gs.einsum('...j,...k->...j', a, b)`? I've started working on the `einsum` function in the backends, to extend it beyond its capacities for pytorch and tf.

geomstats/geometry/poincare_ball.py
```diff
@@ -290,7 +298,7 @@ def retraction(self, tangent_vec, base_point):
         return base_point - retraction_factor * tangent_vec
 
     def inner_product_matrix(self, base_point=None):
-        """Compute the inner product matrix, independent of the base point.
+        """Compute the inner product matrixx.
```
ninamiolane, line 301: NIT: Typo.

tests/test_poincare_ball.py
```diff
@@ -23,7 +23,7 @@ def setUp(self):
         self.hyperboloid_manifold = Hyperboloid(2)
         self.hyperboloid_metric = self.hyperboloid_manifold.metric
 
-    @geomstats.tests.np_and_pytorch_only
+    @geomstats.tests.np_only
```
ninamiolane, line 26: What happened for the pytorch test here?

tgeral68: My bad, it should works with pytorch.

PR #543, opened 2020-04-08
Clean up backends

nkoep:
This removes a lot of boilerplate and unused/unnecessary code in the backends, and fixes a few minor inconsistencies between the different backends.


geomstats/_backend/numpy/linalg.py
```diff
@@ -12,14 +12,13 @@
     svd
 )
 
-# TODO(nina): Clean this import.
-from geomstats._backend.numpy.__init__ import to_ndarray as _to_ndarray
+from .common import to_ndarray
```
ninamiolane, line 15: Cool, thanks! 🙏

geomstats/_backend/numpy/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Numpy based computation backend."""
 
-import autograd.numpy as _np
+import autograd.numpy as np
```
ninamiolane, line 3: Comment for a next PR/ to start a discussion: using autograd.np as np actually causes the code to be ~1.3 times slower, as noticed by @nguigs on some examples.

So, we should probably create another autograd backend, and disentangle it from numpy?

geomstats/_backend/pytorch/__init__.py
```diff
@@ -4,15 +4,19 @@
 import torch
 from torch import (  # NOQA
     abs,
+    acos as arccos,
```
ninamiolane, line 7: Nice!

geomstats/_backend/numpy/__init__.py
nkoep: I'm not sure what the best way to go forward here is. It seems a bit overkill to add an entirely new backend considering that we only rely on automatic differentiation in two places so far.

nguigs: Yes so far we only have a PoC of integrating geodesics with autograd, in a very limited example. I haven't tried it with the other backends. 
Do you think Jax could perform better? It also has an automatic vectorization function which would be helpful. Otherwise we can drop automatic differentiation with numpy backend and see how it works with the others..

nkoep: Jax might have a performance edge over autograd because it includes a JIT (compiler), which can also be applied to autodiff'd code. However, Jax uses an asynchronous execution model, which means you have to call the `block_until_ready` method on Jax expressions, see https://jax.readthedocs.io/en/latest/async_dispatch.html. This also has consequences for the RNG used in Jax, see https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#JAX-PRNG. So unfortunately things aren't as easy as replacing `import autograd.numpy as np` with `import jax.numpy as np`.

Issue #545, opened 2020-04-08
Rename 'diag' function in backends


nkoep:
The `diag` function in the backends behaves very differently from numpy's `diag` function. Given a matrix of size m-by-n, the function currently returns an array/tensor of size m-by-n-by-n, where each n-by-n subarray is a diagonal matrix formed by the rows of the input matrix.

Since the backends are supposed to follow numpy's API, the function should be renamed to emphasize this difference in behavior.

ythanwerdas:
I agree with you. Moreover, that is exactly what `from_vector_to_diagonal_matrix` does now (and for the three backends).

PR #544, opened 2020-04-08
Remove TF1 artifacts

nkoep:
This PR removes the now unnecessary compatibility layer in the Numpy/PyTorch backends we had to maintain due to TF's execution model prior to TF2.

PR #546, opened 2020-04-09
Replace 'diag' function in backend

nkoep:
This addresses the issue raised in #545.


geomstats/utils.py
```diff
@@ -0,0 +1,8 @@
+import geomstats.backend as gs
```
ninamiolane, line 1: NIT: Could we give `utils` a more explicit name? Maybe `algebra.py` or `algebra_utils.py`? Or could this function go to `vectorization.py`?

Calling a file `utils.py` might lead us to put there everything that we do not know where to place, without forcing us to think about the repo's logic.

geomstats/_backend/numpy/__init__.py
```diff
@@ -211,31 +211,10 @@ def vectorize(x, pyfunc, multiple_args=False, signature=None, **kwargs):
...
-    if pred:
-        return true_fn()
-    return false_fn()
-
-
 def cast(x, dtype):
     return x.astype(dtype)
 
 
-def diag(x):
```
ninamiolane, line 225: We don't need this anymore? How about the vectorization of 'diag`?

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -98,10 +96,6 @@ def diagonal(a, axis1=0, axis2=1):
     return tf.linalg.diag_part(a)
 
 
-def diag(a):
```
ninamiolane, line 101: No need for this?

geomstats/utils.py
nkoep: That's a fair point. I had thought of adding it to vectorization, but at least judging by the function name it didn't really fit. I would prefer `algebra_utils` in that case.

geomstats/_backend/numpy/__init__.py
nkoep: Yes, sorry, I created a bit of a mess with the issues and PRs relating to this. @ythanwerdas pointed out that his `from_vector_to_diagonal_matrix` function does exactly what `diag` does, but in a backend-independent fashion. Since we don't seem to be using numpy's `diag` function anywhere, I removed it altogether (at least until someone actually needs it again).

nkoep:
> Great, thanks! Some comments:
> 
>     * There seem to be several PRs in this, maybe rebase?
> 
>     * Coud you address the remaining DeepSource errors (note that many overlap with your other PR)?
> 
>     * We lose the vectorization of the `diag` function with this: is this intentional?
> 
> 
> Thanks!

Good catch re the mixture of PRs. I guess I branched off of another PR. Rebasing tidied things up again.

As in #544, three of the remaining DS issues are already being addressed in #536. The one regarding the `shape` argument we discussed before as something we should probably ignore, given that the argument name `follows` numpy's convention. Maybe it's be best to ignore the `Re-defined variable from outer scope (PYL-W0621)` warnings in DS altogether.

ninamiolane:
Thanks! I'll ignore PYL-W0621 in `_backend` folder as a first step.

PR #549, opened 2020-04-09
Start vectorization of inner_product and downstream methods

ninamiolane:
This PR follows PR #547. It starts to vectorize the method `inner_product` from `riemannian_metric.py` and the corresponding downstream methods.

Many files are modified as the method `inner_product` is central to many geometries.


examples/plot_geodesics_h2.py
```diff
@@ -49,10 +50,11 @@ def plot_geodesic_with_initial_tangent_vector(initial_point,
 
 def main():
     """Plot the geodesics."""
-    initial_point = [np.sqrt(2), 1., 0.]
+    initial_point = gs.array([np.sqrt(2), 1., 0.])
```
nguigs, line 53: would it still work if the user inputs lists though?

ninamiolane: No... :( it was the only place where I found this, so I converted to np.array, but we could systematically convert lists into arrays: that would mean a decorator on top of every method (or, alternatively, copy-pasting the line of code doing the conversion everywhere...)

Issue #538, opened 2020-04-07
Uniformize the Error messages


ninamiolane:
There are several ValueErrors, NotImplementedErrors, etc, across the codebase.

We should create a mechanism to get clean, and uniform error messages.

zigzagthad:
@ninamiolane Can I take up this issue if it's okay?

ninamiolane:
With great pleasure! 🎉

zigzagthad:
@ninamiolane I was looking at the errors and like the bug says there are multiple "types" of errors thrown, so should I change them to a more generic type for example "Exception" ?

ninamiolane:
Hi @zigzagthad , apologies for the late reply!

Oh I meant making the error __messages__ more uniform and explanatory, not the error types. Sorry, this wasn't extremely clear.

Currently, our error messages are not very well written, and I don't think that they are very helpful to anyone trying out geomstats. We should make clearer error messages. Let me know if my explanation helps?

zigzagthad:
Hi @ninamiolane, I guess you're talking about errors like   "belongs is not implemented."  ? Such errors need to be more verbose, correct me if I'm wrong. However, to me not all errors seem improper, for example the error    "The bandwidth should be a positive real number."  seems good to me, what do you think?

Issue #501, opened 2020-04-02
Address issues from deep source


ninamiolane:
Deep source is a tool for automatic code reviews. 

Running it on `geomstats` detects many issues, that we should address.

https://deepsource.io/gh/geomstats/geomstats/

PR #550, opened 2020-04-09
Remove asserts from geomstats' main files

ninamiolane:
This PR addresses the DeepSource error "Assert statement used outside of tests", by transforming `asserts` into errors, see issue #501.

By doing so, it also starts to address issue #538 and creates a file to uniformize the error messages throughout the code base.


geomstats/geometry/matrices.py
```diff
@@ -178,14 +180,10 @@ def __init__(self, m, n):
...
         tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=3)
-        n_tangent_vecs_a, _, _ = tangent_vec_a.shape
-
         tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=3)
-        n_tangent_vecs_b, _, _ = tangent_vec_b.shape
-
-        assert n_tangent_vecs_a == n_tangent_vecs_b
 
-        inner_prod = gs.einsum("nij,nij->n", tangent_vec_a, tangent_vec_b)
+        inner_prod = gs.einsum(
```
nguigs, line 185: :heart_eyes:

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -98,7 +98,9 @@ def is_intrinsic(self, point):
         intrinsic: bool
             Whether intrinsic coordinates are used for all manifolds.
         """
-        assert self.default_point_type == 'vector'
+        if self.default_point_type != 'vector':
```
nguigs, line 101: in fact if we reach that point there really is a problem, it's not that it was not implemented

ninamiolane: Modified!

ninamiolane:
@nguigs thanks for looking at it! Yes, code on default_point_type could be factorized, and actually I was thinking about it in the context of code vectorization. Because so far, we have 3 mechanisms that have to do with vectorization:
- decorator,
- einsums,
- point_types,
and it could be great to converge to the minimal one, both in terms of code efficiency and code readability.

I've put a TODO for now.

PR #552, opened 2020-04-10
Random sampling in GeneralLinear

pchauchat:
Random sampling based on the fact that a random matrix is invertible with probability 1. Similarly to what is done for the hypersphere, a check on the determinant is made to have a numerically acceptable matrix.

The value of the determinant threshold is completely arbitrary, and is surely not the best one. I guess it should vary with `n`.

### DIFF
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -40,7 +40,7 @@ class GeneralLinear(Matrices):
 
         Parameters
         ----------
-        n_samples: int, optional
+        n_samples : int, optional
             Number of samples.
         tol: float, optional
             Threshold for the absolute value of the determinant of the
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -35,7 +35,7 @@ class GeneralLinear(Matrices):
         """Return the inverse of a matrix."""
         return gs.linalg.inv(point)
 
-    def random_uniform(self, n_samples=1, tol=1e-8):
+    def random_uniform(self, n_samples=1, tol=1e-6):
         """Sample in GL(n) from the uniform distribution.
 
         Parameters
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -17,8 +17,9 @@ class GeneralLinear(Matrices):
         point = gs.to_ndarray(point, to_ndim=3)
         _, mat_dim_1, mat_dim_2 = point.shape
         det = gs.linalg.det(point)
-        return gs.logical_and(mat_dim_1 == self.n, gs.where(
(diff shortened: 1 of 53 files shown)


pchauchat:
About the remaining DeepSource error, I am conflicted: the `belongs` in `matrices` does not need to do any check, and therefore does not need any tolerance value. To have matching signatures would mean add an unused parameter to one, or remove a potentially useful parameter to the other (unless it is specified in the core of the function).
Any preferrences @ninamiolane @nkoep  ?

geomstats/geometry/general_linear.py
```diff
@@ -30,6 +35,33 @@ def inv(point):
         """Return the inverse of a matrix."""
         return gs.linalg.inv(point)
 
+    def random_uniform(self, n_samples=1, tol=1e-8):
+        """Sample in GL(n) from the uniform distribution.
+
+        Parameters
+        ----------
+        n_samples: int, optional
```
ninamiolane, line 43: super-NIT: "n_samples : "
-> weirdly, there needs to be a space before the : in numpy's conventions for docstrings, which we are following.

geomstats/geometry/general_linear.py
```diff
@@ -30,6 +35,33 @@ def inv(point):
...
+            Number of samples.
+        tol: float, optional
```
ninamiolane, line 45: Ditto.

geomstats/geometry/general_linear.py
```diff
@@ -30,6 +35,33 @@ def inv(point):
         """Return the inverse of a matrix."""
         return gs.linalg.inv(point)
 
+    def random_uniform(self, n_samples=1, tol=1e-8):
```
ninamiolane, line 38: Numpy uses 1e-6 as a tolerance default, for example in np.isclose. You could use the same?

geomstats/geometry/general_linear.py
```diff
@@ -30,6 +35,33 @@ def inv(point):
...
+        samples = gs.random.rand(n_samples, self.n, self.n)
+        while True:
+            dets = gs.linalg.det(samples)
+            indcs = gs.isclose(dets, 0.0, atol=tol)
+            num_bad_samples = gs.sum(indcs)
+            if num_bad_samples == 0:
+                break
+            samples[indcs, :] = gs.random.rand(num_bad_samples, self.n, self.n)
+
+        return samples
```
ninamiolane, line 63: Great, thanks! Could you squeeze axis 0, for when there is only one sample?

ninamiolane:
@pchauchat (In addition to the PR's comments,) could you solve the GitHub conflicts? Thnaks!

pchauchat:
It's getting confused, I redid it in #558

PR #563, opened 2020-04-13
Adapt cos/sin from pytorch's backend to match numpy's behavior

ninamiolane:
Currently, pytorch's cos and sin only accept tensors as inputs. However, numpy accepts both arrays and values as inputs.

This PR:
- adapts the behavior of cos/sin from pytorch's backend to match numpy's behavior,
- adapts special_orthogonal's unit tests to use the new syntax.

@pchauchat @nguigs

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -1,5 +1,7 @@
 """Pytorch based computation backend."""
 
+from functools import wraps
+
 import numpy as _np
 import torch
 from torch import (  # NOQA
@@ -12,6 +14,7 @@ from torch import (  # NOQA
     atan2 as arctan2,
     ceil,
     clamp as clip,
+    cos,
     cosh,
     div as divide,
     empty_like,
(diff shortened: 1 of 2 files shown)


geomstats/_backend/pytorch/__init__.py
```diff
@@ -66,6 +64,18 @@ def _raise_not_implemented_error(*args, **kwargs):
 vectorize = _raise_not_implemented_error
 
 
+def sin(x):
```
nkoep, line 67: I was thinking that for primitives like this (and all other unary functions), we might be better off using a decorator which wraps scalars, e.g.,
```python
def _box_scalar(function):
    @functools.wraps(function)
    def wrapper(x):
        if not torch.is_tensor(x):
            x = torch.tensor(x)
        return function(x)
    return wrapper

sin = _box_scalar(sin)
sinh = _box_scalar(sinh)
```
We could also make this a little more flexible by writing a "decorator factory" to support a variable number of arguments.

ninamiolane: Great idea, let me do this!

ninamiolane:
@nguigs also added exp, log and others ;)

Issue #511, opened 2020-04-03
Inconsistent behavior of gs.array


pchauchat:
It seems gs.array behaves differently in Pytorch compared to Numpy and Tensorflow :
https://github.com/geomstats/geomstats/blob/6cf2aba8f917718028e445207eb3705fb56c98ab/geomstats/geometry/special_orthogonal.py#L298

Numpy and Tensorflow will transform the line into a (usually) 3D array, while Pytorch creates a 2D array. This is, as it is a list of tensors, it concatenates them along the first dimension.

I though ot calling to_ndarray before, but since it starts by calling array, it will loop. 
Edit on a possible solution: replace `val = concatenate(val)` by `val = stack(val)`, which looks more appropriate.

Another question is also whether that was intended, and if yes, what does it break ?

nguigs:
It would be helpful to pass a `dtype` argument to `gs.array` as well! It's already the case with numpy and tf, it would be nice to add it to pytorch

pchauchat:
I forgot to mention this issue again in the last two PRs, which propose a solution for the question of the `array` function, and add the `dtype` argument. This however required some machinery, which you may want to discuss : #574  #575

lpereira95:
I believe #1550 solved all the inconsistencies. If in the meantime we find an unexpected behavior, we can reopen this.

PR #566, opened 2020-04-14
Pytorch's array function (almost ?) equivalent to numpy's and tensorflow's

pchauchat:
This PR aims at narrowing the differences between all backends' `array` function, to address #511 .

A `dtype` optional argument is also added.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -139,7 +139,7 @@ def vstack(seq):
 
 
 def array(val, dtype=None):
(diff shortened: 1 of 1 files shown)


pchauchat:
Interestingly, arrays do have an `__iter__` attribute, so this does not distinguish them from lists and tuples.
The tests were running fine on my side (at least, `nose2 tests` showed no problem), I'll check whether there are conflicts with the new vectorization policy.

pchauchat:
Considering the number of commits that will be induced by rebasing with your current work, I think it is better to close this PR and reopen a new one once everything is settled.

PR #565, opened 2020-04-14
Vectorization of Stiefel, Grassmanian, SymmetricMatrices and SPDMatrices

ninamiolane:
This PR follows PR #568 and continues to address issue #338.

It uses the decorator in `geomstats/vectorization.py` to:
- factorize the reshapings with `gs.to_ndarray` from methods' bodies,
- ensure that the shape of the output matches the user's expectations.

I explain the strategy behind the refactorization/vectorization PR in a post in the associated issue #338.

Issue #472, opened 2020-03-31
Uniformize (default_)point_type and (default_)coords_type across manifolds and metrics


ninamiolane:
The parameter `point_type` sometimes refers to the type of ndarray ('vector` or `matrix`), and sometimes to the type of coordinates (`intrinsic`, `extrinsic`, `poincare`, etc).

We should create two parameters:
- point_type: for the type of ndarray,
- coords_type: for the type of coordinates.

These parameters would have default values at the instantiation of each manifold and metric, i.e. Manifold and RiemannianMetric objects should both have:
- default_point_type
- default_coords_type
as attributes.

PR #569, opened 2020-04-14
Vectorization of SpecialOrthogonal and SpecialEuclidean

ninamiolane:
[WIP]

This PR follows PR #568 and PR #565 and continues to address issue #338.

It uses the decorator in `geomstats/vectorization.py` to:
- factorize the reshapings with gs.to_ndarray from methods' bodies,
- ensure that the shape of the output matches the user's expectations,
in special_orthogonal.py and special_euclidean.py.

I explain the strategy behind the refactorization/vectorization PR in a post in the associated issue #338.


geomstats/geometry/invariant_metric.py
```diff
@@ -339,8 +341,12 @@ def left_log_from_identity(self, point):
         inner_prod_mat = self.inner_product_mat_at_identity
         inv_inner_prod_mat = gs.linalg.inv(inner_prod_mat)
         sqrt_inv_inner_prod_mat = gs.linalg.sqrtm(inv_inner_prod_mat)
-        aux = gs.squeeze(sqrt_inv_inner_prod_mat, axis=0)
-        log = gs.matmul(point, aux)
+        # aux = gs.squeeze(sqrt_inv_inner_prod_mat, axis=0)
```
ninamiolane, line 344: leftover comment

geomstats/geometry/invariant_metric.py
```diff
@@ -279,13 +282,13 @@ def exp(self, tangent_vec, base_point=None):
             Point in the group equal to the Riemannian exponential
             of tangent_vec at the base point.
         """
+        identity = gs.to_ndarray(self.group.identity, to_ndim=2)
```
nguigs, line 285: did you think about the `point_type='matrix'` case?

geomstats/geometry/invariant_metric.py
```diff
@@ -387,10 +391,12 @@ def log(self, point, base_point=None):
             Tangent vector at the base point equal to the Riemannian logarithm
             of point at the base point.
         """
+        identity = gs.to_ndarray(self.group.identity, to_ndim=2)
```
nguigs, line 394: idem

geomstats/geometry/invariant_metric.py
```diff
@@ -387,10 +391,12 @@ def log(self, point, base_point=None):
...
         if base_point is None:
-            base_point = self.group.identity
+            base_point = identity
         base_point = self.group.regularize(base_point)
-        if gs.allclose(base_point, self.group.identity):
+
+        if gs.allclose(base_point, identity):
```
nguigs, line 399: Did you check that this works when there are n > 1 base points ? I think I ran into errors with pytorch or tf in those cases

geomstats/geometry/lie_group.py
```diff
@@ -255,22 +253,14 @@ def exp(self, tangent_vec, base_point=None, point_type=None):
...
             The exponentiated tangent vector
         """
-        if point_type is None:
-            point_type = self.default_point_type
-
         identity = self.get_identity(point_type=point_type)
-        identity = self.regularize(identity, point_type=point_type)
+        identity = gs.to_ndarray(
+            identity,
+            to_ndim=geomstats.vectorization.POINT_TYPES_TO_NDIMS[point_type])
```
nguigs, line 259: Maybe that's what needs be done above as well (see previous comments)?

geomstats/geometry/lie_group.py
```diff
@@ -255,22 +253,14 @@ def exp(self, tangent_vec, base_point=None, point_type=None):
...
         if base_point is None:
             base_point = identity
         base_point = self.regularize(base_point, point_type=point_type)
 
-        if point_type == 'vector':
-            tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=2)
-            base_point = gs.to_ndarray(base_point, to_ndim=2)
-        if point_type == 'matrix':
-            tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=3)
```
nguigs, line 271: What about the case of one matrix as tangent vector? You would have n_tangent_vec = n > 1 here?

geomstats/geometry/lie_group.py
```diff
@@ -323,9 +315,6 @@ def log_not_from_identity(self, point, base_point, point_type=None):
         -------
         tangent_vec : array-like, shape=[n_samples, {dimension,[n,n]}]
         """
-        if point_type is None:
```
nguigs, line 326: how come we don't need this anymore?

nguigs: Is it handled in the decorator?

geomstats/geometry/special_orthogonal.py
```diff
@@ -399,21 +390,20 @@ def skew_matrix_from_vector(self, vec):
...
-            id_skew = gs.array([[[0., 1.], [-1., 0.]]] * n_vecs)
+            id_skew = gs.array(
+                gs.tile([[[0., 1.], [-1., 0.]]], (n_vecs, 1, 1)))
             skew_mat = gs.einsum(
                 '...ij,...i->...ij', gs.cast(id_skew, gs.float32), vec)
 
         elif self.n == 3:  # SO(3)
             # This avois dividing by 0.
-            levi_civita_symbol = gs.array([[
+            levi_civita_symbol = gs.tile([[
```
nguigs, line 406: typo avois -> avoids

geomstats/geometry/lie_group.py
ninamiolane: Yes!

geomstats/geometry/lie_group.py
ninamiolane: Yes, good eye!

geomstats/geometry/invariant_metric.py
ninamiolane: Argh, good point! And the default_point_type is in self.group.default_point_type and not in self.default_point_type so the logic is not handled by the decorator for this case 😭

geomstats/geometry/invariant_metric.py
ninamiolane: Done.

geomstats/geometry/invariant_metric.py
ninamiolane: Done.

geomstats/geometry/invariant_metric.py
ninamiolane: Not yet. `invariant_metric` is actually not refactorized in this PR: I only did SO and SE to keep the PR reasonable, just fixing what was breaking in invariant_metric for it.

The next PR will focus on invariant_metric.

geomstats/geometry/lie_group.py
ninamiolane: See above comment: the next PR will be on lie_groups and invariant_metric

geomstats/geometry/lie_group.py
ninamiolane: Right now the decorator does:
```
if point_type is None:
```
it needs to be generalized to handle the case that we have in invariant_metric, i.e.:
```
if point_type is None:
    point_type = self.group.default_point_type
```

geomstats/geometry/special_orthogonal.py
ninamiolane: Done.

geomstats/geometry/lie_group.py
nguigs: Don't worry about it, we decided in #472 that metrics would also have a default_point_type attribute

PR #573, opened 2020-04-15
Vectorization of Landmarks, DiscretizedCurves and ProductManifold

ninamiolane:
This PR continues to address issue #338. The strategy behind the refactorization/vectorization PRs is explained in a post in the associated issue #338.

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -162,6 +162,7 @@ class ProductManifold(Manifold):
             point_type = self.default_point_type
         geomstats.error.check_parameter_accepted_values(
             point_type, 'point_type', ['vector', 'matrix'])
+
         if point_type == 'vector':
             data = self.manifolds[0].random_uniform(n_samples)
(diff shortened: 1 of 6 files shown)


geomstats/geometry/product_manifold.py
```diff
@@ -172,4 +172,7 @@ def random_uniform(self, n_samples, point_type=None):
         point = [
             space.random_uniform(n_samples)
             for space in self.manifolds]
-        return gs.stack(point, axis=1)
+        samples = gs.stack(point, axis=1)
+        if n_samples == 1 and samples.shape[0] == 1:
```
nguigs, line 176: why the second condition? if there is only one sample don't we have `samples.shape[0] = n_manifolds`?

ninamiolane: Not necessarily, we could have samples.shape = (dim,) since most of the random_uniform(1) functions now squeeze the 0th dimension.

But actually, since the random_uniform functions are properly vectorized now, this condition is not needed at all.

PR #575, opened 2020-04-15
Include dtype option in array unit test

pchauchat:
This PR adds various `dtype` options to the `array` unit test. It also adds `uint8` to the backend functions accessible outside. This is for the test only at the moment.

### DIFF
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -27,12 +27,16 @@ class TestBackends(geomstats.tests.TestCase):
         np_mat = _np.array([_np.ones(3), _np.ones(3)])
         self.assertAllCloseToNp(gs_mat, np_mat)
 
-        gs_mat = gs.array([[gs.ones(3)], [gs.ones(3)]])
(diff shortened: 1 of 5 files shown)


tests/test_backends.py
```diff
@@ -27,12 +27,16 @@ def test_array(self):
         np_mat = _np.array([_np.ones(3), _np.ones(3)])
         self.assertAllCloseToNp(gs_mat, np_mat)
 
-        gs_mat = gs.array([[gs.ones(3)], [gs.ones(3)]])
-        np_mat = _np.array([[_np.ones(3)], [_np.ones(3)]])
+        gs_mat = gs.array([gs.ones(3), gs.ones(3)], dtype=gs.float64)
+        np_mat = _np.array([_np.ones(3), _np.ones(3)], dtype=_np.float64)
```
ninamiolane, line 31: Nice! Although I'm not sure that assertAllCloseToNp also checks that the results are of the same type. Could you maybe add a separate assert that checks the output's types?

Issue #570, opened 2020-04-15
Frechet Mean could implement a transform method


nguigs:
It could be useful that `FrechetMean` exposed a `transform` method that would return the logs of each data point in the tangent space at the mean, as 1d vectors, as is done in PCA:
https://github.com/geomstats/geomstats/blob/8992dd0b238cb840cf785a4b7a96d923b4e10253/geomstats/learning/pca.py#L238-L247

This would allow then to use this estimator as a preprocessing tool and use any off-the-shelf machine learning algorithm in a pipeline. What do you think @ninamiolane?

ninamiolane:
Yes, we could have `transform`, `fit_transform` and `inverse_transform`, that'd be great! But, in order to stay close to scikit-learn's logic, maybe the transform method could first take the logarithm and then apply the dimension reduction (thus in effect giving the loadings on the different eigenvectors).

https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

nguigs:
That’s what is done currently in pca isn’t it? I was talking those 3 methods in the FréchetMean estimator without svd whatsover

ninamiolane:
Oh sorry, I read too fast and thought you meant TangentPCA for some reason. Yes, let's do the log in FrechetMean.

PR #577, opened 2020-04-16
Transformer for Frechet Mean

nguigs:
Add `transform, fit_transform, inverse_transform` to the Frechet Mean estimator, addressing #570.
The next step will be to refactor these methods in the PCA estimator to use the ones in Frechet Mean, and do the same for exponential barycenter

### DIFF
--- a/geomstats/learning/frechet_mean.py
+++ b/geomstats/learning/frechet_mean.py
@@ -388,6 +388,7 @@ class FrechetMean(BaseEstimator, TransformerMixin):
         -------
         X_new : array-like, shape=[n_samples, dim]
         """
+        # TODO(nguis): put this in a dedicated class
         if base_point is None:
             base_point = self.estimate_
 
--- a/geomstats/learning/frechet_mean.py
+++ b/geomstats/learning/frechet_mean.py
@@ -3,11 +3,14 @@
 import logging
 import math
 
-from sklearn.base import BaseEstimator
+from sklearn.base import BaseEstimator, TransformerMixin
 
 import geomstats.backend as gs
 from geomstats.geometry.euclidean import EuclideanMetric
+from geomstats.geometry.matrices import Matrices
 from geomstats.geometry.minkowski import MinkowskiMetric
+from geomstats.geometry.skew_symmetric_matrices import SkewSymmetricMatrices
+from geomstats.geometry.symmetric_matrices import SymmetricMatrices
 
 EPSILON = 1e-4
 
@@ -291,7 +294,7 @@ def _adaptive_gradient_descent(points,
     return gs.to_ndarray(current_mean, to_ndim=2)
 
 
-class FrechetMean(BaseEstimator):
+class FrechetMean(BaseEstimator, TransformerMixin):
     """Empirical Frechet mean.
 
     Parameters
@@ -302,7 +305,7 @@ class FrechetMean(BaseEstimator):
     def __init__(self, metric,
                  max_iter=32,
                  epsilon=EPSILON,
-                 point_type='vector',
+                 point_type=None,
                  method='default',
                  verbose=False):
         self.metric = metric
@@ -311,6 +314,9 @@ class FrechetMean(BaseEstimator):
         self.point_type = point_type
         self.method = method
         self.verbose = verbose
+
+        if point_type is None:
+            self.point_type = metric.default_point_type
 
     def fit(self, X, y=None, weights=None):
         """Compute the empirical Frechet mean.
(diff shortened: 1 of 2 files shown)


geomstats/learning/frechet_mean.py
```diff
@@ -305,12 +309,19 @@ def __init__(self, metric,
                  point_type='vector',
                  method='default',
                  verbose=False):
+        error.check_parameter_accepted_values(
```
ninamiolane, line 312: We/I should probably factorize this into a `check_point_type` method, or the standalone decorator that also checks if point_type is None.

geomstats/learning/frechet_mean.py
```diff
@@ -355,3 +366,87 @@ def fit(self, X, y=None, weights=None):
         self.estimate_ = mean
 
         return self
+
+    def transform(self, X, y=None, base_point=None):
+        """Lift data to a tangent space.
+
+        Compute the logs of all data point and reshapes them to
+        1d vectors if necessary. By default the logs are taken at the mean
```
ninamiolane, line 374: Is it weird to let have the user be able to pass another base_point, knowing that this class if FrechetMean and passing another base_point now has nothing to do with the mean? They can use the logs directly if they want this, right?

geomstats/learning/frechet_mean.py
```diff
@@ -355,3 +366,87 @@ def fit(self, X, y=None, weights=None):
...
+            vectors at the base point.
+
+        Returns
+        -------
+        X_new : array-like, shape=[n_samples, dim]
+        """
+        if base_point is None:
+            base_point = self.estimate_
+
+            if self.estimate_ is None:
```
ninamiolane, line 394: Needs to be unindented?

geomstats/learning/frechet_mean.py
```diff
@@ -355,3 +366,87 @@ def fit(self, X, y=None, weights=None):
...
+
+        Returns
+        -------
+        X_original : array-like, shape=[n_samples, {dim, [n, n]}
+            Data lying on the manifold.
+        """
+        if base_point is None:
+            base_point = self.estimate_
+
+            if self.estimate_ is None:
```
ninamiolane, line 436: Same?

geomstats/learning/frechet_mean.py
```diff
@@ -355,3 +366,87 @@ def fit(self, X, y=None, weights=None):
...
+                raise RuntimeError('fit needs to be called first or a '
+                                   'base_point passed.')
+
+        if self.point_type == 'matrix':
+            if gs.all(Matrices.is_symmetric(base_point)):
+                tangent_vecs = SymmetricMatrices(
+                    base_point.shape[-1]).symmetric_matrix_from_vector(X)
+            elif gs.all(Matrices.is_skew_symmetric(base_point)):
```
ninamiolane, line 444: The two last elif and else are not tested.

geomstats/learning/frechet_mean.py
```diff
@@ -355,3 +366,87 @@ def fit(self, X, y=None, weights=None):
...
+
+        tangent_vecs = self.metric.log(X, base_point=base_point)
+
+        if self.point_type == 'vector':
+            return tangent_vecs
+
+        if gs.all(Matrices.is_symmetric(tangent_vecs)):
+            X = SymmetricMatrices.vector_from_symmetric_matrix(
+                tangent_vecs)
+        elif gs.all(Matrices.is_skew_symmetric(tangent_vecs)):
```
ninamiolane, line 406: elif and else are not tested.

geomstats/learning/frechet_mean.py
nguigs: I thought it could be useful as one might want to investigate wether using the mean compared to a point (e.g. identity if spd matrices) improves classification.. Otherwise we could put this in a standalone preprocessor or a centralizer transformer... It's not only the log but also the reshaping to 1d that can be useful...

geomstats/learning/frechet_mean.py
nguigs: indent was intended: you don't need to fit if base_point is given

geomstats/learning/frechet_mean.py
nguigs: No

geomstats/learning/frechet_mean.py
ninamiolane: But what about base_point=self.estimate_ is given? Unlikely?

geomstats/learning/frechet_mean.py
nguigs: I knoooow, but FrechetMean doesn't work with point_type='matrix' for SOn.bi_invariant_metric (maybe it's fixed with your vectorization...)

geomstats/learning/frechet_mean.py
ninamiolane: I agree, but I'm not sure these should be put as a method of `FrechetMean`. Maybe these should be put as auxiliary functions to preprocess data in the other learning algorithms?

nguigs: The point is any machine learning algorithm from sklearn can be used after this, so I was thinking we should have this preprocessor instead of implementing other learning algorithms that work in a tangent space

nguigs: And it would finally hide all the geometry for ML users. If your point is that this should not be in Frechet mean I don't disagree, but I was thinking it's convenient here and instead of creating yet another class

geomstats/learning/frechet_mean.py
nguigs: aaaah stg like `mean.transform(X, base_point=mean.estimate_)`? True...

geomstats/learning/frechet_mean.py
ninamiolane: Agreed that a preprocessor would be nice, but we should probably put it in a separate `Projector` class?

geomstats/learning/frechet_mean.py
ninamiolane: how about `default_gradient_descent`, I thought this one was working with matrices?

nguigs: yes if metric.log works, but last time I tried with SOn it didn't

ninamiolane: Got it :/

geomstats/learning/frechet_mean.py
nguigs: But then mean.estimate_ == None so it works?

ninamiolane: yes, ok!

Issue #567, opened 2020-04-14
Inconsitent behavior of gs.split


nguigs:
With numpy backend:
```python
>>> gs.split(gs.arange(13), gs.array([2, 5]))
[array([0, 1]), array([2, 3, 4]), array([ 5,  6,  7,  8,  9, 10, 11, 12])]
```
This behavior can be useful e.g. to split according to a cumulative sum (in product_manifold). Whereas with pytorch:
```python
>>> gs.split(gs.arange(13), gs.array([2, 5]))
TypeError: split_with_sizes(): argument 'split_sizes' (position 1) must be tuple of ints, not Tensor
```
And even
```python
>>> gs.split(gs.arange(13), (2, 5))
RuntimeError: split_with_sizes expects split_sizes to sum exactly to 13 (input tensor's size at dimension 0)
```
Same with tf:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Determined shape must either match input shape along split_dim exactly if fully specified, or be less than the size of the input along split_dim if not fully specified.```

pchauchat:
If I understand correctly, , in the Numpy version the second argument provides the indices at which the split occurs ?

I had a similar problem when I wanted to adapt `hsplit` and make it consistent, I'll see if I can adapt it to `split`.

nguigs:
Yes, but I think it can also accept a single int n to split in n equal batches

pchauchat:
Yep, that was the main problem with `hsplit`, one version takes as input the size of the returned batches, while the other takes the number of batches, and I had to put a conversion in. I'm working on it right now, it should be done tomorrow morning

PR #579, opened 2020-04-17
Adapt split of tf and pytorch to Numpy's

pchauchat:
This PR adresses #567 : `split` functions are defined in Pytorch and Tensorflow to match the behavior of Numpy's one. This is done by inferring the right arguments to pass to them from the arguments that are normallygiven when using Numpy.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -92,8 +92,17 @@ def empty(shape, dtype=float64):
     return torch.empty(*shape, dtype=dtype)
 
 
-def split(ary, indices_or_sections, axis=0):
-    return torch.split(ary, indices_or_sections, dim=axis)
+def split(x, indices_or_sections, axis=0):
+    if isinstance(indices_or_sections, int):
+        indices_or_sections = indices_or_sections // x.shape[axis]
+        return torch.split(x, indices_or_sections, dim=axis)
+    indices_or_sections = _np.array(indices_or_sections)
+    intervals_length = indices_or_sections[1:] - indices_or_sections[:-1]
+    last_interval_length = x.shape[axis] - indices_or_sections[-1]
+    if last_interval_length > 0:
+        intervals_length = _np.append(intervals_length, last_interval_length)
+    intervals_length = _np.insert(intervals_length, 0, indices_or_sections[0])
+    return torch.split(x, tuple(intervals_length), dim=axis)
 
 
 def logical_or(x, y):
@@ -604,6 +613,8 @@ def copy(x):
 
 
 def cumsum(x, axis=None):
+    if not torch.is_tensor(x):
+        x = array(x)
     if axis is None:
         return x.flatten().cumsum(dim=0)
     return torch.cumsum(x, dim=axis)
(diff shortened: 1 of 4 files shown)


geomstats/_backend/pytorch/__init__.py
```diff
@@ -92,8 +92,17 @@ def empty(shape, dtype=float64):
     return torch.empty(*shape, dtype=dtype)
 
 
-def split(ary, indices_or_sections, axis=0):
-    return torch.split(ary, indices_or_sections, dim=axis)
+def split(x, indices_or_sections, axis=0):
+    if isinstance(indices_or_sections, int):
+        indices_or_sections = indices_or_sections // x.shape[axis]
```
nguigs, line 97: shouldn't it be the inverse? `torch.split(x, k)` will return n // k (+1) arrays of size k (except the last one) while `numpy.split(x, k)` would return k arrays of size n // k, where `n = x.shape[axis]`, and throw an error if `n % k != 0`

tests/test_product_manifold.py
```diff
@@ -28,15 +28,13 @@ def test_dimension(self):
         result = self.space_vector.dim
         self.assertAllClose(result, expected)
 
-    @geomstats.tests.np_and_pytorch_only
```
nguigs, line 31: :tada:

geomstats/_backend/pytorch/__init__.py
pchauchat: True, I had mixed them up, and everything was working fine since we were in a case were `indices_or_sections // x.shape[axis] = 1` ...

geomstats/_backend/pytorch/__init__.py
```diff
@@ -604,6 +613,8 @@ def copy(x):
 
 
 def cumsum(x, axis=None):
+    if not torch.is_tensor(x):
+        x = array(x)
```
ninamiolane, line 617: Yes, I had to add a similar condition in less_equal for torch. Why using `x = array(x)` here and not `x = torch.tensor(x)`, which could be faster? Do we need the logic of the `array` function?

geomstats/geometry/hypersphere.py
```diff
@@ -301,6 +301,7 @@ def random_uniform(self, n_samples=1):
         while True:
             norms = gs.linalg.norm(samples, axis=1)
             indcs = gs.isclose(norms, 0.0)
+            indcs = gs.cast(indcs, gs.int32)
```
ninamiolane, line 304: Since `np.sum` works with booleans, could we modify `gs.sum`, in order to do the cast to `gs.int32` in the tf/torch's backend function and not in the main code?

tests/test_product_manifold.py
```diff
@@ -28,15 +28,13 @@ def test_dimension(self):
...
     def test_random_and_belongs_matrix(self):
         n_samples = 5
         data = self.space_matrix.random_uniform(n_samples)
         result = self.space_matrix.belongs(data)
         expected = gs.array([True] * n_samples)
         self.assertAllClose(result, expected)
 
-    @geomstats.tests.np_only
```
ninamiolane, line 39: Wooohoo! 🎉

geomstats/_backend/pytorch/__init__.py
pchauchat: It depends on what kind of input this should accept. Say you want to do something like `np.cumsum([np.ones(3), np.ones(3)])`, then you will need the `array` function. Otherwise, `x = torch.tensor(x)` would be fine.

ninamiolane: Got it, thanks for explaining.

PR #580, opened 2020-04-18
Vectorization of Connection, PoincareBall, Hyperbolic coords, and PoincarePolydisk

ninamiolane:
This PR continues to address issue #338. The strategy behind the refactorization/vectorization PRs is explained in a post in the associated issue #338.


geomstats/geometry/general_linear.py
```diff
@@ -108,13 +108,13 @@ def log(cls, point, base_point=None):
 
         Parameters
         ----------
-        point :         array-like, shape=[..., n, n]
-        base_point :    array-like, shape=[..., n, n]
+        point : array-like, shape=[..., n, n]
```
nguigs, line 111: do we use ... now as well as n_samples?

geomstats/geometry/hyperbolic.py
```diff
@@ -108,9 +109,7 @@ def _extrinsic_to_intrinsic_coordinates(point_extrinsic):
         -------
         point_intrinsic : array-like, shape=[n_samples, dim]
         """
-        point_extrinsic = gs.to_ndarray(point_extrinsic, to_ndim=2)
-
-        point_intrinsic = point_extrinsic[:, 1:]
+        point_intrinsic = point_extrinsic[..., 1:]
```
nguigs, line 112: :heart_eyes:

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -174,8 +174,12 @@ def inner_product(
...
         if point_type == 'matrix':
+            # TODO(ninamiolane): Vectorize this more efficiently
+            tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=2)
             tangent_vec_a = gs.to_ndarray(tangent_vec_a, to_ndim=3)
+            tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=2)
             tangent_vec_b = gs.to_ndarray(tangent_vec_b, to_ndim=3)
+            base_point = gs.to_ndarray(base_point, to_ndim=2)
             base_point = gs.to_ndarray(base_point, to_ndim=3)
             inner_products = [metric.inner_product(tangent_vec_a[:, i],
                                                    tangent_vec_b[:, i],
```
nguigs, line 185: indentation

nguigs: what about using `tangent_vec_a[..., i, :]` here?

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -261,11 +265,14 @@ def log(self, point, base_point=None, point_type=None):
             return logs
 
         if point_type == 'matrix':
-            point = gs.to_ndarray(point, to_ndim=3)
-            base_point = gs.to_ndarray(base_point, to_ndim=3)
-            return gs.stack(
+            point = gs.to_ndarray(point, to_ndim=2, axis=0)
```
nguigs, line 268: shouldn't we use `riemannian_metric` if there is only one metric in the product to rule out this case?

geomstats/geometry/general_linear.py
ninamiolane: Good question. I reverted to shape=[n_samples, n, n] or shape=[n, n] since this is the convention of most of the codebase, this is more explicit anyway, and the goal of this PR is not to change this.

geomstats/geometry/product_riemannian_metric.py
ninamiolane: done and done

geomstats/geometry/product_riemannian_metric.py
ninamiolane: Yes! Although I'll leave that for a next PR if this is OK, this the goal of this one is not to clean this file but not use vectorization.

Issue #412, opened 2020-01-29
Harmonising method names 


opeltre:
It would be nice to harmonise method names! Following an overall common scheme and finding short transparent names would make object's interfaces be easier to learn and remember.

For instance, we could start using "to" and "from" more systematically (actual names in comment):
```python
manifold.is_point                     # belongs
manifold.to_point                    #  regularize?

manifold.is_tangent                 # is_tangent_vector 
manifold.to_tangent               # projection_to_tangent_space

matrices.is_symmetric             
matrices.to_symmetric            # make_symmetric
```
Any suggestion welcome below :)

nguigs:
Definitely agree that is_tangent/to_tangent are necessary and should be added to as many manifolds as possible. Also agree that regularize is not informative, so belongs can be changed to is_point to match the others.

stefanheyder:
I also agree! `is_X` and `to_X` are good naming conventions 

* you can see the return type from the method name
* `regularize` is ambiguous 
* uniform naming is a good thing to have!

opeltre:
For SO, SE, SPD: 
+ `from_vector`  ? 
+ `to_vector` ? 
Converting symmetric/skew-symmetric matrices to 1D-array using a basis

nguigs:
`from_1d` ? `to_1d`?

ninamiolane:
Agreed that uniformizing methods names is great, and `from_`, `to_` naming is explicit! 

Some remarks:

1. What about the methods that can be in several manifolds: `matrices.is_symmetric, matrices.to_symmetric` can also be in symmetric_matrices: `sym_matrices.is_point`, `sym_matrices.to_point`?
    - I think we should keep both, but factorize the code in auxiliary functions, in order to avoid copy-pasting and changing the code at 2 locations when we want to change its logic.

2. The methods `.to_vector`, `.from_vector` are great to change `point_type`, i.e. moving back and forth between `vector` and `matrix` representations. Can we adapt this naming scheme to `coords_types`, i.e. the different coordinate systems with which we can represent a point? 

For example, a point in hyperbolic space can be represented in `explicit`, `implicit`, or `ball` coordinates, which are all `point_type='vector'` representations. Writing `hyperbolic.to_ball`, is not explicit enough, and we do not know from which coords_type we are converting. 

Several solutions:
    a. Add `from` in `hyperbolic.to_ball(x, from='extrinsic')`, and `to` in `hyperbolic.from_ball(x, to='intrinsic')`
[...]

PR #582, opened 2020-04-19
Change SOn to SO3

nguigs:
WIP

This PR updates #484. Use different classes for SOn (matrices), SO3 (vector) and SO2(vector). We can discuss which methods should remain in SO3. I think we could leave both classes in the same file to make it easier for the user to import either (instead of knowing you have to import from special_euclidean3 and not special_euclidean)

Another class could be created for SO(2) in vector representation, and SE(2). SO(3), SO(n), SE(3) and SE(n) are now fixed.

The examples and learning algs still need to be adapted.

Any help for the docstrings and addressing DeepSource is welcome!


examples/plot_bch_so3.py
```diff
@@ -19,7 +19,7 @@
 
 import geomstats.backend as gs
 from geomstats.geometry.skew_symmetric_matrices import SkewSymmetricMatrices
-from geomstats.geometry.special_orthogonal import SpecialOrthogonal
+from geomstats.geometry.special_orthogonal3 import SpecialOrthogonal
```
ninamiolane, line 22: Do we want to name it `SpecialOrthogonal3` so that there is no conflict when a user wants to import both `SpecialOrthogonal3` and `SpecialOrthogonal` in the same file?

Or should they always import `SpecialOrthogonal`, and entering `n=3` calls `SpecialOrthogonal3` behind the scenes? (probably preferred?)

nguigs: how would behind the scene work? return self in the `__init__`?

ninamiolane: Have a `if n == 3` in the SpecialOrthogonal's `__init__` that calls `SpecialOrthogonal3.__init__()`, and `else` that runs the usual constructor's code?

nguigs:
For DeepSource issues, what should we do when we want to match the overidden method's signature but this results in unused arguments?

geomstats/geometry/special_euclidean.py
```diff
@@ -453,277 +431,224 @@ def jacobian_translation(
...
-        point_type: str, {'vector', 'matrix'}, optional
-            default: self.default_point_type
 
         Returns
         -------
         random_point: array-like,
-            shape=[n_samples, {dim, [n + 1, n + 1]}]
-            An array of random elements in SE(n) having the given point_type.
+            shape=[n_samples, 3]
+            An array of random elements in SE(3) having the given.
```
ninamiolane, line 668: missing point_type

geomstats/geometry/special_orthogonal.py
```diff
@@ -26,104 +27,105 @@
...
+
+    def _is_in_lie_algebra(self, tangent_vec, atol=ATOL):
+        return self.lie_algebra.belongs(tangent_vec, atol=atol)
+
+    @classmethod
+    def _to_lie_algebra(cls, tangent_vec):
+        """Project vector onto skew-symmetric matrices."""
+        return cls.make_skew_symmetric(tangent_vec)
+
+    def random_uniform(self, n_samples=1, tol=1e-6):
```
ninamiolane, line 57: Need docstring

geomstats/geometry/special_orthogonal.py
```diff
@@ -26,104 +27,105 @@
...
+        if n_samples == 1:
+            random_mat = gs.random.rand(self.n, self.n)
+        else:
+            random_mat = gs.random.rand(n_samples, self.n, self.n)
+        skew = self.to_tangent(random_mat)
+        return self.exp(skew)
+
+
+class SpecialOrthogonal3(LieGroup):
```
ninamiolane, line 80: private class?

geomstats/geometry/special_orthogonal.py
```diff
@@ -1278,246 +1139,219 @@ def tait_bryan_angles_from_rotation_vector(
...
         """
-        if point_type == 'vector':
-            tangent_vec = self.regularize(point, point_type=point_type)
-        elif point_type == 'matrix':
-            point = self.rotation_vector_from_matrix(point)
-            tangent_vec = self.skew_matrix_from_vector(point)
-        return tangent_vec
+        return self.regularize(point)
+
+    def lie_bracket(
```
ninamiolane, line 1350: Same comment: if we keep the class LieAlgebra, then we should create the lie algebra of SO and SE and put these methods there. Alternatively, we could delete the `LieAlgebra` class for the moment, and keep all of these methods in the groups.

geomstats/learning/exponential_barycenter.py
```diff
@@ -61,12 +61,12 @@ def _default_gradient_descent(
             break
         inv_mean = group.inverse(mean)
         centered_points = group.compose(inv_mean, points)
-        logs = group.log_from_identity(point=centered_points)
+        logs = group.log(point=centered_points)
```
ninamiolane, line 64: 🎉

geomstats/visualization.py
```diff
@@ -5,12 +5,12 @@
...
-from geomstats.geometry.special_euclidean import SpecialEuclidean
-from geomstats.geometry.special_orthogonal import SpecialOrthogonal
+from geomstats.geometry.special_euclidean import SpecialEuclidean3
+from geomstats.geometry.special_orthogonal import SpecialOrthogonal3
 from mpl_toolkits.mplot3d import Axes3D  # NOQA
 
-SE3_GROUP = SpecialEuclidean(n=3)
-SO3_GROUP = SpecialOrthogonal(n=3)
+SE3_GROUP = SpecialEuclidean3()
+SO3_GROUP = SpecialOrthogonal3()
```
ninamiolane, line 13: Ditto: this syntax is definitely not user-friendly. Let's keep SpecialEuclidean(n=3) and modify SpecialEuclidean's constructor, to let it handle the n=3 and 'vector' case.

ninamiolane:
> For DeepSource issues, what should we do when we want to match the overidden method's signature but this results in unused arguments?

Yeah, I got this problem too :( How about we just ignore this issue in DeepSource altogether?

ninamiolane:
I've ignored this DeepSource issue for the whole codebase, as it seems to be something that can be avoided:
https://stackoverflow.com/questions/27045439/how-to-define-a-method-in-the-superclass-so-that-it-will-be-compatible-with-any

geomstats/geometry/special_orthogonal.py
nguigs: yes although it's not really the same bracket: in lie algebra it's the usual commutator, but in Lie group it's the lie bracket for left-invariant vector fields

geomstats/geometry/invariant_metric.py
```diff
@@ -363,12 +363,49 @@ def log(self, point, base_point=None):
...
-            '...j,...jk->...k',
-            log_from_id,
-            Matrices.transpose(jacobian))
+        log = gs.einsum('...ij,...j->...i', jacobian, log_from_id)
         return log
+
+
+class BiInvariantMetric(InvariantMetric):
+    def __init__(self, group):
+        super(BiInvariantMetric, self).__init__(
```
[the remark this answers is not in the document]
nguigs: it's the only one to my knowledge... But I put it here in case we implement other ones in the future. Yes we could do a check

geomstats/geometry/lie_group.py
```diff
@@ -207,17 +220,17 @@ def exp_not_from_identity(self, tangent_vec, base_point, point_type=None):
         """
         if point_type == 'vector':
             jacobian = self.jacobian_translation(
-                point=base_point, left_or_right='left', point_type=point_type)
+                point=base_point, left_or_right='left')
```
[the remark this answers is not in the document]
nguigs: not sure to understand, do you mean there would be a problem if the user passes a different point_type from default_point_type?

geomstats/geometry/lie_group.py
```diff
@@ -368,19 +381,48 @@ def lie_bracket(
...
+        is_tangent = self._is_in_lie_algebra(tangent_vec_at_id, atol)
+        return is_tangent
+
+    def _to_lie_algebra(self, tangent_vec):
+        """Project a vector onto the lie algebra."""
+        raise NotImplementedError(
+            'The Lie Algebra belongs method is not implemented'
+        )
+
+    def to_tangent(self, tangent_vec, base_point=None):
```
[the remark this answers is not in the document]
nguigs: I know but this was discussed in #412

geomstats/geometry/matrices.py
```diff
@@ -149,6 +150,21 @@ def make_symmetric(cls, mat):
         """
         return 1 / 2 * (mat + cls.transpose(mat))
 
+    @classmethod
+    def make_skew_symmetric(cls, mat):
```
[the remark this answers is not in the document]
nguigs: I think they could be called as projection or to_space in symmetric_matrices.py and skew_symmetric_matrices.py but could be useful in other contexts so should stay in that class

examples/plot_bch_so3.py
nguigs: calling or returning `SpecialOrthogonal3.__init__()` or `SpecialOrthogonal3.__init__(self)` does not work. I agree with you, I just don't know how to do, maybe a wrapper deciding which class to instantiate?

ninamiolane: It needs to inherit from SpecialOrthogonal3 in order to work. Then, this syntax should work:
https://github.com/geomstats/geomstats/blob/e44f44a968fbfa963313cc4fe71d326b6f66e9cd/geomstats/geometry/special_orthogonal.py#L60

What fails when you try it?

geomstats/geometry/invariant_metric.py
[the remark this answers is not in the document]
ninamiolane: Riemannian metric are allowed to be pseudo metrics. Therefore, BiInvariantMetric includes bi-invariant pseudo-metrics. So we have at least SE(3) also, with a bi-invariant pseudo-metric. See eq 31 here: https://www.cis.upenn.edu/~cis610/SE3-Croke-Kumar.pdf

geomstats/geometry/lie_group.py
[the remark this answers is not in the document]
ninamiolane: Yes. I meant, would this work for example:

```
so = SpecialOrthogonal(n=3, default_point_type='matrix')
# [...] defining vec and pt in vector representation
so.exp_not_from_identity(vec, pt, point_type='vector')
```
because it will go get the `jacobian_translation` in the `matrix` representation, whereas `vec` and `pt` are in vector representation.

I guess, my confusion is maybe on what we mean with `point_type` for the method `jacobian_translation`. I understood it as: `point_type` of the vectors on which the Jacobian acts. Therefore, it `point_type`=='vector' then jacobian_translation returns a matrix, if `point_type`=='matrix', then jacobian_translation returns... also a matrix?

examples/plot_bch_so3.py
nguigs: But does it make sense to have `_SpecialOrthogonal3` as a parent class of `SpecialOrthogonal`? And will it actually work with MRO (because SpecialOrthogonal must inherit from LieGroup and GeneralLinear)? (I will try)

geomstats/geometry/invariant_metric.py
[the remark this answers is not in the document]
nguigs: ok I didn't know we had pseudo-metrics

geomstats/geometry/lie_group.py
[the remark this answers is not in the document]
nguigs: Ok I see the problem so we want to remove the point_type arg here as well? I think the best would also to implement direcly the push forward of a tangent vector by the left (or right) translation rather than just the jacobian (which always returns a matrix) because it will allow to deal with all combinations of left, right, vector, matrix in one method instead of everytime we want to go to identity!

geomstats/geometry/special_euclidean.py
```diff
@@ -34,15 +36,101 @@
...
+
+        last_term = point[..., self.n:, self.n:]
+        belongs = gs.logical_and(
+            belongs, gs.all(last_term == 1, axis=(-2, -1)))
+
+        if point.ndim == 2:
+            return gs.squeeze(belongs)
+        return gs.flatten(belongs)
+
+    def _is_in_lie_algebra(self, tangent_vec, atol=TOLERANCE):
```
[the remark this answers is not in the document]
nguigs: it could be useful in a next refactorization maybe

geomstats/geometry/special_euclidean.py
```diff
@@ -34,15 +36,101 @@
                         - 1. / 362880.]
 
 
-class SpecialEuclidean(LieGroup):
+class SpecialEuclidean(GeneralLinear, LieGroup):
+    """Class for special orthogonal groups."""
+
+    def __init__(self, n):
+        super(SpecialEuclidean, self).__init__(
```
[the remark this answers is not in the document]
nguigs: SpecialOrthogonal3 cannot be initilazied this way: because, then in the call to super in its init you get
`TypeError: super(type, obj): obj must be an instance or subtype of type`

geomstats/geometry/invariant_metric.py
```diff
@@ -363,12 +363,63 @@ def log(self, point, base_point=None):
...
+    ----------
+    group : LieGroup
+        The group to equip with the bi-invariant metric
+    """
+
+    def __init__(self, group):
+        super(BiInvariantMetric, self).__init__(
+            group=group, inner_product_mat_at_identity=gs.eye(group.dim),
+            default_point_type=group.default_point_type)
+        if 'SpecialOrthogonal' not in group.__str__():
```
ninamiolane, line 389: doesn't `isinstance` work for classes?

geomstats/geometry/lie_group.py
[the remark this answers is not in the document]
ninamiolane: Yes, removing the point_type in this function sounds great: did we also decide to move to one `point_type` per object or one `point_type` per class? (maybe for another PR) removing the point_type of this function would definitely go in this direction.

Yes, agreed for implementing the push-forward directly (provided that the `jacobian_translation` matrix itself is not needed somewhere else in the code? even for another Lie group for example?)

geomstats/geometry/lie_group.py
```diff
@@ -368,19 +381,63 @@ def lie_bracket(
         if point_type is None:
```
ninamiolane, line 381: How come this code is not tested? I thought we had tests for the lie_bracket.

geomstats/geometry/lie_group.py
```diff
@@ -368,19 +381,48 @@ def lie_bracket(
...
+
+        if gs.allclose(base_point, self.identity):
+            tangent_vec_at_id = tangent_vec
+        else:
+            tangent_vec_at_id = self.compose(
+                self.inverse(base_point), tangent_vec)
+        is_tangent = self._is_in_lie_algebra(tangent_vec_at_id, atol)
+        return is_tangent
+
+    def _to_lie_algebra(self, tangent_vec):
```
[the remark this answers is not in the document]
ninamiolane: Ok, I replied on issue #412 .

geomstats/geometry/matrices.py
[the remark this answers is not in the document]
ninamiolane: Yes, I also replied to issue #412 on this. Such conversion code could be in two methods (as it ultimately depends on the user's preferences). Yet, the body of the function should be factorized in a private function somewhere. This way, we avoid duplicate code.

geomstats/geometry/skew_symmetric_matrices.py
```diff
@@ -9,7 +9,7 @@
 from geomstats.geometry.matrices import Matrices
 
 
-TOLERANCE = 1e-12
+TOLERANCE = 1e-8
```
ninamiolane, line 12: Maybe name this `TOL`? we have `tol`, `atol`, `rtol` as naming in the code. Unless it is the tolerance for a particular function, in this case `***_TOL` could be more explicit?

Side remark: numpy default tolerances are `rtol=1e-5, atol=1e-8`, for example: 
https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html. I wonder if we could factorize our tolerance levels somewhere?

geomstats/geometry/special_euclidean.py
```diff
@@ -22,6 +23,7 @@
 
 
 ATOL = 1e-5
+TOLERANCE = 1e-8
```
ninamiolane, line 26: more explicit name, see above?

geomstats/geometry/special_euclidean.py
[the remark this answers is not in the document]
ninamiolane: Is this because `SpecialOrthogonal3` is not added as a parent class of `SpecialOrthogonal`, therefore its constructor cannot be called?

There might be an error of class diagrams if we add it as a parent class of `SpecialOrthogonal`. In this case, we might want to create two privates classes `_SpecialOrthogonal3vector` and `_SpecialOrthogonaln` that will be then wrapped into the public class `SpecialOrthogonal`.

geomstats/geometry/special_euclidean.py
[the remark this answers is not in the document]
ninamiolane: Yes.

geomstats/geometry/general_linear.py
```diff
@@ -7,7 +7,7 @@
 class GeneralLinear(Matrices):
     """Class for the general linear group GL(n)."""
 
-    def __init__(self, n):
+    def __init__(self, n, **kwargs):
```
ninamiolane, line 10: If this was to get rid of the DeepSource issue, this is not needed anymore.

geomstats/geometry/invariant_metric.py
```diff
@@ -30,7 +30,8 @@ class InvariantMetric(RiemannianMetric):
 
     def __init__(self, group,
                  inner_product_mat_at_identity=None,
-                 left_or_right='left'):
+                 left_or_right='left', **kwargs):
```
ninamiolane, line 33: Same DeepSource issue comment.

geomstats/geometry/special_orthogonal.py
```diff
@@ -26,105 +27,119 @@
...
         ----------
-        point_type : str, {'vector', 'matrix'}, optional
-            the point_type of the returned value
+        point_type : str,
+            The point_type of the returned value. Unused here.
 
         Returns
         -------
-        identity : array-like, shape={[dim], [n, n]}
+        identity : array-like, shape={3,}
```
ninamiolane, line 114: NIT: shape=[3,]

geomstats/geometry/special_orthogonal.py
ninamiolane: But the space of left-invariant vector fields is isomorphic to the Lie algebra, and therefore can be identified to it, and to any tangent space.

As a matter of fact, this lie_bracket function operates on tangent_vectors, and not tangent vector fields, so it amounts to the lie bracket on the lie algebra when `base_point=None`?

ninamiolane:
Great! Let's address the more pressing aspects and keep the next refactorization steps for another PR.

Could you look at the remaining DeepSource issues? In particular, this seems to be a problem:
`result = self.group._is_in_lie_algebra(vec_2)`

https://docs.quantifiedcode.com/python-anti-patterns/correctness/accessing_a_protected_member_from_outside_the_class.html

geomstats/geometry/lie_group.py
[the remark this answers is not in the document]
nguigs: it can only be usefull for a Lie group represented by vectors, because if we represent by matrices, jacobian_translation(point) just returns point (and the left_right info is actually missing)

geomstats/geometry/special_euclidean.py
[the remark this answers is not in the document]
nguigs: And should the public class inherit from the private ones?

geomstats/geometry/special_orthogonal.py
nguigs: yes but the question is, is this function useful when base_point not equal to None? Should that function (with possible base_point) be in LieAlgebra? This corresponds to the question, does the LieAlgebra represent TeG or the set of Left-inv vect fields? Although isomorphic, they are not represented the same way, and having base_point argument implies the LieAlgebra is not only TeG but rather L(G)

geomstats/geometry/invariant_metric.py
nguigs: Yes, but I cannot import the class SpecialOrthogonal in invariant_metric.py because special_orthogonal.py also imports invariant_metric and so it loops and doesn't work!

geomstats/geometry/lie_group.py
nguigs: I did fix a test for lie bracket in so3 (with cross product), so it overrides this code

geomstats/geometry/skew_symmetric_matrices.py
nguigs: it would be helpful indeed

geomstats/geometry/general_linear.py
nguigs: no this is for initializing all parent classes with one call of super

geomstats/geometry/invariant_metric.py
nguigs: not deepsource but multiple inheritance

nguigs:
> Great! Let's address the more pressing aspects and keep the next refactorization steps for another PR.
> 
> Could you look at the remaining DeepSource issues? In particular, this seems to be a problem:
> `result = self.group._is_in_lie_algebra(vec_2)`
> 
> https://docs.quantifiedcode.com/python-anti-patterns/correctness/accessing_a_protected_member_from_outside_the_class.html

Yes let's try to merge tonight

PR #584, opened 2020-04-19
Update documentation website

ninamiolane:
This PR aims to make the documentation website more user-friendly:
- Add description of geomstats to the home page,
- Add an example.rst page that points to geomstats/example folder,
- Add details to the page describing the API reference.


docs/api-reference.rst
```diff
@@ -2,6 +2,21 @@
 API Reference
 *************
 
+The API reference gives an overview of Geomstats implementation.
+
+The module `geometry` implements concepts in differential geometry,
+such as manifolds and Riemannian metrics, with associated exponential
+and logarithmic maps, geodesics, and parallel transport.
```
nguigs, line 9: I think this can be misleading, we don't implement manifolds themselves (as triangular meshes or I don't know how) but tools to compute with manifold-valued data.

docs/api-reference.rst
```diff
@@ -2,6 +2,21 @@
...
+
+The module `learning` implements statistics and learning algorithms for data
+on manifolds, such as estimation, clustering and dimension reduction.
+The code is object-oriented and classes inherit from
+scikit-learn's base classes and mixins.
+
+In both modules, the operations are vectorized for batch computation and provide
+support for different execution backends---namely NumPy, PyTorch, and TensorFlow---
+enabling GPU acceleration. The module `backend` implements the operations needed to use Geomstats seemlessly with any backend.
```
nguigs, line 18: Is GPU acceleration tested anywhere?

docs/examples.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Learn differential geometry
+===========================
+
+Assume that your data naturally belongs to the
+`hyperbolic plane H2 <https://en.wikipedia.org/wiki/Hyperbolic_geometry#Models_of_the_hyperbolic_plane>`_
+and you want to get intuition on the geometry of this space.
+The space H2 has a negative curvature. The geodesics - i.e.
+the curves of shortest length - on H2 are not straight lines.
+How do they look like? To answer this question, you can run
```
nguigs, line 25: What do they look like? (help from a native speaker @cshewmake2 !)

docs/first-steps.rst
```diff
@@ -19,14 +19,17 @@ You can choose your backend by setting the environment variable ``GEOMSTATS_BACK
...
 
-    export GEOMSTATS_BACKEND=numpy
+    export GEOMSTATS_BACKEND=pytorch
 
 **FIRST EXAMPLES**
 
-We illustrate here the use of `geomstats` to generalize learning
-algorithms to Riemannian manifolds.
+To use `geomstats` for learning
+algorithms to Riemannian manifolds, you need to follow three steps:
```
nguigs, line 27: to Riemannian--> on Riemannian

docs/first-steps.rst
```diff
@@ -19,14 +19,17 @@ You can choose your backend by setting the environment variable ``GEOMSTATS_BACK
...
+- instantiate the manifold of interest,
+- instantiate the learning algorithm of interest,
+- run the algorithm.
```
nguigs, line 30: maybe mention that the data should be in a numpy array/ tensor ?

docs/examples.rst
cshewmake2: Yeah!

"How do they look?"
or 
"What do they look like?"

Either would do :)

docs/api-reference.rst
ninamiolane: Not really, although if pytorch and tf backends work, then we should be able to use GPU and therefore use their GPU acceleration. I'll remove it here and in the JMLR paper although, to avoid being criticized on this.

docs/api-reference.rst
ninamiolane: Done.

docs/examples.rst
ninamiolane: Cool thanks!

docs/first-steps.rst
ninamiolane: done.

docs/first-steps.rst
ninamiolane: done.

PR #585, opened 2020-04-20
Adapt code to pass more unit tests in tf and pytorch

ninamiolane:
Pass tests in hypersphere, hyperbolic, stiefel, and grassmanian.

### DIFF
--- a/geomstats/_backend/tensorflow/linalg.py
+++ b/geomstats/_backend/tensorflow/linalg.py
@@ -13,9 +13,8 @@ sqrtm = tf.linalg.sqrtm
 diagonal = tf.linalg.diag_part
 
 
-def norm(x, **kwargs):
-    if x.dtype != tf.float32 or x.dtype != tf.float64:
-        x = tf.cast(x, tf.float32)
+def norm(x, dtype=tf.float32, **kwargs):
+    x = tf.cast(x, dtype)
     return tf.linalg.norm(x, **kwargs)
 
 
--- a/geomstats/_backend/tensorflow/linalg.py
+++ b/geomstats/_backend/tensorflow/linalg.py
@@ -9,9 +9,14 @@ eigh = tf.linalg.eigh
 eigvalsh = tf.linalg.eigvalsh
 expm = tf.linalg.expm
 inv = tf.linalg.inv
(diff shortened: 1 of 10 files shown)


geomstats/_backend/tensorflow/linalg.py
```diff
@@ -9,11 +9,16 @@
...
 expm = tf.linalg.expm
 inv = tf.linalg.inv
-norm = tf.linalg.norm
 sqrtm = tf.linalg.sqrtm
 diagonal = tf.linalg.diag_part
 
 
+def norm(x, **kwargs):
+    if x.dtype != tf.float32 or x.dtype != tf.float64:
+        x = tf.cast(x, tf.float32)
```
nguigs, line 18: maybe allow the user to explicitely pass dtype? If working in float64 elsewhere this will break

geomstats/geometry/hypersphere.py
```diff
@@ -452,12 +452,28 @@ def exp(self, tangent_vec, base_point):
...
         norm4 = norm2**2
         norm6 = norm2**3
-        coef_1[mask_0] = 1. - norm2 / 2. + norm4 / 24. - norm6 / 720.
-        coef_2[mask_0] = 1. - norm2 / 6. + norm4 / 120. - norm6 / 5040.
 
-        coef_1[mask_non0] = gs.cos(norm_tangent_vec[mask_non0])
-        coef_2[mask_non0] = gs.sin(norm_tangent_vec[mask_non0]) / \
-            norm_tangent_vec[mask_non0]
+        if gs.sum(mask_0) > 0:
+            coef_1 = gs.assignment(
```
nguigs, line 457: 🎉

geomstats/_backend/tensorflow/linalg.py
ninamiolane: Yes, but the tf.linalg.norm does not work with tf.int32:
```
a  = tf.constant([1, 2, 3])
tf.linalg.norm(a)
```
fails, so we should remember that putting tf.int32 as dtype will fail.

ninamiolane: Done.

PR #592, opened 2020-04-20
Enable tests on python 3.5, 3.6, 3.7 and 3.8 with travis

ninamiolane:


nguigs:
Are you sure now the time? We still have 10 active PR every day!

ninamiolane:
Ok @nguigs let's wait a bit more then.

Issue #555, opened 2020-04-10
Remove warnings with nose2


ninamiolane:
Running nose2 with np, pytorch and tf prints a lot of warnings, specifically for the examples with np.

We should fix the issue, or prevent these warnings to be printed when using nose2.

ninamiolane:
Duplicate of #280.

PR #598, opened 2020-04-21
Disable warnings when running tests

ninamiolane:
Running tests with nose2 often prints a lot of warnings, see issue #555.

This PR corrects it by:
- changing the code that was printing unnecessary warnings, for example a scikit-learn import
- correcting an error divide by zero in Poincare ball,
- disabling warnings in tests.


geomstats/geometry/poincare_ball.py
```diff
@@ -166,11 +166,15 @@ def log(self, point, base_point):
...
-        log = gs.einsum('...i,...j->...j', log, (add_base_point / norm_add))
+        log = gs.einsum(
+            '...i,...j->...j', log, add_base_point)
 
         mask_0 = gs.isclose(gs.squeeze(norm_add, axis=-1), 0.)
+        mask_non0 = ~mask_0
         if gs.any(mask_0):
             log[mask_0] = 0
+        if gs.any(mask_non0):
+            log[mask_non0] /= norm_add[mask_non0]
```
nguigs, line 177: consider using `np.divide(a, b, where= b != 0)`

ninamiolane: Good point! I've put a TODO on these lines anyway, I'll look at it in a next PR.

PR #593, opened 2020-04-20
Assignment with booleans + follow-up on random_uniform for GLn and hypersphere

pchauchat:
This PR allows `gs.assignment` to work with booleans (and indices given as tensors) in tf.

It is also a follow-up on #583

PR #608, opened 2020-04-22
Move commits of old PR #593 on assignment to a new one post-rebase

pchauchat:
This PR allows gs.assignment to work with booleans (and indices given as tensors) in tf.

It is a follow-up on #583 and #593 

gs.assignment was heavily refactored to take into account arrays and booleans. For numpy and pytorch, I tried to take as much advantage as possible of their slicing opportunities, to keep the cost low (although I don't guarantee anything).

I defined more general check functions, to see if an object is some kind of boolean (mostly a native one or an array of bools), and if it is iterable (list, tuple, array of dim > 0). Maybe it will be useful for other functions too.

Some lines are left untested, I will add the corresponding tests later.

### DIFF
--- a/geomstats/_backend/numpy/__init__.py
+++ b/geomstats/_backend/numpy/__init__.py
@@ -118,6 +118,24 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
+def _is_boolean(x):
+    if isinstance(x, bool):
+        return True
+    if isinstance(x, (tuple, list)):
+        return _is_boolean(x[0])
+    if isinstance(x, np.ndarray):
+        return x.dtype == bool
+    return False
+
+
+def _is_iterable(x):
+    if isinstance(x, (list, tuple)):
+        return True
+    if isinstance(x, np.ndarray):
+        return ndim(x) > 0
+    return False
+
+
 def assignment(x, values, indices, axis=0):
     """Assign values at given indices of an array.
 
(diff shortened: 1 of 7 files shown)


geomstats/geometry/hypersphere.py
```diff
@@ -300,12 +307,13 @@ def random_uniform(self, n_samples=1):
...
             norms = gs.linalg.norm(samples, axis=1)
-            indcs = gs.isclose(norms, 0.0)
+            indcs = gs.isclose(norms, 0.0, atol=tol)
             num_bad_samples = gs.sum(indcs)
             if num_bad_samples == 0:
                 break
-            samples[indcs, :] = gs.random.normal(
+            new_samples = gs.random.normal(
                 size=(num_bad_samples, self.dim + 1))
+            samples = self._replace_values(samples, new_samples, indcs)
```
nguigs, line 316: I came up with an idea:
`samples = gs.concatenate([samples[~indcs], new_samples])`
what do you think?

pchauchat: That's smart, since we don't care about the order !

However, to have that part of code tested, we either have to do "seed mining", or define it in an auxiliary function. Any idea of how it could be named ? (replace_values would not be completely appropriate anymore)

nguigs: True, maybe `_enforce_non_zeros`

pchauchat: It does not explictely deal with the zeros, simply the replacement. Maybe _replace_unordered_values ?
I don't know whether it could be interesting to have both (ordered and unordered alternatives).

geomstats/_backend/numpy/__init__.py
```diff
@@ -118,6 +118,24 @@ def get_mask_i_float(i, n):
     return mask_i_float
 
 
+def _is_boolean(x):
+    if isinstance(x, bool):
+        return True
+    if isinstance(x, (tuple, list)):
+        return _is_boolean(x[0])
```
ninamiolane, line 125: Why looking at only the first element of the tuple/list? It is assumed that all elements have the same type?

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -341,11 +365,19 @@ def _assignment_single_value(x, value, indices, axis=0):
         Copy of x where value was assigned at all indices (and possibly
         along an axis).
     """
+    if _is_boolean(indices):
+        indices = [index for index, val in enumerate(indices) if val]
```
ninamiolane, line 369: This line comes up often, and could be factorize into a `_from_bools_to_indices` aux function.

geomstats/_backend/numpy/__init__.py
ninamiolane: Just saw your answer in the former PR:

I made the assumption that either all elements are booleans or no one is. I don't think it makes sense to make the code heavier and longer to run to be able to account for cases where booleans would be mixed with non-boolean values. I am not even sure Numpy handles them.

pchauchat: I made the assumption that either all elements are booleans or no one is. I don't think it makes sense to make the code heavier and longer to run to be able to account for cases where booleans would be mixed with non-boolean values. I am not even sure Numpy handles them.

geomstats/_backend/tensorflow/__init__.py
pchauchat: True ! Do you think this should stay here, or could be useful elsewhere and go to the backend ? (In which case, a subsequent PR could be appropriate.)

PR #610, opened 2020-04-22
Add the Kernel Density Estimation Classifier.

YannCabanes:
Add the Kernel Density Estimation Classifier and a file containing usual radial kernel functions.


YannCabanes:
When this PR will be merged, I will create another PR with an example of the Kernel Density Estimation Classifier on the sphere S2.

PR #612, opened 2020-04-22
Debug assignment for booleans in tf and use it for Poincare ball

ninamiolane:
[WIP]

### DIFF
--- a/geomstats/_backend/tensorflow/__init__.py
+++ b/geomstats/_backend/tensorflow/__init__.py
@@ -192,8 +192,8 @@ def _mask_from_indices(indices, mask_shape, dtype=float32):
         if len(index) != len(mask_shape):
             raise ValueError('Indices must have the same size as shape')
 
-    for i_index, index in enumerate(indices):
-        np_mask[i_index] = 1
+    for index in indices:
+        np_mask[index] = 1
     tf_mask = array(np_mask, dtype=dtype)
     return tf_mask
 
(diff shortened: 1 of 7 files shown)


tests/test_hypersphere.py
```diff
@@ -420,7 +420,9 @@ def test_exp_and_dist_and_projection_to_tangent_space(self):
         expected = gs.linalg.norm(tangent_vec) % (2 * gs.pi)
         self.assertAllClose(result, expected)
 
+    @geomstats.tests.np_and_pytorch_only
     def test_exp_and_dist_and_projection_to_tangent_space_vec(self):
+        # TODO(ninamiolane): Fix hypersphere's exp with assignment
```
ninamiolane, line 425: @pchauchat

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -322,10 +328,19 @@ def assignment_by_sum(x, values, indices, axis=0):
     If a list is given, it must have the same length as indices.
     """
     if _is_boolean(indices):
-        indices = [index for index, val in enumerate(indices) if val]
-
+        if ndim(array(indices)) > 1:
+            indices = tf.where(indices)
+        else:
+            if ndim(x) > 1:
```
pchauchat, line 334: Found the problem which makes hypersphere's exp fail: this if condition is useless. (And even if it was, there lacks an else counterpart). 
It can be removed in `assignment` and `assignment_by_sum`

tests/test_hypersphere.py
pchauchat: Found the problem, see comment https://github.com/geomstats/geomstats/pull/612/files#r413596996

ninamiolane: Yess, it works now! @pchauchat you're the GOAT!

PR #614, opened 2020-04-23
Move FrechetMean transform to a preprocessing module

nguigs:
This addresses comments made in the #577 to address #570 in a specific module rather than in FrechetMean, thus leaving the possibility to use any base point and not only the FrechetMean as base_point to lift data points to a tangent space.


nguigs:
aha nice list, I'd say `ToTangentSpace` or `LogTransformer`

ninamiolane:
@nguigs `ToTangentSpace` then? It also comes closer to the naming discussed in #412.

PR #616, opened 2020-04-23
fix bug in example learning graph

tgeral68:
-Bug was caused by the difference of dimensions returned by squared dist in riemannian_metric in newer version of geomstats


ninamiolane:
@tgeral68 for example, using einsums?

PR #619, opened 2020-04-23
Start to harmonize methods names across manifolds

ninamiolane:
This PR changes:
- symmetric_matrix_from_vector -> from_vector
- symmetric_matrix_to_vector -> to_vector
- projection_to_tangent_space -> to_tangent
- make_skew_symmetric -> to_skew_symmetric
- make_symmetric -> to_symmetric
- _SpecialEuclideanMatrix -> _SpecialEuclideanMatrices
- _SpecialOrthogonalMatrix -> _SpecialOrthogonalMatrices
- _SpecialEuclidean3Vector -> _SpecialEuclidean3Vectors
- _SpecialOrthogonal3Vector -> _SpecialOrthogonal3Vectors

and:
- adds to_tangent, is_tangent as abstract methods of Manifold,
- adds some docstrings.

@opeltre @stefanheyder @nguigs , following issue #412

### DIFF
--- a/geomstats/geometry/embedded_manifold.py
+++ b/geomstats/geometry/embedded_manifold.py
@@ -54,34 +54,3 @@ class EmbeddedManifold(Manifold):
         """
         raise NotImplementedError(
             'projection is not implemented.')
-
-    def is_tangent(self, vector, base_point=None):
-        """Check whether the vector is tangent at base_point.
-
-        Parameters
-        ----------
-        vector : array-like, shape=[n_samples, dim_embedding]
-            Vector at the tangent space of the embedding manifold.
-        base_point : array-like, shape=[n_samples, dim_embedding]
-            Point on the embedded manifold, in extrinsic coordinates.
-
-        Returns
-        -------
-        is_tangent : bool
-            Boolean denoting if vector is a tangent vector at the base point.
-        """
-        raise NotImplementedError(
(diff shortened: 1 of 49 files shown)


geomstats/geometry/embedded_manifold.py
```diff
@@ -55,7 +55,25 @@ def projection(self, point):
         raise NotImplementedError(
             'projection is not implemented.')
 
-    def projection_to_tangent_space(self, vector, base_point):
+    def is_tangent(self, vector, base_point=None):
```
nguigs, line 58: is it necessary to add it here, as it is inherited from the parent class Manifold?

ninamiolane: Good point, I removed it.

ninamiolane:
Good question. I think it relates to the coordinate system used to represent the points:
- we should use `EmbeddedManifold` when we work with embedding coordinates (eg rotation **matrices**...), and operates on them to get a point/tangent_vector for the embedded manifold. In this case, to_tangent is a projection from T_embedding to T_embedded.
- we should use `Manifold` when we use intrinsic coordinates, with same dim than the manifold. In this case, to_tangent is not really a projection, but a function to R^dim to T_manifold. The Manifold is not really embedded in R^dim though.

PR #621, opened 2020-04-24
Improve vectorization of Beta class

nguigs:
- change belong method in SPD to check the size of the input matrix: I think belong methods should not be static method, as the dimension must be checked.
- use the polygamma implemented by each backend 
- import vectorization of some methods thanks to the vectorization of the polygamma function
- remove some deepsource errors

Issues:
- we could use the beta from np.random and torch.random but tensorflow doesn't expose it, only tensorflow_probability does. In pytorch it is vectorized and differentiable so it could improve the actual code.
- the exp and log method rely on scipy and are not expected to work with pytorch/numpy, should we raise a not implemented error if they are called with these backends? This could apply elsewhere in the code base


geomstats/_backend/tensorflow/__init__.py
```diff
@@ -70,6 +70,7 @@
 matmul = tf.linalg.matmul
 mod = tf.math.mod
 power = tf.math.pow
+polygamma = tf.math.polygamma
```
ninamiolane, line 73: NIT polygamma < power

geomstats/_backend/pytorch/__init__.py
```diff
@@ -43,6 +43,7 @@
     ones,
     ones_like,
     pow as power,
+    polygamma,
```
ninamiolane, line 46: NIT: polygamma < pow

geomstats/geometry/beta_distributions.py
```diff
@@ -41,16 +42,14 @@ def belongs(self, point):
...
         """
-        point = gs.to_ndarray(point, to_ndim=2)
-        n_points, point_dim = point.shape
+        point_dim = point.shape[-1]
         belongs = point_dim == self.dim
-        belongs = gs.to_ndarray(belongs, to_ndim=1)
-        belongs = gs.tile(belongs, n_points)
-        belongs = belongs * gs.greater(point, 0).all(axis=1)
-        return belongs[0] if n_points == 1 else belongs
+        belongs = gs.logical_and(
```
ninamiolane, line 47: Nice!

geomstats/geometry/beta_distributions.py
```diff
@@ -65,6 +64,8 @@ def random_uniform(n_samples=1, bound=10.0):
         -------
         samples : array-like, shape=[n_samples, 2]
         """
+        if n_samples == 1:
```
ninamiolane, line 67: Or in two lines:

```
size = (2,) if n_samples == 1 else (n_samples, 2)
return bound * gs.random.rand(*size)
```

geomstats/geometry/beta_distributions.py
```diff
@@ -115,8 +117,9 @@ def maximum_likelihood_fit(data, loc=0, scale=1):
         -------
         parameter : array-like, shape=[n_samples, 2]
         """
+        data = gs.cast(data, gs.float32)
         data = gs.to_ndarray(
-            gs.where(data == 1., 1 - EPSILON, data), to_ndim=2)
+            gs.where(data == 1., gs.array(1. - EPSILON), data), to_ndim=2)
```
ninamiolane, line 122: Any chance we could factorize the `gs.array` within the `gs.where` backend function? I assume the `np.where` function does not require a `np.array` as its second argument, so we could mimic this flexibility?

geomstats/geometry/beta_distributions.py
```diff
@@ -159,20 +162,18 @@ def inner_product_matrix(self, base_point=None):
         -------
         base_point : array-like, shape=[n_samples, 2, 2]
         """
-        if base_point is not None:
+        if base_point is None:
             raise ValueError('The metric depends on the base point.')
```
ninamiolane, line 166: We should start unit testing the errors: it would be an "easy" code coverage improvement, e.g.:
https://github.com/geomstats/geomstats/blob/1e33eb3841cda03348a9640e247f86a244edaf75/tests/test_manifold.py#L20

I can help on this today, let's just make sure we don't address the same ones.

geomstats/geometry/spd_matrices.py
```diff
@@ -23,10 +23,9 @@ def __init__(self, n):
             dim=int(n * (n + 1) / 2),
             embedding_manifold=GeneralLinear(n=n))
 
-    @staticmethod
-    def belongs(mat, atol=TOLERANCE):
+    def belongs(self, mat, atol=TOLERANCE):
         """Check if a matrix is symmetric and invertible."""
-        is_symmetric = GeneralLinear.is_symmetric(mat)
+        is_symmetric = super(SPDMatrices, self).belongs(mat, atol)
```
ninamiolane, line 28: Why this change?

tests/test_beta_distributions.py
```diff
@@ -16,7 +16,6 @@ def setUp(self):
         self.n_samples = 10
         self.dim = self.beta.dim
 
-    @geomstats.tests.np_and_pytorch_only
     def test_random_uniform_and_belongs(self):
```
ninamiolane, line 19: Nice!

geomstats/geometry/spd_matrices.py
nguigs: to check the size of the matrix corresponds to the `n` the object was instantiated with

geomstats/geometry/beta_distributions.py
nguigs: that's probably a good thing to do!

geomstats/geometry/beta_distributions.py
nguigs: nice thanks

geomstats/_backend/pytorch/__init__.py
nguigs: :confused:

geomstats/geometry/beta_distributions.py
nguigs: Sure, I'll do some of these in the coming days. I'll start by the end (alphabetical order)

nguigs: btw using a lambda function as in the example you pointed out raises a DeepSource antipattern issue... Should we keep it or define a function to do it?

geomstats/geometry/beta_distributions.py
nguigs: I'll open an issue for this

Issue #624, opened 2020-04-25
harmonize pytorch's behavior of where method


nguigs:
Pytorch doesn't take float as input in its second argument, whereas numpy does. This would factor out the following change:
```python
gs.where(data == 1., 1 - EPSILON, data), to_ndim=2)
gs.where(data == 1., gs.array(1. - EPSILON), data), to_ndim=2)
```

Issue #127, opened 2018-08-17
In pytorch backend a lot of functions are actually wrappers to numpy


oleg-kachan:
Currently a lot of functions in pytorch backend and all, expept one in pytorch.linalg are just wrappers to numpy, which is suboptimal. It seems that it is because geomstats development started before pytorch 0.4, the release prior to that advanced linear algebra functions were virtually non-existent in pytorch.

Solution: update backend by replacing wrappers to numpy with native pytorch functions, if and when they are become available.

nguigs:
This is the least of functions actually wrapped:

- linalg.sqrtm
- linalg.logm
- linalg.solve_sylvester
- argmax for bools
- diag_indices

None of these exist in Pytorch 1.10

PR #627, opened 2020-04-26
Use native torch functions in backend 

ninamiolane:
Follow-up from #625.

Many functions in pytorch backend are wrappers from numpy functions, see issue #127 .

This issue starts to address it, by looking at the file `pytorch/__init__.py` and:
- gs.dot,
- gs.any,
- gs.all,
- gs.trace,
- gs.isclose,
- gs.where (also addressing issue #624  ).

The same type of PR remains to be done for the other files in `/_backend/pytorch/` folder.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -127,19 +127,31 @@ def any(x, axis=None):
     if not torch.is_tensor(x):
         x = torch.tensor(x)
     if axis is None:
-        return x.bool().any()
+        return torch.any(x)
     if isinstance(axis, int):
         return torch.any(x.bool(), axis)
-    if len(axis) == 2:
+    if len(axis) == 1:
+        return torch.any(x, *axis)
         axis = list(axis)
         for i_axis, one_axis in enumerate(axis):
             if one_axis < 0:
                 axis[i_axis] = ndim(x) + one_axis
-        return torch.any(
-            torch.any(x.bool(), axis[1]), axis[0])
-    raise NotImplementedError(
-        'any not implemented for more than two axes.')
-
+    new_axis = tuple(k - 1 if k >= 0 else k for k in axis[1:])
+    return any(torch.any(x.bool(), axis[0]), new_axis)
+#     if axis is None:
+#         return x.bool().any()
+#     if isinstance(axis, int):
+#         return torch.any(x.bool(), axis)
+#     if len(axis) == 2:
+#         axis = list(axis)
+#         for i_axis, one_axis in enumerate(axis):
+#             if one_axis < 0:
+#                 axis[i_axis] = ndim(x) + one_axis
+#         return torch.any(
+#             torch.any(x.bool(), axis[1]), axis[0])
+#     raise NotImplementedError(
+#         'any not implemented for more than two axes.')
+#
 
 def cast(x, dtype):
     if torch.is_tensor(x):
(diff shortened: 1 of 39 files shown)


geomstats/_backend/pytorch/__init__.py
```diff
@@ -125,10 +124,21 @@ def logical_and(x, y):
...
+        return torch.any(x.bool(), axis)
+    if len(axis) == 2:
+        axis = list(axis)
+        for i_axis, one_axis in enumerate(axis):
+            if one_axis < 0:
+                axis[i_axis] = ndim(x) + one_axis
+        return torch.any(
+            torch.any(x.bool(), axis[1]), axis[0])
+    raise NotImplementedError(
+        'any not implemented for more than two axes.')
```
nguigs, line 141: what about recursively calling any, like:
```python

def any(x, axis=None):
    if axis is None:
        return torch.any(x)
    if isinstance(axis, int):
        return torch.any(x, axis)
    if len(axis) == 1:
        return torch.any(x, *axis)
    else:
        new_axis = tuple(k - 1 if k >= 0 else k for k in axis[1:])
        print(new_axis)
        return any(torch.any(x, axis[0]), new_axis)
```
Just something to think about, the above doesn't work if the axes are not ordered...

geomstats/_backend/pytorch/__init__.py
```diff
@@ -331,20 +350,25 @@ def einsum(*args, **kwargs):
...
     input_tensors_list = args[1:]
 
+    if len(input_tensors_list) == 1:
+        return torch.einsum(einsum_str, input_tensors_list)
+
     einsum_list = einsum_str.split('->')
     input_str = einsum_list[0]
-    output_str = einsum_list[1]
+    if len(einsum_list) > 1:
+        output_str = einsum_list[1]
```
nguigs, line 359: great, but won't line 411 crash if `output_str` is not defined?

geomstats/_backend/pytorch/__init__.py
```diff
@@ -501,6 +535,14 @@ def mean(x, axis=None):
     return torch.mean(x, dim=axis)
 
 
+def where(condition, x, y):
+    if not torch.is_tensor(x):
+        x = torch.tensor(x)
+    if not torch.is_tensor(y):
+        y = torch.tensor(y)
+    return torch.where(condition, x, y)
```
nguigs, line 543: I ran into errors trying to do it like this the other day:
- if x and y are None
-

geomstats/_backend/pytorch/__init__.py
ninamiolane: Thanks for spotting this! I have added the recursive logic.

ninamiolane: Also added it for `all`.

geomstats/_backend/pytorch/__init__.py
ninamiolane: True. I have added the condition there as well.

geomstats/_backend/pytorch/__init__.py
ninamiolane: Good points!

- I added the case where x and y are None
- This is because torch.where only work if x and y are of same `dtype`. Same with a lot of torch functions actually. The function now casts y with the`dtype` of x.

I have added your tests and they pass. Thanks!

PR #629, opened 2020-04-26
Change docstring convention for shape=

ninamiolane:
This PR suggests changing the way we indicate shape of array-like inputs in docstrings.

Instead of:
```
tangent_vec : array-like, shape=[n_samples, dim] or shape=[dim,]
```
we use:
```
tangent_vec : array-like, shape=[..., dim]
```

In addition, this allows avoiding ambiguities in the BetaDistribution manifold, where `n_points` and `n_samples` notations were conflicting.

However, using `[...,` instead of `[n_samples,` could be slightly less explicit for users looking at the documentation. What do you think? @nguigs @stefanheyder @cshewmake2

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -457,14 +457,14 @@ Here's a generic docstring template::
 
       Parameters
       ----------
-      my_param_1 : array-like, shape=[n_samples, dimension]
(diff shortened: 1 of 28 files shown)


ninamiolane:
Cool thanks! I'll change the remaining files and merge.

PR #630, opened 2020-04-26
Clean geomstats.errors

ninamiolane:
Clean geomstats.errors module:
- geomstats.error -> geomstats.errors to be consistent with geomstats.tests (plural)
- clean check_belongs method, no need for manifold's name as parameter
- add unit tests

### DIFF
--- a/geomstats/errors.py
+++ b/geomstats/errors.py
@@ -5,6 +5,9 @@ import math
 import geomstats.backend as gs
 
 
+TOL = 1e-6
+
+
 def check_integer(n, n_name):
     """Raise an error if n is not a > 0 integer.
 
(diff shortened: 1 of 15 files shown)


geomstats/errors.py
```diff
@@ -23,7 +23,7 @@ def check_integer(n, n_name):
                 ' got {}.'.format(n_name, n))
 
 
-def check_belongs(point, manifold, manifold_name):
+def check_belongs(point, manifold):
```
nguigs, line 26: should we allow to pass a `tol` kwarg in this method, as belong is sometimes subject to numerical precision?

ninamiolane: True. And this is where the DeepSource error "Arguments differ from overridden method" was helpful: in order to do this, we should harmonize the arguments of parents' and children's methods - even if this means that some args will be unused.

What do you think?

nguigs: In the long term yes, but this function would work anyway, if tol is only specified in classes where it needs to be, and in this method you just get **kwargs in the args and feed **kwargs to belong?

ninamiolane: Good point. I've modified it and added a test with atol.

PR #631, opened 2020-04-26
Hot fix in assignment_by_sum tests: remove randomness

ninamiolane:
Some values of theta and phi make this test fails (? @pchauchat). 

This PR removes the randomness in sampling theta and phi to fix master.

[Edit]: Removing the randomness is not enough. This test keeps failing 15% of the time (?). But inverting the computations `gs` and `_np` fixes it.


pchauchat:
I had had that problem, but I thought it was fine by now (That said, if it's 15% of the time, I might have not tested enough). I'll check on that. I hope it's not related to some obscure mutable/immutable thing somewhere ...

Issue #83, opened 2018-05-08
Add canonical metric for each space and allow: space.exp and space.log


ninamiolane:

nguigs:
Closed by #632 that allows `space.metric.exp`

PR #632, opened 2020-04-26
Add logic to allow Manifolds having the attribute metric

ninamiolane:
There were discussions about whether:
- a Manifold should have an attribute `metric`,
- a Connection/RiemannianMetric should have an attribute `space`,
since having both results in `maximum depth exceeded` error, since both objects would call themselves.

This PR suggests to address this with the following logic (only on Hypersphere for now):
- add a private class `_Hypersphere`, with *no* `self.metric`,
- add a `self._space` private attribute in the class `HypersphereMetric`. 
    - For example, this private attribute will allow to:
        -  use `geomstats.errors.check_belongs(point, self._space)` in order to consistently check that input parameters of `exp`, `log`, etc., do belong to the manifold on which the metric is used, 
        - use space's methods that convert coordinate types,
        - etc.
   - This private attribute is not intended to be accessed by the user.
- add a wrapper class `Hypersphere` that inherits from `_Hypersphere` and adds a `self.metric` attribute.

If you agree with this logic, we could apply it to other manifolds, and in particular address issue #83 . 

Additionally, we could define `exp` and `log` methods in a `Manifold` class, which would call the `exp` and `log` of the default metric in `self.metric`.

@nguigs @stefanheyder @opeltre

### DIFF
--- a/geomstats/geometry/hypersphere.py
+++ b/geomstats/geometry/hypersphere.py
@@ -607,7 +607,8 @@ class HypersphereMetric(RiemannianMetric):
         """
         return self.dist(point_a, point_b) ** 2
 
-    def parallel_transport(self, tangent_vec_a, tangent_vec_b, base_point):
+    @staticmethod
+    def parallel_transport(tangent_vec_a, tangent_vec_b, base_point):
         """Compute the parallel transport of a tangent vector.
 
         Closed-form solution for the parallel transport of a tangent vector a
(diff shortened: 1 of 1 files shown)


nguigs:
Would the docstrings be accessible from the public classes?

ninamiolane:
Yes, they are accessible. I also found that we can inherit docstrings, in the sense that if the docstring of the child class is not provided, then the docstring from the parent class is applied.

I wonder if we could use this to factorize our docstring writing.

inspect.getdoc(object)
Get the documentation string for an object, cleaned up with cleandoc(). If the documentation string for an object is not provided and the object is a class, a method, a property or a descriptor, retrieve the documentation string from the inheritance hierarchy.

from:https://docs.python.org/3/library/inspect.html

PR #635, opened 2020-04-27
Refactor assignment and assignment_by_sum

pchauchat:
This PR defines tensorflow's `assignment` and `assignment_by_sum` as two variants of the same internal `_assignment` function, avoiding significant redundancy.

### DIFF
--- a/geomstats/_backend/tensorflow/__init__.py
+++ b/geomstats/_backend/tensorflow/__init__.py
@@ -301,8 +301,7 @@ def _assignment_single_value(x, value, indices, mode='replace', axis=0):
         return x + -x * mask + value * mask
     if mode == 'sum':
         return x + value * mask
-    else:
-        raise ValueError('mode must be one of \'replace\' or \'sum\'')
+    raise ValueError('mode must be one of \'replace\' or \'sum\'')
 
 
 def _assignment(x, values, indices, mode, axis):
--- a/geomstats/_backend/tensorflow/__init__.py
+++ b/geomstats/_backend/tensorflow/__init__.py
@@ -250,141 +250,24 @@ def _vectorized_mask_from_indices(
     return _duplicate_array(mask, n_samples, axis=axis)
 
 
-def _assignment_single_value_by_sum(x, value, indices, axis=0):
-    """Add a value at given indices of an array.
-
-    Parameters
-    ----------
-    x: array-like, shape=[dim]
-        Initial array.
-    value: float
-        Value to be added.
-    indices: {int, tuple, list(int), list(tuple)}
-        Single int or tuple, or list of ints or tuples of indices where value
-        is assigned.
-        If the length of the tuples is shorter than ndim(x), value is
-        assigned to each copy along axis.
-    axis: int, optional
-        Axis along which value is assigned, if vectorized.
-
-    Returns
-    -------
-    x_new : array-like, shape=[dim]
-        Copy of x where value was added at all indices (and possibly along
-        an axis).
-    """
-    single_index = not isinstance(indices, list)
-    if tf.is_tensor(indices):
-        single_index = ndim(indices) <= 1 and sum(indices.shape) <= ndim(x)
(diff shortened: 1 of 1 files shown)


geomstats/_backend/tensorflow/__init__.py
```diff
@@ -250,28 +250,31 @@ def _vectorized_mask_from_indices(
...
 
-def _assignment_single_value_by_sum(x, value, indices, axis=0):
-    """Add a value at given indices of an array.
+def _assignment_single_value(x, value, indices, mode='replace', axis=0):
+    """Assign a value at given indices of an array.
 
     Parameters
     ----------
-    x: array-like, shape=[dim]
+    x : array-like, shape=[dim]
```
ninamiolane, line 258: 🙏

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -294,50 +297,15 @@ def _assignment_single_value_by_sum(x, value, indices, axis=0):
...
-    indices: {
-    int, tuple(int), array-like({int, tuple, boolean})
-        Single index or array of indices where values are assigned.
-        If the length of the tuples is shorter than ndim(x) by one, values are
-        assigned to each copy along axis.
-        If indices is a list of booleans and ndim(x) > 1, values are assigned
-        across all dimensions.
-    axis: int, optional
-        Axis along which values are assigned, if vectorized.
+    if mode == 'replace':
```
ninamiolane, line 300: Nice!

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -294,50 +297,15 @@ def _assignment_single_value_by_sum(x, value, indices, axis=0):
...
+        return x + -x * mask + value * mask
+    if mode == 'sum':
+        return x + value * mask
+    else:
```
ninamiolane, line 304: DeepSource error comes from here: unnecessary if/else after return

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -294,50 +297,15 @@ def _assignment_single_value_by_sum(x, value, indices, axis=0):
...
+        raise ValueError('mode must be one of \'replace\' or \'sum\'')
```
ninamiolane, line 305: This line is not tested.

pchauchat: True, but how do I test it ? I chose to add the `mode` parameter to switch between cases, keeping it as a string to allow creating new cases later. However, as it is an internal function, specific to a backend, I don't really see how to explicitely test this ValueError case.
Any idea ?

ninamiolane: Would `assertRaises( ... with gs._assignment_single_value(mode='not_a_mode') ...)` work?

pchauchat: Yes, but only in tf backend, as this internal function does not exist for numpy and pytorch

ninamiolane: Ah ok! `if geomstats.tests.tf_backend():` ... do the test.

pchauchat: The only thinkg (I didn't think of that until now), is that since it is a backend function, it would need to be added to the list of function that are imported.
Otherwise, we'd need to add a `from geomstats._backend.tensorflow import _assignment_single_value` ... Sounds a bit ugly.

ninamiolane: Calling `gs._assignment_single_value` only `if geomstats.tests.tf_backend()` wouldn't do it?

pchauchat: It won't find it if it's not explicitely put in the function exported by the backend, sadly

ninamiolane: Ok! Well, too bad I guess! Let's merge!

PR #645, opened 2020-04-30
Fix pymanopt example

ninamiolane:
[WIP]

This example has bugs, maybe following latest updates of pymanopt?


nkoep:
I've updated the example for the new pymanopt API.

PR #725, opened 2020-04-30
Vectorize parallel transport on the hypersphere

nguigs:
This addresses #701 so that parallel transport can be applied to:

1. n base_points, n tan_a, n tan_b
2. 1 bp, n tan_a, n tan_b
3. 1 bp, 1 tan_a, n tan_b
4. 1 bp, n tan_a, 1 tan_b
5. 1 bp, 1 tan_a, 1 tan_b

The five cases are tested and work with numpy and tensorflow. Only cases 1 and 5 work with pytorch: it seems einsum is not entirely consistent with that of numpy?


ninamiolane:
Nice! I'll check the einsum problem.

Issue #377, opened 2020-01-20
Refactor code on product manifolds/metrics, landmarks sets, discretized curves


ninamiolane:
Currently, the following classes are implemented:

- `ProductManifold`: 
`n_samples` of `n_points` on `n_manifolds` manifolds M_1 x...x M_n_manifolds, where M_1 x...x M_n_manifolds are potentially different manifolds.
- `Landmarks`: `n_samples` of `n_landmarks` on `n_manifolds` M x ... x M where the same manifold M is used, but the number of landmarks may vary.
- `DiscretizedCurves`: `n_samples` of curves on a manifold M, where a curve is a set of landmarks which are called `sampling_points` in this class. The number of landmarks/sampling_points may vary.

With associated `ProductMetric`, `L2Metric`, etc.

There seem to be methods of `Landmarks`/`DiscretizedCurves` that can be refactorized into `ProductManifold`, and methods of` L2Metric` that can be refactorized into `ProductMetric`.

Notes:
- We should agree on the point representation for these spaces, depending on the cases:
    - either [n_samples, dim1 + .. + dim2]
    - or [n_samples, n_landmarks, dim]?
- We could fix the number of landmarks n_landmarks in the class `Landmarks`, and n_sampling_points in the class `DiscretizedCurves`? Methods that do not depend on the number of landmarks / sampling_points could be @classmethod. Alternatively, we could use `ProductManifold` each time n_landmarks is fixed and `Landmarks` each time we need a flexible number of landmarks (for example to study asymptotic when the number of landmarks goes to infinity, without having to instantiate an infinite number of `Landmarks` objects).

ninamiolane:
@YannCabanes you could be interested in this discussion.

nguigs:
I think `n_landmarks` should be the same for all samples in `LandmarkSpace` (and in theory should correspond to the same feature in e.g. an image) but could vary in `DiscretizedCurves`. Thus [n_samples, n_landmarks, dim] is adapted to `Landmarks` but not to `DiscretizedCruves`.

nguigs:
Landmarks uses ProductManifold while it is not necessary that DiscreteCurve does.

PR #727, opened 2020-04-30
Refactorize Discretized Cruves and Landmarks

nguigs:
This addresses #377 and add the documentation. Also addressed #720

What about renaming discretized curves to discrete curves?

### DIFF
--- a/geomstats/geometry/discretized_curves.py
+++ b/geomstats/geometry/discretized_curves.py
@@ -14,17 +14,33 @@ R2 = Euclidean(dim=2)
 R3 = Euclidean(dim=3)
 
 
-class DiscretizedCurves(Manifold):
-    """Space of discretized curves sampled at points in ambient_manifold.
+class DiscreteCurves(Manifold):
+    r"""Space of discrete curves sampled at points in ambient_manifold.
+
+    Each individual curve is represented by a 2d-array of shape `[
+    n_sampling_points, ambient_dim]`. A Batch of curves can be passed to
+    all methods either as a 3d-array if all curves have the same number of
+    sampled points, or as a list of 2d-arrays, each representing a curve.
 
     Parameters
     ----------
-    ambiant_manifold : Manifold
+    ambient_manifold : Manifold
         Manifold in which curves take values.
+
+    Attributes
+    ----------
+    ambient_manifold : Manifold
+        Manifold in which curves take values.
+    l2_metric : callable
+        Function that takes as argument an integer number of sampled points
+        and returns the corresponding L2 metric (product) metric,
+        a RiemannianMetric object
+    square_root_velocity_metric : RiemannianMetric
+        Square root velocity metric.
     """
 
     def __init__(self, ambient_manifold):
-        super(DiscretizedCurves, self).__init__(dim=math.inf)
+        super(DiscreteCurves, self).__init__(dim=math.inf)
         self.ambient_manifold = ambient_manifold
         self.l2_metric = lambda n: L2Metric(
             self.ambient_manifold, n_landmarks=n)
@@ -33,21 +49,26 @@ class DiscretizedCurves(Manifold):
     def belongs(self, point):
         """Test whether a point belongs to the manifold.
 
-        Test that all points on the curve belong to the manifold
+        Test that all points of the curve belong to the ambient manifold.
 
         Parameters
         ----------
-        point : array-like, shape=[n_sampling_points, ambiant_dim]
-            Point representing a discretized curve.
+        point : array-like, shape=[..., n_sampling_points, ambient_dim]
+            Point representing a discrete curve.
 
         Returns
         -------
         belongs : bool
-            Boolean evaluating if point belongs to the space of discretized
+            Boolean evaluating if point belongs to the space of discrete
             curves.
         """
-        belongs = gs.all(self.ambient_manifold.belongs(point))
-        return belongs
+        def each_belongs(pt):
+            return gs.all(self.ambient_manifold.belongs(pt))
+
+        if isinstance(point, list):
+            return gs.vectorize(point, each_belongs)
+
+        return gs.vectorize(point, each_belongs, signature='(n,k)->()')
 
 
 class SRVMetric(RiemannianMetric):
@@ -57,9 +78,12 @@ class SRVMetric(RiemannianMetric):
 
     Parameters
     ----------
-    ambiant_manifold : Manifold
+    ambient_manifold : Manifold
         Manifold in which curves take values.
-
+    metric : RiemannianMetric
+        Metric to use on the ambient manifold. If None is passed, ambient
+        manifold should have a metric attribute, which will be used.
+        Optional, default : None.
 
     References
     ----------
@@ -69,10 +93,17 @@ class SRVMetric(RiemannianMetric):
     vol. 33, no. 7, pp. 1415-1428, July 2011.
     """
 
-    def __init__(self, ambient_manifold):
+    def __init__(self, ambient_manifold, metric=None):
         super(SRVMetric, self).__init__(dim=math.inf,
                                         signature=(math.inf, 0, 0))
-        self.ambient_metric = ambient_manifold.metric
+        if metric is None:
+            if hasattr(ambient_manifold, 'metric'):
+                self.ambient_metric = ambient_manifold.metric
+            else:
+                raise ValueError('Either a metric must be passed or an '
+                                 'ambient manifold equipped with a metric.')
+        else:
+            self.ambient_metric = metric
         self.l2_metric = lambda n: L2Metric(ambient_manifold, n_landmarks=n)
 
     def pointwise_inner_product(self, tangent_vec_a, tangent_vec_b,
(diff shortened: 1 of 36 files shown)


geomstats/geometry/discretized_curves.py
```diff
@@ -15,25 +15,36 @@
...
 
     def belongs(self, point):
         """Test whether a point belongs to the manifold.
 
+        Test that all points on the curve belong to the manifold
+
         Parameters
         ----------
-        point : array-like
+        point : array-like, shape=[n_sampling_points, ambiant_dim]
```
ninamiolane, line 40: Typo ambient_dim

geomstats/geometry/landmarks.py
```diff
@@ -1,229 +1,102 @@
...
+    """Class for space of landmarks.
+
+    The landmark space is a product manifold where all manifolds in the
+    product are the same. The default metric is the product metric and
+    is often referred to as the L2 metric.
+
+    Parameters
+    ----------
+    ambient_manifold : Manifold
+        Manifold to which landmarks belong
```
ninamiolane, line 19: Nit: missing end "."

geomstats/geometry/landmarks.py
```diff
@@ -1,229 +1,102 @@
...
+    n_landmarks: int
+            Number of landmarks.
```
ninamiolane, line 21: NIT: space before : and indentation too large.

tests/test_poincare_polydisk.py
```diff
@@ -46,7 +46,7 @@ def test_product_distance_extrinsic_representation(self):
 
         single_disk = PoincarePolydisk(n_disks=1, coords_type=coords_type)
         two_disks = PoincarePolydisk(n_disks=2, coords_type=coords_type)
-
+        print(point_b.shape)
```
ninamiolane, line 49: remaining print

geomstats/geometry/discretized_curves.py
```diff
@@ -15,25 +15,36 @@
...
-    """Space of discretized curves sampled at points in ambient_manifold."""
+    """Space of discretized curves sampled at points in ambient_manifold.
+
+    Parameters
+    ----------
+    ambiant_manifold : Manifold
+        Manifold in which curves take values.
+    """
 
     def __init__(self, ambient_manifold):
```
ninamiolane, line 26: Why not having n_landmarks as a parameter of the constructor, like for the class `Landmarks`?
Or having `lambda n : L2Metric(manifold, n)` for both?

tests/test_discretized_curves.py
```diff
@@ -256,7 +257,9 @@ def test_srv_metric_dist_and_geod(self):
         srv = self.srv_metric_r3.square_root_velocity(geod)
 
         srv_derivative = self.n_discretized_curves * (srv[1:, :] - srv[:-1, :])
-        norms = self.l2_metric_r3.norm(srv_derivative, geod[:-1, :-1, :])
+        l2_metric = self.space_curves_in_euclidean_3d.l2_metric(
+            self.n_sampling_points - 1)
```
ninamiolane, line 261: Why do we need `self.n_sampling_points -1` here, and in setup we have:
```
self.l2_metric_r3 = self.space_curves_in_euclidean_3d.l2_metric(
            self.n_sampling_points)
```

geomstats/geometry/discretized_curves.py
```diff
@@ -207,23 +229,22 @@ def exp(self, tangent_vec, base_curve):
...
         coef_1 = 1 / gs.sqrt(base_curve_velocity_norm)
         coef_2 = -1 / (2 * base_curve_velocity_norm**(5 / 2)) * inner_prod
 
         term_1 = gs.einsum('ij,ijk->ijk', coef_1, tangent_vec_derivative)
         term_2 = gs.einsum('ij,ijk->ijk', coef_2, base_curve_velocity)
         srv_initial_derivative = term_1 + term_2
 
-        end_curve_srv = self.l2_metric.exp(tangent_vec=srv_initial_derivative,
-                                           base_landmarks=base_curve_srv)
+        end_curve_srv = self.l2_metric(n_sampling_points - 1).exp(
```
ninamiolane, line 246: Why `n_sampling_points - 1`?

geomstats/geometry/discretized_curves.py
```diff
@@ -44,73 +55,83 @@ class SRVMetric(RiemannianMetric):
...
+        base_curve : array-like, shape=[..., n_sampling_points, ambiant_dim]
+            Point representing a discretized curve.
 
         Returns
         -------
-        norm :
+        norm : array-like, shape=[..., n_sampling_points]
+            Point-wise norms.
         """
         # TODO: Revise this to refer to action on single elements
```
ninamiolane, line 131: Do you understand this TODO? x)

geomstats/geometry/discretized_curves.py
ninamiolane: (also at other places in the PR)

geomstats/geometry/landmarks.py
ninamiolane: (also at other places in the PR)

geomstats/geometry/landmarks.py
ninamiolane: (also at other places in the PR)

nguigs:
> Great! Do we want to have different logics for Landmarks and DiscretizedCurves? Here, one takes n_landmarks as a parameter, and the other doesn't. Is this because, in applications, we may want to change the sampling along the curve, while the number of landmarks rarely changes?
> 
> Yes to change to DiscreteCurves if you think this makes more sense.
> 
> Why do we have sometimes `n_sampling_points` and then `n_sampling_points - 1` in the code?

Have a look at the discussion in #377, the whole point of having both Landmarks and DiscreteCurves classes is that they don't behave the same: in the landmark geometry literature the number of landmarks is usually fixed and landmarks are supposed to represent features, whereas in the discrete curve the number of sampling points is not, and results from a random process. That being said, I am not sure the current implementation of discrete curve works for different number of sampling points, but it is what we should be aiming for.

About `n_sampling_points` and `n_sampling_points - 1`, it results from the method `square_root_velocity` in the SRV metric and a different coefficient is applied to the initial point in e.g. distance, but I am not familiar with the theory, I just put the parameter so that the tests pass! Maybe @AliceLeBrigant can tell us more about it!

https://github.com/geomstats/geomstats/blob/c70315cca98ff604b2137a9cc676c1555d1910bd/geomstats/geometry/discretized_curves.py#L368-L373

geomstats/geometry/discretized_curves.py
nguigs: No...

geomstats/geometry/discretized_curves.py
nguigs: it was already the case
https://github.com/geomstats/geomstats/blob/932fe5a05da9d0f1622b51c152d3ab34d3e6ae3a/geomstats/geometry/discretized_curves.py#L230

tests/test_discretized_curves.py
nguigs: in the setup it is used as the L2 metric, here it is used to test the SRV metric, which results in `n_sampling_points - 1` points

nguigs:
The vectorization raises a NotImplementedError with Pytorch, and the SRV metric only works with numpy. Should this be mentionned in the doc?

ninamiolane:
Thanks for the explanation on the different logic for DiscreteCurves and Landmarks.

Which not implemented error are you referring to for vectorization? OK to mention it in the doc or raise an issue.

geomstats/geometry/discretized_curves.py
```diff
@@ -14,120 +14,172 @@
...
 
         Returns
         -------
         belongs : bool
+            Boolean evaluating if point belongs to the space of discrete
+            curves.
         """
-        belongs = gs.all(self.ambient_manifold.belongs(point))
-        return belongs
+        def each_belongs(pt):
```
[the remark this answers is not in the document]
nguigs: but this is a function within a method, and point is already used in the method

nguigs:
> Which not implemented error are you referring to for vectorization? OK to mention it in the doc or raise an issue.

https://github.com/geomstats/geomstats/blob/c70315cca98ff604b2137a9cc676c1555d1910bd/geomstats/_backend/pytorch/__init__.py#L71

geomstats/geometry/discretized_curves.py
```diff
@@ -14,120 +14,172 @@
...
+            if hasattr(ambient_manifold, 'metric'):
+                self.ambient_metric = ambient_manifold.metric
+            else:
+                raise ValueError('Either a metric must be passed or an '
+                                 'ambient manifold equipped with a metric.')
+        else:
+            self.ambient_metric = metric
+        self.l2_metric = lambda n: L2Metric(ambient_manifold, n_landmarks=n)
 
     def pointwise_inner_product(self, tangent_vec_a, tangent_vec_b,
```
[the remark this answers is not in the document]
nguigs: I had the feeling their was a difference in the output shape (in L2 it is summed over the dimension representing the sampling points), I'll check

PR #772, opened 2020-05-03
Address issues in hypersphere, hyperbolic and backend

ninamiolane:


geomstats/_backend/tensorflow/__init__.py
```diff
@@ -79,9 +79,8 @@ def _raise_not_implemented_error(*args, **kwargs):
     raise NotImplementedError
 
 
-# TODO (nkoep): The 'repeat' function was added in TF 2.1. Backport the
-#              implementation from tensorflow/python/ops/array_ops.py.
-repeat = _raise_not_implemented_error
+def repeat(*args, **kwargs):
```
nkoep, line 82: Did we bump the minimal version dependency of TF to 2.1? If yes, I think we should still make sure the argument names follow the numpy convention, i.e., the first argument name should be `a`, not `input` as in TF's case.

ninamiolane: Yes, we did bump it, see : https://github.com/geomstats/geomstats/blob/master/opt-requirements.txt.

Good point, I will change the argument name. On that matter, do you want to merge PR #540?

PR #564, opened 2020-04-14
Add EM on Poincaré Ball

Hzaatiti:
Add Expectation-Maximization on Poincare Ball

Left to do:
- Optimise code to avoid using repeats
- Apply EM to atleast one other manifold (the actual code is not generic to apply EM on other manifolds)
- Reorganise the PoincareBall class (to be discussed)
- Add further tests for EM

PR #773, opened 2020-05-03
PoincareBall dist vectorization

Hzaatiti:
Address issue #338 for PoincareBall geodesic. 

Previously, the PoincareBallMetric dist function does not handle point_a of shape(m,dim) and point_b of shape(u,dim) with u different than m. 
With this code, dist can handle such shapes by broadcasting arrays point_a and point_b (which is not expensive because no new memory is allocated) then a resulting array with shape (m, u, dim) contains the  distances of each point in point_a with each point in point_b. 
This will avoid using repeats and reshapes for my other EM pull request.

Todo: Add further tests to test the logging errors ?


geomstats/_backend/tensorflow/__init__.py
```diff
@@ -532,6 +532,39 @@ def cast(x, dtype):
...
+        if len(shape) == max_rank:
+            continue
+
+        tensor = tensors[index]
+        for _ in range(max_rank - len(shape)):
+            shape.insert(0, 1)
+            tensor = tf.expand_dims(tensor, axis=0)
+        tensors[index] = tensor
+
+    from collections import Counter
```
ninamiolane, line 551: Why an import here?

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -532,6 +532,39 @@ def cast(x, dtype):
...
+    broadcast_shape = []
+    for index in range(max_rank):
+        dimensions = [s[index] for s in shapes]
+        repeats = Counter(dimensions)
+        if len(repeats) > 2 or (len(repeats) == 2 and
+                                1 not in list(repeats.keys())):
+            raise Exception("Broadcasting not possible")
```
ninamiolane, line 558: Maybe give the reason why broadcasting is not possible, in the spirit of np's error message:

```ValueError: operands could not be broadcast together with shapes (2,3) (4,3)```

File: geomstats/geometry/poincare_ball.py [hunk 3]
```diff
@@ -227,32 +229,73 @@ def mobius_add(self, point_a, point_b):
         return mobius_add
 
     @geomstats.vectorization.decorator(['else', 'vector', 'vector'])
-    def dist(self, point_a, point_b):
-        """Compute the geodesic distance between two points.
+    def dist(self, points_a, points_b):
```
ninamiolane, line 232: We chose the singular form to be the convention of parameters' names and docstring: see methods' signature and docstrings in the codebase.

geomstats/geometry/poincare_ball.py
```diff
@@ -227,32 +229,73 @@ def mobius_add(self, point_a, point_b):
...
+            Second set of points in hyperbolic space.
 
         Returns
         -------
-        dist : array-like, shape=[...,]
-            Geodesic distance between the two points.
+        dist : array-like,
+            shape=[n_samples_a, dim] or [n_samples_a, n_samples_b, dim]
+            Geodesic distance between the two set of points.
+            If n_samples_a == n_samples_b then dist is the element-wise
```
ninamiolane, line 247: This explanation could go into the description of the method, and not of the output.

geomstats/geometry/poincare_ball.py
```diff
@@ -227,32 +229,73 @@ def mobius_add(self, point_a, point_b):
...
+            distance result of a point in points_a with the point from
+            points_b of the same index.
+            If n_samples_a not equal to n_samples_b then dist is the result
+            of applying geodesic distance for each point from points_a to all
+            points from points_b.
         """
-        point_a_norm = gs.clip(gs.sum(point_a ** 2, -1), 0., 1 - EPSILON)
-        point_b_norm = gs.clip(gs.sum(point_b ** 2, -1), 0., 1 - EPSILON)
+        if points_a.shape[-1] != points_b.shape[-1]:
+            logging.error('Manifold Dimensions not equal')
```
ninamiolane, line 255: NIT: dimensions

geomstats/geometry/poincare_ball.py
```diff
@@ -227,32 +229,73 @@ def mobius_add(self, point_a, point_b):
...
+                point_b_flatten ** 2, -1), 0., 1 - EPSILON)
+
+            square_diff = (point_a_flatten - point_b_flatten) ** 2
+
+            diff_norm = gs.sum(square_diff, -1)
+            norm_function = 1 + 2 * \
+                diff_norm / ((1 - point_a_norm) * (1 - point_b_norm))
+
+            dist = gs.log(norm_function + gs.sqrt(norm_function ** 2 - 1))
+            dist = gs.to_ndarray(dist, to_ndim=1)
```
ninamiolane, line 279: Are these two `to_ndarray` needed?

geomstats/geometry/poincare_ball.py
```diff
@@ -227,32 +229,73 @@ def mobius_add(self, point_a, point_b):
...
+
+            diff_norm = gs.sum((points_a - points_b) ** 2, -1)
+            norm_function = 1 + 2 * \
+                diff_norm / ((1 - point_a_norm) * (1 - point_b_norm))
 
-        diff_norm = gs.sum((point_a - point_b) ** 2, -1)
-        norm_function = 1 + 2 * \
-            diff_norm / ((1 - point_a_norm) * (1 - point_b_norm))
+            dist = gs.log(norm_function + gs.sqrt(norm_function ** 2 - 1))
+            dist = gs.to_ndarray(dist, to_ndim=1)
```
ninamiolane, line 295: Same question

Hzaatiti:
> Great idea! Will the use of the vectorized dist method shorten the code somewhere else in the codebase? (Maybe in PR #564 , on which I have commented several times about the vectorization/use of `to_ndarray`?).
> 
> My only concern is that the logic of the `PoincareBall`'s dist method is now different than the logic of all other `dist` methods in the codebase, which do not allow this broadcasting.
> 
> Therefore, I see two options:
> 
> 1. If this `dist` is only a convenient method to shorten the number of code lines in Poincare ball computations, maybe make it a private method `_dist` and have the `dist` be consistent with the behavior of the other dist methods?
> 2. Use this logic in all `dist` methods: the code from this PR would go in the method `dist` of the class `RiemannianMetric`. We could think about doing the same thing for `exp`, `log`, `inner_product`, etc.
> 
> I am just worried that integrating this broadcasting logic everywhere will make the codebase harder to maintain for us. Thus, I would prefer 1. What do you think?
> 
> @nguigs do you have an opinion on this?

Thanks for the comments. Indeed, for PR #564 will allow me to remove alot of the code that is used to expand/repeat/reshapes (and repeat in numpy reallocates memory so it is an expensive operation). I haven't checked if code somewhere else can be shortened too. If additionally, for exp and log this logic is continued this could also help shorten the code even more. 
For option 1, on a short term adding this broadcasting as a private method is I guess a good idea to avoid confusion. However if we wish to apply a same learning algorithm for different distances and avoid expensive repeats then on the long term option 2 can be beneficial.

ninamiolane:
Great, let's move forward with option 1 then, and open an issue to discuss option 2 with others?

geomstats/_backend/pytorch/__init__.py
```diff
@@ -154,6 +154,10 @@ def flip(x, axis):
     return torch.flip(x, dims=axis)
 
 
+def broadcast_arrays(*args, **kwargs):
+    return torch.broadcast_tensors(*args, **kwargs)
```
nguigs, line 158: Couldn't this be imported as broadcast_arrays directly?

Hzaatiti: Yes! thank you for that

Hzaatiti:
Making it a private method _dist causes DeepSource error (accessing a protected member from the test function), should I call _dist from within the dist function ?

geomstats/geometry/poincare_ball.py
```diff
@@ -226,6 +228,74 @@ def mobius_add(self, point_a, point_b):
...
+            Second set of points in hyperbolic space.
+
+        Returns
+        -------
+        dist : array-like,
+            shape=[n_samples_a, dim] or [n_samples_a, n_samples_b, dim]
+            Geodesic distance between the two points.
+        """
+        if point_a.shape[-1] != point_b.shape[-1]:
+            logging.error('Manifold dimensions not equal')
```
ninamiolane, line 256: Do you want to raise ValueError here instead?

geomstats/geometry/poincare_ball.py
```diff
@@ -226,6 +228,74 @@ def mobius_add(self, point_a, point_b):
...
+            diff_norm = gs.sum(square_diff, -1)
+            norm_function = 1 + 2 * \
+                diff_norm / ((1 - point_a_norm) * (1 - point_b_norm))
+
+            dist = gs.log(norm_function + gs.sqrt(norm_function ** 2 - 1))
+            dist *= self.scale
+            dist = gs.reshape(dist, (point_a.shape[0], point_b.shape[0]))
+            dist = gs.squeeze(dist)
+
+        elif point_a.shape == point_b.shape:
```
ninamiolane, line 285: These lines are not tested.

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -553,6 +554,39 @@ def cast(x, dtype):
...
+    tensors = [x, y]
+    shapes = [t.get_shape().as_list() for t in tensors]
+    max_rank = max(len(s) for s in shapes)
+
+    for index, value in enumerate(shapes):
+        shape = value
+        if len(shape) == max_rank:
+            continue
+
+        tensor = tensors[index]
```
ninamiolane, line 567: These lines are not tested.

Hzaatiti: Added test_broadcast_arrays and error tests as well

Hzaatiti: > Very nice! 🎉 Good to go when the remaining unit tests are added!
> 
> About the DeepSource error: it is fine to have that error from a test file: we do want to test the private functions. So if you prefer having `_dist` and the error only comes from the test file, that is fine! However, `dist_broadcast` is a good explicit name: maybe it's even better to keep it like this?

Yes dist_broadcast is kept

Issue #769, opened 2020-05-02
Enhance einsum function


ninamiolane:
The `einsum` function in the pytorch and tf backends needs to be generalized to handle these cases:
- x, y not of the same dtype,
- when there is no '->' in the einsum str,
- ellipsis of the type 'n...k'
- work for more than two input tensors, eg einsum(str, x, y, z), useful in `connection.py` for example.

PR #776, opened 2020-05-04
Allow einsum to work with arrays of different dtypes

ninamiolane:
This PR starts to address issue #769 .

This PR adds the function `convert_to_wider_dtype` in the backends. This function casts arrays from a list of arrays into arrays of dtype the "wider" dtype. For example, a list of int32 and float32 arrays is cast to float32 and float32 arrays.

This allows einsum to work with arrays of different dtypes.

Note: The code of `convert_to_wider_dtype` is essentially the same for all backends. Could we factorize it somewhere? I tried to put it in a file `_backend/casting.py` but this leads to failure of backend's imports. @nkoep

### DIFF
--- a/geomstats/_backend/numpy/__init__.py
+++ b/geomstats/_backend/numpy/__init__.py
@@ -27,6 +27,7 @@ from autograd.numpy import (  # NOQA
     diagonal,
     divide,
     dot,
+    dtype,
     einsum,
     empty,
     empty_like,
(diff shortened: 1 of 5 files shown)


geomstats/_backend/numpy/__init__.py
```diff
@@ -93,6 +93,21 @@
...
+    dtype_list = [x.dtype for x in tensor_list]
+    wider_dtype = dtype_list[0]
+    wider_dtype_index = DTYPES.index(wider_dtype)
+    for dtype in dtype_list[1:]:
+        index = DTYPES.index(dtype)
+        if index > wider_dtype_index:
+            wider_dtype = dtype
+            wider_dtype_index = index
+    tensor_list = [cast(x, dtype=wider_dtype) for x in tensor_list]
+    return tensor_list
```
nguigs, line 109: Your basically looking for a max with respect to the order defined by the list `DTYPES`, so maybe
DTYPES could be a dict with keys int32 etc, and values 1..4 and here you could have
dtype_list = [DTYPES[x.dtype] for x in tensor_list]
so that the wider_dtype_index is just max(dtype_list) and you don't need to look for it by hand below

ninamiolane: Good point, thanks! This is done.

PR #777, opened 2020-05-04
Run geometric examples with all backends

ninamiolane:
This follows PR #776 .

The visualization module can work with all 3 backends, as matplotlib.pyplot can plot torch/tf tensors.

This PR allows the examples calling the `geomstats/geometry` module to run - and be tested - with the 3 backends.

However, the examples calling the `geomstats/learning` module cannot be easily ported to the 3 backends so far. This is because they call scikit-learn functions that cannot take tensors as inputs.


examples/plot_geodesics_so3.py
```diff
@@ -1,41 +1,33 @@
 """Plot a geodesic of SO(3).
 
-SO(3) is equipped with its left-invariant canonical METRIC.
+SO(3) is equipped with its left-invariant canonical metric.
```
nguigs, line 3: bi-invariant

geomstats/_backend/numpy/__init__.py
```diff
@@ -93,6 +93,25 @@
...
 
+DTYPES = [int32, int64, float32, float64]
+
+
+def to_numpy(x):
+    return x
+
+
+def convert_to_wider_dtype(tensor_list):
+    dtype_list = [x.dtype for x in tensor_list]
```
nguigs, line 104: `DTYPES` could be a dict with keys int32 etc, and values 1..4 and here you could have
`dtype_list = [DTYPES[x.dtype] for x in tensor_list]`
so that the `wider_dtype_index` is just `max(dtype_list)` and you don't need to look for it by hand below

geomstats/_backend/pytorch/__init__.py
```diff
@@ -88,6 +90,29 @@ def wrapper(x):
...
+    for dtype in dtype_list[1:]:
+        index = DTYPES.index(dtype)
+        if index > wider_dtype_index:
+            wider_dtype = dtype
+            wider_dtype_index = index
+    tensor_list = [cast(x, dtype=wider_dtype) for x in tensor_list]
+    return tensor_list
+
+
+def ceil(x):
```
nguigs, line 110: well spotted! What about using the `_box_scalar` wrapper though?

geomstats/_backend/numpy/__init__.py
ninamiolane: Done.

examples/plot_geodesics_so3.py
ninamiolane: Done.

geomstats/_backend/pytorch/__init__.py
ninamiolane: I had forgotten about that decorator, thank you!

Issue #453, opened 2020-02-21
Create tutorials of geometric statistics with geomstats


ninamiolane:
Getting inspiration from the notebooks at: https://github.com/ageron/handson-ml,
we could create a series of 5 to 10 notebooks teaching geometric statistics through the use of geomstats.

This would re-use the code from the examples, preferably applied to real datasets.

@AliceLeBrigant you could be interested in this!

AliceLeBrigant:
Yes, great idea. The difficulty would mostly reside in finding real datasets. @ninamiolane did you have any specific in mind ?

Hzaatiti:
Could we start by providing jupiter notebooks (.ipynb) for each of the existing examples in geomstats ?  this could enhance the readability of the examples given that github displays notebooks really well. With some modification to sphinx, the notebooks can be displayed in the online documentation it seems as well (https://nbsphinx.readthedocs.io/en/0.5.1/)

ninamiolane:
Yes, notebooks would be the best! And knowing that we can have them directly on the documentation with sphinx is amazing, thanks for the link.

The only reason we have not provided notebooks is that continuous integration can be tricky, and we didn't want to have potentially buggy notebooks on the GitHub. However, it seems that notebooks can be tested on Travis, for example with: 

https://github.com/ghego/travis_anaconda_jupyter

ninamiolane:
For the datasets we have started to discuss it in issue #302 . So far, we have a `datasets` directory, but only a real dataset on graphs in provided. @AliceLeBrigant

ninamiolane:
Let's divide the work on writing the tutorials/notebooks? @AliceLeBrigant @Hzaatiti 

- [] `01_data_on_manifolds.ipynb`: 
    - what: data examples and forbidden linear operations on manifolds
    - who: @ninamiolane 
- [] `02_from_vector_spaces_to_manifolds.ipynb`: 
    - what: metric, exp, log, geodesics, retraction, lifting
   - who: @ninamiolane

AliceLeBrigant:
Yes ! I can do something like
- [ ] `03_basic_statistics_on_manifolds.ipynb`:
      - what: interpolation (geodesics), Fréchet mean, pca
      - who: @AliceLeBrigant 
- [ ]  `xx_information_geometry.ipynb`:
      - what: Fisher-Rao geometry of Gaussian and beta distributions
      - who: @AliceLeBrigant

ninamiolane:
[...]

Issue #302, opened 2020-01-16
Add a dataset directory


stefanheyder:
This should contain common datasets of Riemannian Geometry, see e.g. https://cran.r-project.org/web/packages/circular/circular.pdf and https://cran.r-project.org/web/packages/CircNNTSR/CircNNTSR.pdf for classical datasets available in R (related to circular / spherical statistics).

ninamiolane:
On which platform should we store the datasets? They cannot be on github for memory reasons.

Maybe OSF? https://osf.io/
Or a platform that scikit-learn / pytorch / tf / keras are using?

stefanheyder:
Good idea, but I am afraid that I am not familiar with publishing datasets.

Also does someone know how licensing works for datasets, especially if we want to reuse some datasets from e.g. R packages?

ninamiolane:
For starters, we could actually follow scikit-learn strategy and create a directory for small datasets directly on github:
https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/datasets

Not sure how licensing works there though.

ninamiolane:
A repository for small datasets has been created here: https://github.com/geomstats/geomstats/tree/master/examples/data.

nguigs:
This was done in #780 or before.

PR #781, opened 2020-05-06
Add functions to load toy datasets

ninamiolane:
This PR continues to address issues #302 and #453 .

- It changes the localization of the `data` folder, which now under `geomstats/datasets`.
- It uses .json format to store data, as it is easier to load data using the dictionary structure.
- It adds utils functions to load datasets.
- It adds the corresponding tests to load datasets in np, tf and torch.

@AliceLeBrigant @Hzaatiti @tgeral68


geomstats/datasets/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    with open(CITIES_PATH) as json_file:
+        data_file = json.load(json_file)
+
+        for row in data_file:
+            lat_in_radians = row['lat'] / 90 * gs.pi / 2
+            lng_in_radians = row['lng'] / 180 * gs.pi
+            data.append(gs.array([lat_in_radians, lng_in_radians]))
+            names.append(row['city'])
+
+    data = gs.array(data)
```
nguigs, line 50: Here's how I would do it using the map function:
```python
names = list(map(lambda x: x['city'], data_file))
data = gs.array(list(map(lamda x:  [x[k] / 180 * gs.pi for k in ['lat', 'lng']], data_file)))
```

tests/test_datasets.py
```diff
@@ -1,23 +1,50 @@
...
-                        '/graph_karate/karate_labels.txt')
+            graph_matrix_path=data_utils.KARATE_PATH,
+            labels_path=data_utils.KARATE_LABELS_PATH)
 
-    @geomstats.tests.np_only
+    def test_load_cities(self):
+        """Test that the cities coordinates belong to the sphere."""
+        sphere = Hypersphere(dim=2)
+        data, _ = data_utils.load_cities()
+        data = sphere.spherical_to_extrinsic(data)
```
nguigs, line 26: Really nice that we can now do this! Did you make sure that our `spherical_to_extrinsic` uses the right reference point for the spherical coordinates ? ([0, 0, 1]  should be the north pole but we used [1, 0, 0] I think). Also I think there is a further conversion from latitude to colatitude to do (x -> pi / 2 - x). This is probably not important here, but in a tutorial we should be faithful to geography :)

tests/test_datasets.py
```diff
@@ -1,23 +1,50 @@
...
+        result = sphere.belongs(data)
```
nguigs, line 27: Another comment: `spherical_to_extrinsic` always returns a norm-1 vector by construction, so the belong check doesn't really check anything. It doesn't really matter here because you just want to make sure load_cities ran smoothly.

tests/test_datasets.py
ninamiolane: 🙏 Thanks! I was battling with this on the tutorial indeed (see PR #782 ).

geomstats/datasets/utils.py
ninamiolane: Done.

tests/test_datasets.py
ninamiolane: I'll address this in the next PR.

Issue #726, opened 2020-04-30
Geodesic method is not vectorized


nguigs:
The following code does not run (where `self.space=Hypersphere(4)`):

```python
n_geodesic_points = 10
initial_point = self.space.random_uniform(2)
vector = gs.array([[2., 0., -1., -2., 1.]] * 2)
initial_tangent_vec = self.space.to_tangent(
    vector=vector, base_point=initial_point)
geodesic = self.metric.geodesic(
    initial_point=initial_point,
    initial_tangent_vec=initial_tangent_vec)

t = gs.linspace(start=0., stop=1., num=n_geodesic_points)
points = geodesic(t)
```

In fact it won't be possible to fully vectorize the path method of geodesics, because it would require exp and log method to take input of shapes `[n_times, n_samples, {dim [n,n]}]`, so there is an extra dimension we are not ready to deal with. We can either
- update the doc to mention it's not vectorized
- call path separately for each sample with a for loop.

PR #783, opened 2020-05-08
Vectorization of geodesic method

nguigs:
This PR adresses
- vectorization of geodesics #726. It now works when multiple initial conditions are passed. The time axis is the first one in the output but this can be discussed
- factorization of left/right translation of tangent vectors #703. The new method should be used to avoid bugs when switching point types


examples/plot_pole_ladder_s2.py
```diff
@@ -62,7 +62,7 @@ def main():
 
     base_point = gs.to_ndarray(base_point, to_ndim=2)
     origin = gs.concatenate(
-        [base_point, base_point, final_geodesic(gs.array([0]))], axis=1)
+        [base_point, base_point, final_geodesic(gs.array([0.]))], axis=1)
```
ninamiolane, line 65: Hopefully, we can get rid of these soon, using `convert_to_wider_dtype` backend functions.

geomstats/geometry/connection.py
```diff
@@ -455,32 +455,39 @@ def geodesic(self, initial_point,
         Returns
         -------
         path : callable
-            Time parameterized geodesic curve.
+            Time parameterized geodesic curve. IF a batch of initial
```
ninamiolane, line 458: Typo: If

geomstats/geometry/invariant_metric.py
```diff
@@ -124,18 +124,10 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
...
-            tangent_vec_b_at_id = self.group.compose(
-                inv_jacobian, tangent_vec_b)
-        elif self.left_or_right == 'right':
-            tangent_vec_a_at_id = self.group.compose(
-                tangent_vec_a, inv_jacobian)
-            tangent_vec_b_at_id = self.group.compose(
-                tangent_vec_b, inv_jacobian)
+        tangent_translation = self.group.tangent_translation_map(
+            base_point, left_or_right=self.left_or_right, inverse=True)
+        tangent_vec_a_at_id = tangent_translation(tangent_vec_a)
```
ninamiolane, line 129: I wonder if we could use the parallel transport naming for the group left-invariant connection? Something like `left_parallel_transport`?

geomstats/geometry/invariant_metric.py
```diff
@@ -124,18 +124,10 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
...
-            tangent_vec_a_at_id = self.group.compose(
-                inv_jacobian, tangent_vec_a)
-            tangent_vec_b_at_id = self.group.compose(
-                inv_jacobian, tangent_vec_b)
-        elif self.left_or_right == 'right':
-            tangent_vec_a_at_id = self.group.compose(
-                tangent_vec_a, inv_jacobian)
-            tangent_vec_b_at_id = self.group.compose(
-                tangent_vec_b, inv_jacobian)
+        tangent_translation = self.group.tangent_translation_map(
```
ninamiolane, line 127: Nice!

tests/test_connection.py
```diff
@@ -153,6 +154,60 @@ def test_log_connection_metric_vectorization(self):
...
+            initial_point=initial_point,
+            initial_tangent_vec=initial_tangent_vec)
+        points = geodesic(t)
+        result = points[-1]
+        expected = space.bi_invariant_metric.exp(
+            initial_tangent_vec, initial_point)
+        self.assertAllClose(expected, result)
+
 
 if __name__ == '__main__':
```
ninamiolane, line 212: We don't need these lines at the end of the test file.

tests/test_connection.py
```diff
@@ -153,6 +154,60 @@ def test_log_connection_metric_vectorization(self):
 
         self.assertAllClose(result, expected, rtol=1e-5, atol=1e-5)
```
ninamiolane, line 155: Sourcegraph spots these lines as not tested. Maybe deleting the geomstats.tests.main() at the end of this file will change this?

geomstats/geometry/invariant_metric.py
nguigs: I would argue not to do that, because this is only true with a certain connection, not necessarily the canonical one: on SOn, with the bi-invariant connection, the parallel transport is dL_h dR_h where h = sqrt(g). Plus parallel transport needs a connection or a metric to be defined, where as the left translation only needs the smooth group structure.

tests/test_connection.py
nguigs: I think the codecov report was generated before travis was done with the numpy tests, and that's why it says coverage will drop.

geomstats/geometry/invariant_metric.py
ninamiolane: Ok!

examples/plot_pole_ladder_s2.py
nguigs: The problem here is more that path except a 1d array for t, a single scalar cannot be passed.

nguigs:
About the time dimension, I think the usecase for geodesic is to be able to access points at different times easily, and by putting t in the first dimension this is easy with e.g.
```python
points = geodesic(t)
point_ti = point[i]
```
rather than having to be careful with the number of samples, if t was in the second dim, but only if there are different initial conditions: `point_ti = point[..., i]`.

Maybe one way to decide would be to make the output of geodesic(t) easily compatible with the visualisation module, `draw_points(geodesic[t])` would be nice if it worked straight away with many initial conditions.

examples/plot_pole_ladder_s2.py
nguigs: I added a to_ndarray to deal with this case

ninamiolane: Yes, but also `gs.array([0])` -> `gs.array([0.])`, right?

PR #784, opened 2020-05-08
Remove assignments in so3

nguigs:
This allows tf tests to pass.

### DIFF
--- a/geomstats/geometry/special_orthogonal.py
+++ b/geomstats/geometry/special_orthogonal.py
@@ -183,13 +183,14 @@ class _SpecialOrthogonal3Vectors(LieGroup):
 
     identity = property(get_identity)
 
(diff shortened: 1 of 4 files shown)


geomstats/geometry/special_orthogonal.py
```diff
@@ -42,21 +43,24 @@ def __init__(self, n):
...
+    def belongs(self, point, atol=ATOL):
         """Check whether point is an orthogonal matrix.
 
         Parameters
         ----------
         point : array-like, shape=[..., n, n]
             Point to check.
+        atol : float
+            Absolute tolerance to check equality of the transpose and the
+            inverse of point.
```
ninamiolane, line 55: Optional, default: 1e-5

PR #788, opened 2020-05-10
Adapt discrete curve to pytorch and tf

nguigs:
- Implement a vectorize method for pytorch. It may not behave exactly like the numpy one (which is flexible thanks to the signature argument)
- Remove decorators so that all methods are tested in the three backends

Remaining problems: The discrete curve class is thought for points on manifolds which have vector representations, it won't work for curves of matrices...


ninamiolane:
Nice!!

PR #813, opened 2020-05-27
Factorize gradient computation in the backends

nguigs:
This PR adds an autograd module to the backend so that automatic differentiation can be used with all backends. This is a proof of concept that it works for the computation of the log on the sphere using Christoffel symbols (already existed before but with numpy only).
For the gradient descent, the scipy solver is used, so we go back and forth between torch/tensorflow tensors and numpy arrays, but it remains simpler than using torch/tensorflow optimizers.

### DIFF
--- a/geomstats/_backend/__init__.py
+++ b/geomstats/_backend/__init__.py
@@ -110,6 +110,7 @@ BACKEND_ATTRIBUTES = {
         'zeros',
         'zeros_like'
     ],
+    'autograd': ['value_and_grad'],
     'linalg': [
         'det',
         'eig',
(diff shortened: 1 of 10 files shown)


nguigs:
there is a new deepsource issue that is not from me...

ninamiolane:
Yes, but tf/torch need a `linalg` module, whereas they do not need an `autograd` module as they have separate automatic differentiation features.

We can ignore the DS issue since it's not from you. 

Let's wait for @johmathe and @nkoep on this?

nguigs:
> Yes, but tf/torch need a `linalg` module, whereas they do not need an `autograd` module as they have separate automatic differentiation features.

ok, linalg is a bad example, have a look at random, all the methods there are native torch/tensorflow methods, but with made the module random to have a common API! This is also what we need for auto diff because the torch/tf/autograd have a different api.

But what else would you suggest?  In all cases, it has to go to the back-end. A different module name?

PR #815, opened 2020-05-28
Kalman filter on Lie group example

pchauchat:
This PR proposes an example of a non-linear Kalman filter on Lie groups. It will most probably not be accepted as is, but this allows asking a few questions.

The example is that of localization problems, with two different setups, a linear and a non-linear one. Both are handled with the proposed KF, all the specificities are enclosed in the associated models (Localization and LocalizationLinear classes).

- The model for the 2D non-linear states estimation on SE(2) is based on Invariant Extended Kalman Filter, from https://arxiv.org/pdf/1410.1465.pdf (it is mentioned in the example docstring). Basically this changes how the innovation is computed and the associated jacobian and covariance. Is it too specific, or is it ok ?

- Should the various classes be defined in the example file, or in a companion file ? If it is the second, where should this file be ?

- Should the tangent map associated to the adjoint (`adjoint_map`) be placed directly in the corresponding group classes ?

- I also have a vectorized version. The code is a bit heavier (for now I explicitly distinguishes the cases based on `ndim` tests). Would it be an interesting contribution ? Maybe in a separate example ?


examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+        gain = self.compute_gain(observation)
+        obs_jac = self.model.observation_jacobian(self.state, observation)
+        self.covariance = (gs.eye(self.model.dim) - gain.dot(obs_jac)).dot(
+            self.covariance)
+        self.state = self.model.group.exp(gain.dot(innovation), self.state)
+        if gs.ndim(self.state) > 1:
+            return gs.squeeze(self.state)
+
+
+def main():
```
johmathe, line 252: It would be great to have this example in a python notebook.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+        Here the linear error observation - expected is brought back to
+        the Lie algebra
+        """
+        theta, _, _ = state
+        rot = self.rotation_matrix(theta)
+        expected = Localization.observation_model(state)
+        return rot.T.dot(observation - expected)
+
+
+class KalmanFilter:
```
johmathe, line 194: Can we bring in some unit tests for this class? Especially edge cases.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+    def observation_model(state):
+        """Model used to create the measurements."""
+        return state[:1]
+
+    def innovation(self, state, observation):
+        """Discrepancy between the measurement and its expected value."""
+        return observation - LocalizationLinear.observation_model(state)
+
+
+class Localization:
```
johmathe, line 91: We need unit testing for this class.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+    def __init__(self, model):
+        self.model = model
+        self.state = model.group.get_identity()
+        self.covariance = gs.zeros((self.model.dim, self.model.dim))
+        self.process_noise = gs.zeros(
+            (self.model.dim_noise, self.model.dim_noise))
+        self.measurement_noise = gs.zeros(
+            (self.model.dim_obs, self.model.dim_obs))
+
+    def initialise_covariances(self, prior_values, process_values, obs_values):
```
johmathe, line 210: initialize or init - geomstats has been using american English for the codebase so far.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+
+        return traj, three_sigmas
+
+    np.random.seed(12345)
+    model = LocalizationLinear()
+    kalman = KalmanFilter(model)
+    n_traj = 1000
+    obs_freq = 50
+    dt = 0.1
+    P0 = gs.array([10., 1.])
```
johmathe, line 302: caps are for global variables IIRC. @ninamiolane can you confirm?

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+
+    @staticmethod
+    def observation_model(state):
+        """Model used to obtain the measurements."""
+        return state[1:]
+
+    def innovation(self, state, observation):
+        """Discrepancy between the measurement and its expected value.
+
+        Here the linear error observation - expected is brought back to
```
johmathe, line 185: Should this be a todo? Hard to understand the comment here.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def propagate(self, state, sensor_input):
+        """Propagate with piece-wise constant acceleration and velocity."""
+        dt, acc = sensor_input
+        pos, speed = state
+        pos = pos + dt * speed
+        speed = speed + dt * acc
+        return gs.array([pos, speed])
+
+    def propagation_jacobian(self, state, sensor_input):
```
johmathe, line 58: why do you pass the state here?

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+
+class LocalizationLinear:
+    """Class for modeling a linear 1D localization problem.
+
+    The state is made of a scalar position and scalar speed, thus a 2D vector.
+    A sensor provides acceleration inputs, while another one provides sparse
+    measurements of the position.
+    """
+    def __init__(self):
+        self.group = Euclidean(2)
```
johmathe, line 45: It seems that this class doens't have any state variable, these look all like static variables and don't need to be initialized in init. Then most of the methods could be classmethods.

examples/kalman_filter.py
```diff
@@ -0,0 +1,374 @@
...
+    plt.figure()
+    plt.plot(true_traj[:, 1], true_traj[:, 2], label='GT')
+    plt.plot(estimate[:, 1], estimate[:, 2], label='Kalman')
+    plt.scatter(observations[:, 0], observations[:, 1], s=2, c='k',
+                label='Observation')
+    plt.legend()
+    plt.axis('equal')
+    plt.title('2D Localization')
+
+    plt.show()
```
johmathe, line 370: the examples are supposed to be able to run from command line IIRC. Instead of show this should be savefig.

See example at:

https://github.com/geomstats/geomstats/blob/master/examples/plot_expectation_maximization_manifolds.py

examples/kalman_filter.py
pchauchat: It is not a todo, more an explanation of why it is different than the linear case and the standard EKF, following the reference given at the beginning. Should I give a more direct reference, then (e.g., equation number) ? Or simply go back to a more simple approach ?

examples/kalman_filter.py
pchauchat: I tried to keep everything as general as possible, as in some cases the state will be involved in the jacobian's computation. If someone wants to create a class for another problem where it is the case, they will not have to change the rest of the code. I definitely realize it is not perfect, and I'd gladly take any advice here !

examples/kalman_filter.py
pchauchat: In the example you linked to, the `main` function ends with `show`. I agree for the command line part, but why is savefig a better choice then ? (Real question there, I'm simply used to work with `show` and did not really consider alternatives)

examples/kalman_filter.py
pchauchat: Should everything be in the notebook ? Or only the part after the classes are defined ?

examples/kalman_filter.py
```diff
@@ -0,0 +1,376 @@
...
+import geomstats.backend as gs
+from geomstats import algebra_utils
+from geomstats.geometry.euclidean import Euclidean
+from geomstats.geometry.special_euclidean import SpecialEuclidean
+
+os.environ['GEOMSTATS_BACKEND'] = 'numpy'  # NOGA
+
+
+class LocalizationLinear:
+    """Class for modeling a linear 1D localization problem.
```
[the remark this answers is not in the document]
pchauchat: Not at all, but since it is supposed to be a toy example, I thought keeping things in 1D could be clearer. What do you think ?

examples/kalman_filter.py
johmathe: I see. Might be the formulation of the sentence. Maybe state clearly that 'the 'expected' vector is brought back to the Lie algebra. (Also can we find something more accurate than 'brought'? It doesn't sound like a proper mathematical term).

examples/kalman_filter.py
johmathe: I would say only the parts after the class @ninamiolane thoughts?

examples/kalman_filter.py
johmathe: If you don't use savefig the mapltolib backend is going to try to show the figure on a screen. These examples should be able to run from the command line without hiccups, hence better use savefig.

examples/kalman_filter.py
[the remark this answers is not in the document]
ninamiolane: Maybe the class could be general (both classes, if possible?) but the examples could be 1D/2D to keep it clearer?

examples/kalman_filter.py
ninamiolane: Agreed!

examples/kalman_filter.py
ninamiolane: Yes! The only exception that we have made is `X`, the design matrix, to follow scikit learn notations.

examples/kalman_filter.py
ninamiolane: Hm, so far we have some `plt.show()` in almost all examples... (We shut them down when we run the examples at test time). Why is it better to save than to show? I would say this is personal preference?

tests/test_examples.py
```diff
@@ -72,6 +73,11 @@ def test_loss_and_gradient_so3():
     def test_loss_and_gradient_se3():
         loss_and_gradient_se3.main()
 
+    @staticmethod
+    @geomstats.tests.np_only
```
ninamiolane, line 77: I know this PR is already huge, so for the next one: it would be great if it can work in tf/torch as well, otherwise this will decrease or test coverage :(

pchauchat: I am also working on it, but since I knew there would be major changes, I was waiting a bit to see where that PR would go (although only the numpy version of `eye` allows for defining subdiagonals, which is a pain)

examples/kalman_filter.py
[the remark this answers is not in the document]
pchauchat: That should be feasible, it may just make some functions a bit uglier (especially the jacobians), let's see how it goes.

tests/test_examples.py
ninamiolane: Makes sense!

For the eye issue, maybe gs.diag or gs.diagonal can help?

examples/kalman_filter.py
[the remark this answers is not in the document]
ninamiolane: Of course, i's also OK to have a first version in 1D/2D if going to nD is significantly more work! But I feel that we should go to nD in the long run!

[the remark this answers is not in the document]
pchauchat: The main problem, for the non linear case, is that it's based on the vector representation of SE(n), which we only have for n = 2 and 3 for now.

pchauchat:
Since I had to rebase to take the last PRs into account, this has just become a monster. Is it still reviewable, or should I close it and make a new one ?

For the new changes :
- Now works with all backends (Note that I had to redefine `matmul` in tf's backend, to mimic the torch's and numpy's versions, which allow both matrix-matrix and matrix-vector products)
- As a side effect, this is also the case for SE(2)
- I kept the 1D/2D version for now, as the vector representation of the nD case is not defined yet.
- Code cleaned a bit (single letter variable, etc.)

What is NOT in there :
- Unit tests (I don't really see how to put one for the KalmanFilter class: show exact computations of some cases ? Also, where should they be put ? Inside the function testing the example ?)
- General nD case
- Notebook version. I still have to get my hands on this thing.

pchauchat:
Seems like the redefinition made some other tests fail. I'll look into it.

PR #816, opened 2020-05-28
Notebook to demonstrate ML in tangent Spaces

nguigs:
Adds a notebook that exemplifies the use of the `ToTangentSpace` transformer in a pipeline with a logistic regression, on brain connectome data. Should this go in applications or to the main repo?


notebooks/04_simple_machine_learning_tangent_spaces.ipynb
```diff
@@ -0,0 +1,266 @@
+{
```
ninamiolane, line 1: <p>NIT: spd -&gt; SPD</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/geomstats/geomstats/pull/816/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

ninamiolane, line 1: <p>Actually, FrechetMean should be covered in tutorial 03. Maybe let's say: covered in the previous tutorial?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/geomstats/geomstats/pull/816/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

ninamiolane:
Awesome! Very well written and very clear! Two main comments:

- Maybe add a function that tests that load_connectomes works in test_datasets? This PR decreases coverage otherwise.

- The tutorial is great to illustrate the use of ToTangentSpace, but it feels like the story is not finished w.r.t. the connectomes dataset -> The only output of the algorithm is a score, we do not know if the AffineInvariant metric performs better or worst than the Euclidean metric or another, we do not classify the brain connetomes... etc
--> Will this be in another notebook? Maybe @danielbrookes is working on this?

ninamiolane:
As for `notebooks` VS `applications`, I think this PR should be in the notebooks. 

However, we should probably adapt the notebooks so that:
- They are not downloaded with pip3 install (probably already the case? 🙏 )
- The unit tests only test a "debug" version of the notebooks, e.g. with 5 data only.

nguigs:
>     * Maybe add a function that tests that load_connectomes works in test_datasets? This PR decreases coverage otherwise.
Yes good catch!
>     * The tutorial is great to illustrate the use of ToTangentSpace, but it feels like the story is not finished w.r.t. the connectomes dataset -> The only output of the algorithm is a score, we do not know if the AffineInvariant metric performs better or worst than the Euclidean metric or another, we do not classify the brain connetomes... etc
We do classify, the score that gets printed is the cross val accuracy! I can compare with Euclidean, but AI is slightly worse.

ninamiolane:
I see, maybe add at least a conclusion sentence: this means that we can classify subjects with schizophrenia versus control with a 70% accuracy, etc, etc.

Or maybe try another metric and find which metric gives the best accuracy? (Coordinating with Daniel depending on what he wants to do?)

PR #821, opened 2020-06-04
Refresh of  #815 : Kalman filter on Lie group example

pchauchat:
This PR is a follow-up of #815, to avoid all the commits that came with several rebasing. 

**Reminder of the original purpose:**

This PR proposes an example of a non-linear Kalman filter on Lie groups. It will most probably not be accepted as is, but this allows asking a few questions.

The example is that of localization problems, with two different setups, a linear and a non-linear one. Both are handled with the proposed KF, all the specificities are enclosed in the associated models (Localization and LocalizationLinear classes).

- The model for the 2D non-linear states estimation on SE(2) is based on Invariant Extended Kalman Filter, from https://arxiv.org/pdf/1410.1465.pdf (it is mentioned in the example docstring).
- Should the tangent map associated to the adjoint (adjoint_map) be placed directly in the corresponding group classes ?
- I also have a vectorized version. The code is a bit heavier (for now I explicitly distinguishes the cases based on ndim tests). Would it be an interesting contribution ? Maybe in a separate example ?

**After the discussion on the previous PR:** 
Recent changes:
- Now works with all backends Note that I had to redefine `matmul` in tf's backend, to mimic the torch's and numpy's versions, which allow both matrix-matrix and matrix-vector products)
- As a side effect, this is also the case for SE(2)
- I kept the 1D/2D version for now, as the vector representation of the nD case is not defined 
- Code cleaned a bit (single letter variable, etc.)

What is NOT in this PR, let me know what you think on some of the points
- Unit tests: I don't really see how to put one for the KalmanFilter class: show exact computations of some cases ? Also, where should they be put ? Inside the function testing the example ?
- Some docstring
-  General nD case (Note that making a general nD version and specifying the dimension when calling it would make all methods not static anymore)
- Notebook version. I still have to get my hands on this thing.


examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+    """
+
+    group = Euclidean(2)
+    dim = group.dim
+    dim_noise = 1
+    dim_obs = 1
+
+    @staticmethod
+    def propagate(state, sensor_input):
+        """Propagate with piece-wise constant acceleration and velocity."""
```
ninamiolane, line 55: Docstrings need parameters' descriptions
+ math equations would be helpful, using :math:

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+        """
+        dt, _ = sensor_input
+        dim = LocalizationLinear.dim
+        position_wrt_noise = gs.zeros((dim // 2, dim // 2))
+        speed_wrt_noise = gs.sqrt(dt) * gs.eye(dim // 2)
+        jac = gs.vstack((position_wrt_noise, speed_wrt_noise))
+        return jac
+
+    @staticmethod
+    def observation_jacobian(state, observation):
```
ninamiolane, line 87: `observation` unused?

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+        self.state = model.group.get_identity()
+        self.covariance = gs.zeros((self.model.dim, self.model.dim))
+        self.process_noise = gs.zeros(
+            (self.model.dim_noise, self.model.dim_noise))
+        self.measurement_noise = gs.zeros(
+            (self.model.dim_obs, self.model.dim_obs))
+
+    def initialize_covariances(self, prior_values, process_values, obs_values):
+        """Set the values of the covariances."""
+        values = [prior_values, process_values, obs_values]
```
ninamiolane, line 245: Maybe use a dict here?

tests/test_examples.py
```diff
@@ -72,6 +73,10 @@ def test_loss_and_gradient_so3():
     def test_loss_and_gradient_se3():
         loss_and_gradient_se3.main()
 
+    @staticmethod
+    def test_kalman_filter():
```
ninamiolane, line 77: Nice!! In all 3 backends 🎉

tests/test_special_euclidean2.py
```diff
@@ -165,7 +165,6 @@ def test_inverse_vectorization(self):
         self.assertAllClose(
             gs.shape(result), (n_samples, *self.group.get_point_type_shape()))
 
-    @geomstats.tests.np_only
     def test_group_exp_from_identity_vectorization(self):
```
ninamiolane, line 168: Nice!

examples/kalman_filter.py
johmathe: +1:
```
 {'covariance': prior_value, ...}
```

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+def main():
+    """Carry out two examples of state estimation on groups.
+
+    Both examples are localization problems, where only a part of the system
+    is observed. The first one is a linear system, while the second one is
+    non-linear.
+    """
+    def create_data(kalman, true_init, true_inputs, obs_freq):
```
johmathe, line 293: let's declare these outside main please.

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+        """Model used to create the measurements."""
+        return state[:1]
+
+    @staticmethod
+    def innovation(state, observation):
+        """Discrepancy between the measurement and its expected value."""
+        return observation - LocalizationLinear.observation_model(state)
+
+
+class Localization:
```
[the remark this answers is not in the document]
pchauchat: @ninamiolane had the same idea, but the main problem is that the vector representation for SE(n) only exists for the 2D and 3D cases (as far as I understood it).

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
...
+    """
+
+    group = SpecialEuclidean(2, 'vector')
+    dim = group.dim
+    dim_rot = group.rotations.dim
+    dim_noise = 3
+    dim_obs = 2
+
+    @staticmethod
+    def split_input(sensor_input):
```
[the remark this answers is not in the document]
pchauchat: It simply retrieves the inputs of various "sensors" from the aggregated vector, e.g. separates the time step from the linear velocity and angular velocity in the non-linear case, or just the time step and the acceleration in the linear one.
I don't really have a better name in mind, maybe retrieve_inputs ?

examples/kalman_filter.py
```diff
@@ -0,0 +1,416 @@
+r"""Illustrate how a Kalman-like filter can be defined on Lie groups.
+
+A generic Kalman filter class is defined for systems on Lie groups for which
+the exponential is subjective. Its use is illustrated on two localization
```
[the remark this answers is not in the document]
pchauchat: Indeed

examples/kalman_filter.py
pchauchat: In this case yes, but since in other cases it might, I preferred keeping it here so that if one changes the model the KalmanFilter class would still work. Don't really know whether it's best practice, but I could not think of an alternative yet.

tests/test_kalman_filter.py
```diff
@@ -0,0 +1,167 @@
...
+        self.obs_cov = 2. * gs.eye(1)
+
+    def test_LocalizationLinear(self):
+        initial_state = gs.array([0.5, 1.])
+        time_step = 0.5
+        acc = 2.
+        increment = gs.array([time_step, acc])
+        measurement = gs.array([0.7])
+
+        # Test propagation
```
ninamiolane, line 31: These different tests should be in different `test_*` functions, in order to have "unit" tests. This way, when travis fails, we directly know which part of the code is failing: using the name `test_*`.

Would it be possible to make these tests unit?

PR #824, opened 2020-06-15
Fix embedding for tensorflow

Hzaatiti:


geomstats/_backend/pytorch/__init__.py
```diff
@@ -39,6 +39,7 @@
     lt as less,
     matmul,
     max as amax,
+    mean,
```
ninamiolane, line 42: Since numpy's and torch's API of mean are different (`axis` is `dim` in torch) we should keep the lines below. Or use https://github.com/geomstats/geomstats/pull/540

If you are interested in using this, let me know and I'll merge the PR.

geomstats/learning/frechet_mean.py
```diff
@@ -167,7 +167,8 @@ def _ball_gradient_descent(points, metric, weights=None, max_iter=32,
 
         iteration = 0
         convergence = math.inf
-        barycenter = points.mean(0, keepdims=True) * 0
+        barycenter = gs.mean(points, axis=0, keepdims=True)
```
ninamiolane, line 170: How does this work in pytorch, since `axis` is not a parameter of `torch.mean`https://pytorch.org/docs/1.4.0/torch.html?highlight=torch%20mean#torch.mean?

tests/test_prepare_graph_data.py
```diff
@@ -58,7 +57,6 @@ def test_loss(self):
         self.assertAllClose(loss_value[0], expected_loss, rtol=1e-3)
         self.assertAllClose(gs.squeeze(loss_grad), expected_grad, rtol=1e-3)
 
-    @geomstats.tests.np_and_pytorch_only
     def test_embed(self):
         """Test embedding function."""
         embeddings = self.embedding_class.embed(self.karate_graph)
```
ninamiolane, line 62: NIT (not so important for the tests, but more important for the core code): we should not use `embedding_class` to call the OBJECT from the class HyperbolicEmbedding, because it is precisely an object and not a class.

ninamiolane:
Good to merge once the conflicts are solved ! (+ small questions on the PR).

PR #825, opened 2020-06-17
Fix references in notebook

Hzaatiti:


ninamiolane:
Great!! Two remarks though:
- Don't you want to cite your paper?
- Would it be possible to use this (https://stackoverflow.com/questions/30061902/how-to-handle-citations-in-ipython-notebook), using the .bib from the scipy-paper? I would be more sustainable, allowing us to build the bibliography of papers that are implemented/referenced in the library. This, in turn, could for a reference biblio that we could link on our soon-to-come wonderful website, etc.

Hzaatiti:
> * Don't you want to cite your paper?
Graph embedding isn't much the idea of our paper, but I'll be preparing a new notebook for community learning on graphs (which is embedding while using the EM algorithm). There I will cite our paper since it's more in the scope :)

> Would it be possible to use this (https://stackoverflow.com/questions/30061902/how-to-handle-citations-in-ipython-notebook), using the .bib from the scipy-paper? I would be more sustainable, allowing us to build the bibliography of papers that are implemented/referenced in the library. This, in turn, could for a reference biblio that we could link on our soon-to-come wonderful website, etc.

I had tried almost everything to make .bib work :/ the link u refer to produces a word/or pdf with the citations but not sure if it can display the .bib within the notebook (tried it). 
The only thing I found working is [cite2c ](https://github.com/takluyver/cite2c). It requires to sync the .bib to zotero and will need authentication to read the bib entry from it. I managed to make it work locally, but not sure how would that go server side ?

ninamiolane:
@Hzaatiti Got it, let's merge it for now! Maybe create an issue so that we remember to clean/factorize the references in the notebooks?

PR #828, opened 2020-06-23
Add kmedoids

Hzaatiti:
Not yet compatible with pytorch/tensorflow

Will add it to notebook 03 once it works with all backends

### DIFF
--- a/geomstats/_backend/pytorch/random.py
+++ b/geomstats/_backend/pytorch/random.py
@@ -2,12 +2,14 @@
 
 import torch
 
+
 def choice(x, a):
     """Generate a random sample from an array of given size."""
     if torch.is_tensor(x):
         return x[torch.randint(len(x), (a,))]
     else:
         return x
+
 
 def rand(*args, **kwargs):
     return torch.rand(*args, **kwargs)
@@ -37,4 +39,3 @@ def uniform(low=0.0, high=1.0, size=(1,)):
     if low >= high:
         raise ValueError('Upper bound must be higher than lower bound')
     return (high - low) * torch.rand(*size) + low
-
--- a/geomstats/_backend/pytorch/random.py
+++ b/geomstats/_backend/pytorch/random.py
@@ -7,8 +7,8 @@ def choice(x, a):
     """Generate a random sample from an array of given size."""
     if torch.is_tensor(x):
         return x[torch.randint(len(x), (a,))]
-    else:
-        return x
+
+    return x
 
 
 def rand(*args, **kwargs):
--- a/geomstats/_backend/pytorch/random.py
+++ b/geomstats/_backend/pytorch/random.py
@@ -2,6 +2,12 @@
 
 import torch
 
+def choice(x, a):
+    """Generate a random sample from an array of given size."""
+    if torch.is_tensor(x):
+        return x[torch.randint(len(x), (a,))]
+    else:
+        return x
 
 def rand(*args, **kwargs):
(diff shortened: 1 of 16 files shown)


geomstats/geometry/hypersphere.py
```diff
@@ -597,6 +597,33 @@ def dist(self, point_a, point_b):
...
+        dist : array-like, shape=[n_samples, n_samples]
+            Pairwise distance matrix between all points.
+        """
+        n_samples, _ = point.shape
+
+        dist = gs.zeros((n_samples, n_samples))
+
+        for i in range(0, n_samples):
+            for j in range(i + 1, n_samples):
+                if i == j:
```
ninamiolane, line 619: j can't be i here, right?

geomstats/geometry/hypersphere.py
```diff
@@ -597,6 +597,33 @@ def dist(self, point_a, point_b):
 
         return dist
 
+    def dist_pairwise(self, point):
```
ninamiolane, line 600: This should be a method of the class RiemannianMetric, since it will be the same code for all the child classes.

geomstats/geometry/hypersphere.py
```diff
@@ -597,6 +597,33 @@ def dist(self, point_a, point_b):
...
+
+        dist = gs.zeros((n_samples, n_samples))
+
+        for i in range(0, n_samples):
+            for j in range(i + 1, n_samples):
+                if i == j:
+                    continue
+
+                dist[i][j] = self.dist(point[i], point[j])
+                dist[j][i] = dist[i][j]
```
nguigs, line 623: Try to avoid assignment to make it more easily compatible with other backends, you can use append:
```python
pairwise_dist = []
for i, x in enumerate(data):
    for y in data[i:]:
        pairwise_dist.append(self.dist(x,y))
pairwise_dist = SymmetricMatrices.from_vector(pairwise_dist)
```

Hzaatiti: Thank you, this helped me make the example work for pytorch!

geomstats/geometry/hypersphere.py
Hzaatiti: True! thanks

geomstats/geometry/hypersphere.py
Hzaatiti: Right sorry!

tests/test_backends.py
```diff
@@ -916,3 +916,18 @@ def test_value_and_grad_numpy_input(self):
...
+
+    def test_choice(self):
+
+        x = gs.array([0.1, 0.2, 0.3, 0.4, 0.5])
+        a = 4
+
+        result = gs.random.choice(x, a)
+
+        for i in result:
+            if i in x:
```
ninamiolane, line 928: maybe `self.assertTrue(i in x)`? and delete the `self.assertTrue(False)`?

geomstats/_backend/pytorch/random.py
```diff
@@ -3,6 +3,14 @@
 import torch
 
 
+def choice(x, a):
+    """Generate a random sample from an array of given size."""
+    if torch.is_tensor(x):
+        return x[torch.randint(len(x), (a,))]
+
+    return x
```
ninamiolane, line 11: This line is not tested.

Hzaatiti:
The tests and examples are fixed now for pytorch, now working on tensorflow

PR #842, opened 2020-07-21
Fix EM issue

tgeral68:
Fix the issue in Expectation Maximization:
- Frechet Mean issue on one element
- Adding projection in Poincaré Ball manifold
- EM issue with Gaussian PDF initialization


geomstats/learning/expectation_maximization.py
```diff
@@ -312,6 +318,8 @@ def fit(self,
                 label_mask = gs.where(labeled_data[:, 0] == label)
                 grouped_by_label = labeled_data[label_mask][:, 1:]
                 v = variance(grouped_by_label, centroid, self.metric)
+                if grouped_by_label.shape[0] == 1:
```
clharris, line 321: Good catch

clharris:
> Very nice, thanks! Only the github conflicts that need to be solved. @clharris @Hzaatiti do you want to have a look?

Looks good 👍

geomstats/geometry/poincare_ball.py
```diff
@@ -173,6 +173,36 @@ def weighted_gmm_pdf(mixture_coefficients,
 
         return weighted_pdf
 
+    @geomstats.vectorization.decorator(['else', 'vector'])
+    def projection(self, point):
```
Hzaatiti, line 177: Could we have a unit test for this function ?

Hzaatiti:
This is great and should solve the discussed issue #791 !

ninamiolane:
Agreed with @Hzaatiti , adding a unit-test would be great, otherwise we are decreasing code coverage (even for numpy).

ninamiolane:
@tgeral68 update on this?

PR #844, opened 2020-07-26
Change symmetric eigenvalue, eigenvector func from np to torch

cshewmake2:
Passed my local tests as well as all nose2 tests (excluding known karate club dataset import issue in a notebook).

### DIFF
--- a/geomstats/_backend/pytorch/linalg.py
+++ b/geomstats/_backend/pytorch/linalg.py
@@ -35,8 +35,8 @@ def eigvalsh(*args, **kwargs):
 
 
 def eigh(*args, **kwargs):
-    eigs = np.linalg.eigh(*args, **kwargs)
-    return torch.from_numpy(eigs[0]), torch.from_numpy(eigs[1])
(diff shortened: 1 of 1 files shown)


cshewmake2:
I just looked; I'm not sure that we have a unit test for this function.

ninamiolane:
@clharris you might be interested in this!

@cshewmake2  thanks!! 🎉  We might have an issue later on if we want to perform eigenvalue decomposition of non-symmetric matrices. But since eigh is only used on symmetric matrices in our code base right now, it's all good!

cshewmake2:
@ninamiolane, as I understand it, np.linalg.eigh is specifically built for Hermitian (conjugate symmetric) matrices. I agree, if we want to do eigenvalue decomposition for non-symmetric matrices, we'll have to use something else (e.g. we couldn't have used eigh either).

One possible issue is that torch.symeig supposedly only works on **real** symmetric matrices, whereas eigh can handle matrices with complex entries. So perhaps this will cause problems.

ninamiolane:
Got it, thanks for the clarification!

PR #849, opened 2020-08-21
Factorize Taylor Approximations

nguigs:
This PR factorizes Taylor approximations around 0. This avoids the use of masks, assignments and calls to to_ndarray, and thus should improve readability of the code. Furthermore, the accuracy should be slightly improved as:
```python
space = Hypersphere(2)
base_point = space.random_uniform()
vect = space.to_tangent(gs.array([1.0, 0, 0]), base_point)
point = space.metric.exp(1e-6 * vect, base_point)
log = space.metric.log(point, base_point)
```
currently returns log=0. It is not longer the case with this factorization, and even works for much smaller vectors. The same could be applied to hyperbolic geometry and the special euclidean group in a future PR.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -96,6 +96,7 @@ exp = _box_scalar(exp)
 log = _box_scalar(log)
 sin = _box_scalar(sin)
 sinh = _box_scalar(sinh)
+tan = _box_scalar(tan)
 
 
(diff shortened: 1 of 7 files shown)


nguigs:
The ResourceVariable problem doesn't occur in #851 so maybe it can be merged instead and this one can be closed.

ninamiolane:
@nguigs Yes let's merge #851

PR #847, opened 2020-08-13
Kristianeschenburg grassmann belongs

nguigs:
This PR adds tools for the Grassmann Manifold, following up on #572:
- remove hard coded dimension
- add belongs, is_tangent, to_tangent methods
- change exp and log to take and return tangent vectors instead of skew-sym matrices
- add random_uniform method
- change the metric to inherit MatricesMetric so that inner products and norms can be computed with the Frobenius dot prod.

@opeltre

PR #850, opened 2020-08-22
Grassmann doc

opeltre:
First effort to improve the docs for Grassmannians...

Putting this here so we can discuss with @nguigs, yet it should be merged after #847 as the code here is old


ninamiolane:
🎉  Very welcome documentation and the doc screenshot looks great! #847 has been merged, so this is ready to go too.

Though, the unit tests need to pass and the github conflicts need to be resolved.

nguigs:
continued in #877

PR #851, opened 2020-08-25
Update parallel transport methods

nguigs:
Thanks to new [results](https://hal.inria.fr/hal-02894783v1) on Schild's and pole ladder.
#849 was merged in this branch because the numerical accuracy is needed:
#849 factorizes Taylor approximations around 0. This avoids the use of masks, assignments and calls to to_ndarray, and thus should improve readability of the code. Furthermore, the accuracy should be slightly improved as:
```python
space = Hypersphere(2)
base_point = space.random_uniform()
vect = space.to_tangent(gs.array([1.0, 0, 0]), base_point)
point = space.metric.exp(1e-6 * vect, base_point)
log = space.metric.log(point, base_point)
```
currently returns log=0. It is not longer the case with this factorization, and even works for much smaller vectors. The same could be applied to hyperbolic geometry and the special euclidean group in a future PR.


geomstats/algebra_utils.py
```diff
@@ -21,3 +58,27 @@ def from_vector_to_diagonal_matrix(vector):
...
     diagonals = gs.einsum('...i,ij->...ij', vector, identity)
     return diagonals
+
+
+def taylor_exp_even_func(
+        point, taylor_function, order=5, tol=EPSILON):
+    """Taylor Approximation of an even function around zero.
+
+    Parameters
+    ----------
```
ninamiolane, line 68: Missing documentation

geomstats/geometry/hypersphere.py
```diff
@@ -454,51 +439,21 @@ def exp(self, tangent_vec, base_point):
...
-            mask_non0)
-
+        hypersphere = Hypersphere(dim=self.dim)
+        proj_tangent_vec = hypersphere.to_tangent(tangent_vec, base_point)
+        norm2 = self.embedding_metric.squared_norm(proj_tangent_vec)
+
+        coef_1 = utils.taylor_exp_even_func(
+            norm2, utils.cos_close_0, order=4)
+        coef_2 = utils.taylor_exp_even_func(
+            norm2, utils.sinc_close_0, order=4)
```
ninamiolane, line 449: Nice! 🎉

geomstats/geometry/hypersphere.py
```diff
@@ -16,21 +16,7 @@
...
-                     + 1.0 / math.factorial(8), 0.]
-INV_SIN_TAYLOR_COEFFS = [0., 1. / 6.,
-                         0., 7. / 360.,
-                         0., 31. / 15120.,
-                         0., 127. / 604800.]
-INV_TAN_TAYLOR_COEFFS = [0., - 1. / 3.,
-                         0., - 1. / 45.,
-                         0., - 2. / 945.,
-                         0., -1. / 4725.]
+EPSILON = 1e-4
```
ninamiolane, line 19: Why this change in EPSILON?

geomstats/geometry/special_euclidean.py
```diff
@@ -46,10 +46,11 @@ class _SpecialEuclideanMatrices(GeneralLinear, LieGroup):
 
     def __init__(self, n):
         super(_SpecialEuclideanMatrices, self).__init__(
-            dim=int((n * (n - 1)) / 2), default_point_type='matrix', n=n + 1)
+            dim=int((n * (n + 1)) / 2), default_point_type='matrix', n=n + 1)
         self.rotations = SpecialOrthogonal(n=n)
         self.translations = Euclidean(dim=n)
         self.n = n
+        self.dim = int((n * (n + 1)) / 2)
```
ninamiolane, line 53: This attribute should have been set while calling super()?

geomstats/geometry/special_orthogonal.py
```diff
@@ -1,9 +1,9 @@
 """Exposes the `SpecialOrthogonal` group class."""
 
+import geomstats.algebra_utils as utils
```
ninamiolane, line 3: We should try to avoid renaming the submodules, for ease of readability of the code.

tests/test_algebra_utils.py
```diff
@@ -0,0 +1,24 @@
+import math
+
+import geomstats.algebra_utils as utils
+import geomstats.tests
+
+
+class TestTaylorExp(geomstats.tests.TestCase):
```
ninamiolane, line 7: name of the class should match the name of the file: TestAlgebraUtils

tests/test_connection.py
```diff
@@ -54,45 +54,54 @@ def test_christoffels(self):
 
         self.assertAllClose(result, expected)
 
-    @geomstats.tests.np_and_pytorch_only
     def test_parallel_transport(self):
```
ninamiolane, line 57: Nice!!

tests/test_hypersphere.py
```diff
@@ -296,16 +311,13 @@ def test_exp_and_log_and_projection_to_tangent_space_edge_case(self):
         # Edge case: tangent vector has norm < epsilon
         base_point = gs.array([10., -2., -.5, 34., 3.])
         base_point = base_point / gs.linalg.norm(base_point)
-        vector = 1e-10 * gs.array([.06, -51., 6., 5., 3.])
+        vector = 1e-4 * gs.array([.06, -51., 6., 5., 3.])
```
ninamiolane, line 314: Why this?

geomstats/geometry/hypersphere.py
nguigs: Because this is the precision of some functions in Pytorch, e.g. gs.cos(1e-4) == 1. results in True

geomstats/geometry/special_euclidean.py
nguigs: For some reason using super does not overwrite the dimension of the parent class GeneralLinear so this results in `SpecialEuclidean(n).dim = n^2` (:fearful: )

tests/test_hypersphere.py
nguigs: For a smaller vector, the test used to pass because the tolerance was much larger than the vector. With Pytorch and Tensorflow, it doesn't work for vectors smaller than 1e-6 with tol 1e-8...

nguigs:
The decrease in coverage was due to a check that alpha >=1. I added a test with assertRaises.

geomstats/geometry/special_euclidean.py
ninamiolane: Hmmm, it'd be great if we can unterstand this, as we might have that problem for other classes too!

ninamiolane:
🎉  Great! Good to go when the comments on `algebra_utils as utils` and `TestAlgebraUtils` are addressed (for some reason they don't show up on the last commit - or maybe you don't want to include these changes? happy to discuss it if so!).

geomstats/geometry/special_orthogonal.py
nguigs: I didn't want the lines to be too long, and it doesn't really make a difference between algebra_utils and utils as 'algebra' is not even relevant in that case.

nguigs:
I did the change for `TestAlgebraUtils`

nguigs:
I don't think having `geomstats.algebra_utils as utils` hinders the understanding of the code, as nothing in `geomstats.algebra_utils` gives any precision. The only indication is the name of the method which is explicit: `taylor_exp_even_func`. And using the entire name `geomstats.algebra_utils` will force breaking lines with \ which is not recommended. So I am for merging this PR as is.

PR #790, opened 2020-05-11
Create Poincare Half Space

AliceLeBrigant:
Create class Poincaré Half Space. The operations are deduced from the ones in the Poincare ball and a change of coordinates (the Cayley transform).

I don't know if the tests run in pytorch and tensorflow, I am currently unable to execute any test using nose2 and another backend than numpy. I don't know why.

PR #852, opened 2020-08-27
Poincarre halfspace

nguigs:
Continuing #790 from @AliceLeBrigant to create a class for the Poincare Half-space representation of the hyperbolic space. However, maybe this representation could fit in the existing class, by adding the change of coordinate functions in the Hyperbolic class?

### DIFF
--- a/examples/kalman_filter.py
+++ b/examples/kalman_filter.py
@@ -73,9 +73,12 @@ def create_data(kalman, true_init, true_inputs, obs_freq):
     observations = [gs.cast(obs, obs_dtype) for obs in observations]
 
     input_dtype = true_inputs[0].dtype
-    inputs = [gs.concatenate(
-        (incr[:1], np.random.multivariate_normal(
-            incr[1:], kalman.process_noise)), axis=0)
+    inputs = [
+        gs.concatenate([
+            incr[:1],
+            gs.array(
+                np.random.multivariate_normal(
+                    incr[1:], kalman.process_noise))], axis=0)
         for incr in true_inputs]
     inputs = [gs.cast(incr, input_dtype) for incr in inputs]
 
--- a/examples/kalman_filter.py
+++ b/examples/kalman_filter.py
@@ -67,10 +67,11 @@ def create_data(kalman, true_init, true_inputs, obs_freq):
         for pose in true_traj[obs_freq::obs_freq]]
 
     obs_dtype = true_obs[0].dtype
-    observations = [
-        np.random.multivariate_normal(obs, kalman.measurement_noise)
-        for obs in true_obs]
-    observations = [gs.cast(obs, obs_dtype) for obs in observations]
+    observations = gs.stack([
+        gs.array(
+            np.random.multivariate_normal(obs, kalman.measurement_noise),
+            dtype=obs_dtype)
+        for obs in true_obs])
 
     input_dtype = true_inputs[0].dtype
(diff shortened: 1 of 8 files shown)


ninamiolane:
This goes back to the discussion on how to implement the different coordinates system of a given manifold. @tgeral68 @Hzaatiti .

Our first implementation was to:
- have methods changing coordinates into one single Hyperbolic class,
and after several discussions @tgeral68 refactorized this into:
- have one hyperbolic class for each coordinate system (these classes could be private, so that the user only has to remember one class: `Hyperbolic`?), and independent conversion functions that can go from one representation to another,
which is easier to read!

Thus it makes sense to have a separate class for the Poincare Half-space representation.

We could also re-use the change of coordinate functions that were in `visualization.py`, and implement the Klein Disk representation, etc.

AliceLeBrigant:
Thanks @nguigs for completing the docstrings ! Does the test run in pytorch and tensorflow ? I still seem to have a problem running tests with ```nose2``` locally with these backends (even though I deleted the old ```backend``` folder and kept only ```_backend```).

nguigs:
I moved the change of coordinates functions to the Hyperbolic class, because the ones there that used the half plane only worked for n=2.
The coverage was decreased because the change of coordinates for tangent vectors were not tested. I also noticed that the methods did not for n_samples > 1, so I wrote them differently.
There are two problems remaining:
- ball_to_half_plane_tangent and half_plane_to_ball_tangent are not inverse
- the test for the exp with vertical initial vector does not pass anymore, whereas the other one does, any idea why @AliceLeBrigant ?

AliceLeBrigant:
I believe I have fixed both problems, but I don't know how to push the modifications to this branch. What I did:
1. I fetched the PR branch and gave it the local branch name ```pr852```
```git fetch upstream pull/852/head:pr852```
```git checkout pr852```
2. I did my modifications on this branch
3. I commited the changes and pushed them
```git push origin pr852```

I think I have pushed them to my fork, but not to the upstream. Do you know what I did wrong ? Thanks !

nguigs:
Thanks for the fixes! I merged your changes locally and just pushed them. For next time, in fact you have to push to my fork, so add it as a remote:
`git remote add nicolas https://github.com/nguigs/geomstats.git`

AliceLeBrigant:
Thanks for the merge and the tip ! In fact I had to specify on which branch to push:
```git push nicolas pr852:poincarre-halfspace```

Now:
- Is there a reason for ```_extrinsic_to_half_space_coordinates``` and ```_half_space_to_extrinsic_coordinates``` methods to be class methods instead of static methods like the rest of the change of coordinates methods ?

- Should ```half_space_to_ball_coordinates``` and  ```ball_to_half_space_coordinates``` begin with an ```_ ``` like the est of the change of coordinates methods ?

nguigs:
>     * Is there a reason for `_extrinsic_to_half_space_coordinates` and `_half_space_to_extrinsic_coordinates` methods to be class methods instead of static methods like the rest of the change of coordinates methods ?
Yes, this allows to call other methods from the same class, which you cannot do in a static method.

>     * Should `half_space_to_ball_coordinates` and  `ball_to_half_space_coordinates` begin with an `_ ` like the rest of the change of coordinates methods ?
In the long term probably yes, but so far we cannot because this method is used in another class (PoincareHalfSpaceMetric) and `_` is for private methods

PR #853, opened 2020-09-09
Poincare half space (continued)

AliceLeBrigant:
Continue with development of Poincare half space.

### DIFF
--- a/tests/test_poincare_half_space.py
+++ b/tests/test_poincare_half_space.py
@@ -25,6 +25,14 @@ class TestPoincareHalfSpace(geomstats.tests.TestCase):
         expected = gs.array([True, False])
         self.assertAllClose(result, expected)
 
+    def test_inner_product_vectorization(self):
+        tangent_vec = gs.array([[1., 2.], [3., 4.]])
+        base_point = gs.array([[0., 1.], [0., 5.]])
+        result = self.metric.inner_product(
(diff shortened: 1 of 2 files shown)


tests/test_poincare_half_space.py
```diff
@@ -25,6 +25,14 @@ def test_belongs(self):
         expected = gs.array([True, False])
         self.assertAllClose(result, expected)
 
+    def test_inner_product_vectorization(self):
```
nguigs, line 28: Did you try other vectorization scenarios? E.g. one pair of tangent_vec + one base_point, one tangant_vec_a, multiple tangent_vec_b, one base_point, or multiple pairs of tangent_vecs at one base_point?

nguigs (commented):
Great!
@ninamiolane what do you think about the Deep Source error? I thought we had turned off these types of errors?

tests/test_poincare_half_space.py
AliceLeBrigant: Yes they also work. Should I include them in ```test_poincare_half_space```?

ninamiolane (approved):
🎉 Awesome, LGTM! Yes, let's ignore the deep source error, this is one that we have chosen to ignore. I tried to change the deep source settings, but couldn't for some reason. They might have changed their interface. I'll try later also, but in the meantime we can merge this!

Issue #854, opened 2020-09-17
Vectorization problem of exp in PoincareBallMetric


AliceLeBrigant:
The ```geodesic``` function of ```PoincareBallMetric``` fails to give results for a series of times including ```0```. 
Consider the following example:
```
import geomstats.backend as gs
from geomstats.geometry.poincare_ball import PoincareBall

B2 = PoincareBall(2)
initial_point = gs.array([0.2, 0.4])
end_point = gs.array([0.3, 0.5])
geod = B2.metric.geodesic(initial_point, end_point)
```
Then the following line of code fails:
```
geod(gs.linspace(0., 1., 10))
```
while these do not:
```
geod(0)
geod(gs.linspace(0.1, 1., 10))
```
This seems to come from ```exp```. A simpler example of the same bug is the following:
```
point = gs.array([0.2, 0.4])
tangent_vecs = gs.array([[0., 0.], [0.5, 0.5], [1., 1.]])
```
Then the following line of code fails:
```
B2.metric.exp(tangent_vecs, point)
```
while these do not:
```
B2.metric.exp(tangent_vecs[0, :], point)
B2.metric.exp(tangent_vecs[1:, :], point)
```
This seems to come from the vectorization decorator of ```exp``` in ```PoincareBallMetric```.

tgeral68:
Yes, it comes from the vectorization, since :
```
import geomstats.backend as gs  
from geomstats.geometry.poincare_ball import PoincareBall 
B2 = PoincareBall(2)  
point = gs.array([0.2, 0.4], [0.2, 0.4],[0.2, 0.4]])  
tangent_vecs = gs.array([[0., 0.], [0.5, 0.5], [1., 1.]])      
B2.metric.exp(tangent_vecs, point)  
```
The error occurs while checking if at least one of the tangent vectors is zero vectors (if it the case return the initial point) to avoid dividing by zero. However, it does not seem still necessary anymore since this case is considered earlier in the function. I will make a quick fix adding into the test this particular case.

PR #855, opened 2020-09-17
Fix vectorization problem with zero values in exp map of Poincare ball

AliceLeBrigant:
Address [issue#854](https://github.com/geomstats/geomstats/issues/854).

### DIFF
--- a/geomstats/geometry/poincare_ball.py
+++ b/geomstats/geometry/poincare_ball.py
@@ -266,8 +266,9 @@ class PoincareBallMetric(RiemannianMetric):
             gs.einsum('...i,...->...i', direction, factor))
 
         if gs.any(zero_tan):
+            base_points = gs.tile(base_point, (len(zero_tan), 1))
             exp = gs.assignment(
-                exp, base_point[zero_tan], zero_tan)
+                exp, base_points[zero_tan], zero_tan)
 
         return exp
 
--- a/geomstats/geometry/poincare_ball.py
+++ b/geomstats/geometry/poincare_ball.py
@@ -266,9 +266,10 @@ class PoincareBallMetric(RiemannianMetric):
             gs.einsum('...i,...->...i', direction, factor))
 
         if gs.any(zero_tan):
-            base_points = gs.tile(base_point, (len(zero_tan), 1))
+            if base_point.shape != tangent_vec.shape:
+                base_point = gs.tile(base_point, (len(zero_tan), 1))
             exp = gs.assignment(
-                exp, base_points[zero_tan], zero_tan)
+                exp, base_point[zero_tan], zero_tan)
 
         return exp
 


ninamiolane:
Thanks for fixing this! Good to merge when the following error is fixed in the unit tests 🎉 

```
======================================================================
ERROR: test_log_exp_ball_null_tangent (tests.test_hyperbolic_coords.TestHyperbolicCoords)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/geomstats/geomstats/tests/test_hyperbolic_coords.py", line 181, in test_log_exp_ball_null_tangent
    exp = self.ball_metric.exp(tangent_vec, x)
  File "/home/travis/build/geomstats/geomstats/geomstats/vectorization.py", line 104, in wrapper
    result = function(*vect_args, **vect_kwargs)
  File "/home/travis/build/geomstats/geomstats/geomstats/geometry/poincare_ball.py", line 271, in exp
    exp, base_points[zero_tan], zero_tan)
IndexError: boolean index did not match indexed array along dimension 0; dimension is 4 but corresponding boolean dimension is 2
```

ninamiolane (changes requested):
Just need to fix the travis error.

AliceLeBrigant:
Sorry I'm not familiar with Travis... ```test_hyperbolic_coords.py``` runs without errors on my machine. Do you know what I should check ? Thanks !

ninamiolane:
@Hzaatiti @tgeral68 any idea on the travis error?

AliceLeBrigant:
My bad, ```test_hyperbolic_coords.py``` was indeed not working. It should be fixed now !

tgeral68:
Hi, I fixed this issue in another pull request. Looking closer at the code this check is not necessary anymore since it is partially checked before. The issue when considering zeros vector is division by zero caused when we use the norm of the tangent vector:
`direction = gs.einsum('...i,...->...i', tangent_vec, 1 / norm_tan)
`
However, looking closer at the code we replace null tangent_vector component by a value epsilon. 

```
if gs.any(zero_tan):
            norm_tan = gs.assignment(norm_tan, EPSILON, zero_tan)
```
We then compute the direction : 

` direction = gs.einsum('...i,...->...i', tangent_vec, 1 / norm_tan)
`
Thus, in case of zero vector direction will be zero vector too (norm is no more null but tangent_vec is). We then compute the exp by 
```
        exp = self.mobius_add(
            base_point,
            gs.einsum('...i,...->...i', direction, factor))

```
the mobius addition will return base_point if added to a zero vector. Thus exp(base_point, (0,....,0)) will return the correct results which is  base_point. To conclude, there is no need to check  zero tangent vectors anymore at this location in the code.

sorry for my late intervention.

AliceLeBrigant:
Ok great ! Thanks for this better fix !

PR #860, opened 2020-09-21
Visualization H2

AliceLeBrigant:
In ```visualization``` for ```space=H2_poincare_half_plane```,
- adapt the axis bounds of the Poincare half plane plot to the points that are plotted
- authorize points to be given in Poincare half plane coordinates, and make it the default representation, instead of the extrinsic coordinates. I would be in favour of doing the same for the other representations, as I think it is more natural for the user to give the points' coordinates in the appropriate representation.

### DIFF
--- a/geomstats/visualization.py
+++ b/geomstats/visualization.py
@@ -355,7 +355,7 @@ class PoincareHalfPlane:
                 raise ValueError(
                     'Points do not belong to the hyperbolic space '
                     '(extrinsic coordinates)')
-            points = self.convert_to_poincare_coordinates(points)
+            points = self.convert_to_half_plane_coordinates(points)
         elif self.point_type == 'half-plane':
             if not gs.all(POINCARE_HALF_PLANE.belongs(points)):
                 raise ValueError(
--- a/geomstats/visualization.py
+++ b/geomstats/visualization.py
@@ -365,10 +365,11 @@ class PoincareHalfPlane:
             points = list(points)
         self.points.extend(points)
 
-    @staticmethod
-    def set_ax(points, ax=None):
+    def set_ax(self, points, ax=None):
         if ax is None:
             ax = plt.subplot()
+        if self.point_type == 'extrinsic':
+            points = self.convert_to_half_plane_coordinates(points)
         xmin = points[gs.argmin(points[:, 0]), 0]
         xmax = points[gs.argmax(points[:, 0]), 0]
         ymax = points[gs.argmax(points[:, 1]), 1]
--- a/geomstats/visualization.py
+++ b/geomstats/visualization.py
@@ -6,6 +6,7 @@ import matplotlib.pyplot as plt
 import geomstats.backend as gs
 from geomstats.geometry.hyperboloid import Hyperboloid
 from geomstats.geometry.hypersphere import Hypersphere
+from geomstats.geometry.poincare_half_space import PoincareHalfSpace
 from geomstats.geometry.special_euclidean import SpecialEuclidean
 from geomstats.geometry.special_orthogonal import SpecialOrthogonal
 from mpl_toolkits.mplot3d import Axes3D  # NOQA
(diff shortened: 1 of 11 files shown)


geomstats/visualization.py
```diff
@@ -342,28 +343,41 @@ def draw(self, ax, **kwargs):
...
+        xmin = points[gs.argmin(points[:, 0]), 0]
+        xmax = points[gs.argmax(points[:, 0]), 0]
+        ymax = points[gs.argmax(points[:, 1]), 1]
+        xmargin = 0.1 * (xmax - xmin)
+        ymargin = 0.1 * ymax
         plt.setp(ax,
-                 xlim=(-ax_s, ax_s),
-                 ylim=(0., ax_s),
+                 xlim=(xmin - xmargin, xmax + xmargin),
+                 ylim=(0., ymax + ymargin),
```
nguigs, line 380: Is this necessary to manually set xlim/ylim? It is the only thing preventing the exemple from running with tensorflow, and matplotlib does a good job at defining the limits automatically!

geomstats/visualization.py
```diff
@@ -486,11 +500,25 @@ def convert_to_trihedron(point, space=None):
...
+         point_type=None, **point_draw_kwargs):
+    """Plot points in one of the implemented manifolds.
+
+    The implemented manifolds are:
+    - the special orthogonal group SO(3)
+    - the special Euclidean group SE(3)
+    - the circle S1 and the sphere S2
+    - the hyperbolic plane (the Poincare disk, the Poincare
+      half plane and the Klein disk)
+    - the Poincare polydisk
```
nguigs, line 512: Much nicer!

tests/test_examples.py
```diff
@@ -117,6 +117,7 @@ def test_plot_square_h2_square_poincare_disk():
         plot_square_h2_poincare_disk.main()
 
     @staticmethod
+    @geomstats.tests.np_and_pytorch_only
```
nguigs, line 120: Could be removed, see above comment

geomstats/visualization.py
```diff
@@ -342,28 +343,41 @@ def draw(self, ax, **kwargs):
...
-        points = self.convert_to_half_plane_coordinates(points)
+        if self.point_type == 'extrinsic':
+            if not gs.all(H2.belongs(points)):
+                raise ValueError(
+                    'Points do not belong to the hyperbolic space '
+                    '(extrinsic coordinates)')
+            points = self.convert_to_half_plane_coordinates(points)
+        elif self.point_type == 'half-plane':
+            if not gs.all(POINCARE_HALF_PLANE.belongs(points)):
+                raise ValueError(
```
nguigs, line 361: These lines are not covered by the unit-tests, maybe add a test with point_type=half-plane?

tests/test_visualization.py
```diff
@@ -56,7 +56,8 @@ def test_plot_points_h2_poincare_disk(self):
     @geomstats.tests.np_only
     def test_plot_points_h2_poincare_half_plane(self):
         points = self.H2.random_uniform(self.n_samples)
-        visualization.plot(points, space='H2_poincare_half_plane')
+        visualization.plot(points, space='H2_poincare_half_plane',
+                           point_type='extrinsic')
```
nguigs, line 60: Also test `point_type=None` to improve coverage

nguigs (commented):
Great, I agree with you, its nicer to plot points in the half-plane coordinate system straight away! Do we want to have `point_type='half-plane'` or `'half-space'` to be consistent with the hyperbolic class?

geomstats/visualization.py
nguigs: (Have a look at https://sourcegraph.com/github.com/geomstats/geomstats to find out which lines are not being tested)

geomstats/visualization.py
AliceLeBrigant: Good point !

AliceLeBrigant:
> Great, I agree with you, its nicer to plot points in the half-plane coordinate system straight away! Do we want to have `point_type='half-plane'` or `'half-space'` to be consistent with the hyperbolic class?

Probably 'half-space' is better. Thanks for the comments !

PR #862, opened 2020-09-27
Normal distributions

alebrigant:
Compute pdf associated to a point in the manifold of normal distributions.

### DIFF
--- a/geomstats/geometry/beta_distributions.py
+++ b/geomstats/geometry/beta_distributions.py
@@ -104,6 +104,47 @@ class BetaDistributions(EmbeddedManifold):
                 beta.rvs(param_a, param_b, size=n_samples)))
         return samples[0] if len(point) == 1 else gs.stack(samples)
 
+    def point_to_pdf(self, point):
+        """Compute pdf associated to point.
+
+        Compute the probability density function of the beta
+        distribution with parameters provided by point.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., 2]
+            Point representing a beta distribution.
+
+        Returns
+        -------
+        pdf : function
+            Probability density function of the beta distribution with
+            parameters provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        point = gs.to_ndarray(point, to_ndim=2)
+        a_params = point[:, 0]
+        b_params = point[:, 1]
+
+        def pdf(x):
+            """Generate parameterized function for normal pdf.
+
+            Parameters
+            ----------
+            x : array-like, shape=[n_points,]
+                Points at which to compute the probability density function.
+            """
+            x = gs.array(x, gs.float32)
+            x = gs.to_ndarray(x, to_ndim=1)
+
+            pdf_at_x = [
+                gs.array(beta.pdf(x, a=a, b=b)) for a, b
+                in zip(a_params, b_params)]
+            pdf_at_x = gs.stack(pdf_at_x, axis=1)
+
+            return pdf_at_x
+        return pdf
+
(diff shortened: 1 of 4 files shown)


alebrigant:
I would like to get started on a jupyter notebook on information geometry, but for some reason I cannot 'git add' notebooks. I get the following error:
```
The following paths are ignored by one of your .gitignore files:
notebooks/06_information_geometry.ipynb
Use -f if you really want to add them.
```
Is this deliberate or is this a problem on my end ?

ninamiolane:
You should be able to add notebooks with our settings @alebrigant 

Can you maybe check your `.gitignore` file? If there is the line `notebooks/06_information_geometry.ipynb` written there, it can be removed (we don't have it in the geomstats' master's .gitignore).

geomstats/geometry/beta_distributions.py
```diff
@@ -104,6 +104,47 @@ def sample(self, point, n_samples=1):
...
+        Returns
+        -------
+        pdf : function
+            Probability density function of the beta distribution with
+            parameters provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        point = gs.to_ndarray(point, to_ndim=2)
+        a_params = point[:, 0]
+        b_params = point[:, 1]
```
nguigs, line 127: The ellipsis notation would work here and avoid using to_ndarray:
```python
a_params = point[..., 0]
b_params = point[..., 1]
```

geomstats/geometry/beta_distributions.py
```diff
@@ -104,6 +104,47 @@ def sample(self, point, n_samples=1):
...
+            x : array-like, shape=[n_points,]
+                Points at which to compute the probability density function.
+            """
+            x = gs.array(x, gs.float32)
+            x = gs.to_ndarray(x, to_ndim=1)
+
+            pdf_at_x = [
+                gs.array(beta.pdf(x, a=a, b=b)) for a, b
+                in zip(a_params, b_params)]
+            pdf_at_x = gs.stack(pdf_at_x, axis=1)
```
nguigs, line 143: Using `axis=-1` works if there is only a single x or if there are many, this avoids using to_ndarray above

geomstats/geometry/normal_distributions.py
```diff
@@ -71,6 +71,47 @@ def sample(self, point, n_samples=1):
...
+            x : array-like, shape=[n_points,]
+                Points at which to compute the probability density function.
+            """
+            x = gs.array(x, gs.float32)
+            x = gs.to_ndarray(x, to_ndim=1)
+
+            pdf_at_x = [
+                gs.array(norm.pdf(x, loc=mean, scale=std)) for mean, std
+                in zip(means, stds)]
+            pdf_at_x = gs.stack(pdf_at_x, axis=1)
```
nguigs, line 110: ditto

geomstats/geometry/normal_distributions.py
```diff
@@ -71,6 +71,47 @@ def sample(self, point, n_samples=1):
...
+        Returns
+        -------
+        pdf : function
+            Probability density function of the normal distribution with
+            parameters provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        point = gs.to_ndarray(point, to_ndim=2)
+        means = point[:, 0]
+        stds = point[:, 1]
```
nguigs, line 94: ditto

nguigs (commented):
Nice! Just a comment on handling vectorization.

alebrigant:
Thanks for the comments @nguigs !

Issue #861, opened 2020-09-25
Front Page Tangent PCA example fails using Pytorch Backend: type error tensor.float64


cshewmake2:
1. Set up a new environment with python 3.7
2. Activate the environment
3. Pip install geomstats, torch, torchvision
4. Run export GEOMSTATS_BACKEND=pytorch
4. Run the following code from the README
```
from geomstats.geometry.special_orthogonal import SpecialOrthogonal
from geomstats.learning.pca import TangentPCA

so3 = SpecialOrthogonal(n=3, point_type='vector')
metric = so3.bi_invariant_metric

data = so3.random_uniform(n_samples=10)

tpca = TangentPCA(metric=metric, n_components=2)
tpca = tpca.fit(data)
tangent_projected_data = tpca.transform(data)
```

I'm getting a type error which seems to be related to a numpy call either in our backend or sklearn's. 
Either mac or linux, the issue persists.
`TypeError: Cannot interpret 'torch.float64' as a data type`

PR #864, opened 2020-09-29
Fix tangent PCA with pytorch

nguigs:
Address #861 

Use native SVD of each back-end
remove check array from sklearn because it returns a numpy array, whichever backend
Works with pytorch but not tensorflow because of another sklearn function.


ninamiolane:
Nice! Just a flake8 error. @cshewmake2 this is good news!

ninamiolane (approved):

cshewmake2:
Works for me! Great job. Let's get it into master and onward to pypi. 👍

cshewmake2 (approved):

Issue #865, opened 2020-09-30
Vectorization problem with Exp and Log maps of beta distributions


alebrigant:
The ```exp``` (```log``` resp.) map of beta distributions does not handle the case where the number of base points does not match the number of tangent vectors (end points resp.), which causes the following code
```
base_point = gs.array([1., 1.])
tangent_vecs = gs.array([[0., 0.], [0.5, 0.5]])
geod = beta.metric.exp(tangent_vecs, base_point)
t = gs.linspace(0., 1., 10)
geod(t)
```
to yield the following output
```
gs.array([1., 1.])
```

PR #866, opened 2020-09-30
Fix vectorization of Exp and Log of beta distributions

alebrigant:
This answers [issue#865](https://github.com/geomstats/geomstats/issues/865).

### DIFF
--- a/geomstats/geometry/beta_distributions.py
+++ b/geomstats/geometry/beta_distributions.py
@@ -294,10 +294,15 @@ class BetaMetric(RiemannianMetric):
 
         n_base_points = base_point.shape[0]
         n_tangent_vecs = tangent_vec.shape[0]
-        if n_base_points < n_tangent_vecs:
-            base_point = gs.tile(base_point, (n_tangent_vecs, 1))
-        elif n_base_points > n_tangent_vecs:
-            tangent_vec = gs.tile(tangent_vec, (n_base_points, 1))
+        if n_base_points > n_tangent_vecs:
+            raise ValueError('There cannot be more base points than tangent '
+                             'vectors.')
+        elif n_tangent_vecs > n_base_points:
+            if n_base_points > 1:
+                raise ValueError('For several tangent vectors, specify either '
+                                 'one or the same number of base points.')
+            else:
+                base_point = gs.tile(base_point, (n_tangent_vecs, 1))
 
         def ivp(state, _):
             """Reformat the initial value problem geodesic ODE."""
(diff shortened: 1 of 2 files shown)


geomstats/geometry/beta_distributions.py
```diff
@@ -292,6 +292,13 @@ def exp(self, tangent_vec, base_point, n_steps=N_STEPS):
         base_point = gs.to_ndarray(base_point, to_ndim=2)
         tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=2)
 
+        n_base_points = base_point.shape[0]
+        n_tangent_vecs = tangent_vec.shape[0]
+        if n_base_points < n_tangent_vecs:
+            base_point = gs.tile(base_point, (n_tangent_vecs, 1))
+        elif n_base_points > n_tangent_vecs:
+            tangent_vec = gs.tile(tangent_vec, (n_base_points, 1))
```
nguigs, line 300: This case is not tested. Does it make sense though, as usually one vector is not tangent at two different points, except in this case of an open set in R^n...?

geomstats/geometry/beta_distributions.py
```diff
@@ -292,6 +292,13 @@ def exp(self, tangent_vec, base_point, n_steps=N_STEPS):
         base_point = gs.to_ndarray(base_point, to_ndim=2)
         tangent_vec = gs.to_ndarray(tangent_vec, to_ndim=2)
 
+        n_base_points = base_point.shape[0]
+        n_tangent_vecs = tangent_vec.shape[0]
+        if n_base_points < n_tangent_vecs:
```
nguigs, line 297: Maybe the condition should be more strict: if n_base_points and n_tangent_vecs differ, one of them should equal 1, otherwise we may run into trouble so I think it should raise an error.

geomstats/geometry/beta_distributions.py
```diff
@@ -333,6 +340,12 @@ def log(self, point, base_point, n_steps=N_STEPS):
         t = gs.linspace(0, stop_time, n_steps)
         point = gs.to_ndarray(point, to_ndim=2)
         base_point = gs.to_ndarray(base_point, to_ndim=2)
+        n_points = point.shape[0]
+        n_base_points = base_point.shape[0]
+        if n_points < n_base_points:
```
nguigs, line 345: idem

nguigs (commented):
Thanks for spotting this! Just two remarks, let me know what you think!

geomstats/geometry/beta_distributions.py
alebrigant: I hadn't thought of it but you're right, it doesn't make much sense to consider one vector and several base points. Maybe just remove this possibility ? Then just allow ```n_base_points = n_tangent_vecs``` and ```n_tangent_vecs > n_base_points = 1```.

geomstats/geometry/beta_distributions.py
alebrigant: I agree.

geomstats/geometry/beta_distributions.py
nguigs: Yes, let's remove this possibility and add a check that `n_base_points == n_tangent_vecs or n_tangent_vecs > n_base_points == 1`

PR #868, opened 2020-10-02
Information geometry notebook

alebrigant:
First draft for the notebook on information geometry. It lacks references, and maybe a real data example.


ninamiolane (approved):
Very nice! Any idea why it fails with pytorch or tf? If it can't be solved quickly, maybe we should add a disclaimer at the beginning of the notebook to tell users to only try it with numpy? Otherwise, I'll let you merge!

alebrigant:
It's linked to the ```BetaDistributions```class and its ```exp``` and ```log``` methods, which seem to only work for numpy. It could maybe be fixed, I haven't really tried. Did you try already @nguigs ?

I added an example on a real dataset at the end, and the corresponding dataset.

ninamiolane:
Great! Maybe also add a test for the new dataset: https://github.com/geomstats/geomstats/blob/master/tests/test_datasets.py

alebrigant:
Noted ! The ```pandas``` package I'm using to read an Excel file is not installed with geomstats I suppose ? Do you think it could be interesting to include it ? It will probably become useful if we add more datasets.

ninamiolane:
Yes, having `pandas` would be great, let's add it in the requirements? However, could you maybe convert the `.xls` into `.csv`? It would allow viewing the `.csv` directly in github (not possible with .xls if i remember correctly!).

nguigs:
> It's linked to the `BetaDistributions` class and its `exp` and `log` methods, which seem to only work for numpy. It could maybe be fixed, I haven't really tried. Did you try already @nguigs ?
I did try, but the polygamma function derivatives are not implemented for Pytorch and Tensorflow, so the only way to make it work would be to wrap scipy functions and gradients like autograd does..

PR #876, opened 2020-10-20
add kendall pre-shape space

emaignant:
wip


geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+        -------
+        is_tangent : bool
+            Boolean denoting if tangent vector is horizontal.
+        """
+        product = gs.matmul(tangent_vec, Matrices.transpose(base_point))
+        is_tangent = self.is_tangent(tangent_vec, base_point, atol)
+        is_symmetric = Matrices.is_symmetric(product, atol)
+        return gs.logical_and(is_tangent, is_symmetric)
+
+    def realign(self, point, base_point):
```
ninamiolane, line 280: NIT: This could just be called `align` or `register`.

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+            Point on the manifold.
+        base_point : array-like, shape=[..., m, k]
+            Point on the manifold.
+
+        Returns
+        -------
+        aligned : array-like, shape=[..., m, k]
+            R.point.
+        """
+        M = gs.matmul(point, Matrices.transpose(base_point))
```
ninamiolane, line 298: Let's try to avoid uppercase variable names when possible: uppercase letters are reserved to constant variables in python.
Maybe:
M -> mat
U, S, Vh -> mat_u, mat_s, mat_vh

https://www.python.org/dev/peps/pep-0008/#constants

ninamiolane: (single letter variables are also a pain when we try to do global renaming, as trying to rename all the "M"'s automatically is not always possible! - can do it with vscode, but not everybody uses it).

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+        U, S, Vh = gs.linalg.svd(M)
+        if gs.any(gs.isclose(S[..., -1] + S[..., -2], 0.)):
+            logging.warning("Alignement matrix is not unique.")
```
ninamiolane, line 301: NIT: Alignment

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+        det = gs.linalg.det(M)
+        if gs.any(det < 0):
+            ones = gs.ones(self.m_ambient)
+            changer = gs.concatenate([ones[:-1], gs.array([-1.])], axis=0)
```
ninamiolane, line 305: changer? maybe `to_change`?

tests/test_pre_shape.py
```diff
@@ -0,0 +1,239 @@
...
+        proj = self.space.projection(point)
+        result = self.space.belongs(proj)
+        expected = True
+
+        self.assertAllClose(expected, result)
+
+    def test_is_centered(self):
+        point = gs.ones((self.m_ambient, self.k_landmarks))
+        result = self.space.is_centered(point)
+        self.assertTrue(~ result)
```
ninamiolane, line 58: NIT: you can also use `assertFalse`

tests/test_pre_shape.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertTrue(result)
+
+        vec = gs.array([tangent_vec, vector])
+        result = self.space.is_tangent(vec, point)
+        expected = gs.array([True, False])
+        self.assertAllClose(result, expected)
+
+    @geomstats.tests.np_and_pytorch_only
+    def test_vertical_projection(self):
+        w = gs.random.rand(self.m_ambient, self.k_landmarks)
```
ninamiolane, line 94: when possible, avoid single-letter variable (hard to automatically rename them afterwards in many code editors)

geomstats/_backend/pytorch/linalg.py
```diff
@@ -59,6 +59,13 @@ def norm(x, ord=2, axis=None):
     return torch.norm(x, p=ord, dim=axis)
 
 
+def solve_sylvester(a, b, q):
+    solution = np.vectorize(
+        scipy.linalg.solve_sylvester,
+        signature='(m,m),(n,n),(m,n)->(m,n)')(a, b, q)
+    return torch.from_numpy(solution)
```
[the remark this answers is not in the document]
ninamiolane: Otherwise, looking at the code here:
https://github.com/scipy/scipy/blob/d286f8525c16b2cd4e179dea2c77b6b09622aff9/scipy/linalg/_solvers.py#L30-L105

Could give ideas on how to re-implement in pytorch.

tests/test_pre_shape.py
```diff
@@ -0,0 +1,239 @@
...
+        tan = self.space.to_tangent(w, point)
+        horizontal = self.space.horizontal_projection(tan, point)
+        transposed_point = Matrices.transpose(point)
+        result = gs.matmul(horizontal, transposed_point)
+        expected = Matrices.transpose(result)
+
+        self.assertAllClose(result, expected)
+
+    @geomstats.tests.np_and_pytorch_only
+    def test_horizontal_and_is_horizontal(self):
```
ninamiolane, line 147: `test_horizontal_and_is_tangent`
-> let's use the name of the function being tested, here it is `is_tangent`

tests/test_pre_shape.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertTrue(gs.all(result))
+
+        base_point = self.space.random_uniform(2)
+        point = self.space.random_uniform()
+        aligned = self.space.realign(point, base_point)
+        alignment = gs.matmul(aligned, Matrices.transpose(base_point))
+        result = Matrices.is_symmetric(alignment)
+        self.assertTrue(gs.all(result))
+
+    def test_inner_product(self):
```
ninamiolane, line 181: This only tests the shape of the inner_product. Maybe call it `test_inner_product_shape` and have another test that tests the result?

tests/test_pre_shape.py
```diff
@@ -0,0 +1,239 @@
...
+        result = self.space.belongs(exp)
+        self.assertTrue(gs.all(result))
+
+        point = point[0]
+        tan = self.space.to_tangent(w, point)
+        exp = self.space.metric.exp(tan, point)
+        result = self.space.belongs(exp)
+        self.assertTrue(gs.all(result))
+
+    def test_log(self):
```
ninamiolane, line 214: This is `test_log_and_exp`

ninamiolane: Could we also add `test_exp_and_log`, i.e. first the Exponential and then the Logarithm?

ninamiolane (approved):
Awesome! 🎉  I'm sure this implementation will be useful to many. I have added several comments, but they are mostly minor. Otherwise it looks good to go!

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+        -------
+        projected_point : array-like, shape=[..., m_ambient, k_landmarks]
+            Point projected on the pre-shape space.
+        """
+        centered_point = self.center(point)
+        frob_norm = self.embedding_metric.norm(centered_point)
+        projected_point = gs.einsum(
+            '...,...ij->...ij', 1. / frob_norm, centered_point)
+
+        return projected_point
```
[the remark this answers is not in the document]
nguigs: but it's not the same number of indices as we're dealing with matrices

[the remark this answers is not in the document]
ninamiolane: Yes... maybe we should think about how to have some code that works for Euclidean space and Matrices space, but not for this PR.

geomstats/_backend/pytorch/linalg.py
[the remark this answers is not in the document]
nguigs: In both cases it calls Fortran code. This might be an issue to look at with the ADT

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+    def belongs(self, point, atol=TOLERANCE):
+        """Test if a point belongs to the pre-shape space.
+
+        This tests whether the point is centered and whether the point's
+        Frobenius norm is 1.
+
+        Parameters
+        ----------
+        atol
+        point : array-like, shape=[..., m_ambient, k_landmarks]
```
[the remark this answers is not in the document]
nguigs: Yes in product manifolds and landmarks spaces, it is `[..., k_landmarks, m_ambient]` but @emaignant wanted to be consistent with Kendall's convention and the representation of vectors as columns...

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+        Returns
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean evaluating if point belongs to the pre-shape space.
+        """
+        frob_norm = self.embedding_metric.norm(point)
+        diff = gs.abs(frob_norm - 1)
+        is_centered = self.is_centered(point, atol)
+        return gs.logical_and(
+            gs.less_equal(diff, atol), is_centered)
```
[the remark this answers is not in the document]
nguigs: We could inherit from Hypersphere, but we would always have to flatten then reshape arguments, so there will be much more duplicate code for reshaping, and the core computations will be hidden from the core of the class, which does not help readibility.

geomstats/geometry/pre_shape.py
```diff
@@ -0,0 +1,405 @@
...
+from geomstats.errors import check_tf_error
+from geomstats.geometry.embedded_manifold import EmbeddedManifold
+from geomstats.geometry.hypersphere import Hypersphere
+from geomstats.geometry.matrices import Matrices, MatricesMetric
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+
+TOLERANCE = 1e-6
+
+
+class PreShapeSpace(EmbeddedManifold):
```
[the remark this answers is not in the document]
nguigs: @emaignant

geomstats/geometry/pre_shape.py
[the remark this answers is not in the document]
ninamiolane: Yes I understand (also saw the `k`, `m` naming to be consistent with Kendall's convention) , but I still think we should be consistent in our codebase, where dimension of the space comes at the end :/ We could add remarks in several docstring to point out that we change the Kendall convention from `m_ambient, k_landmarks` to `k_landmarks, m_ambient` so that the users are fully aware?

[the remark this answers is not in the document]
nguigs: Ok I made the change and added a remark in the main docstring

ninamiolane:
Should we merge?

tests/test_pre_shape.py
```diff
@@ -18,6 +18,16 @@ def setUp(self):
         self.shape_metric = KendallShapeMetric(
             self.k_landmarks, self.m_ambient)
 
+    def test_bleongs(self):
+        point = gs.random.rand(self.m_ambient - 1, self.k_landmarks)
+        result = self.space.belongs(point)
```
ythanwerdas, line 23: test_belongs is written test_bleongs

ythanwerdas: Great work!

PR #881, opened 2020-11-10
Template for lddmm kernel metric

nguigs:
WIP
This is to help creating a kernel metric on the space of landmarks in R^d. Some work remains to adapt the existing integration modules to the hamiltonian equations.


ninamiolane:
Nice! I know this is WIP, but we could also add the (template) unit-tests to be passed with it?

PR #882, opened 2020-11-11
Harmonize integration schemes

nguigs:
These can be used in three different settings:
- Classic geodesic equation with Christoffel symbols
- Geodesic equation on a Lie group, where the acceleration is computed in the Lie algebra
- Hamiltonian systems (yet to come in geomstats #881)


ninamiolane (approved):
Very nice! Good to go as soon as the tests pass. If you have time, maybe add slightly more documentation, equations and/or references.

PR #883, opened 2020-11-11
Template for lddmm metric on landmarks

nguigs:
WIP (continuation of #881)
This is to help creating a kernel metric on the space of landmarks in R^d. Some work remains to adapt the existing integration modules to the hamiltonian equations.

### DIFF
--- a/geomstats/geometry/landmarks.py
+++ b/geomstats/geometry/landmarks.py
@@ -3,7 +3,7 @@
 import geomstats.backend as gs
 from geomstats.geometry.product_manifold import ProductManifold
 from geomstats.geometry.product_riemannian_metric import \
-    ProductRiemannianMetric
+    ProductRiemannianMetric, RiemannianMetric
 
 
 class Landmarks(ProductManifold):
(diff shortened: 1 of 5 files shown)


nguigs:
Comments from @philipp-harms:

* We tried the code on a couple of corpus callosum shapes and could not verify that \exp(\log(x))=x. This might be due to convergence issues

* There are some naming inconsistencies between inverse metrics and cometrics, and between momenta and covelocities etc. I opt for the terminology vector-covector and metric-cometric with one exception: The term momentum should be reserved for the covector excluding the volume form. The distinction between covectors and momenta is relevant in LDDMM for images, densities, curves, surfaces, etc but disappears in the landmark case, where the volume form is just the counting measure on the set 1, ..., n.

* The code for solving ODEs should be kept general. Right now it only works for vector fields on TM rather than general manifolds M. Specifically, to change this, one would have to modify the function integrate() such that states are not split into positions and velocities.

* It would be nice to have a good application for the LDDMM code. The corpus callosum data which we used in the exercise class are ill-suited because they should be treated as curves rather than pseudo-landmarks.

ninamiolane:
@all-contributors please add @nguigs  for tests and code

Issue #879, opened 2020-11-04
Implement Lie algebras


nguigs:
Classes should be created for lie algebras of SE(n) and SO(2), SO(3) in vector representation

ninamiolane:
Or just a field self.lie_algebra that refers to the corresponding SkewSymmetric() object (for SO(n) for example)

PR #884, opened 2020-11-16
Add lie algebra for matrix SE(n)

nguigs:
This addresses #879 and will allow to use #878 on the Special Euclidean group

### DIFF
--- a/geomstats/geometry/lie_algebra.py
+++ b/geomstats/geometry/lie_algebra.py
@@ -8,7 +8,6 @@ in that base. This base will be provided in child classes
 """
 import geomstats.backend as gs
 import geomstats.errors
-from geomstats.geometry.matrices import Matrices
 from ._bch_coefficients import BCH_COEFFICIENTS
 
 
@@ -135,54 +134,6 @@ class MatrixLieAlgebra:
             raise NotImplementedError("basis not implemented")
 
         return gs.einsum("...i,ijk ->...jk", basis_representation, self.basis)
-
-    def reshape_metric_matrix(self, metric_matrix):
-        """Reshape diagonal metric matrix to a symmetric matrix of size n.
-
-        Reshape a diagonal metric matrix of size `dim x dim` into a symmetric
-        matrix of size `n x n` where :math: `dim= n (n -1) / 2` is the
-        dimension of the space of skew symmetric matrices. The
-        non-diagonal coefficients in the output matrix correspond to the
-        basis matrices of this space. The diagonal is filled with ones.
-        This useful to compute a matrix inner product.
-
-        Parameters
-        ----------
-        metric_matrix : array-like, shape=[dim, dim]
-            Diagonal metric matrix.
-
-        Returns
-        -------
-        symmetric_matrix : array-like, shape=[n, n]
-            Symmetric matrix.
-        """
-        if Matrices.is_diagonal(metric_matrix):
-            metric_coeffs = gs.diagonal(metric_matrix)
-            metric_mat = gs.abs(
-                self.matrix_representation(metric_coeffs))
-            return metric_mat
-        raise ValueError('This is only possible for a diagonal matrix')
-
-    def orthonormal_basis(self, metric_matrix):
-        """Orthonormalize the basis with respect to the given metric.
-
-        This corresponds to a renormalization.
-
-        Parameters
-        ----------
-        metric_matrix : array-like, shape=[dim, dim]
-            Matrix of a metric.
-
-        Returns
-        -------
-        basis : array-like, shape=[dim, n, n]
-            Orthonormal basis.
-        """
-        metric_matrix = self.reshape_metric_matrix(metric_matrix)
-        norms = gs.sum(
-            metric_matrix * self.basis * self.basis, (-2, -1))
-
-        return gs.einsum('i, ikl->ikl', 1. / gs.sqrt(norms), self.basis)
 
     def projection(self, mat):
         """Project a matrix to the Lie Algebra.
--- a/geomstats/geometry/lie_algebra.py
+++ b/geomstats/geometry/lie_algebra.py
@@ -178,9 +178,11 @@ class MatrixLieAlgebra:
         basis : array-like, shape=[dim, n, n]
             Orthonormal basis.
         """
-        metric_matrix = self.reshape_metric_matrix(metric_matrix) + gs.eye(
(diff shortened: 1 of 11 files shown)


geomstats/geometry/skew_symmetric_matrices.py
```diff
@@ -52,6 +52,22 @@ def belongs(self, mat, atol=TOLERANCE):
         """
         return Matrices(self.n, self.n).is_skew_symmetric(mat=mat, atol=atol)
 
+    @staticmethod
+    def projection(mat):
+        """Make matrix skew-symmetric, by averaging with minus its transpose.
```
ninamiolane, line 57: NIT: Maybe here an equation would be clearer than the sentence in English

geomstats/geometry/lie_algebra.py
```diff
@@ -135,17 +136,66 @@ def matrix_representation(self, basis_representation):
 
         return gs.einsum("...i,ijk ->...jk", basis_representation, self.basis)
 
+    def reshape_metric_matrix(self, metric_matrix):
```
ninamiolane, line 139: Would this be better as a function within the module `matrices.py` ?

geomstats/geometry/lie_algebra.py
```diff
@@ -135,17 +136,66 @@ def matrix_representation(self, basis_representation):
...
+            Symmetric matrix.
+        """
+        if Matrices.is_diagonal(metric_matrix):
+            metric_coeffs = gs.diagonal(metric_matrix)
+            metric_mat = gs.abs(
+                self.matrix_representation(metric_coeffs))
+            return metric_mat
+        raise ValueError('This is only possible for a diagonal matrix')
+
     def orthonormal_basis(self, metric_matrix):
```
ninamiolane, line 166: would this be better within the module `riemannian_metric.py`? Having metric-related functions within the Lie algera module could be confusing?

geomstats/geometry/special_euclidean.py
```diff
@@ -9,7 +9,9 @@
 from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.general_linear import GeneralLinear
 from geomstats.geometry.invariant_metric import InvariantMetric
+from geomstats.geometry.lie_algebra import MatrixLieAlgebra
```
ninamiolane, line 12: Should we distinguish MatrixLieGroup and LieGroup then?

geomstats/geometry/special_euclidean.py
```diff
@@ -36,13 +38,46 @@
                         - 1. / 362880.]
 
 
+def embed(rotation, translation, output_shape, constant=1.):
+    r"""Embed rotation, translation couples into n+1 square matrices.
```
ninamiolane, line 42: I think this is called the homogeneous matrix representation. Maybe rename the function and add a link?

geomstats/geometry/special_euclidean.py
```diff
@@ -36,13 +38,46 @@
...
+    output_shape : tuple of int
+        Desired output shape. This is need for vectorization.
+    constant : float
+        Constant to use at the last line and column of the square matrix.
+        Optional, default: 1
+
+    Returns
+    -------
+    mat: array-like, shape=[..., n + 1, n + 1]
+        Squre Matrix of size n + 1. It can represent an element of the
```
ninamiolane, line 62: Squre -> Square

ninamiolane: Not sure I understand this docstring? special euclidean group of the Lie algebra?

tests/test_skew_symmetric_matrices.py
```diff
@@ -61,7 +61,7 @@ def test_bch_up_to_fourth_order_works(self):
             )
             self.assertAllClose(expected, result)
 
-    @geomstats.tests.np_only
+    @geomstats.tests.np_and_pytorch_only
```
ninamiolane, line 64: 🎉 always exciting!

ninamiolane (approved):
Very nice!

ninamiolane:
On a high-level, in terms of the structure and inheritance of classes:

- Since there is now MatrixLieAlgebra, do we want to have MatrixLieGroup? It could help both for consistency in our naming, but also acknowledging that all of our groups are matrix lie groups so far.

- Should we factorize Euclidean and Matrices into the same (more abstract) class, as in `pymanopt`:

Screen Shot 2020-11-17 at 9 07 00 AM

geomstats/geometry/lie_algebra.py
nguigs: It needs access to `self.matrix_representation` and `self.basis`

nguigs: So both should go there? But the MatrixLieAlgbera class doesn't inherit the Matrices class

geomstats/geometry/special_euclidean.py
nguigs: I think LieGroup with the point_type argument works ok.

geomstats/geometry/special_euclidean.py
nguigs: True

geomstats/geometry/special_euclidean.py
nguigs: Nit: it can represent an element of SE(3) or its Lie aglebra

nguigs:
>     * Since there is now MatrixLieAlgebra, do we want to have MatrixLieGroup? It could help both for consistency in our naming, but also acknowledging that all of our groups are matrix lie groups so far.

MatrixLieAlgebra was here before (since the first hackathon).

>     * Should we factorize Euclidean and Matrices into the same (more abstract) class, as in `pymanopt`:

This makes sense, however, the point of the Matrices class was to gather all the current matrix operations in a well vectorized way, more as an extention of the backend than as mathematical space. In Euclidean we won't have the transpose, mul, to_symmetric etc... which are the useful methods of the current Matrices class

geomstats/geometry/lie_algebra.py
nguigs: Yes, but we need a basis!

ninamiolane: The basis could be passed in argument?

nguigs: Yes makes sense actually, and to also move reshape metric matrix

PR #885, opened 2020-11-17
Create dirichlet distributions

alebrigant:
Generalizing code from ```beta_distributions``` to arbitrary dimensions. It is quite straightforward thanks to the clean code of @nguigs !

### DIFF
--- a/geomstats/geometry/dirichlet_distributions.py
+++ b/geomstats/geometry/dirichlet_distributions.py
@@ -113,7 +113,9 @@ class DirichletMetric(RiemannianMetric):
         super(DirichletMetric, self).__init__(dim=dim)
 
     def metric_matrix(self, base_point=None):
-        """Compute inner-product matrix of the Fisher information metric
+        """Compute the inner-product matrix.
+
+        Compute the inner-product matrix of the Fisher information metric
         at the tangent space at base point.
 
         Parameters
--- a/geomstats/geometry/dirichlet_distributions.py
+++ b/geomstats/geometry/dirichlet_distributions.py
@@ -12,14 +12,13 @@ from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
 N_STEPS = 100
-EPSILON = 1e-6
 
 
 class DirichletDistributions(EmbeddedManifold):
     """Class for the manifold of Dirichlet distributions.
 
-    This is :math: Dirichlet = `(R_+^*)^n`, the positive quadrant of the
-    n-dimensional Euclidean space.
+    This is :math: Dirichlet = `(R_+^*)^dim`, the positive quadrant of the
+    dim-dimensional Euclidean space.
 
     Attributes
     ----------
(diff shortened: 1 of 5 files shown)


geomstats/geometry/dirichlet_distributions.py
```diff
@@ -0,0 +1,322 @@
...
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
+        n_points = base_point.shape[0]
+
+        def coefficients(ind_k):
+            param_k = base_point[..., ind_k]
+            param_sum = gs.sum(base_point, -1)
+            C1 = 1 / gs.polygamma(1, param_k) / (
+                1 / gs.polygamma(1, param_sum)
+                - gs.sum(1 / gs.polygamma(1, base_point), -1))
+            C2 = - C1 * gs.polygamma(2, param_sum) / gs.polygamma(1, param_sum)
```
nguigs, line 170: Use lower-case for local variable names

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -0,0 +1,322 @@
...
+
+            mat_ones = gs.ones((n_points, self.dim, self.dim))
+            mat_diag = gs.zeros((n_points, self.dim, self.dim))
+            diag = gs.einsum('ijj->ij', mat_diag)
+            diag[:] = - gs.polygamma(2, base_point) \
+                / gs.polygamma(1, base_point)
```
nguigs, line 176: Are mat_diag and diag used?

nguigs: If the point is to modify `mat_diag` by filling `diag`, you could also use `geomstats.algebra_utils.from_vector_to_diagonal_matrix(diag_vector)` instead.

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -0,0 +1,322 @@
...
+
+            mat_k = gs.einsum('i,ijk->ijk', C2, mat_ones)\
+                - gs.einsum('i,ijk->ijk', C1, mat_diag)
+            mat_k[:, ind_k, ind_k] += gs.polygamma(2, param_k)\
+                / gs.polygamma(1, param_k)
+            return 1 / 2 * mat_k
+
+        christoffels = gs.zeros((n_points, self.dim, self.dim, self.dim))
+        for ind_k in range(self.dim):
+            christoffels[:, ind_k, :, :] = coefficients(ind_k)
```
nguigs, line 186: In case this could work for tensorflow in the future, it would be nice to avoid assignment already: for example construct `christoffels` as a list by appending the new entry in the loop, then stack along the second dimension.

nguigs (approved):
Great! Just a few typos (beta -> Dirichlet) and a comment on filling diagonal matrices. More broadly, shouldn't this class replace the beta class as it is more general and coincides exactly for dim=2?

geomstats/geometry/dirichlet_distributions.py
alebrigant: Did not know this function existed, nice !

alebrigant:
> Great! Just a few typos (beta -> Dirichlet) and a comment on filling diagonal matrices. More broadly, shouldn't this class replace the beta class as it is more general and coincides exactly for dim=2?

Maybe keep a ```BetaDistributions``` class, since someone who does not know the Dirichlet distribution might think it is missing, but make it point to ```DirichletDistributions(dim=2)```?

alebrigant:
Good to merge ? @nguigs @ninamiolane

ninamiolane (approved):
Ready to merge! Maybe just remove the 2 leftover comments? (see my comments)

PR #875, opened 2020-10-19
Invariant Metrics

nguigs:
Compute Riemann curvature on lie groups with an invariant metric:
- renamed inner_product_matrix to metric_matrix, because it's shorter
- Add invariant metrics represented by a diagonal matrix (#702 )
This is currently tested on SO(n), but will be tested on SE(n) in a next PR: need to create its Lie Algebra first

PR #886, opened 2020-11-17
Right invariant metric on Lie Groups

nguigs:
This continues #875 to include right-invariant metrics on Lie groups. This will be tested with #884.

### DIFF
--- a/geomstats/geometry/invariant_metric.py
+++ b/geomstats/geometry/invariant_metric.py
@@ -549,7 +549,7 @@ class _InvariantMetricMatrix(RiemannianMetric):
         sign = 1. if self.left_or_right == 'left' else 1.
 
         def lie_acceleration(point, vector):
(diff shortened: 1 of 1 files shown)


geomstats/geometry/invariant_metric.py
```diff
@@ -543,16 +544,19 @@ def exp(self, tangent_vec, base_point, n_steps=10, step='rk4',
         """
         group = self.group
         basis = self.lie_algebra.orthonormal_basis(self.metric_mat_at_identity)
+        sign = 1. if self.left_or_right == 'left' else 1.
 
         def lie_acceleration(point, vector):
-            velocity = self.group.compose(point, vector)
+            velocity = self.group.tangent_translation_map(
```
ninamiolane, line 550: Docstring for this function? Even if it is a private function, an equation could help the user.

ninamiolane (approved):
Awesome!

PR #888, opened 2020-11-20
Visualisation on SE(2) and Canonical left-invariant metric

nguigs:
This PR continues #878, #884 and #886. It
- adds a module for visualisation of SE(2) in the plane
- adds a class for the left-invariant metric on matrix SE(n). It is a product metric with closed-form formulas. It is checked that they coincide with the vector implementations.
- adds an example where the geodesics with the different structures are plotted.
- Harmonize 2 and 3 dimensional bases of the space of skew symmetric matrices with conventions

### DIFF
--- a/examples/plot_geodesics_se2.py
+++ b/examples/plot_geodesics_se2.py
@@ -14,6 +14,7 @@ from geomstats.geometry.special_euclidean import SpecialEuclidean
 SE2_GROUP = SpecialEuclidean(n=2, point_type='matrix')
 N_STEPS = 40
 METRIC = SE2_GROUP.left_canonical_metric
+SE2_VEC = SpecialEuclidean(n=2, point_type='vector')
 
 
 def main():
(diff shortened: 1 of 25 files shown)


examples/plot_geodesics_se2.py
```diff
@@ -0,0 +1,50 @@
...
+from geomstats.geometry.special_euclidean import SpecialEuclidean
+
+SE2_GROUP = SpecialEuclidean(n=2, point_type='matrix')
+N_STEPS = 40
+LEFT_METRIC = SE2_GROUP.left_canonical_metric
+RIGHT_METRIC = SE2_GROUP.right_canonical_metric
+
+
+def main():
+    """Plot geodesics on SE2 with different structures."""
```
ninamiolane, line 23: NIT: SE(2)

tests/test_skew_symmetric_matrices.py
```diff
@@ -68,3 +68,12 @@ def test_basis_representation_is_correctly_vectorized(self):
             shape = gs.shape(skew.basis_representation(skew.basis))
             dim = int(n * (n - 1) / 2)
             self.assertEqual(shape, (dim, dim))
+
+    @geomstats.tests.np_and_pytorch_only
```
ninamiolane, line 72: what is the limitation for tf here?

ninamiolane (approved):
Awesome!

geomstats/visualization.py
```diff
@@ -20,7 +21,7 @@
 
 AX_SCALE = 1.2
 
-IMPLEMENTED = ['SO3_GROUP', 'SE3_GROUP', 'S1', 'S2',
+IMPLEMENTED = ['SO3_GROUP', 'SE3_GROUP', 'SE2_GROUP', 'S1', 'S2',
```
ninamiolane, line 24: (this `IMPLEMENTED` list is not very clean, it could probably be replaced by something else.)

tests/test_skew_symmetric_matrices.py
nguigs: The limitation is in `skew.basis_representation`: https://github.com/geomstats/geomstats/blob/f908c173064773a888dfc61e1510acf1e4546eae/geomstats/geometry/skew_symmetric_matrices.py#L94-L97
This doesn't work with tf because indexing by an array doesn't work like in numpy. The method tf.gather_nd could be used, but it would have to go in the backend..

nguigs: Something simpler and cleaner also works with pytorch and numpy but not tf:
````python
rows, cols = gs.triu_indices(self.n, k=1)
result_ = matrix_representation[..., rows, cols]

nguigs: I created a triu_to_vec function in the back-ends

PR #889, opened 2020-11-20
Yt spdbureswasserstein

ythanwerdas:
Finally! The Bures-Wasserstein metric on SPD matrices: inner_product (modified), exp, log, squared_dist.

PR #890, opened 2020-11-24
Add sylvester solve for tf in backend

nguigs:
In the particular case where A=B is a Symmetric Non-Negative Definite matrix, the solution to AX + XB = Q can be computed with an eigenvalue decomposition of A. It turns out that this is much faster than vectorizing the scipy solver.
This allows to extend all the methods in the Kendall shape spaces to run with tensorflow, and could be used in #889.


ninamiolane (approved):
Awesome! 🎉

PR #891, opened 2020-12-14
add abstract class for quotient spaces

nguigs:
This PR defines an abstract class for fiber bundles and quotient metrics. There are different scenarios depending on the available methods (whether the submersion is explicitly defined or not). This is tested in two different cases: 
- kendall shape space: the quotient space is abstract and computations are carried out in a section of the total space
- Bures Wasserstein metric on spd: everything is explicit
@ythanwerdas

### DIFF
--- a/geomstats/_backend/numpy/linalg.py
+++ b/geomstats/_backend/numpy/linalg.py
@@ -1,7 +1,9 @@
 """Numpy based linear algebra backend."""
 
 import autograd.numpy as np
+import autograd.scipy.linalg as asp
 import scipy.linalg
+from autograd.extend import defvjp, primitive
 from autograd.numpy.linalg import (  # NOQA
     cholesky,
     det,
@@ -33,18 +35,37 @@ def _expsym(x):
     return result
 
 
+@primitive
 def expm(x):
     ndim = x.ndim
     new_x = to_ndarray(x, to_ndim=3)
     if _is_symmetric(new_x):
         result = _expsym(new_x)
     else:
-        result = np.vectorize(scipy.linalg.expm,
+        result = np.vectorize(asp.expm,
                               signature='(n,m)->(n,m)')(new_x)
 
     if ndim == 2:
         return result[0]
     return result
+
+
+def _expm_vjp(_ans, x):
+    vectorized = x.ndim == 3
+    axes = (0, 2, 1) if vectorized else (1, 0)
+
+    def vjp(g):
+        n = x.shape[-1]
+        size_m = x.shape[:-2] + (2 * n, 2 * n)
+        mat = np.zeros(size_m)
+        mat[..., :n, :n] = x.transpose(axes)
+        mat[..., n:, n:] = x.transpose(axes)
+        mat[..., :n, n:] = g
+        return expm(mat)[..., :n, n:]
+    return vjp
+
+
+defvjp(expm, _expm_vjp)
 
 
 def logm(x):
--- a/geomstats/_backend/numpy/linalg.py
+++ b/geomstats/_backend/numpy/linalg.py
@@ -2,7 +2,6 @@
 
 import autograd.numpy as np
 import autograd.scipy.linalg as asp
-import scipy.linalg
 from autograd.extend import defvjp, primitive
 from autograd.numpy.linalg import (  # NOQA
     cholesky,
@@ -15,39 +14,14 @@ from autograd.numpy.linalg import (  # NOQA
     svd
 )
 
-from .common import to_ndarray
 
 _TOL = 1e-10
 
 
-def _is_symmetric(x, tol=_TOL):
-    new_x = to_ndarray(x, to_ndim=3)
-    return (np.abs(new_x - np.transpose(new_x, axes=(0, 2, 1))) < tol).all()
-
-
-def _expsym(x):
-    eigvals, eigvecs = np.linalg.eigh(x)
-    eigvals = np.exp(eigvals)
-    eigvals = np.vectorize(np.diag, signature='(n)->(n,n)')(eigvals)
-    transp_eigvecs = np.transpose(eigvecs, axes=(0, 2, 1))
-    result = np.matmul(eigvecs, eigvals)
-    result = np.matmul(result, transp_eigvecs)
-    return result
-
-
 @primitive
 def expm(x):
-    ndim = x.ndim
-    new_x = to_ndarray(x, to_ndim=3)
-    if _is_symmetric(new_x):
-        result = _expsym(new_x)
-    else:
-        result = np.vectorize(asp.expm,
-                              signature='(n,m)->(n,m)')(new_x)
-
-    if ndim == 2:
-        return result[0]
-    return result
+    return np.vectorize(
+        asp.expm, signature='(n,m)->(n,m)')(x)
 
 
 def _expm_vjp(_ans, x):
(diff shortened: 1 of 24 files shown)


ythanwerdas (approved):
Great great work! I think you managed to allow different ways of defining quotient spaces.

I suggest to implement a function `vector_lifts` which would provide all the lifts of a vector. It could be an isomorphism from the Lie algebra to the vertical space. It provides another way of computing the horizontal lift by composing with the horizontal projection.

geomstats/_backend/numpy/linalg.py
```diff
@@ -1,8 +1,11 @@
 """Numpy based linear algebra backend."""
 
 import autograd.numpy as np
+import autograd.scipy.linalg as asp
```
ninamiolane, line 4: Do we need scipy.linalg if it has been replaced by autograd's version?

geomstats/_backend/pytorch/linalg.py
```diff
@@ -20,12 +25,6 @@ def sqrtm(x):
...
 
 
-def expm(x):
-    np_expm = np.vectorize(
-        scipy.linalg.expm, signature='(n,m)->(n,m)')(x)
-    return torch.from_numpy(np_expm)
-
-
 def inv(*args, **kwargs):
     return torch.from_numpy(np.linalg.inv(*args, **kwargs))
```
ninamiolane, line 29: does `torch.inverse` work here?

geomstats/geometry/pre_shape.py
```diff
@@ -40,10 +41,12 @@ class PreShapeSpace(EmbeddedManifold):
     """
 
     def __init__(self, k_landmarks, m_ambient):
+        embedding_manifold = Matrices(k_landmarks, m_ambient)
         super(PreShapeSpace, self).__init__(
```
ninamiolane, line 45: Is the constructor of FiberBundle called here?

geomstats/geometry/pre_shape.py
```diff
@@ -426,10 +408,10 @@ def log(self, point, base_point):
         return log
 
 
-class KendallShapeMetric(ProcrustesMetric):
+class KendallShapeMetric(QuotientMetric):
     """Quotient metric on the shape space.
 
-    The Kendall shape space is obtained by taking the caution of the
+    The Kendall shape space is obtained by taking the quotient of the
```
ninamiolane, line 414: Nice edit! :D

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
+"""Classes for fiber bundles and quotient metrics."""
```
ninamiolane, line 1: 🎉 So good!!

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
...
+
```
ninamiolane, line 2: How about having 2 files: quotient.py and fiber_bundle.py?

For users who are browsing through the geomstats' geometry (e.g. directory `geometry`), they would more directly see that fiber bundle is available. Otherwise, it is a bit hidden.

tests/test_backends.py
```diff
@@ -1001,3 +1003,31 @@ def test_eigvalsh(self):
...
         expected = _np.linalg.eigvalsh(mat)
         self.assertAllCloseToNp(result, expected)
+
+    def test_cholesky(self):
+        mat = SPDMatrices(3).random_uniform(2)
+        result = gs.linalg.cholesky(mat)
+        expected = _np.linalg.cholesky(mat)
+        self.assertAllClose(result, expected)
+
+    @geomstats.tests.np_only
```
ninamiolane, line 1013: this test has torch code, what doesn't work with pytorch?

tests/test_quotient_manifold.py
```diff
@@ -0,0 +1,157 @@
+"""Unit tests for the quotient space."""
```
ninamiolane, line 1: 🎉

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
...
+        so the lift is the identity map.
+
+        Parameters
+        ----------
+        point: array-like, shape=[..., {dim, [n, n]}]
+            Point of the base manifold.
+
+        Returns
+        -------
+        lift: array-like, shape=[..., {ambient_dim, [n, n]}]
```
ninamiolane, line 122: NIT: maybe not use "lift", since there is no variable called lift here.

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
...
+
+        This is a section of the fiber bundle, defined on the base manifold,
+        with values in the total space. It means that submersion applied after
+        lift results in the identity map. By default, the base manifold
+        is not explicit but is identified with a section of the fiber bundle,
+        so the lift is the identity map.
+
+        Parameters
+        ----------
+        point: array-like, shape=[..., {dim, [n, n]}]
```
ninamiolane, line 117: nit: point : (space before :)

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
...
+        map.
+
+        Parameters
+        ----------
+        point: array-like, shape=[..., {ambient_dim, [n, n]}]
+            Point of the total space.
+
+        Returns
+        -------
+        projection: array-like, shape=[..., {dim, [n, n]}]
```
ninamiolane, line 100: NIT: maybe not use "projection" since no var is called projection here

geomstats/geometry/quotient_manifold.py
```diff
@@ -0,0 +1,480 @@
...
+
+        This is the projection of the fiber bundle, defined on the total
+        space, with values in the base manifold. This map is surjective.
+        By default, the base manifold  is not explicit but is identified with a
+        local section of the fiber bundle, so the submersion is the identity
+        map.
+
+        Parameters
+        ----------
+        point: array-like, shape=[..., {ambient_dim, [n, n]}]
```
ninamiolane, line 95: NIT: point :

tests/test_quotient_manifold.py
```diff
@@ -0,0 +1,157 @@
...
+            return 2 * GeneralLinear.to_symmetric(product)
+
+        def horizontal_lift(tangent_vec, point, base_point=None):
+            if base_point is None:
+                base_point = submersion(point)
+            sylvester = gs.linalg.solve_sylvester(
+                base_point, base_point, tangent_vec)
+            return GeneralLinear.mul(sylvester, point)
+
+        self.bundle.submersion = submersion
```
ninamiolane, line 40: It is not very natural for a user to modify an object's method with this syntax.

Could we add:
- either a method `define_submersion` or `set_submersion` that takes a callable as input and basically performs the operation self.submersion = callable,
- or indicate in the docstring of `submersion` that this method can be replaced by the user?

Same for tangent_submersion, horizontal_lift and lift

ninamiolane (commented):
Awesome additions!! 🎉  Sorry for the delay in the review

ninamiolane:
Travis needs to pass with pytorch and tf before we merge

geomstats/geometry/pre_shape.py
nguigs: `super(..).__init__` calls all the constructors of the parent classes

geomstats/geometry/quotient_manifold.py
nguigs: True

tests/test_backends.py
nguigs: It does pass with torch but the goal was actually to use torch as reference and test numpy implementation here.

geomstats/geometry/quotient_manifold.py
nguigs: don't we do this very often?

tests/test_quotient_manifold.py
nguigs: Yes I agree the user should be guided here:
- this might be a good solution if this class is to be instanciated as is, but I was thinking of it as an abstract class, and other child classes should be created which override these methods.
- in any case, either the docstring or a notebook will explain the different ways to use these classes: there is a minimum set of methods the user needs to explicitly define (e.g. align and vertical_projection as in Kendall)

geomstats/geometry/quotient_manifold.py
nguigs: Should I split the tests in two files as well? There would be redunduncies

geomstats/geometry/quotient_manifold.py
nguigs: same as above

geomstats/_backend/pytorch/linalg.py
nguigs: Yes! I was worried about numerical precision but it seems to be fine

geomstats/_backend/numpy/linalg.py
nguigs: Yes, I had troubles with the log but it works now

geomstats/geometry/quotient_manifold.py
ninamiolane: Yes, split into two would be good. For the redundant tests, maybe just put it in one of the tests? (if it is testing the same code anyway, and would not increase the code coverage)?

geomstats/geometry/quotient_manifold.py
ninamiolane: Yes we do do it... it's confusing though.

tests/test_quotient_manifold.py
ninamiolane: Got it, yes we can make it clear in the docstrings and notebooks.

Btw, we have not used the structure of Abstract Class (in Python) - using "raise NotImplementedError" instead - but we could have a look in the future:
https://www.tutorialspoint.com/abstract-base-classes-in-python-abc#:~:text=A%20class%20is%20called%20an,be%20implemented%20by%20its%20subclasses.

ninamiolane:
LGTM! Impressive contributions!! 🎉

ninamiolane (approved):

PR #896, opened 2021-01-05
Issue 893 and 894

tgeral68:
fixing issue 893 and 894 :
-> projection now use the TOLERANCE value defined in PoincareBall class
-> default behaviour of mobius_add project points to the ball according to TOLERANCE and EPSILON  values

It fixes the issue when using metric.geodesic when path steps are not carefully chosen. User now get the choice to project points on the ball when using mobius_add. However, further changes can be considered such as using **kwargs to pass arguments to the function.


tgeral68:
Got an issue in Travis, however it seems to fail on SPD matrix (not for every runs) and not on the current changes.

tests/test_poincare_ball.py
```diff
@@ -237,3 +237,18 @@ def test_ball_retraction(self):
...
         ball_metric.retraction(tangent_vec, x)
+
+    def test_ball_geodesic(self):
+        path_function =\
+            self.manifold.metric.geodesic(gs.array([0.1, 0.1]),
+                                          gs.array([0.2, 0.2]))
+        steps = gs.array(gs.linspace(-1000., 1000., 10000))
+        path_function(steps)
+
+    def test_mobius_out_of_the_ball(self):
```
ninamiolane, line 248: To check that a method raises the error that is expected, use:
`assertRaises(ValueError)`
e.g. 

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

ninamiolane:
Thanks @tgeral68 🙏! 

I don't know why pytorch's tests fail now... could you try with `torch==1.7.1` in `opt-requirements.txt`? In case the latest version of PyTorch does not care about Double versus Float? I saw that @nguigs has modified the version in his latest PR (not yet merged), I wonder if it could come from there.

@nguigs @ythanwerdas do you have other ideas there?

nguigs:
> @nguigs @ythanwerdas do you have other ideas there?

Yes it is a dtype problem in pytorch's wrapping of sqrtm, I fixed it in #891

ninamiolane:
@tgeral68 #891 has been merged: if you rebase, travis should pass and we can merge! 🎉 thanks for your patience!

PR #897, opened 2021-01-11
Make geodesic faster and more accurate in Dirichlet Distributions

alebrigant:
When boundary conditions are given, the geodesic method for Dirichlet Distributions solves 2 ODEs: first to compute the log, then to compute exp of the log. This is slow and the resulting geodesic does not end at the desired end point. This PR aims to fix this.

### DIFF
--- a/geomstats/geometry/dirichlet_distributions.py
+++ b/geomstats/geometry/dirichlet_distributions.py
@@ -235,13 +235,77 @@ class DirichletMetric(RiemannianMetric):
 
         return gs.squeeze(christoffels)
 
+    def _geodesic_ivp(self, initial_point, initial_tangent_vec):
+        """Solve geodesic initial value problem.
+
+        Compute the parameterized function for the geodesic starting at
+        initial_point with initial velocity given by initial_tangent_vec.
+        This is acheived by integrating the geodesic equation.
+
+        Parameters
+        ----------
+        initial_point : array-like, shape=[..., dim]
+            Initial point.
+
+        initial_tangent_vec : array-like, shape=[..., dim]
+            Tangent vector at initial point.
+
+        Returns
+        -------
+        path : function
+            Parameterized function for the geodesic curve starting at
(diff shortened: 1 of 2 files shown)


geomstats/geometry/dirichlet_distributions.py
```diff
@@ -235,71 +235,82 @@ def coefficients(ind_k):
...
+    def _geodesic_ivp(self, initial_point, initial_tangent_vec):
+        """Solve geodesic initial value problem.
 
-        Comute the exponential map associated to the Fisher information
-        metric at base_point of tangent_vec. This is achieved  by integration
-        of the geodesic equation (initial value problem), using the Christoffel
-        symbols.
+        Compute the parameterized function for the geodesic starting at
+        initial_point with initial velocity given by initial_tangent_vec.
+        This is acheived by integrating the geodesic equation.
```
nguigs, line 243: typo: achieved

tests/test_dirichlet_distributions.py
```diff
@@ -228,6 +228,7 @@ def test_log_and_exp(self):
         log = self.metric.log(points, base_points, n_steps=500)
         expected = points
         result = self.metric.exp(tangent_vec=log, base_point=base_points)
+        # print(gs.abs(result - expected) / expected)
```
nguigs, line 231: NIT: remove print

nguigs (approved):
Very nice! It will indeed avoid redundant computations! I think this could be extended to all cases where geodesics are computed by integration but this might be trickier if times are not sampled on a regular grid as we don't use scipy's odeint.

nguigs:
There are three errors with Pytorch that are caused by a previous commit, I fixed them in #891, so don't worry about it and just merge (or rebase) the master branch into your branch once #891 is merged.

ninamiolane:
@alebrigant #891 has been merged so some travis error should go away thanks to @nguigs 🎉

ninamiolane:
Closing this since #900 was just merged.

Issue #880, opened 2020-11-05
Parallelize geometry.riemannian_metric.dist_pairwise


oleg-kachan:
dist_pairwise is implemented using for loop, thus is single-threaded and slow, this could be fixed to use all available processor threads.

Also method signature contains a hint "Set of points in _hyperbolic_ space" - to be corrected to points of the arbitrary manifold which metric is specified.

ninamiolane:
@tgeral68 @Hzaatiti any thoughts on this?

Issue #775, opened 2020-05-04
Np, tf and torch initialize arrays with different default dtype


ninamiolane:
The following code:
```
import numpy as np
import tensorflow as tf
import torch

a = np.array([1.])
b = tf.constant([1.])
c = torch.tensor([1.])
print(a.dtype, b.dtype, c.dtype)
```
gives:
```
float64 torch.float32 <dtype: 'float32'>
```

Similarly:
```
import geomstats.backend as gs

a = gs.array([1.])
print(a.dtype)
```
outputs float64, float32, float32 with np, tf and torch respectively.

Do we want to uniformize this and if so, use float64 or float32 as a default?

nguigs:
That's probably why we always have to adapt tol for tests in pytorch and tensorflow! I would have said float32 but we still need to gain on numerical accuracy...

ninamiolane:
We could have a `DEFAULT_FLOAT_DTYPE = ...` in each backend, and initially set it to `gs.float32` (?). We could revert it if `gs.float64` seems more convenient?

lpereira95:
Unless it has changed in recent years (which I don't believe), `numpy` does not have a way of setting `float32` as default (see [this](https://github.com/numpy/numpy/issues/6860)).

So in my opinion we should set `float64` as default for all backends (for consistency) and let the user decide to reduce precision if it applies to his/her use case (it will be specially relevant when using gpu).

lpereira95:
Solved by #1615.

PR #901, opened 2021-01-19
Parallelize pairwise_dist

nguigs:
address #880

### DIFF
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -466,11 +466,11 @@ class Connection:
         if end_point is not None:
             shooting_tangent_vec = self.log(
                 point=end_point, base_point=initial_point)
-            all_close = gs.allclose(shooting_tangent_vec, initial_tangent_vec)
-            if initial_tangent_vec is not None and not all_close:
-                raise RuntimeError(
-                    'The shooting tangent vector is too'
-                    ' far from the input initial tangent vector.')
+            if initial_tangent_vec is not None:
+                if not gs.allclose(shooting_tangent_vec, initial_tangent_vec):
+                    raise RuntimeError(
+                        'The shooting tangent vector is too'
+                        ' far from the input initial tangent vector.')
             initial_tangent_vec = shooting_tangent_vec
 
         if point_type == 'vector':
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -466,11 +466,11 @@ class Connection:
         if end_point is not None:
             shooting_tangent_vec = self.log(
                 point=end_point, base_point=initial_point)
-            if initial_tangent_vec is not None:
-                if not gs.allclose(shooting_tangent_vec, initial_tangent_vec):
-                    raise RuntimeError(
-                        'The shooting tangent vector is too'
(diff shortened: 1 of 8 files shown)


geomstats/geometry/connection.py
```diff
@@ -178,6 +178,7 @@ def objective(velocity):
 
         tangent_vec = gs.array(res.x)
         tangent_vec = gs.reshape(tangent_vec, max_shape)
+        tangent_vec = gs.cast(tangent_vec, dtype=base_point.dtype)
```
ninamiolane, line 181: NIT: why using the dtype from base_point, and not from point for example?

On a higher level, this could be related to issue #775

geomstats/geometry/invariant_metric.py
```diff
@@ -595,7 +595,7 @@ def lie_acceleration(point, vector):
                             step=step, **kwargs)
         return flow[-1]
 
-    def log(self, point, base_point, n_steps=15, step='k4',
+    def log(self, point, base_point, n_steps=15, step='rk4',
```
ninamiolane, line 598: From the docstring it seems that step should be in group_rk2, group_rk4, and that the default is group_rk4?

Maybe add a check:
```
if step not in ['group_rk2', 'group_rk4']:
....
```
?

tests/test_hypersphere.py
```diff
@@ -432,6 +433,15 @@ def test_dist_pairwise(self):
 
         self.assertAllClose(result, expected, rtol=1e-3)
 
+    def test_dist_pairwise_parallel(self):
```
ninamiolane, line 436: You do not have any test decorator here. Does that mean that we can remove `@geomstats.tests.np_and_pytorch_only` from the above `test_dist_pairwise`?

ninamiolane (approved):
Yesss, thanks for doing this!! 🎉  How much speed up did you experience? Did you give more thoughts about potentially using Keops?

Just small comments.

geomstats/geometry/connection.py
nguigs: Because tangent_vec is tangent at base point, and they might be used together in .e.g norm (for a distance computation)

geomstats/geometry/invariant_metric.py
nguigs: I think the docstring is not up to date, I harmonized all the steps so that there is no longer group specific ones

tests/test_hypersphere.py
nguigs: Yes!

geomstats/geometry/invariant_metric.py
nguigs: And there is `check_parameter_accepted_values(step, 'step', STEP_FUNCTIONS)` in integrator

nguigs:
> Yesss, thanks for doing this!! tada How much speed up did you experience? Did you give more thoughts about potentially using Keops?
I didn't try Keops, I have cuda issues on my laptop :(
For long distance computation (with invariant metric), I get at least 30% speed up when using parallel processes, and with more samples the difference increases. For quick computations (with the hypersphere), multi-threading/processing is longer

PR #902, opened 2021-01-20
Full-rank correlation matrices

ythanwerdas:
New class of full-rank correlation matrices.
New class of quotient metric on it, defined by the quotient of the affine-invariant metric on SPD matrices by the action of diagonal matrices.

As this is a new instanciation of a quotient metric, I have some new comments about the recent classes QuotientMetric and FiberBundle @nguigs.
1. In `FiberBundle`, it seems that there is an inconsistent use of `base_point` and `point`: in the functions `vertical_projection` and `horizontal_projection`, the `base_point` is a point of the total space whereas in the function `horizontal_lift`, the `base_point` is a point of the quotient manifold and `point` is a point of the total space. The ladder seems more logical to me.
2. As the functions `vertical_projection`, `horizontal_projection` and `horizontal_lift` depend on the choice of the quotient metric, I think they should be implemented in the class `QuotientMetric` instead of the class `FiberBundle`.
3. Later on, I would like to implement some other quotient metrics on correlation matrices based on the same fiber bundle of SPD matrices. It seems to me that I cannot right now because I can only implement one quotient structure on the SPD manifold. A possibility would be to implement fiber bundles directly as submersions (functions) that could have the total space and the quotient space as attributes.


ythanwerdas:
No unit tests yet, coming soon.

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -0,0 +1,143 @@
...
+        atol : float
+            Tolerance.
+            Optional, default: TOLERANCE.
+
+        Returns
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean denoting if mat is an SPD matrix."""
+        is_spd = super(FullRankCorrelationMatrices, self).belongs(mat, atol)
+        diagonal = gs.einsum('...ii->...i', mat)
```
nguigs, line 47: Use `gs.diagonal(mat, axis1=-2, axis2=-1)` to be more explicit

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -0,0 +1,143 @@
...
+        is_diag_one = gs.all(diagonal == 1)
+        belongs = is_spd and is_diag_one
```
nguigs, line 49: use `gs.logical_and` for vectorization (otherwise you'll get `ValueError: The truth value of an array with more than one element is ambiguous`)

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -0,0 +1,143 @@
...
+from geomstats.geometry.general_linear import GeneralLinear
+from geomstats.geometry.matrices import Matrices
+from geomstats.geometry.quotient_metric import QuotientMetric
+from geomstats.geometry.spd_matrices import SPDMatrices, SPDMetricAffine
+from geomstats.geometry.symmetric_matrices import SymmetricMatrices
+
+EPSILON = 1e-6
+TOLERANCE = 1e-12
+
+class FullRankCorrelationMatrices(SPDMatrices, EmbeddedManifold):
```
nguigs, line 16: What I had in mind was that this class should inherit FiberBundle with total space SPD, instead of SPD inheriting FIberBundle

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -0,0 +1,143 @@
...
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean denoting if mat is an SPD matrix."""
+        is_spd = super(FullRankCorrelationMatrices, self).belongs(mat, atol)
+        diagonal = gs.einsum('...ii->...i', mat)
+        is_diag_one = gs.all(diagonal == 1)
+        belongs = is_spd and is_diag_one
+        return belongs
+
+    def random_uniform(self, n_samples=1):
```
nguigs, line 52: This method would be inherited from FiberBundle

geomstats/geometry/spd_matrices.py
```diff
@@ -30,7 +31,7 @@ def __init__(self, n):
             embedding_manifold=GeneralLinear(n=n))
 
     def belongs(self, mat, atol=TOLERANCE):
-        """Check if a matrix is symmetric and invertible.
+        """Check if a matrix is symmetric and positive definite.
```
nguigs, line 34: Thanks!

geomstats/geometry/spd_matrices.py
```diff
@@ -108,6 +109,105 @@ def random_tangent_vec_uniform(self, n_samples=1, base_point=None):
 
         return tangent_vec
 
+    @staticmethod
+    def submersion(point):
```
nguigs, line 113: This shouldn't be in the SPD class, otherwise, as you said you can only have one quotient strucutre with total space SPD

geomstats/geometry/spd_matrices.py
```diff
@@ -108,6 +109,105 @@ def random_tangent_vec_uniform(self, n_samples=1, base_point=None):
...
+                -------
+                cor : array_like, shape=[..., n, n]
+                    Full rank correlation matrix.
+                """
+        diagonal = gs.einsum('...ii->...i', point)
+        cor = gs.einsum('...i,...ij->...ij', diagonal, point)
+        cor = gs.einsum('...ij,...j->...ij', cor, diagonal)
+        return cor
+
+    def tangent_submersion(self, tangent_vec, base_point):
```
nguigs, line 131: same as above

nguigs (changes requested):
>     1. In `FiberBundle`, it seems that there is an inconsistent use of `base_point` and `point`: in the functions `vertical_projection` and `horizontal_projection`, the `base_point` is a point of the total space whereas in the function `horizontal_lift`, the `base_point` is a point of the quotient manifold and `point` is a point of the total space. The ladder seems more logical to me.
You're right I tried respecting the `base_point` convention of geomstats manifolds in the case of the total space, but maybe we should consistently use base_point for a point in the base manifold and point for a point in the total space.

>     2. As the functions `vertical_projection`, `horizontal_projection` and `horizontal_lift` depend on the choice of the quotient metric, I think they should be implemented in the class `QuotientMetric` instead of the class `FiberBundle`.
Yes I thought about that but I think we should each child class of FiberBundle with a metric on the total space given so that align, horizontal, etc are defined.

>     3. Later on, I would like to implement some other quotient metrics on correlation matrices based on the same fiber bundle of SPD matrices. It seems to me that I cannot right now because I can only implement one quotient structure on the SPD manifold. A possibility would be to implement fiber bundles directly as submersions (functions) that could have the total space and the quotient space as attributes.
This is exactly what I had in mind for the class FiberBundle, (it has a submersion method, and total/base spaces as attributes). I think you understood the FiberBundle class to necessarily represent the total space, but for me it is really the triple (E, B, Pi). Maybe the choice of vocabulary is not optimal?

ninamiolane:
@ythanwerdas  any update on this? 😃

nguigs:
@ythanwerdas since #932 you can specify the metric of the total space with the attribute `ambient_metric`.

nguigs:
Hi @ythanwerdas, I got this branch back on tracks and added tests, the `vertical_projection` does not return something in the ker of the `tangent_riemannian_submersion`. Any idea what's wrong?

nguigs (approved):

nguigs:
@ythanwerdas this is now ready!

geomstats/geometry/quotient_metric.py
```diff
@@ -61,17 +62,19 @@ def inner_product(
         inner_product : float, shape=[...]
             Inner products
         """
-        if point is None:
+        if point_fiber is None:
```
ninamiolane, line 65: if statements can be merged?

geomstats/geometry/spd_matrices.py
```diff
@@ -348,11 +371,16 @@ def logm(cls, mat):
         """
         n = mat.shape[-1]
         dim_3_mat = gs.reshape(mat, [-1, n, n])
-        logm = cls.apply_func_to_eigvals(
+        logm = SymmetricMatrices.apply_func_to_eigvals(
             dim_3_mat, gs.log, check_positive=True)
         logm = gs.reshape(logm, mat.shape)
         return logm
 
+    expm = SymmetricMatrices.__dict__['expm']
```
ninamiolane, line 379: What is this? Why not just `expm = SymmetricMatrices.expm`?

tests/tests_geomstats/test_full_rank_correlation_matrices.py
```diff
@@ -0,0 +1,144 @@
...
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.full_rank_correlation_matrices import \
+    CorrelationMatricesBundle, FullRankCorrelationAffineQuotientMetric
+from geomstats.geometry.general_linear import GeneralLinear
+from geomstats.geometry.matrices import Matrices
+from geomstats.geometry.spd_matrices import SymmetricMatrices
+
+
+class TestCorrelationMatrices(geomstats.tests.TestCase):
```
ninamiolane, line 12: -> `TestFullRankCorrelationMatrices`: the name of the test class is meant to match the name of the test file.

tests/tests_geomstats/test_quotient_metric.py
```diff
@@ -26,17 +26,17 @@ def tangent_riemannian_submersion(self, tangent_vec, base_point):
             base_point, Matrices.transpose(tangent_vec))
         return 2 * Matrices.to_symmetric(product)
 
-    def horizontal_lift(self, tangent_vec, point=None, base_point=None):
+    def horizontal_lift(self, tangent_vec, base_point=None, point_fiber=None):
         if base_point is None:
-            if point is not None:
-                base_point = self.riemannian_submersion(point)
+            if point_fiber is not None:
```
ninamiolane, line 31: if statements could be merged?

ninamiolane (commented):
Very nice!!

geomstats/geometry/quotient_metric.py
nguigs: The else condition would need to be replaced by `elif fiber_point is None and base_point is None` so it's just as clear?

geomstats/geometry/spd_matrices.py
nguigs: Try it if you want :stuck_out_tongue_closed_eyes: 

SymmetricMatrices.expm is a class method so what you suggest works, but it does not for `from_vector` and `to_vector` that are static,  because the first argument `self=SPD(n)` is not dealt with.

PR #905, opened 2021-02-02
Add jacobian of christoffels in Dirichlet distributions and a timer

alebrigant:
* Add the possibility to use the explicit Jacobian of the function to optimize when solving the boundary value problem. This speeds up the computations in most cases by a factor of 2.
* Add a timer when solving the boundary value problem. When the time is out, the process is terminated. This can be interesting in practice as the solver can take several minutes in some cases.


geomstats/_backend/__init__.py
```diff
@@ -111,7 +111,7 @@
         'zeros',
         'zeros_like'
     ],
-    'autograd': ['value_and_grad'],
+    'autograd': ['jacobian', 'value_and_grad'],
```
nguigs, line 114: You have to add a `jacobian` function to the other backends, in the [autograd](https://github.com/geomstats/geomstats/blob/master/geomstats/_backend/pytorch/autograd.py) module.

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -1,5 +1,7 @@
 """Statistical Manifold of Dirichlet distributions with the Fisher metric."""
 
+import multiprocessing
```
nguigs, line 3: Can you try with joblib, as in [pairwise_dist](https://github.com/geomstats/geomstats/blob/544debcf81dc40d7e5d584c5606792c2cfa9f4eb/geomstats/geometry/riemannian_metric.py#L298) to be consistent?

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -235,6 +238,78 @@ def coefficients(ind_k):
...
+        Parameters
+        ----------
+        base_point : array-like, shape=[..., dim]
+            Base point.
+
+        Returns
+        -------
+        jac : array-like, shape=[..., dim, dim, dim, dim]
+            Jacobian of the Christoffel symbols.
+            :math: jac[..., i, j, k, l] = 'dGamma^i_{jk} / dx_l'
```
nguigs, line 256: You have to use ticks ` after :math:

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -407,6 +491,41 @@ def boundary_cond(
...
+
+            gamma = self.christoffels(gs.transpose(position))
+            df_dvelocity = - 2 * gs.einsum(
+                '...ijk,k...->ij...', gamma, velocity)
+
+            jac = gs.zeros((2 * self.dim,) + state.shape)
+            jac[:self.dim, self.dim:, ...] = gs.squeeze(gs.transpose(gs.tile(
+                gs.eye(self.dim), (n_times, 1, 1))))
+            jac[self.dim:, :self.dim, ...] = df_dposition
+            jac[self.dim:, self.dim:, ...] = df_dvelocity
```
nguigs, line 525: Could we avoid assignment in case one day this works with tensorflow or automatic differentiation?

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -441,16 +560,39 @@ def initialize(point_0, point_1):
...
+                    target=process_function, args=(return_dict,))
+                process.start()
+
+                process.join(max_time)
+                if process.is_alive():
+                    process.terminate()
+                    print('Maximum time of {} seconds reached. '
+                          'Process terminated. '
+                          'Result is inaccurate.'.format(max_time))
+                    geod.append(gs.zeros((n_steps, self.dim)))
```
nguigs, line 586: Use logging.info instead of print

nguigs: Maybe returning NaNs would be better than zeros?

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -516,10 +676,72 @@ def geodesic(self, initial_point, end_point=None,
...
+            path = self._geodesic_bvp(initial_point, end_point, n_steps,
+                                      jacobian=jacobian, max_time=max_time)
 
         if initial_tangent_vec is not None:
             path = self._geodesic_ivp(
                 initial_point, initial_tangent_vec, n_steps)
 
         return path
+
+    def squared_dist(self, point_a, point_b, n_steps=N_STEPS, jacobian=False,
```
nguigs, line 688: Maybe we can add a `**kwargs` in the signature of squared_dist and propagate it to the log in the base class Riemannian metric so that you don't have to duplicate the code? It will be useful for other metrics where the log is an optimization problem with more key-word arguments.

nguigs (commented):
Nice contribution! Two main points and a few remarks:
- The jacobian function needs to be added to all backends
- Use joblib instead of multiprocessing if possible

geomstats/_backend/__init__.py
alebrigant: I'm not sure I get it: do you mean that I should write a function that does the same thing as ```jacobian``` of autograd but for the other backends ?

geomstats/geometry/dirichlet_distributions.py
alebrigant: Searching the documentation and looking for examples, I haven't found a way to do this with joblib...

geomstats/geometry/dirichlet_distributions.py
alebrigant: You're right.

geomstats/geometry/dirichlet_distributions.py
alebrigant: Good idea.

geomstats/_backend/pytorch/__init__.py
```diff
@@ -767,3 +767,7 @@ def triu_to_vec(x, k=0):
     n = x.shape[-1]
     rows, cols = triu_indices(n, k=k)
     return x[..., rows, cols]
+
+
+def nan():
```
ninamiolane, line 772: Why did you add this function `nan`? I don't see it used in this PR. 

We should also avoid using `numpy`'s functions in tensorflow and pytorch backends: do tf and pytorch have the function that you are looking for?

Additionally, `numpy` does not seem to have a `nan` function, only a `nan` attribute:

Screen Shot 2021-03-28 at 11 34 50 AM

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -235,6 +239,78 @@ def coefficients(ind_k):
 
         return gs.squeeze(christoffels)
 
+    def jac_christoffels(self, base_point):
```
ninamiolane, line 242: Would it be possible to write `jacobian` in full? Just because we have started to write it in full in other locations in the code, e.g.:
https://github.com/geomstats/geomstats/blob/d895d33c83ba44229c6f8b9b6deebe457b1b5964/geomstats/geometry/special_euclidean.py#L552

Thanks 🙏

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -235,6 +239,78 @@ def coefficients(ind_k):
...
+        """Compute the Jacobian of the Christoffel symbols.
+
+        Compute the Jacobian of the Christoffel symbols of the
+        Fisher information metric.
```
ninamiolane, line 246: Would you have a reference that explains the computations performed in this function? See e.g. here to add a reference to a docstring:
https://github.com/geomstats/geomstats/blob/d895d33c83ba44229c6f8b9b6deebe457b1b5964/geomstats/geometry/stiefel.py#L211

ninamiolane: It would be great if it could also explain/define the notations `g_y, g_t, df_y, df_t`, that are hard to understand just by reading the code (at least for me!). Alternatively, the variable names `g_y` etc could be replaced by names that are more self-explanatory. Thanks!

tests/test_dirichlet_distributions.py
```diff
@@ -348,3 +348,48 @@ def test_geodesic_vectorization(self):
         result = geod(time).shape
         expected = (self.n_points, self.dim)
         self.assertAllClose(expected, result)
+
+    @geomstats.tests.np_only
+    def test_jac_christoffels(self):
```
ninamiolane, line 353: Nice! It would be great to test if the value of the jacobian is what is expected on a single well-chosen point, i.e. not only testing the shape and the vectorization but also the result itself.

I see that you added `@geomstats.tests.np_only`: what are the problems for tf and pytorch, did you try running it with these backends?

tests/test_dirichlet_distributions.py
```diff
@@ -348,3 +348,48 @@ def test_geodesic_vectorization(self):
...
+        and without.
+        """
+        point_a = self.dirichlet.random_point()
+        point_b = self.dirichlet.random_point()
+        result = self.dirichlet.metric.dist(point_a, point_b, jacobian=True)
+        expected = self.dirichlet.metric.dist(point_a, point_b)
+        self.assertAllClose(expected, result)
+
+    @geomstats.tests.np_only
+    def test_geodesic_bvp_timer(self):
```
ninamiolane, line 386: If you have an easy example on this: would it be possible to also write a test for `jac` of bvp function?

ninamiolane (approved):
So sorry for the very late code review! But this is great, I have just minor comments.

geomstats/_backend/pytorch/__init__.py
alebrigant: I use ```nan``` in ```_geodesic_bvp```: when the process takes too long, it is terminated and returns an array full of ```nan```s. This attribute seems to be particular to numpy. An other possibility is to return an array full of zeros, but that doesn't seem very satisfactory either. What do you think ?

tests/test_dirichlet_distributions.py
alebrigant: That's what I did at first using the ```jacobian``` from autograd. But then I needed to add a ```jacobian``` function to the autograd file of the other backends, and I didn't know how. If you have insights on this, I am interested !

geomstats/geometry/dirichlet_distributions.py
alebrigant: The notations ```f_y, f_t, g_y, g_t``` etc. simply referred to intermediary terms in the computations. I have renamed them in a more neutral way.

alebrigant: I haven't put these computations in any publication yet. But I have added a reference for the computation of the Christoffel symbols.

tests/test_dirichlet_distributions.py
ninamiolane: @nguigs any idea there?

geomstats/geometry/dirichlet_distributions.py
ninamiolane: Ok perfect, thank you! 🙏

geomstats/_backend/pytorch/__init__.py
ninamiolane: Oh, got it! Sorry, I hadn't seen the line in `_geodesic_bvp`. In this case, do you think that we could use `math.nan` since the `nan` is used to populate an array anyway. This would avoid adding/maintaining it within the backends. Sorry to ask yet another change! :/ We're almost there though.

geomstats/_backend/__init__.py
nguigs: Sorry I didn't see this earlier

tests/test_dirichlet_distributions.py
nguigs: Yes, but where did you need it? I don't see any call to gs.autograd.jacobian in the PR?

nguigs: In both pytorch-tensorflow/autograd files, a function `jacobian` needs to be added that behaves like the one from the `autograd` package.
For tensorflow, it should look like:
```python
def jacobian(f, x):
    if isinstance(x, np.ndarray):
        x = tf.Variable(x)
    with tf.GradientTape() as g:
      g.watch(x)
      y = f(x)
    return g.jacobian(y, x)
```
Here is a discussion on how to do it with pytorch: https://gist.github.com/sbarratt/37356c46ad1350d4c30aefbd488a4faa

nguigs: If you can show your original code using `jacobian` for the numpy backend I can make sur the above solutions work

nguigs:
I made the changes but am not able to push, is the box allow edits from maintainers ticked?

alebrigant:
Great thanks ! I was in the middle of trying the changes you suggested. Yes the box is ticked, strange that you can't push...

alebrigant:
Since it was only a few lines I copied your changes and pushed them @nguigs. Thanks !

geomstats/_backend/pytorch/__init__.py
alebrigant: Yes good idea to use ```math.nan``` instead, perfect ! Thanks !

tests/test_dirichlet_distributions.py
alebrigant: Yes it would be nice to test this, with autograd for example, but I would need to make it a class method, right ? As it is, I don't see how I can test it since I can't access it. What I did for now, is I tested that the bvp solver yields the same result, with and without using the jacobian.

alebrigant:
@ninamiolane @nguigs do you have remaining comments/corrections or is this good to merge ? Thanks !

alebrigant:
The last commits are just destined to make ```test_jacobian_christoffels``` work with pytorch and tensorflow. It now works with pytorch, but I wasn't able to fix an error occurring with ```einsum``` in tensorflow.

alebrigant:
@ninamiolane @nguigs sorry for insisting, is this good to merge or do you see other changes to make ? Thanks !

nguigs:
Any chance DeepSource could pass? Otherwise it will show in other PRs...
The rest LGTM :)

alebrigant:
Ah yes I see what you mean @nguigs, there are now 45 missing module/function docstrings in the general backend and tensorflow backend... I have fixed the DeepSource error linked to this PR.

Should DeepSource be configured to ignore the missing docstring errors ? Or I can add docstrings if needed.

alebrigant:
There are also new errors linked to ```SpecialEuclidean```,  that I have also found in #1000 and #999. Any idea how this happened ?

PR #918, opened 2021-03-03
Run automatic formatter: black

ninamiolane:
Instead of using flake8 to verify that the code is formatted correctly (i.e. following Pep8), we can run an automatic formatter.

I suggest using black: https://pypi.org/project/black/

This PR is the output of the command:
```black --skip-string-normalization .    ```

We could move to double quotes convention for strings, so that we can just run:
```black .```
easily. What do you think?

Finally, we could automate black, by having it run automatically each time a contributor submits a PR.


ninamiolane:
Travis tests are failing because of flake8, which has different style conventions than black. Will look into this.

PR #932, opened 2021-03-10
add O'Neill's formula for curvature

nguigs:
Add O'Neill's formula to compute the curvature of Riemannian submersions, and application in Kendall shape space.

### DIFF
--- a/geomstats/geometry/fiber_bundle.py
+++ b/geomstats/geometry/fiber_bundle.py
@@ -38,7 +38,7 @@ class FiberBundle(Manifold):
     def __init__(
             self, total_space: Manifold = None, base: Manifold = None,
             group: LieGroup = None, ambient_metric: RiemannianMetric = None,
-            group_action=None, dim:  int = None, **kwargs):
+            group_action=None, dim: int = None, **kwargs):
 
         if dim is None:
             if base is not None:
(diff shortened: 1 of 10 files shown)


geomstats/geometry/fiber_bundle.py
```diff
@@ -33,8 +39,10 @@ class FiberBundle(Manifold):
         group. Either dim, base or group must be given as input.
     """
 
-    def __init__(self, total_space, base=None, group=None, group_action=None,
-                 dim=None, **kwargs):
+    def __init__(
```
ninamiolane, line 42: Yes, we need to decide if we use typing everywhere! @MariusGuerard was also thinking about this.

tests/test_quotient_metric.py
```diff
@@ -153,3 +153,14 @@ def test_squared_dist(self):
             points[1], points[0], tol=1e-10)
         expected = self.base_metric.squared_dist(points[1], points[0])
         self.assertAllClose(result, expected, atol=1e-5)
+
+    def test_fundamental_tensor(self):
```
ninamiolane, line 157: NIT: test_fundamental_a_tensor

ninamiolane (approved):
Very nice! Why the decrease in code coverage?

Also, we should decide if we want to add typing in the library. I think @MariusGuerard and @tristancabel could be pro-typing.

nguigs:
I am surprised by the codecov report too, I used sourcegraph but didn't find the missing hits...

PR #934, opened 2021-03-12
remove redundant call to variance function

nguigs:
The variance function uses `metric.squared_dist`, which can be replaced by `metric.norm(logs)`, where logs have already been computed. This avoids computing the logs twice.
Also fixed `test_weighted_frechet_mean` which was wrong as the `weights` arguments was not used...


ninamiolane (approved):
🙏 thanks for spotting this!

Issue #920, opened 2021-03-05
1. As a end user I need to have better performance


tristancabel:
Performance et passage à l’échelle par rapport au nombre de données et la dimension :
 - Moyenne géométrique 100 matrices SPD. ok avec nilearn pour 10x10→pas 15x15. Données Réelles ≥ 60x60
https://travis-ci.org/github/geomstats/geomstats/builds/759593811

Une piste serait de regardé Numba (regarder https://ct.gitlabpages.inria.fr/gallery/kepler-py/kepler-py.html)

nguigs:
> * Moyenne géométrique 100 matrices SPD. ok avec nilearn pour 10x10→pas 15x15.
see #913 for more details

nguigs:
```python
import geomstats.backend as gs
from geomstats.geometry.hypersphere import HypersphereMetric
from timeit import timeit
x = gs.array([1., 0., 0., 0.])
v = gs.array([0., 2., 1.5, .4])
metric = HypersphereMetric(3)
print(timeit(lambda: metric.norm(v, x), number=1000)) #  0.11182429000018601
print(timeit(lambda: gs.linalg.norm(v), number=1000)) # 0.008933541999795125
```

tristancabel:
- [ ] profiling
- [ ] numba

ninamiolane:
- [ ] interaction with joblib and OpenBLAS

ninamiolane:
Could using cache be an option?

https://gist.github.com/Susensio/61f4fee01150caaac1e10fc5f005eb75

mvesin:
tests & basic benchmarks report for issue 913 :
https://github.com/mvesin/geomstats/blob/feature/test_profiling/profiling/out.issue913/README.md
* confidence in our profiling instrumentation with cProfile
* OMP_NUM_THREADS=1 good practice when using PyPi numpy package (why ? to be investigated)
* ... but does not fill the performance gap between geomstats and nilearn for this example
* better performance and scaling for this example with nilearn, tests suggest to investigate implementation and algorithmic aspects

mvesin:
more tests & benchmarks for issue 913 : 
https://github.com/mvesin/geomstats/blob/feature/test_profiling/profiling/out.issue_913.commit_eac607b8/README.md

xpennec:
[...]

Issue #913, opened 2021-02-25
SPD means don't coincide with nilearn


nguigs:
For dim > 10, the Frechet mean does not coincide with the one computed by nilearn
```python
from nilearn.connectome.connectivity_matrices import _geometric_mean

from geomstats.geometry.spd_matrices import SPDMatrices, SPDMetricAffine
from geomstats.learning.frechet_mean import FrechetMean
import geomstats.backend as gs

gs.random.seed(0)
n_points = 100
dim = 15
space = SPDMatrices(dim)
data = space.random_uniform(n_samples=n_points)

metric = SPDMetricAffine(dim)
mean = FrechetMean(metric=metric, method='default', point_type='matrix',
                   max_iter=1000, verbose=True, epsilon=1e-10)
mean.fit(data)
mean_estimate = mean.estimate_

nilearn_mean = _geometric_mean(data, max_iter=1000, tol=1e-7)
print(metric.dist(nilearn_mean, mean_estimate))
```

ninamiolane:
@nguigs what does the print give? Is the distance large?

PR #935, opened 2021-03-16
Adapt step size in Frechet Mean

nguigs:
As an outcome of discussions aroud #913 and #920, this PR divides the step size by 2 in the case the gradient descent starts diverging.


ninamiolane (approved):
Great!

PR #937, opened 2021-03-16
architecture

tristancabel:
- dummy_archi2 and spdMatrices inheritance


examples/dummy_archi2.py
```diff
@@ -0,0 +1,153 @@
...
+        if len(metrics) > 1:
+            res = {}
+            for metric in super().getMetrics():
+                res[metric.name()] = metric() # TODO tangent_vec
+            return res
+        elif len(metrics) == 1:
+            return metrics[0]()
+        else :
+            print("no metric")
+            return 0
```
nguigs, line 107: Would these conditions have to be added to every method of the `Metric` class?

examples/dummy_archi2.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def __init__(self, metrics : List[Metric]):
+        self.metrics = metrics
+
+    def getMetrics(self) -> List[Metric]:
+        return self.metrics
+
+    @abstractmethod
+    def exp(self, tangent_vec) -> Union[int, Dict[str, int]]:  # TODO return also only str 
+        pass
```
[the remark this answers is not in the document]
tristancabel: done in last commit

examples/dummy_archi2.py
tristancabel: solved

tristancabel:
@nguigs , what do you think?

nguigs:
Hi @tristancabel, thanks for the notebook! It is nice from the developer's point of view. There is also the user point of view, e.g. I want to compare two metrics on some manifold, how do I do? Maybe you could re use this [example on se2](https://github.com/geomstats/geomstats/blob/master/examples/plot_geodesics_se2.py) and see how it adapts to the new architecture?

tristancabel:
> Hi @tristancabel, thanks for the notebook! It is nice from the developer's point of view. There is also the user point of view, e.g. I want to compare two metrics on some manifold, how do I do? Maybe you could re use this [example on se2](https://github.com/geomstats/geomstats/blob/master/examples/plot_geodesics_se2.py) and see how it adapts to the new architecture?

done [here](https://github.com/geomstats/geomstats/pull/937/files#diff-ca7230c96e75ce84def5bcda4c18c62b95a2151b8b5712941769a4c7463e2ffbR36)

ninamiolane:
Closing this stalled PR: @nguigs has refactor the architecture of the `geometry` module.

PR #939, opened 2021-03-17
Nguigs adaptive gd

nguigs:
Following discussions around #913


nguigs:
Commits from two different projects got mixed up

PR #940, opened 2021-03-17
Adaptive gradient descent for point_type=matrix

nguigs:
Update adaptive gradient descent to work with matrices


nguigs:
@xpennec

tests/test_frechet_mean.py
```diff
@@ -149,6 +149,18 @@ def test_estimate_and_belongs_default_gradient_descent_so_matrix(self):
         expected = True
         self.assertAllClose(result, expected)
 
+    @geomstats.tests.np_and_tf_only
```
ninamiolane, line 152: What is the problem with pytorch?

ninamiolane (approved):
Very nice!

ninamiolane:
@nguigs what was the problem with tests in pytorch on this one?

tests/test_frechet_mean.py
nguigs: This test is new, it cannot be used with pytorch because the matrix log is used but not implemented in pytorch

Issue #407, opened 2020-01-27
Rename `random_uniform` for Euclidean and SpecialEuclidean spaces


stefanheyder:
The names of these methods are misleading, as you cannot sample uniform from Euclidean space. As this will affect more than one space (also: non-compact Riemannian manifolds) and we should have a uniform (pun intended) handling of this, let's discuss how to handle this.

nguigs:
I see two solutions:
- Give more details in the docstring: uniform in Euclidean space is actually uniform on [0,1]^dim and can be scaled by the parameter `bound`, in other non-compact it is uniform on a ball of a tangent space (identity for SPDs)
- Or we use a Gaussian distribution (or wrapped Gaussian) when the manifold is non compact

ninamiolane:
(love the pun @stefanheyder !)

Let's do both recommendations from @nguigs ?
- solution 1, as per the function `random_uniform`.
- implement (wrapped) Gaussian distribution, and call it `random_gaussian`?

nguigs:
In fact the notion of probability distribution depends on the Riemannian metric (because the volume form does). We should keep a `random_point` because it's handy, but should `random_unfirom`, `random_gaussian` (upcoming) be in the metric class @ninamiolane ?

PR #941, opened 2021-03-17
Sampling on Manifolds + Stiefel Log

nguigs:
- Start to address #407 by renaming random_uniform to random_point on non-compact manifolds
- Fix the log in Stiefel space: when negative determinant is encountered, 1 or 2 eigen vectors from the svd need to be flipped.


ninamiolane:
Great, thank you. A cleaning was indeed much needed on the random_* functions. However, don't you think that random_point is a bit vague? How about `random_uniform(radius=...)` to generate random points on a geodesic sphere for a given radius?

ninamiolane:
Merging 🎉  yet we should think about refactoring the `sampling` functions.

PR #947, opened 2021-03-22
Avoid repetitive eigen-value decompositions

nguigs:
- allow to pass list as argument to `powerm` and `apply_func_to_eigvals`
- simplify code in `aux_differential_power`
- use methods from `Matrices` instead of backend functions with custom arguments
- remove `point_type` argument in method `geodesic`: point_type is fixed by default for every metric instance
- Consistently use `Matrices.frobenius_product`, which uses einsum to compute the Frobenius inner product. This is faster than computing matrix product and trace, or element-wise product then sum.


geomstats/geometry/spd_matrices.py
```diff
@@ -307,12 +279,10 @@ def inverse_differential_log(cls, tangent_vec, base_point):
             cls.aux_differential_power(0, tangent_vec, base_point)
         power_operator = denominator / numerator
         result = power_operator * temp_result
-        result = gs.matmul(result, transp_eigvectors)
-        result = gs.matmul(eigvectors, result)
+        result = Matrices.mul(eigvectors, result, transp_eigvectors)
         return result
 
     @classmethod
```
ninamiolane, line 285: No need for the vectorization decorator anymore? Nice! 🎉

geomstats/geometry/spd_matrices.py
```diff
@@ -644,43 +601,18 @@ def log(self, point, base_point, **kwargs):
...
             log = self._aux_log(
-                power_point,
-                power_sqrt_base_point,
-                power_inv_sqrt_base_point)
-            log = self.space.inverse_differential_power(power_affine, log,
-                                                        base_point)
+                power_point, powers[0], powers[1])
+            log = self.space.inverse_differential_power(
+                power_affine, log, base_point)
         return log
```
ninamiolane, line 615: where did geodesic go?

geomstats/geometry/spd_matrices.py
```diff
@@ -1053,25 +977,3 @@ def log(self, point, base_point):
             log_point - log_base_point, log_base_point)
 
         return log
```
ninamiolane, line 979: where did this geodesic go?

ninamiolane (approved):
Nice! Could you also document (both in the documentation of the functions and the contributing guidelines) the choices of linear algebra computations that were made to improve efficiency? Thank you.

geomstats/geometry/spd_matrices.py
nguigs: It's inherited

geomstats/geometry/spd_matrices.py
nguigs: same

PR #952, opened 2021-03-29
Sampling on Manifolds

nguigs:
- Fix the random uniform method of SO(n)
- Generalize the von Mises-Fisher distribution to any dimension and any mean point

### DIFF
--- a/geomstats/geometry/special_orthogonal.py
+++ b/geomstats/geometry/special_orthogonal.py
@@ -773,8 +773,7 @@ class _SpecialOrthogonal3Vectors(_SpecialOrthogonalVectors):
         term_1 = (gs.eye(self.dim)
                   + gs.einsum('...,...jk->...jk', coef_1, skew_rot_vec))
 
-        squared_skew_rot_vec = gs.einsum(
-            '...ij,...jk->...ik', skew_rot_vec, skew_rot_vec)
+        squared_skew_rot_vec = GeneralLinear.mul(skew_rot_vec, skew_rot_vec)
 
         term_2 = gs.einsum('...,...jk->...jk', coef_2, squared_skew_rot_vec)
 
--- a/geomstats/geometry/special_orthogonal.py
+++ b/geomstats/geometry/special_orthogonal.py
@@ -259,7 +259,6 @@ class _SpecialOrthogonalVectors(LieGroup):
         """
         return -self.regularize(point)
 
-    @geomstats.vectorization.decorator(['else', 'vector', 'output_point'])
     def exp_from_identity(self, tangent_vec):
         """Compute the group exponential of the tangent vector at the identity.
 
@@ -279,7 +278,6 @@ class _SpecialOrthogonalVectors(LieGroup):
         """
         return self.regularize(tangent_vec)
 
-    @geomstats.vectorization.decorator(['else', 'vector', 'output_point'])
     def log_from_identity(self, point):
         """Compute the group logarithm of the point at the identity.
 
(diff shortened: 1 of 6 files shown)


geomstats/geometry/special_orthogonal.py
```diff
@@ -335,6 +336,56 @@ def vector_from_skew_matrix(self, skew_mat):
         """
         return SkewSymmetricMatrices(self.n).basis_representation(skew_mat)
 
+    def to_tangent(self, vector, base_point=None):
+        return self.regularize_tangent_vec(vector, base_point)
+
+    @geomstats.vectorization.decorator(
+        ['else', 'vector', 'else', 'output_point'])
+    def regularize_tangent_vec_at_identity(
```
ninamiolane, line 344: Hm I know this is not your code, but I don't remember the purpose of `regularize`, `regularize_tangent_vec_at_identity` and `regularize_tangent_vec`: is it expected that they all call `regularize`?

geomstats/geometry/special_orthogonal.py
```diff
@@ -335,6 +336,56 @@ def vector_from_skew_matrix(self, skew_mat):
...
+            Tangent vector at base point.
+        base_point : array-like, shape=[..., 1]
+            Point on the manifold.
+
+        Returns
+        -------
+        regularized_tangent_vec : array-like, shape=[..., 1]
+            Regularized tangent vector.
+        """
+        return self.regularize_tangent_vec_at_identity(tangent_vec)
```
ninamiolane, line 387: Could this be `self.regularize(tangent_vec)` to save one call?

tests/test_hypersphere.py
```diff
@@ -7,7 +7,7 @@
 from geomstats.geometry.hypersphere import Hypersphere
 from geomstats.geometry.matrices import Matrices
 
-MEAN_ESTIMATION_TOL = 1e-6
+MEAN_ESTIMATION_TOL = 1e-3
```
ninamiolane, line 10: Why this change in tolerance?

ninamiolane (approved):
Awesome, thanks! 🎉  Just one question on some code that isn't yours (about `regularize`), and another small comment but it's already good to go and I'll let you merge!

geomstats/geometry/special_orthogonal.py
nguigs: In vector representation, points are in fact tangent vectors so there's no surprise all these methods are the same.

geomstats/geometry/special_orthogonal.py
nguigs: Yes

tests/test_hypersphere.py
nguigs: Because it doesn't work that well when changing the mean... But I guess this is expected from properties of an estimator, given that according to the central limit theorem, the sample mean has variance in sqrt(n), and n is 10^6 here.

geomstats/geometry/special_orthogonal.py
nguigs: DeepSource goes crazy when I try, because of the unused arguments

nguigs: In fact Deepsource didn't raise an error for the unused arguments before because they were used by the decorator... What should we do?

ninamiolane:
@nguigs sorry that you had to go through this DeepSource problem... Let's forget about DeepSource, I think we should remove it and autoformat the codebase with `black`, see issue #918

PR #953, opened 2021-03-29
install from docs/requirements.doc.txt as well

SaitejaUtpala:
Travis.yml is not installing requirements from  docs/requirements.doc.txt. This will fix the 'build-failing' because the sphinx-gallery is added as dependency in  docs/requirements.doc.txt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 install:
   - pip install --upgrade pip setuptools wheel
   - |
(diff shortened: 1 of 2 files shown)


ninamiolane:
NIT: requirements should be in alphabetic order in the ci-requirements.txt file :/ could you just adapt this? Thanks! @Maya95assal i think this will solve the build issue

PR #967, opened 2021-04-09
Features/precision

Maya95assal:
deals with precision


nguigs:
Great! For `test_expm_backward` it seems that one of the outputs is in float32, maybe using float64 would allow to reach the desired precision? For `test_quaternion_and_rotation_vector` and `test_quaternion_and_matrix` I will investigate. For `test_right_exp_coincides`, not sure we can reach 10e-12, 10e-6 is enough in that case

geomstats/_backend/pytorch/__init__.py
```diff
@@ -354,7 +358,7 @@ def sqrt(x):
     return torch.sqrt(x)
 
 
-def isclose(x, y, rtol=1e-5, atol=1e-8):
+def isclose(x, y, rtol=1e-5, atol=atol):
```
ninamiolane, line 361: Would it make sense to have `rtol=rtol` there? @Maya95assal @nguigs

ninamiolane: (Same question for everywhere else in the PR)

tests/test_algebra_utils.py
```diff
@@ -1,10 +1,11 @@
 import math
 
-import geomstats.algebra_utils as utils
 import geomstats.tests
+import geomstats.algebra_utils as utils
```
ninamiolane, line 4: imports should come in alphabetical order as per PEP8 guidelines :)

tests/test_algebra_utils.py
```diff
@@ -27,4 +28,4 @@ def test_all(self):
                 expected = taylor_function['function'](math.sqrt(x))
                 result = utils.taylor_exp_even_func(
                     x, taylor_function, order=4)
-                self.assertAllClose(result, expected, atol=1e-15)
+                self.assertAllClose(result, expected)
```
ninamiolane, line 31: yayyyy 🎉 so much cleaner!

tests/test_dirichlet_distributions.py
```diff
@@ -310,7 +310,7 @@ def test_geodesic(self):
             velocity_norm.max() - velocity_norm.min())
         expected = 0.
 
-        self.assertAllClose(expected, result, atol=1e-4, rtol=1.)
+        self.assertAllClose(expected, result, rtol=1.)
```
ninamiolane, line 313: Aoutch for the rtol here 😆 (it's not the problem of this PR though)

tests/test_general_linear.py
```diff
@@ -127,8 +127,12 @@ def test_exp_vectorization(self):
                              [[2.718281828, 0., 0.],
                               [0., 148.413159, 0.],
                               [0., 0., 403.42879349]]])
+
+        expected = gs.cast(expected, gs.float64)
```
ninamiolane, line 131: 🎉

tests/test_hyperbolic.py
```diff
@@ -70,7 +70,7 @@ def test_projection_to_tangent_space(self):
         result = self.metric.inner_product(tangent_vec, base_point)
         expected = 0.
 
-        self.assertAllClose(result, expected)
+        self.assertAllClose(result, expected, atol=1e-2)
```
ninamiolane, line 73: why this `atol`? wasn't it working before?

tests/test_hyperbolic.py
```diff
@@ -418,6 +418,7 @@ def test_exp_and_dist_and_projection_to_tangent_space(self):
         tangent_vec = self.space.to_tangent(
             vector=vector,
             base_point=base_point)
+        print(tangent_vec)
```
ninamiolane, line 421: Remove this leftover print

tests/test_hyperbolic_coords.py
```diff
@@ -70,21 +70,21 @@ def test_extrinsic_half_plane_extrinsic(self):
 
         x2 = Hyperbolic.change_coordinates_system(
             x_up, 'half-space', 'extrinsic')
-        self.assertAllClose(x, x2)
+        self.assertAllClose(x, x2, gs.atol)
```
ninamiolane, line 73: do we need this `gs.atol` here? Isn't it used by default in `assertAllClose`?

tests/test_spd_matrices.py
```diff
@@ -528,14 +523,18 @@ def test_parallel_transport_affine_invariant(self):
         tan_a = self.space.random_tangent_vec(n_samples, point)
         tan_b = self.space.random_tangent_vec(n_samples, point)
 
+        point = gs.cast(point, gs.float64)
```
ninamiolane, line 526: Why do we need these casts?

tests/test_spd_matrices.py
```diff
@@ -561,13 +560,16 @@ def test_squared_dist_bureswasserstein_vectorization(self):
                             [0., 5., 0.],
                             [0., 0., 1.]])
 
+        point_a = gs.cast(point_a, gs.float64)
```
ninamiolane, line 563: Same question

ninamiolane (commented):
🎉 yessss! very glad to see this issue tackled, thank you!! It is much more clean!

I've put comments directly in the code (very minor ones). We need to get the tests to pass before being able to merge - even if that means temporarily putting some hardcoded `atol=...` in the tests.

tests/test_general_linear.py
Maya95assal: In some cases, casting float32 to float64 allows to reach the desires tolerance

tests/test_hyperbolic_coords.py
Maya95assal: you're right, it's an oversight !

tests/test_hyperbolic.py
Maya95assal: Unfortunately yes.

Maya95assal: It was a mistake. Fixed !

Maya95assal: It was a mistake. Fixed !

geomstats/_backend/pytorch/__init__.py
nguigs: Yes I think we should have a custom value for rtol according to the backend, because the threshold is in fact atol + b * rtol, so if b >> atol, and rtol is not so small, rtol will dominate. See https://numpy.org/doc/stable/reference/generated/numpy.isclose.html
```python
>>> import geomstats.backend as gs # numpy backend here
>>> gs.isclose(gs.array([1. + 1e-5, 1. - 1e-9]), 1., atol=1e-12)
array([ True,  True])
```

geomstats/geometry/poincare_ball.py
```diff
@@ -10,7 +10,7 @@
 from geomstats.geometry.hyperbolic import Hyperbolic
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
-TOLERANCE = 1e-6
+TOLERANCE = gs.atol
```
nguigs, line 13: Maybe nicer to write gs.atol instead of TOLERANCE where it is used to avoid having a global variable (that won't be visible in the documentation whereas gs.atol should be)

geomstats/geometry/pre_shape.py
```diff
@@ -12,7 +12,7 @@
 from geomstats.geometry.quotient_metric import QuotientMetric
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
-TOLERANCE = 1e-6
+TOLERANCE = gs.atol
```
nguigs, line 15: same here

tests/test_kernel_density_estimation_classifier.py
```diff
@@ -10,7 +10,7 @@
     import KernelDensityEstimationClassifier
 from geomstats.learning.radial_kernel_functions import triangular_radial_kernel
 
-TOLERANCE = 1e-4
+TOLERANCE = gs.atol
```
nguigs, line 13: same here

tests/test_stiefel.py
```diff
@@ -84,7 +84,7 @@ def test_belongs_shape(self):
 
     def test_random_uniform_and_belongs(self):
         point = self.space.random_uniform()
-        result = self.space.belongs(point, tolerance=1e-4)
```
nguigs, line 87: Could you also change the kwarg `tolerance` to atol for consistency?

nguigs (commented):
Very nice, so many atol removed!

tests/test_stiefel.py
Maya95assal: Done !

geomstats/geometry/poincare_ball.py
Maya95assal: Done !

Issue #924, opened 2021-03-05
5. As a researcher I need to be able to use double precision with numpy


tristancabel:
PyTorch and Tensorflow use simple precision (32 bits) by default whereas numpy uses double precision (64 bits).
Currently, using geomstat, even with numpy simple precision is used.

We would like to be able to impose double precision on some computations. 
A solution with double precision only supported by numpy is acceptable (other backends would say that they don't support double precision). 

the precision (epsilon) shall be defined by the backend.

  1. ajuster les tests suivant les backends (qq tests 64 bits pour numpy).
  2. Tests sur matrices SPD : validés sur matrices 3x3 mais échouent surmatrices 5x5 (erreur absolue : 1e-2).
   https://github.com/geomstats/geomstats/blob/master/tests/test_spd_matrices.py
  3. Idem sur les espaces hyperboliques (exemple).
   https://github.com/nguigs/geomstats/blob/2021f6dc8253be1a6084f7b58e99dbdee0dc1c4e/tests/test_hyperbolic.py#L415
  4. Passage en double difficile : Runtime Error si DoubleTensor en entrée

PR #968, opened 2021-04-09
Use taylor approx for precision in quaternion to rot vec

nguigs:
+ rewrite inner_product @mvesin 
+ Use taylor approx in quaternion to rot vec and back, this should help #924 @Maya95assal


ninamiolane (commented):
LGTM as long as the tests pass! 🎉  thanks for this!!

PR #970, opened 2021-04-11
GitHub actions test

SaitejaUtpala:
Removed linting and testing from travis and moved it github-actions


.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
+on:
+  push:
+    branches: [master,github-actions-test]
```
ninamiolane, line 3: Oh, nice idea to have a dedicated github-actions-test!

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+    branches: [master]
+
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
```
ninamiolane, line 14: Yes to add more versions: 3.6, 3.7, 3.8 (3.9)? With more recent versions of python, we might lose dependencies: for example, TensorFlow is not yet available in python 3.9: https://www.tensorflow.org/install/pip

.travis.yml
```diff
@@ -19,51 +19,8 @@ before_script:
   - export PYTHONPATH=$PYTHONPATH:$(pwd)
```
ninamiolane, line 19: Do we need this? Can't we remove .travis.yml completely?

ninamiolane (commented):
Wooooohoooo! 🎉 Comments in the review directly.

It'd be great to decide on a solution for the test_notebooks, and add it to this PR, before merging it. 

Also adding the several versions of python, and maybe systems too.

That's all really good, thanks!!

.github/workflows/notebooks_test.yml
```diff
@@ -0,0 +1,47 @@
...
+        run:  |
+          pip install --upgrade pip setuptools wheel
+          for req in ci-requirements.txt dev-requirements.txt opt-requirements.txt requirements.txt; do
+            pip install -q -r $req
+          done  
+          pip install -e .      
+
+      - name: unit testing [nose2]
+        run: |
+          nose2 --with-coverage tests.notebook_test
```
ninamiolane, line 47: Does code-coverage know how to integrate the code coverage of this workflow with the codecoverage of the build workflow?

ninamiolane (approved):
Great, thank you @SaitejaUtpala for this very welcome contribution! 🎉 This is something the package needed for quite some time.

Before merging:

- I put a question on the codecoverage in my comment - just wondering if you had an opinion on this.

- Is it possible to rename the tests subfolders and have: `tests/test_geomstats/test_*.py` and `tests/test_notebooks.py` as the names of folders instead? 
    - It would just be more consistent in the use of `test_` (as opposed to having `_tests` - in plural - or having `test` at the beginning or at the end of a name).
    - It would follow the organization of the package (folders `geomstats` and `notebooks` respectively). 
    - The `test_notebooks.py` doesn't need to be in its own folder in this case: `nose2 --with-coverage tests.test_notebooks` should run well just the same when test_notebooks.py is just a file.

.github/workflows/notebooks_test.yml
SaitejaUtpala: @ninamiolane which codecov should we upload btw. 3.7,3.8,3.9 ?

ninamiolane:
@SaitejaUtpala you could manually increase the rtol/atol to get the last failing checks to pass. Can't wait to merge this!

SaitejaUtpala:
> @SaitejaUtpala you could manually increase the rtol/atol to get the last failing checks to pass. Can't wait to merge this!

@ninamiolane  you can merge it now

PR #971, opened 2021-04-12
Parallel Transport for hyperboloid

nguigs:


geomstats/geometry/hyperboloid.py
```diff
@@ -150,6 +150,28 @@ def to_tangent(self, vector, base_point):
         tangent_vec = vector - gs.einsum('...,...j->...j', coef, base_point)
         return tangent_vec
 
+    def is_tangent(self, vector, base_point=None, atol=TOLERANCE):
```
ninamiolane, line 153: Do we want to wait for Maya's PR and use `atol=gs.atol` there? :/

tests/helper.py
```diff
@@ -87,3 +87,52 @@ def group_exp_then_log(group, tangent_vec, base_point):
     result = group.log(
         point=aux, base_point=base_point)
     return result
+
+
+def test_parallel_transport(space, metric, shape):
```
ninamiolane, line 92: Nice! Good idea - maybe we should think about refactoring more tests like this? Like the `_vectorization` tests...

ninamiolane (approved):
Very nice! Two remarks, but they don't prevent the merge of this PR:
- in `is_tangent`, TOLERANCE could become atol=gs.atol within @Maya95assal 's PR : @Maya95assal if you rebase on `master`, maybe you could try it?
- Very good idea on factoring code within tests, as you did with the helper.py file: we should maybe put more code there?

PR #983, opened 2021-04-18
Corrected a + - sign typo in curvature doc-string (connection.py / in…

xpennec:
…variant_metric.py)

Edited the sectional_curvature doc-string in riemannian_metric.py to display the full formula with non-orthonormal vectors
Added directional_curvature (or tidal operator) in connection.py
Corrected a sign error for top_curvature in quotient_metric.py curvature
Added a unit-test for the sectional curvature of Kendal shape metric in test_pre_shape.py. It add code coverage for the directional_curvature by verifying that the formula gives the same results as sectional_curvature and it verifies that the sectional curvature is greater than the one of the top space (1 in this case because we are on the sphere). A sufficient number of samples is needed to cover statistically all 2-planes.

### DIFF
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -453,11 +453,11 @@ class Connection:
 
     def directional_curvature(
             self, tangent_vec_a, tangent_vec_b, base_point):
-        r"""Compute the directional curvature, better known in relativity as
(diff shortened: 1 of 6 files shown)


ninamiolane (approved):
Great, thank you! LGTM.

PR #985, opened 2021-04-20
test without atol

Maya95assal:


nguigs (approved):
Great! It's nice to have all constants at the same place!
I spotted a last ATOL in [special_orthogonal](https://github.com/Maya95assal/geomstats/blob/d4b9dd0a56bb8b81da474f4bf0aa44ff62603986/geomstats/geometry/special_orthogonal.py#L14), is it intentional to leave it there?

Maya95assal:
Yes, Otherwise it don't pass

PR #986, opened 2021-04-20
numpy backend with non local and local autograd package

tristancabel:


nguigs:
@nkoep what do you think about this? The aim is to add an autograd module to the numpy backend, that relies on the autograd package, but with refactorization of some functions to match tf and pytorch's syntax, as was done for those backends ([here](https://github.com/geomstats/geomstats/blob/master/geomstats/_backend/tensorflow/autograd.py) and [here](https://github.com/geomstats/geomstats/blob/master/geomstats/_backend/pytorch/autograd.py)).

nguigs:
This is dealt by #1139

PR #987, opened 2021-04-20
Spherical Normal

nguigs:
This PR adds the spherical normal distribution and addresses a few design issues:
- test shape of matrices in general_linear, skew_symmetric, special_orthogonal and special_euclidean to avoid errors when comparing arrays
- Make GeneralLinear a LieGroup, and remove inheritance from child classes
- factorize code to rotate points on a sphere : usefull for sampling around a mean that is not the north pole


geomstats/_backend/tensorflow/random.py
```diff
@@ -42,3 +42,7 @@ def uniform(low=0.0, high=1.0, size=None):
     if size is None:
         size = (1,)
     return tf.random.uniform(shape=size, minval=low, maxval=high)
+
+
+def multivariate_normal(*args, **kwargs):
```
ninamiolane, line 47: How about this:
https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/MultivariateNormalFullCovariance

geomstats/algebra_utils.py
```diff
@@ -157,3 +157,33 @@ def flip_determinant(matrix, det):
                 + (1. - mask)[..., None] * ones)
         return gs.einsum('...ij,...j->...ij', matrix, sign)
     return matrix
+
+
+def rotate_points(points, end_point):
```
ninamiolane, line 162: Should this be factorized in a module that implements fundamental actions of Matrix Lie Groups on Rn?

geomstats/geometry/invariant_metric.py
```diff
@@ -708,7 +707,7 @@ def metric_matrix(self, base_point=None):
         jacobian = self.group.jacobian_translation(
             point=base_point, left_or_right=self.left_or_right)
 
-        inv_jacobian = GeneralLinear.inverse(jacobian)
+        inv_jacobian = gs.linalg.inv(jacobian)
```
ninamiolane, line 710: Which logic do we want to use with the linear algebraic operations? We prefer using `Matrices.mul`, `Matrices.transpose`, etc - but we recommend `gs.linalg.inv` instead of `GeneralLinear.inverse`?

geomstats/geometry/matrices.py
```diff
@@ -242,7 +247,7 @@ def is_diagonal(cls, mat, atol=gs.atol):
         return is_diagonal
 
     def random_point(self, n_samples=1, bound=1.):
-        """Sample from a uniform distribution.
+        """Sample from a uniform distribution in a cube.
```
ninamiolane, line 250: Should we rename them to `random_uniform` or `random_uniform_cube` then?

geomstats/geometry/special_euclidean.py
```diff
@@ -7,7 +7,7 @@
 import geomstats.backend as gs
 import geomstats.vectorization
 from geomstats.geometry.euclidean import Euclidean
-from geomstats.geometry.general_linear import GeneralLinear
+from geomstats.geometry.general_linear import GeneralLinear, Matrices
```
ninamiolane, line 10: why not: `from geomstats.geometry.matrices import Matrices`? Written like this it seems that `Matrices` is in the general_linear file? Is it because you were dealing with circular imports?

ninamiolane (approved):
Very nice!! Thanks for this. A few comments.

geomstats/geometry/invariant_metric.py
nguigs: No in general we can use GeneralLinear.inverse, but it lead to circular imports in this case

geomstats/geometry/matrices.py
nguigs: No I think it is just useful to say how points are sample but let's keep `random_point` so that all manifolds have a random_point method

geomstats/geometry/special_euclidean.py
nguigs: No just to avoid adding a line...

geomstats/_backend/tensorflow/random.py
nguigs: Yes, but we have to add the dependence on the tensorflow probability package, it is a separate one

PR #988, opened 2021-04-20
Re arange Riemannian metric's inner_product to speed-up the usual cases

nguigs:
In most cases, the embedding Euclidean inner product is used, whereas it is computed as in a much more general case, involving a metric matrix and resulting in a sum-product of n^2 numbers whereas only n are needed.
This would address #920 while keeping a flexible interface.
For the following dummy example,
```python
from geomstats.geometry.hypersphere import HypersphereMetric
from timeit import timeit
x = gs.array([1., 0., 0., 0.])
v = gs.array([0., 2., 1.5, .4])
metric = HypersphereMetric(3)
print(timeit(lambda: metric.norm(v, x), number=1000))
print(timeit(lambda: gs.linalg.norm(v), number=1000))
```
we obtain 0.007 (metric) against 0.004 (pure backend), whereas on the master branch we have 0.013 vs 0.004

My only concern is, would it raise any issue to define `inner_product` in the `__init__`?
What do you think @ninamiolane, @mvesin, @tristancabel ?

### DIFF
--- a/geomstats/geometry/euclidean.py
+++ b/geomstats/geometry/euclidean.py
@@ -118,8 +118,7 @@ class EuclideanMetric(RiemannianMetric):
 
     def __init__(self, dim, default_point_type='vector'):
         super(EuclideanMetric, self).__init__(
-            dim=dim, signature=(dim, 0, 0),
-            default_point_type=default_point_type)
+            dim=dim, default_point_type=default_point_type)
 
     def metric_matrix(self, base_point=None):
         """Compute the inner-product matrix, independent of the base point.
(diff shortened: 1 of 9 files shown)


mvesin:
@nguigs wouldn't you then also need to change the constructors for sub-classes with non default metric ? eg :
```
class MinkowskiMetric(RiemannianMetric):
    def __init__(self, dim):
        super(MinkowskiMetric, self).__init__(
            dim=dim,
            signature=(dim - 1, 1, 0),
            metric_matrix=self.metric_matrix)
```

nguigs:
> @nguigs wouldn't you then also need to change the constructors for sub-classes with non default metric ?

You're right, I think the signature argument can be used as well

nguigs:
@tristancabel my problem now is that `inner_product` is not overriden in the child classes of `Riemannian_metric`

PR #989, opened 2021-04-21
Override Minkowski and Euclidean inner product for speed

nguigs:
This continues #988 but is much simpler


geomstats/geometry/euclidean.py
```diff
@@ -118,7 +119,7 @@ class EuclideanMetric(RiemannianMetric):
 
     def __init__(self, dim, default_point_type='vector'):
         super(EuclideanMetric, self).__init__(
-            dim=dim, signature=(dim, 0, 0),
+            dim=dim, signature=(dim, 0),
```
ninamiolane, line 122: Don't we use signatures with three elements: number of positive, negative, and 0 eigenvalues of the metric matrix? Why this change?

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -33,10 +33,9 @@ def __init__(self, metrics, default_point_type='vector', n_jobs=1):
 
         sig_0 = sum(sig[0] for sig in signatures)
         sig_1 = sum(sig[1] for sig in signatures)
-        sig_2 = sum(sig[2] for sig in signatures)
         super(ProductRiemannianMetric, self).__init__(
             dim=sum(dims),
-            signature=(sig_0, sig_1, sig_2),
+            signature=(sig_0, sig_1),
```
ninamiolane, line 38: NIT: instead of sig_0, sig_1 we should rename this into `sig_pos` and `sig_neg` (or `sig_neg` and sig_pos`, I forget the order) for clarity.

geomstats/learning/frechet_mean.py
```diff
@@ -286,7 +286,7 @@ def _adaptive_gradient_descent(points,
     if n_points == 1:
         return points[0]
 
-    current_mean = points[0] if init_point is None else init_point
+    current_mean = points[0] if (init_point is None) else init_point
```
ninamiolane, line 289: Do we need the parenthesis?

ninamiolane (approved):
Great! One comment on the signature changes

geomstats/geometry/euclidean.py
nguigs: We actually never use 0 eingenvalues, and I think we should only support non degenerate bilinear forms

nguigs: By the way, I think Minkowski should be renamed Lorentz because Minkowski seems specific to 4d ([wikipedia](https://en.wikipedia.org/wiki/Minkowski_space#Minkowski_metric))

geomstats/geometry/product_riemannian_metric.py
nguigs: done

PR #990, opened 2021-04-22
Dimension of Kendall shape metric

nguigs:
- Fix dimension in KendallShapeMetric
- In most cases `base_point` is not optional in `is_tangent` and `to_tangent`, and a missing argument error is much more informative than one raised by manipulating None

### DIFF
--- a/geomstats/geometry/_my_manifold.py
+++ b/geomstats/geometry/_my_manifold.py
@@ -80,7 +80,7 @@ class MyManifold(Manifold):
         return belongs
 
     # Another example of method of MyManifold.
-    def is_tangent(self, vector, base_point=None):
+    def is_tangent(self, vector, base_point, atol=gs.atol):
         """Check whether vector is tangent to the manifold at base_point.
 
         In what follows, the ellipsis ... indicates either nothing
@@ -96,7 +96,8 @@ class MyManifold(Manifold):
             Vector.
         base_point : array-like, shape=[..., dim]
             Point on the manifold.
-            Optional, default: None.
+        atol : float
+            Absolute tolerance threshold
 
         Returns
         -------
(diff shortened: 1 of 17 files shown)


geomstats/geometry/_my_manifold.py
```diff
@@ -80,7 +80,7 @@ def belongs(self, point, atol=gs.atol):
         return belongs
 
     # Another example of method of MyManifold.
-    def is_tangent(self, vector, base_point=None):
+    def is_tangent(self, vector, base_point=None, atol=gs.atol):
```
ninamiolane, line 83: Do you want to remove `None` from here too?

ninamiolane (approved):
Good, thanks! 

Do you think though that the user might not find it weird if they get a missing `base_point` error, even though they know from the math that the `base_point` does not matter to find out if the vector is tangent (for e.g. Euclidean space)? I let you decide there, but maybe for these few manifolds, we could just remove the `base_point` as an argument?

geomstats/geometry/_my_manifold.py
nguigs: > Do you think though that the user might not find it weird if they get a missing `base_point` error, even though they know from the math that the `base_point` does not matter to find out if the vector is tangent (for e.g. Euclidean space)? I let you decide there, but maybe for these few manifolds, we could just remove the `base_point` as an argument?

So deepsource does not complaint if the overriding method has `base_point=None` where as the overriden one only has `base_point`, so we can adjust to those where it is mathematically true that base point is not required, which was not the case until now as e.g. hypersphere had bp=None.
And no we shouldn't remove the `base_point` argument otherwise learning algorithm with fail when we passe a base_point!

nguigs:
Yes we can keep `base_point=None` when relevant, e.g. in spd_matrices or Lie groups -> default=identity

nguigs:
The last commit only changed Euclidean, yet DeepSource and some tests now fail :angry:

PR #1006, opened 2021-05-18
Architecture of the abstract classes

nguigs:
WIP @ythanwerdas @emaignant 
Clarify abstract classes in geomstats: this is ensured using the ABC built-in class ([info here](https://www.tutorialspoint.com/abstract-base-classes-in-python-abc#:~:text=A%20class%20is%20called%20an,be%20implemented%20by%20its%20subclasses))
The manifolds of geomstats belong to two fundamental classes:
- Open sets of some R^N (GL(n), SPD matrices, beta distrbutions, Poincare ball and half-space etc)
- Pre-image of some submersion (hypersphere, SO(n), etc)

So far, the second case is referred to as `EmbeddedManifold`. I propose to define an abstract class for the first case. So far I have been calling it `OpenSet` we can use `ImmersedManifold` or stg else.
 Then with quotients and products we will cover all the manifolds of the package.


geomstats/geometry/fiber_bundle.py
```diff
@@ -8,7 +9,7 @@
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
 
-class FiberBundle(Manifold):
+class FiberBundle(Manifold, ABC):
```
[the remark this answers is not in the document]
alebrigant: Also, maybe this class should be renamed ```PrincipalBundle``` at some point ? Or if this is supposed to be more general, allow for other types of fibers such as vector spaces ?

geomstats/geometry/hyperbolic.py
```diff
@@ -5,11 +5,10 @@
 
 
 import geomstats.backend as gs
-from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 
 
-class Hyperbolic(Manifold):
+class Hyperbolic:
```
alebrigant, line 11: Why doesn't ```Hyperbolic``` inherit from ```Manifold``` anymore ?

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -84,6 +84,25 @@ def random_point(self, n_samples=1, bound=5.):
...
+        Parameters
+        ----------
+        point : array-like, shape=[..., dim_embedding]
+            Point in ambient space.
+        atol : float
+            Tolerance to evaluate positivity.
+
+        Returns
+        -------
+        projected : array-like, shape=[..., dim_embedding]
```
alebrigant, line 101: Shouldn't the projected point have dimension ```dim``` instead of ```dim_embedding```?

alebrigant (commented):
Great reorganization ! I like the term ```OpenSet```, it's straightforward. I have a few questions.

geomstats/geometry/fiber_bundle.py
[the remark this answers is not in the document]
nguigs: For your first question, I hesitated and tried both. 
The first thing to decide is whether we want FiberBundle to represent a space at all. In the previous versions I had it identified with the total space (but it was not very clear). This avoids creating another object: one for the total space, one for the base, one for the bundle etc. The current implementation of the `PreShapeSpace` does this identification. In this case we need `FiberBundle` to inherit from `Manifold`, to make sure it has the attributes (dim, default_point_type) and implements the usual methods a manifold must have (belongs, to_tangent etc)

geomstats/geometry/hyperbolic.py
nguigs: This class doesnt actually implement any manifold, but wraps the different representations of hyperbolic spaces (`Hyperboloid`, `PoincareBall`, `PoincareHalfSpace`), I need to clarify however how it should interact with these. What I've done so far works to instantiate one of `Hyperboloid`, `PoincareBall`, `PoincareHalfSpace` but `Hyperbolic(2)` does not work.

geomstats/geometry/dirichlet_distributions.py
nguigs: Yes you're right, because we're in an `OpenSet` :)

geomstats/geometry/fiber_bundle.py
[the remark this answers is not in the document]
nguigs: For the second question, I think `FiberBundle` is more general and could exist if we have a submersion that is not necessarily defined by a group action

geomstats/geometry/discrete_curves.py
```diff
@@ -14,7 +15,7 @@
 R3 = Euclidean(dim=3)
 
 
-class DiscreteCurves(Manifold):
+class DiscreteCurves(OpenSet):
```
ninamiolane, line 18: Why didn't we have this as a ProductManifold?

geomstats/geometry/embedded_manifold.py
```diff
@@ -70,7 +128,102 @@ def projection(self, point):
...
+            Vector.
+        base_point : array-like, shape=[..., dim]
+            Point on the manifold.
+
+        Returns
+        -------
+        tangent_vec : array-like, shape=[..., dim]
+            Tangent vector at base point.
+        """
+        return self.ambient_manifold.projection(vector)
```
ninamiolane, line 191: Can the vector space have a projection that is different from the identity function? If not, maybe we should replace this with `return vector` to save some calls?

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
...
         -------
         belongs : array-like, shape=[...,]
-            Boolean evaluating if point belongs to the Minkowski space.
+            Boolean evaluating if point belongs to the manifold.
         """
-        point_dim = point.shape[-1]
-        belongs = point_dim == self.dim
-        if gs.ndim(point) == 2:
-            belongs = gs.tile([belongs], (point.shape[0],))
+        return self.ambient_manifold.belongs(point, atol)
```
ninamiolane, line 39: Yes, but will this be slower?

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
     """
 
     def __init__(self, dim):
-        super(Minkowski, self).__init__(dim=dim)
+        super(Minkowski, self).__init__(dim=dim,
+                                        ambient_manifold=Euclidean(dim))
```
ninamiolane, line 21: Should we rename Euclidean in VectorSpace? or Have a VectorSpace class?

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
...
         self.metric = MinkowskiMetric(dim)
 
-    def belongs(self, point):
-        """Evaluate if a point belongs to the Minkowski space.
+    def belongs(self, point, atol=gs.atol):
+        """Evaluate if a point belongs to the manifold.
```
ninamiolane, line 25: to the Minkowski space? It would make the documentation website more specific when users browse to it.

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
...
         point : array-like, shape=[..., dim]
             Point to evaluate.
+        atol : float
+            Unused here.
 
         Returns
         -------
         belongs : array-like, shape=[...,]
-            Boolean evaluating if point belongs to the Minkowski space.
+            Boolean evaluating if point belongs to the manifold.
```
ninamiolane, line 37: Same.

geomstats/geometry/discrete_curves.py
```diff
@@ -70,6 +75,23 @@ def each_belongs(pt):
...
 
+    def projection(self, point):
+        """Project a point in ambient manifold on manifold.
+
+        This method is for compatibility and returns the input point unchanged.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., dim_embedding]
+            Point in embedding manifold.
```
ninamiolane, line 86: Maybe we should avoid using "embedding manifold" when nothing is inheriting from EmbeddedManifold?

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
...
-        if gs.ndim(point) == 2:
-            belongs = gs.tile([belongs], (point.shape[0],))
+        return self.ambient_manifold.belongs(point, atol)
 
-        return belongs
-
-    def random_point(self, n_samples=1, bound=1.):
-        """Sample in the Minkowski space from the uniform distribution.
+    def projection(self, point):
+        """Project a point in embedding manifold on embedded manifold.
```
ninamiolane, line 42: Maybe we should avoid using "embedding manifold" when nothing is inheriting from EmbeddedManifold?

geomstats/geometry/minkowski.py
```diff
@@ -16,51 +17,40 @@ class Minkowski(Manifold):
...
 
         Parameters
         ----------
-        n_samples: int
-            Number of samples.
-            Optional, default: 1
-        bound : float
-            Side of hypercube support of the uniform distribution.
-            Optional, default: 1
+        point : array-like, shape=[..., dim_embedding]
```
ninamiolane, line 46: dim_embedding -> dim?

geomstats/geometry/poincare_half_space.py
```diff
@@ -29,19 +31,20 @@ class PoincareHalfSpace(Hyperbolic):
...
-    def belongs(self, point):
+    def belongs(self, point, atol=gs.atol):
         """Evaluate if a point belongs to the upper half space.
 
         Parameters
         ----------
         point : array-like, shape=[..., dim]
             Point to be checked.
+        atol : float
+            Absolute tolerance to evaluate positivity of the last coordinate
```
ninamiolane, line 47: NIT: missing end period "."

geomstats/geometry/pre_shape.py
```diff
@@ -33,52 +33,28 @@ class PreShapeSpace(EmbeddedManifold, FiberBundle):
...
+                v, x),
             default_point_type='matrix',
-            total_space=embedding_manifold,
             ambient_metric=PreShapeMetric(k_landmarks, m_ambient))
-        self.embedding_metric = self.embedding_manifold.metric
         self.k_landmarks = k_landmarks
         self.m_ambient = m_ambient
         self.ambient_metric = PreShapeMetric(k_landmarks, m_ambient)
 
-    def belongs(self, point, atol=gs.atol):
```
ninamiolane, line 56: Nice!

geomstats/geometry/riemannian_metric.py
```diff
@@ -72,7 +73,7 @@ def grad(y_pred, y_true, metric):
     return loss_grad
 
 
-class RiemannianMetric(Connection):
+class RiemannianMetric(Connection, ABC):
```
ninamiolane, line 76: Same, why ABC here?

geomstats/geometry/spd_matrices.py
```diff
@@ -4,13 +4,14 @@
...
 import geomstats.vectorization
+from geomstats.geometry.embedded_manifold import OpenSet
 from geomstats.geometry.general_linear import GeneralLinear
 from geomstats.geometry.matrices import Matrices
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 from geomstats.geometry.symmetric_matrices import SymmetricMatrices
 
 
-class SPDMatrices(SymmetricMatrices):
+class SPDMatrices(SymmetricMatrices, OpenSet):
```
ninamiolane, line 14: Do we need SPDMatrices to inherit from SymmetricMatrices now? The corresponding methods could be accessed by `self.ambient_manifold.my_method`? Same question for the other manifolds in this case.

geomstats/geometry/embedded_manifold.py
```diff
@@ -60,6 +117,7 @@ def extrinsic_to_intrinsic_coords(self, point_extrinsic):
         raise NotImplementedError(
             'extrinsic_to_intrinsic_coords is not implemented.')
```
ninamiolane, line 119: Do we want to add to_tangent as an abstractmethod too?

geomstats/geometry/stiefel.py
```diff
@@ -172,6 +144,31 @@ def to_tangent(self, vector, base_point):
         sym_aux = Matrices.to_symmetric(aux)
         return vector - Matrices.mul(base_point, sym_aux)
 
+    def projection(self, point):
```
ninamiolane, line 147: Nice!

geomstats/geometry/stiefel.py
```diff
@@ -172,6 +144,31 @@ def to_tangent(self, vector, base_point):
...
+        projected : array-like, shape=[..., n, p]
+
+        References
+        ----------
+        ..[Absil]   Absil, Pierre-Antoine, and Jérôme Malick.
+                    “Projection-like Retractions on Matrix Manifolds.”
+                    SIAM Journal on Optimization 22, no. 1 (January 2012):
+                     135–58. https://doi.org/10.1137/100802529.
+        """
+        u, _, v = gs.linalg.svd(point)
```
ninamiolane, line 169: NIT: I'd rather avoid single letters as variable names (when it's not n, p, etc). For vim users, single variable names are more complicated to find and replace automatically. `mat_u`, `mat_v`?

geomstats/geometry/stiefel.py
```diff
@@ -343,7 +340,7 @@ def _procrustes_preprocessing(p, matrix_v, matrix_m, matrix_n):
             j_matrix = algebra_utils.from_vector_to_diagonal_matrix(sign)
         return matrix_v
 
-    def log(self, point, base_point, max_iter=30, tol=1e-6):
+    def log(self, point, base_point, max_iter=30, tol=1e-6, **kwargs):
```
ninamiolane, line 343: use gs.atol?

geomstats/geometry/symmetric_matrices.py
```diff
@@ -5,11 +5,11 @@
 import geomstats.backend as gs
 import geomstats.vectorization
 from geomstats import algebra_utils
-from geomstats.geometry.embedded_manifold import EmbeddedManifold
+from geomstats.geometry.embedded_manifold import EmbeddedManifold, OpenSet
 from geomstats.geometry.matrices import Matrices
 
 
-class SymmetricMatrices(EmbeddedManifold):
+class SymmetricMatrices(EmbeddedManifold, OpenSet):
```
ninamiolane, line 12: Why do we need OpenSet here? It seems that its methods are not called?

tests/helper.py
```diff
@@ -141,3 +141,22 @@ def is_isometry(tan_a, trans_a, endpoint):
     results.append(gs.all(result))
 
     return results
+
+
+def test_projection_and_belongs(space, shape):
```
ninamiolane, line 146: Nice!!

geomstats/geometry/symmetric_matrices.py
```diff
@@ -19,9 +19,13 @@ class SymmetricMatrices(EmbeddedManifold):
...
         super(SymmetricMatrices, self).__init__(
             dim=int(n * (n + 1) / 2),
-            embedding_manifold=Matrices(n, n))
+            embedding_manifold=matrices,
+            ambient_manifold=matrices,
+            submersion=matrices.to_skew_symmetric,
+            tangent_submersion=matrices.to_skew_symmetric)
         self.n = n
 
     def belongs(self, mat, atol=gs.atol):
```
ninamiolane, line 31: Shouldn't `belong` be called through `EmbeddedManifold` here?

tests/tests_geomstats/test_quotient_metric.py
```diff
@@ -11,154 +11,163 @@
 from geomstats.geometry.special_orthogonal import SpecialOrthogonal
 
 
+class BuresWassersteinBundle(GeneralLinear, FiberBundle):
```
ninamiolane, line 14: Should this go into the module `geometry`?

ninamiolane (commented):
🎉 🎉 🎉  Refactoring, so courageous! 🤣  🚀  Thanks for the great work, guys!

If I understand correctly, we want to factor the code from the manifolds in the package. The class `EmbeddedManifold` had a bit of this role, but unfortunately does not cover all our manifolds. Thus, `OpenSet` is created for "the other manifolds" that cannot be formulated as an `EmbeddedManifold`, as it turns out that they can be formulated as open sets. Sounds great!

Should any manifold in geomstats inherits from EmbeddedManifold **or** OpenSet only (no inheritance from Manifold, from now on? -- should we just remove the class Manifold?)? EmbeddedManifold and OpenSet essentially have the same methods (or could have): is the **or** above, an exclusive or?

I think the classes `EmbeddedManifold` and `OpenSet` might be difficult to grasp for newcomers. We, as maintainers, may have to show new contributors which class to inherit from each time.

What's happening with the unit tests?

geomstats/geometry/connection.py
```diff
@@ -54,24 +56,6 @@ def christoffels(self, base_point):
         raise NotImplementedError(
```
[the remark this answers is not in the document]
nguigs: Warning: you cannot instantiate a class with an abc.abstractmethod that is not implemented!
As connection is never implemented (except for invariant metric), it's better to delete it. And in fact it's not a tensor so we could never actually implement it

geomstats/geometry/connection.py
```diff
@@ -557,17 +530,6 @@ def path(t):
                 points_at_time_t
         return path
 
-    def torsion(self, base_point):
```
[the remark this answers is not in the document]
nguigs: All our connections are Riemannian...

geomstats/geometry/connection.py
```diff
@@ -413,17 +397,6 @@ def ladder_parallel_transport(
                 'end_point': current_point,
                 'trajectory': trajectory}
 
-    def riemannian_curvature_tensor(self, base_point):
```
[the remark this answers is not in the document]
nguigs: It's never implemented, and a complicated object. I think having curvature that takes as input 3 vectors and returns one is better

geomstats/geometry/dirichlet_distributions.py
```diff
@@ -11,15 +11,15 @@
 import geomstats.backend as gs
 import geomstats.errors
 from geomstats.algebra_utils import from_vector_to_diagonal_matrix
-from geomstats.geometry.embedded_manifold import EmbeddedManifold
+from geomstats.geometry.embedded_manifold import OpenSet
```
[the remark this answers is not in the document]
nguigs: Yes, or we could have a base.py with all abstract classes?

geomstats/geometry/embedded_manifold.py
```diff
@@ -1,9 +1,11 @@
 """Manifold embedded in another manifold."""
+from abc import ABC, abstractmethod
 
+import geomstats.backend as gs
 from geomstats.geometry.manifold import Manifold
 
 
-class EmbeddedManifold(Manifold):
+class EmbeddedManifold(Manifold, ABC):
```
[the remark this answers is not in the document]
nguigs: Yes, two reasons:
- it is also an abstract class; i.e. nothing is actually implemented inside this class
- it defines other abstractmethods specific to embeddedManifold that Manifold does not expose

geomstats/geometry/discrete_curves.py
```diff
@@ -70,6 +75,23 @@ def each_belongs(pt):
...
         return each_belongs(point)
 
+    def projection(self, point):
+        """Project a point in ambient manifold on manifold.
+
+        This method is for compatibility and returns the input point unchanged.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., dim_embedding]
```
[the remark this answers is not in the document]
nguigs: Yes, thanks :)

geomstats/geometry/embedded_manifold.py
```diff
@@ -20,12 +22,67 @@ class EmbeddedManifold(Manifold):
...
+    def __init__(self, dim, embedding_manifold, submersion, tangent_submersion,
+                 default_point_type='vector',
                  default_coords_type='intrinsic', **kwargs):
         super(EmbeddedManifold, self).__init__(
             dim=dim, default_point_type=default_point_type,
             default_coords_type=default_coords_type, **kwargs)
         self.embedding_manifold = embedding_manifold
+        self.embedding_metric = embedding_manifold.metric
+        self.submersion = submersion
+        self.tangent_submersion = tangent_submersion
```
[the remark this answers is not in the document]
nguigs: True that it is redudant, I will try with autodiff

geomstats/geometry/embedded_manifold.py
```diff
@@ -20,12 +22,67 @@ class EmbeddedManifold(Manifold):
...
+            Absolute tolerance.
+            Optional, default: backend atol.
+
+        Returns
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean evaluating if point belongs to the manifold.
+        """
+        belongs = self.embedding_manifold.belongs(point, atol)
+        constraint = gs.isclose(self.submersion(point), 0., atol=atol)
```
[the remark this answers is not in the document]
nguigs: Yes, I made this implicit assumption, but in fact, you're right, we could not make it and have a second argument specifying the singleton of which the manifold is the preimage of. This would solve the shape problem in isclose

geomstats/geometry/embedded_manifold.py
```diff
@@ -20,12 +22,67 @@ class EmbeddedManifold(Manifold):
...
+        atol : float
+            Absolute tolerance.
+            Optional, default: backend atol.
+
+        Returns
+        -------
+        is_tangent : bool
+            Boolean denoting if vector is a tangent vector at the base point.
+        """
+        belongs = self.embedding_manifold.belongs(vector, atol)
```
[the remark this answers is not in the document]
nguigs: I think you're right. This check is mostly to check the shape of the input, and in most cases the embedding_manifold is an openset so it is the same, but for mathematical coherence I will make the change

geomstats/geometry/connection.py
```diff
@@ -1,5 +1,7 @@
 """Affine connections."""
 
+from abc import ABC
```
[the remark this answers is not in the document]
nguigs: ok didn't know

geomstats/geometry/embedded_manifold.py
```diff
@@ -70,7 +128,102 @@ def projection(self, point):
...
+    dim: int
+        Dimension of the manifold. It is often the same as the ambient space
+        dimension but may differ in some cases.
+    ambient_manifold: Manifold
+        Ambient manifold that contains the manifold.
+    """
+
+    def __init__(self, dim, ambient_manifold, **kwargs):
+        super().__init__(dim, **kwargs)
+        self.ambient_manifold = ambient_manifold
```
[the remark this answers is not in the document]
nguigs: Yes, but all our vector spaces are also `Manifold` so far, and it is more general with manifold. Otherwiese we could have a base class for vector spaces (like the matrix one)

geomstats/geometry/discrete_curves.py
nguigs: I don't think it would be efficient if `n_sampling_points` is large

nguigs: I don't think it would be efficient if `n_sampling_points` is large

geomstats/geometry/embedded_manifold.py
nguigs: I think I did this for SPD matrices, where `ambient_manifold=SymmetricMatrices` whose projection is not exactly identity

geomstats/geometry/minkowski.py
nguigs: I can check. It avoids duplicating code

geomstats/geometry/minkowski.py
nguigs: Maybe yes

geomstats/geometry/minkowski.py
nguigs: yes

geomstats/geometry/symmetric_matrices.py
nguigs: Its a particular case because this is a vector space. I thought OpenSet would be useful for tangent vecs?

geomstats/geometry/symmetric_matrices.py
nguigs: Yes :)

tests/tests_geomstats/test_quotient_metric.py
nguigs: Yann implemented the closed-form formulas for the BW metric, that's why I didn't think relevant to add it, but it could be in a notebook to exemplify the construction of a quotient metric

nguigs:
Thanks for the thorough review!

> 🎉 🎉 🎉 Refactoring, so courageous! 🤣 🚀 Thanks for the great work, guys!
> 
> If I understand correctly, we want to factor the code from the manifolds in the package. The class `EmbeddedManifold` had a bit of this role, but unfortunately does not cover all our manifolds. Thus, `OpenSet` is created for "the other manifolds" that cannot be formulated as an `EmbeddedManifold`, as it turns out that they can be formulated as open sets. Sounds great!

Yes I think there really are three cases to construct manifolds: preimage of a submersion, immersion (here always the inclusion) and quotient. The first two really are distince that's why Embedded manifold cannot be used everywhere

> Should any manifold in geomstats inherits from EmbeddedManifold **or** OpenSet only (no inheritance from Manifold, from now on? -- should we just remove the class Manifold?)? EmbeddedManifold and OpenSet essentially have the same methods (or could have): is the **or** above, an exclusive or?

Almost, we can have quotients that are neither open sets or embedded manifolds :)


> Should we put (Manifold, ) Embedded Manifold, and OpenSet in a `base_manifold.py` or `abstract_manifold.py`? I like the name OpenSet too, although having the word `Manifold` in it could be easier for users (they could be confused otherwise as to why this is presumably not a manifold?)?

Yes!

> I think the classes `EmbeddedManifold` and `OpenSet` might be difficult to grasp for newcomers. We, as maintainers, may have to show new contributors which class to inherit from each time.

The point is that those classes are invisible to users. For contributors I will explain everything (in my thesis)

> What's happening with the unit tests?

So as I mentionned you cannot instantiate a subclass if there remains not implemented abstract methods. I still need to add a few of them :)

geomstats/geometry/connection.py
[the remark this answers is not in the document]
ninamiolane: 👍

geomstats/geometry/connection.py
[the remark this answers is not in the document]
ninamiolane: 👍

geomstats/geometry/dirichlet_distributions.py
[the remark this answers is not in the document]
ninamiolane: Yes, I like `base.py`: it will allow to describe the 3 base cases of abstract manifolds classes at once, and refer to your PhD thesis.

We should make a note that "base" does not refer to the base manifold of a fiber bundle... x)

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: Got it!

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: Whichever you prefer, but the docstring states: "In this case, tangent vectors are identified with vectors of the ambient space.", which is the case for `is_tangent`, so this wouldn't work for an ambient manifold?

geomstats/geometry/discrete_curves.py
```diff
@@ -70,6 +75,23 @@ def each_belongs(pt):
 
         return each_belongs(point)
 
+    def projection(self, point):
+        """Project a point in ambient manifold on manifold.
+
+        This method is for compatibility and returns the input point unchanged.
```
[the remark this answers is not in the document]
ninamiolane: Ah, my bad.

For discrete curves, I don't think the `ambient_manifold` has the same meaning as the `ambient_manifold` of OpenSet. For discrete curves, `ambient_manifold` is the manifold to which each of the sampled point belongs.

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: 👍

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: 👍

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: 👍

geomstats/geometry/connection.py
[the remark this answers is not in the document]
ninamiolane: I don't really know either... ^^

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
ninamiolane: Whichever you prefer, but the docstring states: "In this case, tangent vectors are identified with vectors of the ambient space.", which is the case for is_tangent, so this wouldn't work for an ambient manifold?

geomstats/geometry/discrete_curves.py
ninamiolane: Good point!

geomstats/geometry/embedded_manifold.py
ninamiolane: 👍

geomstats/geometry/minkowski.py
ninamiolane: True

geomstats/geometry/symmetric_matrices.py
ninamiolane: Gooot it, for `is_tangent`, `to_tangent`, etc. Thanks! (see my other comment on `is_tangent`, `to_tangent` and the fact that we have to impose that the ambient manifold is a vector space with the current implementation)

tests/tests_geomstats/test_quotient_metric.py
ninamiolane: Yessss 🎉 !

geomstats/geometry/fiber_bundle.py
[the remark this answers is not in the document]
alebrigant: To me, if ```FiberBundle``` were to represent a space, it should be the base manifold (and I see that ```belongs``` checks in priority that a point belongs to the base manifold), because the fiber bundle structure allows to define a geometry on the base space (from (1) the total space, (2) a (group) action and (3) a Riemannian metric on the total space that is invariant wrt the action). I'm guessing this is what you had in mind ? Maybe renaming ```FiberBundle``` into ```QuotientSpace``` would make it more clear ? Then we would have the 3 abstract or base classes that you and @ninamiolane were discussing: ```EmbeddedManifold```, ```OpenSet``` and ```QuotientSpace``` ?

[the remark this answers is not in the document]
alebrigant: Or that would actually make 4 abstract classes with ```Manifold``` ?

geomstats/geometry/hyperbolic.py
alebrigant: Ok I see. Then ```Hyperbolic``` is another abstract class (like ```Manifold```, ```EmbeddedManifold``` etc.) that cannot be instantiated ?

geomstats/geometry/discrete_curves.py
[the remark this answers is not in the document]
nguigs: Yes, but in fact as `n_sampling_points` is not fixed at instantiation, this manifold can be considered of infinite dimension and should not be an `OpenSet`

geomstats/geometry/connection.py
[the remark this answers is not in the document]
nguigs: https://softwareengineering.stackexchange.com/questions/187403/import-module-vs-from-module-import-function

geomstats/geometry/minkowski.py
nguigs: Running time is similar so I will make the change

geomstats/geometry/spd_matrices.py
nguigs: The other manifolds don't inherit from the ambient manifold in general. the methods like expm, powerm are inherited here, which can be useful.. I don't know if you should keep this

geomstats/geometry/embedded_manifold.py
nguigs: It is already from the parent class `Manifold` :)

geomstats/geometry/minkowski.py
nguigs: Minkowski space is really just Euclidean with a different metric, we should have something to create that class without actually writing a new class

geomstats/geometry/embedded_manifold.py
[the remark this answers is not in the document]
nguigs: Ok I changeed to ambiant_space

geomstats/geometry/stiefel.py
nguigs: done

nguigs:
The recent changes:
- Inheritance should only be for abstract classes, not subsets
- I created a MatrixLieGroup class to pass matrix operations to matrix lie groups, but it does not represent GL(n). We will have to clean the original LieGroup class in the future.
- I created a VectorSpace base class
- Hyperbolic is now a common interface to the different models, but not a class on its own as before, it returns the chosen model. There might be some cleaning to do there as well

There should be a new release once we merge this PR as many things would have changed.

PR #1012, opened 2021-05-25
Fix vectorization of geodesic

nguigs:
The previous method `geodesic` did not work when only one base_point was passed.
I am wondering whether the shape of the output should be `(n_initial_conditions, n_times, {dim, [n, n]})` or  `(n_times, n_initial_conditions, {dim, [n, n]})`? So far the second option is returned

### DIFF
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -532,9 +532,8 @@ class Connection:
         n_initial_conditions = initial_tangent_vec.shape[0]
 
         if n_initial_conditions > 1 and len(initial_point) == 1:
-            initial_point = gs.tile(
-                initial_point, n_initial_conditions).reshape(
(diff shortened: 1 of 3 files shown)


ninamiolane (approved):
Thank you for spotting this and fixing it! 🎉  I think `(n_initial_conditions, n_times, {dim, [n, n]})` would be more intuitive indeed, since the geometric object here is the geodesic, and we usually index a list of geometric objects by the 0th dimension. Would you mind adapting this?

nguigs:
> I think `(n_initial_conditions, n_times, {dim, [n, n]})` would be more intuitive indeed, since the geometric object here is the geodesic, and we usually index a list of geometric objects by the 0th dimension. Would you mind adapting this?

Ok done

PR #1015, opened 2021-05-29
Projection to closed curves

Florent-Michel:
Addition of classes for closed curves with a projection method.


geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,150 @@ def dist(self, point_a, point_b):
...
+        if not is_planar:
+            raise AssertionError('The projection is only implemented '
+                                 'for discrete curves embedded in a'
+                                 '2D Euclidean space.')
+
+        n_sampling_points = srv.shape[-2]
+        srv_metric = self.l2_metric(n_sampling_points)
+        ambient_inner_prod = self.ambient_metric.inner_product
+
+        def G(srv, srv_norms):
```
ninamiolane, line 555: Avoid upper case letters fur function names: upper case letters are used for objects or classes (they get colored differently too).

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,150 @@ def dist(self, point_a, point_b):
...
+            return gs.sum(srv * srv_norms[:, None], axis=0)
+
+        initial_norm = srv_metric.norm(srv)
+        proj = srv
+        proj_norms = self.ambient_metric.norm(proj)
+        criteria = atol + 1
+        residual = G(proj, proj_norms)
+
+        while(criteria >= atol):
```
ninamiolane, line 564: You probably don't need the parentheses here

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,150 @@ def dist(self, point_a, point_b):
...
+            J = gs.zeros((2, 2))
```
ninamiolane, line 565: 2 refers to the ambient dimension right? maybe use `ambient_manifold.dim` here, so that 2 is not hard coded.

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,150 @@ def dist(self, point_a, point_b):
...
+        atol : float
+            Tolerance of the projection algorithm.
+            Optional, default: backend atol.
+
+        Returns
+        -------
+        proj : array-like, shape=[..., n_sampling_points, ambient_dim]
+        """
+        is_euclidean = isinstance(self.ambient_metric, EuclideanMetric)
+        is_planar = is_euclidean and self.ambient_metric.dim == 2
```
ninamiolane, line 544: Just curious: Would the code work for any Euclidean space? I thought the paper was giving an algorithm for curves in Rn?

geomstats/geometry/discrete_curves.py
ninamiolane, line 565: Same: avoid uppercase letters for variables: uppercase letters are used for objects and classes in python.

Also, it is good practice to user understandable english names for variables (as opposed to the notations used in the paper), it will improve the readability of your code for the next users and collaborator. For example:

- `G` could be called: `close_criterion`
- `J` could be called: `jacobian_of_criterion`

or maybe some better names, but you get the idea?

ninamiolane (commented):
Very nice!! This is a much-needed addition to the codebase. I have made very minor comments, we will wait for the unit tests. Thanks!

geomstats/geometry/discrete_curves.py
ninamiolane: or - `g_criterion` and `jacobian_of_g` and then explain in a docstring that you are using the notations of the paper.

geomstats/geometry/discrete_curves.py
Florent-Michel: The code doesn't work for Rn yet. But indeed, the paper provides an algorithm for curves in Rn and it shouldn't be too long to modify the code to make it work in dimension n.

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,155 @@ def dist(self, point_a, point_b):
...
+        proj = srv
+        proj_norms = self.ambient_metric.norm(proj)
+        criteria = atol + 1
+        residual = g_criterion(proj, proj_norms)
+
+        while criteria >= atol:
+            g_jacobian = gs.zeros((dim, dim))
+            for i in range(dim):
+                for j in range(dim):
+                    g_jacobian[i, j] = 3 * inner_prod(proj[:, i], proj[:, j])
```
alebrigant, line 572: Maybe avoid computing twice the same inner products by computing for ```j in range(i+1, dim)``` and then transposing would be more efficient ?

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,155 @@ def dist(self, point_a, point_b):
...
+            proj_squared_norm = srv_metric.squared_norm(proj)
+            g_jacobian += proj_squared_norm * gs.array([[1, 0], [0, 1]])
```
alebrigant, line 574: Is the shape of this array linked to the dimension of the ambient space ? If it is, it would be nice to use ```dim``` to make it more general.

geomstats/geometry/discrete_curves.py
```diff
@@ -440,3 +440,155 @@ def dist(self, point_a, point_b):
...
+            e_1, e_2 = gs.array([1, 0]), gs.array([0, 1])
+            grad_1 = proj_norms[:, None] * e_1
+            grad_1 = grad_1 + (proj[:, 0] / proj_norms)[:, None] * proj
+            grad_2 = proj_norms[:, None] * e_2
+            grad_2 = grad_2 + (proj[:, 1] / proj_norms)[:, None] * proj
+
+            b_1 = grad_1 / srv_metric.norm(grad_1)
+            b_2 = grad_2 - srv_metric.inner_product(grad_2, b_1) * grad_2
+            b_2 = b_2 / srv_metric.norm(b_2)
+            b = gs.array([b_1, b_2])
```
alebrigant, line 586: As Nina said, we try to avoid single letter names, so here something like ```basis_vector_1```, ```basis_vector_2```, ```basis``` would be more instructive.

alebrigant (commented):
Great that we finally have this ! Like Nina, a few minor comments while we wait for the unit tests.

geomstats/geometry/discrete_curves.py
nguigs: And please avoid assignment (that won't work with the tensorflow backend and cannot be automatically differentiated) :pray: 
```python
jacobian_vec = []
for i ...:
    for j ...
          jacobian_vec.append(....) # fills in the list with the computed values
jacobian_vec = gs.stack(jacobian_vec) # makes the list a gs.array
g_jacobian = SymmetricMatrices.from_vector(jacobian) # transforms the vector into a symmetric matrix
```
Note that in the snippet, as Alice mentioned, only the values for j >= i need to be computed

alebrigant: Right, j >= i and not j > i as I wrote !

ninamiolane:
@Florent-Michel do you have updates on this? :D

Florent-Michel:
Hi @ninamiolane, sorry for the delay, I just wrote the unit tests related to this PR. However, when I tried to update my branch with git rebase upstream/master as said in the "Contributing" page, it seems like I did something wrong.

nguigs:
> Hi @ninamiolane, sorry for the delay, I just wrote the unit tests related to this PR. However, when I tried to update my branch with git rebase upstream/master as said in the "Contributing" page, it created a lot of commits, which made the PR a bit messy.

Hi @Florent-Michel, I've created a new branch and cherry-picked your commits only so that we can review your PR properly in #1053. If you want to edit it, you can pull the branch with 

```
git checkout -b Florent-curves-new master
git pull https://github.com/nguigs/geomstats.git florent-curves-clean
```
and then open a new PR.

Florent-Michel:
Thanks @nguigs, I pulled the branch, edited it and opened a new PR.

PR #1016, opened 2021-05-31
Add tests for product manifolds

nguigs:
Following up #1006, this PR adds 
- some missing tests 
- a projection method for the Hyperboloid
- GL+(n): positive determinant matrices, within existing GLn to avoid code duplication


geomstats/geometry/general_linear.py
```diff
@@ -1,27 +1,34 @@
...
+    """Class for the general linear group GL(n) and its identity component.
+
+    If `positive_det=True`, this is the connected component of the identity,
+    i.e. the space of matrices with positive determinant.
 
     Parameters
     ----------
     n : int
         Integer representing the shape of the matrices: n x n.
+    positive_det: bool
```
ninamiolane, line 20: NIT: missing space before :

geomstats/geometry/general_linear.py
```diff
@@ -1,27 +1,34 @@
...
+        Whether to restrict to the identity connected component of the
+        general linear group, i.e. matrices with positive determinant.
+        Optional, default: False.
     """
 
-    def __init__(self, n, **kwargs):
+    def __init__(self, n, positive_det=False, **kwargs):
```
ninamiolane, line 26: Why not True by default?

geomstats/geometry/product_riemannian_metric.py
```diff
@@ -108,9 +108,9 @@ def is_intrinsic(self, point):
             raise ValueError(
                 'Invalid default_point_type: \'vector\' expected.')
 
-        if point.shape[1] == self.dim:
+        if point.shape[-1] == self.dim:
```
ninamiolane, line 111: This could be shorten in: `return True` directly, same for below

tests/tests_geomstats/test_hyperbolic.py
```diff
@@ -559,3 +559,15 @@ def test_parallel_transport_vectorization(self):
         results = helper.test_parallel_transport(space, metric, shape)
         for res in results:
             self.assertTrue(res)
+
+    def test_projection_and_belongs(self):
```
ninamiolane, line 563: Not necessarily for this PR, but the unit tests could be refactored too. For example, we could have a TestHyperbolic inherit from TestEmbeddedManifold, and directly run a `test_projection_and_belongs` function on the hyperbolic space. Same for the other embedded manifolds.

ninamiolane (approved):
Nice! LGTM, just a few comments!

geomstats/geometry/general_linear.py
nguigs: So that GeneralLinear is actually the whole of GeneralLinear by default

PR #1026, opened 2021-06-10
Generalise EM algorithms

nguigs:
WIP
This PR aims at making the current implementation of the EM algorithm more general, in order to make it work on other manifolds.
- [x] generalise `dist_broadcast` to all metrics
- [x] move `gmm_pdf` and `weighted_gmm_pdf` to the EM module
- [x]  move `normalization_factor_init`, `find_normalization_factor`, `find_variance_from_index` to the EM module
- [x] Implement the normalisation constant for the hypersphere and other (symmetric) spaces.
- [x] generalise and rename frechet-poincarre-ball method to point_type=matrix


examples/plot_expectation_maximization_ball.py
```diff
@@ -14,7 +14,8 @@
 
 import geomstats.backend as gs
 from geomstats.geometry.poincare_ball import PoincareBall
-from geomstats.learning.expectation_maximization import RiemannianEM
+from geomstats.learning.expectation_maximization import RiemannianEM, \
+    weighted_gmm_pdf
```
ninamiolane, line 18: Should `weighted_gmm_pdf` go in the `sampling` module?

geomstats/_backend/numpy/__init__.py
```diff
@@ -368,14 +370,14 @@ def array_from_sparse(indices, data, target_shape):
...
 
-def erf(x):
-    cst_erf = 8.0 / (3.0 * np.pi) * (np.pi - 3.0) / (4.0 - np.pi)
-    return \
-        np.sign(x) * \
-        np.sqrt(1 - np.exp(-x * x *
-                           (4 / np.pi + cst_erf * x * x) /
-                           (1 + cst_erf * x * x)))
-
+# def erf(x):
```
ninamiolane, line 373: leftover?

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -78,6 +78,7 @@
 cross = tf.linalg.cross
 erf = tf.math.erf
 isnan = tf.math.is_nan
+imag = tf.math.imag
```
ninamiolane, line 81: NIT: imag before isnan

geomstats/geometry/hypersphere.py
```diff
@@ -737,6 +737,98 @@ def curvature(
...
+            sum_term = gs.sum(
+                gs.stack([summand(k)] for k in range(half_dim_2)))
+        else:
+            sum_term = summand(0)
+        coef = area * (- 1.) ** half_dim_2 / 2 ** (dim - 2) * gs.sqrt(
+            gs.pi / 2 / variances)
+
+        return coef * sum_term
+
+    def normalization_factor(self, variances):
```
ninamiolane, line 795: Maybe call it `norm_factor` everywhere, since this is what is used in the name of the gradient?

geomstats/geometry/riemannian_metric.py
```diff
@@ -290,6 +290,53 @@ def dist(self, point_a, point_b, **kwargs):
         dist = gs.sqrt(sq_dist)
         return dist
 
+    def dist_broadcast(self, point_a, point_b):
```
ninamiolane, line 293: Why not just including this in the `dist` function?

ninamiolane (commented):
Very nice until now!! Thanks for doing this! 🎉

examples/plot_expectation_maximization_ball.py
nguigs: It does not sample points from a distribution but evaluates the density function of given points. If the sampling module intends to follow tensorflow/pytorch distribution modules then yes

geomstats/geometry/hypersphere.py
nguigs: But it's less explicit as it can be confused with the norm of the metric

geomstats/geometry/riemannian_metric.py
nguigs: To keep things simple: the output of `dist` is a scalar or 1d, where as `dist_broadcast` is always 2d. This would be more prone to errors

PR #1027, opened 2021-06-15
clean pytorch backend.

SaitejaUtpala:
Pytorch 1.9 has just been released. This PR upgrades torch version to 1.8.1 and cleans pytorch background.


geomstats/_backend/pytorch/__init__.py
```diff
@@ -28,13 +31,15 @@
...
-    gt as greater,
+    outer,
+    greater,
+    hstack,
     int32,
     int64,
     isnan,
     log,
-    lt as less,
+    logical_or,
```
nguigs, line 41: `torch.logical_and` seems to exist as well: https://pytorch.org/docs/1.8.1/generated/torch.logical_and.html?highlight=logical_and#torch.logical_and

geomstats/tests.py
```diff
@@ -68,14 +68,29 @@ def np_and_pytorch_only(test_item):
     import tensorflow as tf
     _TestBaseClass = tf.test.TestCase
 
+if pytorch_backend():
+    import torch
+
+def pytorch_error_msg(a, b):
```
nguigs, line 74: Good idea!

nguigs (approved):
Great!

geomstats/_backend/pytorch/__init__.py
SaitejaUtpala: > `torch.logical_and` seems to exist as well: https://pytorch.org/docs/1.8.1/generated/torch.logical_and.html?highlight=logical_and#torch.logical_and

@nguigs yup. But somehwere in the codebase, `logical_and` is used for non-torch tensor type and so we have to keep what is currently. 

```
def logical_and(x, y):
    if torch.is_tensor(x):
        return x & y
    return x and y

```

SaitejaUtpala: Actually. Quite a bit of code can be removed if we don't use these backend related functions in heterogeneous way  (mixing numpy arrays, torch tensors, lists and passing to backend functions)

geomstats/tests.py
ninamiolane: Yay!! 🎉

opt-requirements.txt
```diff
@@ -1,3 +1,3 @@
 tensorflow>=2.1.2; python_version < '3.8'
+torch==1.8.1
```
ninamiolane, line 2: import order

geomstats/_backend/pytorch/__init__.py
```diff
@@ -155,13 +161,9 @@ def split(x, indices_or_sections, axis=0):
     return torch.split(x, tuple(intervals_length), dim=axis)
 
 
-def logical_or(x, y):
-    return x or y
-
-
 def logical_and(x, y):
```
ninamiolane, line 164: what is happening with codecov? the coverage is shown to decrease on this PR, and for example it seems that the __init__.py file has coverage 0%... :/

ninamiolane (approved):
Awesome work, thank you @SaitejaUtpala !! 🎉 

I'm just confused about the codecoverage: do you know what is happening there?

SaitejaUtpala:
@ninamiolane  RTM.

PR #1032, opened 2021-06-17
Parallel transport on the Grassmann manifold

nguigs:


geomstats/geometry/grassmannian.py
```diff
@@ -185,63 +185,6 @@ def to_tangent(self, vector, base_point):
         sym = Matrices.to_symmetric(vector)
         return Matrices.bracket(base_point, Matrices.bracket(base_point, sym))
 
-    @staticmethod
-    def _check_square(point):
```
ninamiolane, line 189: We don't need these?

ninamiolane (approved):
Nice! LGTM. You removed functions that are not needed?

geomstats/geometry/grassmannian.py
nguigs: Yes they are not needed since #1006

PR #1040, opened 2021-06-22
ToTangentSpace for the Stiefel manifold

nguigs:
The log sometimes results in complex numbers, preventing computation of the Frechet mean


ninamiolane:
Good catch!! I'll wait for the unit tests to pass before looking at the code.

ninamiolane:
@nguigs any update on this? :D

nguigs:
The problem is a 4 x 4 orthogonal matrix with determinant 1, whose matrix log is complex. I couldn't solve it.

Issue #1048, opened 2021-07-01
Inconsistent way of handling metric attribute of  manifold


SaitejaUtpala:
currently, when we pass metric with dimension not equal to dimension of manifold. It would just set dimension of metric to manifold ( But this is not ideal, better way would be throwing error and leaving it to user)

```
@metric.setter
    def metric(self, value):
        if not isinstance(value, RiemannianMetric):
            raise ValueError('The argument must be a RiemannianMetric object')
        if value.dim != self.dim:
            value.dim = self.dim
        self._metric = value
```

So we have
```
from geomstats.geometry.spd_matrices import SPDMatrices, SPDMetricAffine
SPDManifold2 = SPDMatrices(5)
SPDManifold2.metric = SPDMetricAffine(6)
print(SPDManifold2.metric.dim)
-----
15
```

But actual problem is  below,
 Now, instead of setting metric like above, if we pass the (invalid)metric as parameter to manifold. It doesn't throw error or correct the dimension like above, which is inconsistent

```
from geomstats.geometry.spd_matrices import SPDMatrices, SPDMetricAffine
SPDManifold1 = SPDMatrices(5, metric=SPDMetricAffine(6))
SPDManifold1.metric.dim
----
21 
```

nguigs:
I will solve the second part of the issue, we just need to use the setter.
For your first remark, we cannot throw an error because a 'valid' metric can have a different dimension, for example the MatrixMetric (Frobenius), whose default dimension would be n^2, used on the SPD or SO(n) or Grassmann manifold.

SaitejaUtpala:
> I will solve the second part of the issue, we just need to use the setter.
> For your first remark, we cannot throw an error because a 'valid' metric can have a different dimension, for example the MatrixMetric (Frobenius), whose default dimension would be n^2, used on the SPD or SO(n) or Grassmann manifold.

@nguigs 
- yup that solves it.   Also, any reason why `metric` is protected ? once you have setter, making metric protected is meaningless right?  . Based on this much simpler solution would be `self.metric = metric` (instead of `self._metric=metric`) in constructor of `Manifold` (setter would be called automatically).
- oooh, didn't knew that before! Then it becomes tricky.

PR #1051, opened 2021-07-12
Minor fixes

nguigs:
- Remove `OpenSet` inheritance in `SymmetricMatrices` (it's not an open set in the set of Matrices)
- Use the setter when initializing the metric (addresses #1048)


ninamiolane (approved):
LGTM!

PR #1052, opened 2021-07-14
Update contributing.rst

nguigs:
Advise to merge instead of rebase when it's been a while, to avoid the situation of #1015


ninamiolane (approved):
LGTM!

PR #1055, opened 2021-07-14
Projection to closed curves - new

Florent-Michel:
Here is a cleaner PR.


Florent-Michel:
Unfortunately I don't know anything about tensorflow and I don't understand the error message of the failed tests. I am thus open to suggestions to know what I have to change.

nguigs:
It seems you have an input tensor that contains only integers and an other one floats, whereas tf can only compute with tensors of the same type. You can check this by looking at the `dtype` attribute of the tensor, and use `cast` to convert them to the same dtypes. Also if you write a tensor by hand, make sure you use floats instead of integers by adding a dot (`1` is `int` but `1.` is float)

geomstats/geometry/discrete_curves.py
```diff
@@ -325,6 +326,48 @@ def square_root_velocity_inverse(self, srv, starting_point):
 
         return curve
 
+    def srv_inner_product(self, srv_1, srv_2):
```
ninamiolane, line 329: Sorry, I forget. Remind me how this function is different from `SRVMetric.inner_product` and from `SRVMetric.pointwise_inner_product`?

ninamiolane (commented):
🎉 this is awesome, and almost ready to merge!! It would be nice to have some tests for `srv_inner_product`, `srv_norm` and `ClosedDiscreteCurves.project`: do you have ideas for these ones? At the very least, checking that it runs and returns the expected shape for the output array. For project, checking that project of project is the original object.

I'll write to you independently so as to solve the tf error.

Thank you for this great contribution!!

geomstats/geometry/discrete_curves.py
Florent-Michel: `SRVMetric.pointwise_inner_product` takes two tangent vectors [u_1, u_2, ...] and [v_1, v_2, ...] at a base curve and returns an array with the inner products (`ambient_metric.inner_product`) between u_1 and v_1, u_2 and v_2 etc. `SRVMetric.inner_product` is inherited from `RiemannianMetric` and is supposed to compute the inner_product between two tangent vectors at a base point by using `RiemannianMetric.metric_matrix`, which is not implemented (by default and in `SRVMetric`). `SRVMetric.srv_inner_product` takes two srv representations and computes their L² inner product since the elastic metric becomes the L² metric in the srv space.

PR #1065, opened 2021-07-22
Categorical distributions with Fisher-Rao metric

alebrigant:
New clean PR.

The manifold of categorical distributions with $n+1$ outcomes is the $n$-simplex. The Fisher-Rao geometry is just the pullback of the hypersphere metric by the map that takes the square root of each component.

### DIFF
--- a/geomstats/geometry/categorical_distributions.py
+++ b/geomstats/geometry/categorical_distributions.py
@@ -160,6 +160,7 @@ class CategoricalMetric(RiemannianMetric):
         mat = from_vector_to_diagonal_matrix(1 / base_point)
         return gs.squeeze(mat)
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


ninamiolane (approved):
LGTM! A few little comments:
- there is a leftover comment at the beginning of the test file,
- would you have references or math formulae that you could put in the docstrings to help understand the computations?
- do you know why nothing is working for tf/torch?

@nguigs I don't know what you think, but it seems that maybe we should try to start merging code even if it's not torch/tf compliant because it slows down so many contributors... :/

ninamiolane:
@alebrigant @nguigs merging this because I'm scared we'll get github conflicts later on. I'll just raise an issue for the documentation.

PR #1075, opened 2021-07-27
Speedup  from_vector function

SaitejaUtpala:
This PR speeds up SymmetricMatrices's `from_vector` using backend functions I have written for log normal sampling. For numpy,  this obtains 500x speedup (110 seconds to 0.220 seconds) on data with shape (100000, 6)

### DIFF
--- a/geomstats/_backend/numpy/__init__.py
+++ b/geomstats/_backend/numpy/__init__.py
@@ -408,9 +408,12 @@ def mat_from_diag_triu_tril(diag, tri_upp, tri_low):
     return mat
 
 def vec_to_sym(vec, mat_dim):
-    shape = (vec.shape[0] , ) + (mat_dim, mat_dim)  
+    shape = (mat_dim, mat_dim)  
+    shape = (vec.shape[0],) + shape if vec.ndim>1 else shape
     mat = np.zeros(shape)
-    i, j = np.triu_indices(mat_dim , k=0)
-    mat[... , i, j] = vec
-    sym = 0.5 * ( mat + np.tranpose(mat, [0,2,1]))
+    i, = np.diag_indices(mat_dim, ndim=1)
+    j, k = np.tril_indices(mat_dim)
+    mat[... , j, k] = vec
+    sym = mat+ mat.swapaxes(-1,-2)
+    sym[..., i, i] *= 0.5
     return sym
--- a/geomstats/_backend/numpy/__init__.py
+++ b/geomstats/_backend/numpy/__init__.py
@@ -406,3 +406,11 @@ def mat_from_diag_triu_tril(diag, tri_upp, tri_low):
     mat[..., j, k] = tri_upp
     mat[..., k, j] = tri_low
     return mat
+
+def vec_to_sym(vec, mat_dim):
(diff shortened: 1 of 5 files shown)


geomstats/_backend/pytorch/__init__.py
```diff
@@ -804,3 +804,15 @@ def mat_from_diag_triu_tril(diag, tri_upp, tri_low):
...
     return mat
+
+
+def vec_to_sym(vec, mat_dim):
+    shape = (mat_dim, mat_dim)  
+    shape = (vec.shape[0],) + shape if vec.ndim>1 else shape
+    mat = torch.zeros(shape)
+    i, i = gs.diag_indices(mat_dim)
+    j, k = gs.tril_indices(mat_dim)
+    mat[... , j, k] = vec
```
nguigs, line 815: will this be differentiable by automatic differentiation?
This is why the former methods in SymmetricMatrix were written without assignment.

geomstats/_backend/numpy/__init__.py
```diff
@@ -406,3 +406,14 @@ def mat_from_diag_triu_tril(diag, tri_upp, tri_low):
...
     mat[..., k, j] = tri_low
     return mat
+
+def vec_to_sym(vec, mat_dim):
+    shape = (mat_dim, mat_dim)  
+    shape = (vec.shape[0],) + shape if vec.ndim>1 else shape
+    mat = np.zeros(shape)
+    i, = np.diag_indices(mat_dim, ndim=1)
+    j, k = np.tril_indices(mat_dim)
+    mat[... , j, k] = vec
```
nguigs, line 416: Is this differentiable?

nguigs (commented):
It would be great but we need to avoid assignment to keep automatic differentiation working!

geomstats/_backend/pytorch/__init__.py
SaitejaUtpala: @nguigs  
It works actually. pytorch is smart. The reason is twofold.

indexes that the diag_indices returns is int64 torch tensors and importantly
There is differentiable operation called <IndexPutBackward> (whose gradient is straight-through). Here is the working example
```
ind, = diag_indices(3, ndim =1)
mat = torch.zeros((3,3))
diag = torch.tensor([2., 3., 4.], requires_grad=True)
squared_diag = diag**2
mat[ind,ind] = squared_diag
sum = mat.sum()
sum.backward()
print(diag.grad)

```

now grad tensor should be 2*diag (we have squared and added) if everything is correct. output is indeed tensor([4., 6., 8.])

geomstats/_backend/numpy/__init__.py
SaitejaUtpala: This, I am not sure. I have never used old `autograd`   :) . I will modify this probably. But pytorch one works

ninamiolane:
@SaitejaUtpala Such a speed-up would be fantastic! Any update on having this pass in the 4 backends?

If the speed-up can only work for numpy/autograd/pytorch, we can also put the faster computations in the corresponding backend, and keep the older implementations in the other backends.

PR #1078, opened 2021-07-28
General elastic metric

Florent-Michel:


geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+            det = gs.linalg.det(gs.array([last_point, point]))
+            orientation = gs.sign(det)
+            arg = args[-1] + orientation * angle
+            args.append(arg)
+
+        args = gs.array(args)
+
+        return norms, args
+
+    def polar_to_cartesian(self, norms, args):
```
ninamiolane, line 914: You can remove `self` and make it a `@staticmethod` (cf remaining DeepSource issue)

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
             nb_iter += 1
 
         return proj
+
+
+class ElasticMetric(RiemannianMetric):
```
ninamiolane, line 847: 🎉 Excellent!!

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+        curve_x = gs.cos(args)
+        curve_y = gs.sin(args)
+        unit_curve = gs.vstack((curve_x, curve_y)).T
+        curve = norms[:, None] * unit_curve
+
+        return curve
+
+    def f_transform(self, curve):
+        """Compute the F_transform of a curve.
+
```
ninamiolane, line 938: Could you add the formula  using :math: in the docstring here? (see other docstrings in geomstats where this has been done).

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+        Parameters
+        ----------
+        curve : array-like, shape=[..., n_sampling_points, ambient_dim]
+            Discrete curve.
+
+        Returns
+        -------
+        f : array-like, shape=[..., n_sampling_points - 1, ambient_dim]
+            F_transform of the curve..
```
ninamiolane, line 949: Nit: only one period after curve.

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+        f_args = args * self.a / (2 * self.b)
+        f_norms = 2 * self.b * gs.sqrt(speeds)
+        f = self.polar_to_cartesian(f_norms, f_args)
+
+        return f
+
+    def f_transform_inverse(self, f, starting_point):
+        """Compute the inverse F_transform of a transformed curve.
+
+        Only works if a/2b <= 1.
```
ninamiolane, line 963: Maybe add a test in this function, that will raise an error if a/2b>1?

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+
+        f_args = args * self.a / (2 * self.b)
+        f_norms = 2 * self.b * gs.sqrt(speeds)
+        f = self.polar_to_cartesian(f_norms, f_args)
+
+        return f
+
+    def f_transform_inverse(self, f, starting_point):
+        """Compute the inverse F_transform of a transformed curve.
+
```
ninamiolane, line 962: If we have an analytic form for the inverse, could you add it in :math: in the docstring too? thank you! 🙌

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+        if rescaled:
+            cosine = l2_prod(f_1, f_2) / (4 * self.b**2)
+            distance = 2 * self.b * gs.arccos(gs.clip(cosine, -1, 1))
+        else:
+            distance = l2_dist(f_1, f_2)
+
+        return distance
+
+
+class ElasticCurves(Manifold):
```
ninamiolane, line 1052: Why do we need a new class here? How about: adding the elastic metric as an attribute of the existing classes of `DiscreteCurves` and/or `ClosedDiscreteCurves` (I don't remember if it would work in the closed curves case?).

ninamiolane (commented):
Excellent!! The code is very clean, thank you 🎉 

There are:
- 1 deepsource error
- 1 comment about whether we really need a new ElasticCurve class
- minor comments here and there
- errors in pytorch and tf: if you can solve them, that would be amazing! otherwise we can use the decorators.

Ready to merge soon IMO! Congrats!

geomstats/geometry/discrete_curves.py
Florent-Michel: Indeed, I just thought DiscreteCurves was reserved for the srv metric.

geomstats/geometry/discrete_curves.py
```diff
@@ -842,3 +842,340 @@ def g_criterion(srv, srv_norms):
...
+        Stretching parameter.
+
+    References
+    ----------
+    .. [KN2018] S. Kurtek and T. Needham,
+    "Simplifying transforms for general elastic metrics on the space of
+    plane curves", arXiv:1803.10894 [math.DG], 29 Mar 2018.
+    """
+
+    def __init__(self, a, b):
```
ninamiolane, line 866: Should we call the SRVMetric if a = 1 and b = 1? Or at least explain in the docstring how the two metrics relate? @Florent-Michel

Florent-Michel: `SRVMetric` and `ElasticMetric(1, 0.5)` are indeed almost the same (up to a factor which depends on the number of sampling points) because I made different normalization choices

alebrigant: I think that for coherence it is important that the results of ```SRVMetric``` and ```ElasticMetric(1, 0.5)``` are the same. If it's just a normalization choice, can it be easily changed in ```SRVMetric``` ?

geomstats/geometry/discrete_curves.py
Florent-Michel: Actually I'm not sure of how I should add the elastic metric as an attribute of `DiscreteCurves` since it only works for planar discrete curves (i.e. `ambient_manifold = Euclidean(dim=2)`.

ninamiolane: You could add `if dim == 2: self.elastic_metric = ...`?

ninamiolane:
@Florent-Michel any update on this?

Issue #1069, opened 2021-07-23
Make RiemannianMetric work when we only specify metric_matrix(self, ...) in child class


ninamiolane:
With the current implementation, if a metric that inherits from `RiemannianMetric` is created, with only the `metric_matrix` method, the implementation of the christoffel symbols will not work because of the automatic differentiation performed with autograd and almost not tested.

PR #1085, opened 2021-08-11
Fix computations of Christoffel symbols for Levi-Civita connection + Add tests for Riemannian metrics

ninamiolane:
This PR:
- fixes the computations of the christoffel symbols of the Levi-Civita connection in the RiemannianMetric class,
- adds the formula corresponding to this computation in the associated docstring,
- creates test_riemannian_metric.py test file,
- moves tests from test_connection.py that belong to test_riemannian_metric.py,
- adds tests for the `RiemannianMetric` class, by building "new" toy euclidean and sphere metrics, providing only the `metric_matrix` function to 2 new instances of the `RiemannianMetric` class.

This PR also relaxes the flake8 linting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04]
-        python-version: [3.6,3.7,3.8]
+        python-version: [3.6,3.7,3.8, 3.9]
         geomstats-backend : ['numpy','pytorch','tensorflow']
         test-folder : ['tests.tests_geomstats' , 'tests.test_notebooks']
       fail-fast: false  


ninamiolane:
This addresses issue #1069

.github/workflows/build.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       - name: linting [flake8]
         if: ${{matrix.test-folder == 'tests.tests_geomstats' && matrix.python-version == 3.8 && matrix.geomstats-backend == 'tensorflow'}}
         run: |
-          flake8 --ignore=D,W503,W504 geomstats examples tests
-          flake8 geomstats/geometry geomstats/learning 
+          flake8 --ignore=D,W292,W503,W504, geomstats examples tests
```
nguigs, line 54: This is not the most cumbersome convention to respect, and usual IDEs also highlight this, so why do you want to remove it?

geomstats/geometry/riemannian_metric.py
```diff
@@ -484,3 +438,61 @@ def sectional_curvature(
...
+    inner_prod_mat = metric.metric_matrix(base_point=y_pred)
+    is_vectorized = inner_prod_mat.ndim == 3
+    axes = (0, 2, 1) if is_vectorized else (1, 0)
+
+    loss_grad = gs.einsum(
+        '...i,...ij->...i',
+        grad_vec,
+        gs.transpose(inner_prod_mat, axes=axes))
+
+    return loss_grad
```
nguigs, line 498: First, I never understood why these methods were here, what is their purpose?
Second, it would be much more useful to have one single method that returns both loss and grad, to avoid computing the loss twice!
Finally, why do we need the contraction with the metric matrix?

geomstats/geometry/riemannian_metric.py
```diff
@@ -161,18 +111,22 @@ def christoffels(self, base_point):
         christoffels: array-like, shape=[..., dim, dim, dim]
             Christoffel symbols.
         """
-        cometric_mat_at_point = self.inner_product_inverse_matrix(base_point)
+        cometric_mat_at_point = self.inner_product_inverse_matrix(
```
nguigs, line 114: Maybe `inner_product_inverse_matrix` should be renamed to `metric_inverse_matrix` for consistency and brevity?

nguigs (approved):
Great that we now know it works! A few comments that are maybe more general than this PR

geomstats/geometry/riemannian_metric.py
ninamiolane: Definitely!

geomstats/geometry/riemannian_metric.py
ninamiolane: It's some legacy code coming from a deep learning pipeline predicting on SO(3) - haha. Yes, they should go into a notebook to showcase this usecase, or be deleted.

.github/workflows/build.yml
ninamiolane: Will put them back.

PR #1086, opened 2021-08-11
Implement pullback metric for an input immersion + tests on the 2-sphere

ninamiolane:
Note: this PR depends on the PR #1085 as it requires the fix for the christoffels of the Levi-Civita connection.

This PR:
- adds a module pullback_metric.py that implements the pullback metric
- adds the corresponding tests in test_pullback_metric.py.

The pullback metric is only defined for manifolds immersed in some Euclidean space. The pullback metric only requires the immersion function as input. The rest of the computations are performed through automatic differentiation with autograd, hence only working with the numpy backend.

The tests compare:
- computations coming from the methods of the class HypersphereMetric(dim=2), written in closed forms
- computations coming from the methods of a PullbackMetric instantiated with the immersion defining the 2-sphere in R^3
and check that the results correspond.


geomstats/geometry/pullback_metric.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    def __init__(
+            self, dim, embedding_dim, immersion,
+            jacobian_immersion=None, tangent_immersion=None):
+        super(PullbackMetric, self).__init__(dim=dim)
+        self.embedding_metric = EuclideanMetric(embedding_dim)
+        self.immersion = immersion
+        if jacobian_immersion is None:
+            jacobian_immersion = autograd.jacobian(immersion)
```
nguigs, line 44: This can work with other backends with `gs.autograd.jacobian`

geomstats/geometry/pullback_metric.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        immersed_base_point = self.immersion(base_point)
+        jacobian_immersion = self.jacobian_immersion(base_point)
+        all_lines = []
+        basis_elements = gs.eye(self.dim)
+        for i in range(self.dim):
+            line_i = []
+            immersed_basis_element_i = gs.matmul(
+                jacobian_immersion, basis_elements[i])
+            for j in range(self.dim):
```
nguigs, line 83: We can loop over the upper triangular part of the matrix, then reconstruct it using `SymmetricMatrices.from_vec` to avoid redundancy. This is done in e.g. dist_pairwise: https://github.com/geomstats/geomstats/blob/90d9954f525fe6e5b6da625205657675a3baa830/geomstats/geometry/riemannian_metric.py#L377-L380

tests/tests_geomstats/test_pullback_metric.py
```diff
@@ -0,0 +1,307 @@
...
+
+class TestPullbackMetric(geomstats.tests.TestCase):
+    def setUp(self):
+        warnings.simplefilter('ignore', category=UserWarning)
+        gs.random.seed(0)
+        self.dim = 2
+        self.sphere = Hypersphere(dim=self.dim)
+        self.sphere_metric = self.sphere.metric
+
+        def _sphere_immersion(spherical_coords):
```
nguigs, line 19: Why didn't you use `Hypersphere.spherical_to_extrinsic` directly?

tests/tests_geomstats/test_pullback_metric.py
```diff
@@ -0,0 +1,307 @@
...
+    def test_jacobian_immersion(self):
+        def _expected_jacobian_immersion(point):
+            theta = point[..., 0]
+            phi = point[..., 1]
+            jacobian = gs.array([
+                [gs.cos(phi) * gs.cos(theta), - gs.sin(phi) * gs.sin(theta)],
+                [gs.sin(phi) * gs.cos(theta), gs.cos(phi) * gs.sin(theta)],
+                [-gs.sin(theta), 0.]
+            ])
+            return jacobian
```
nguigs, line 68: Why didn't you use `Hypersphere.tangent_spherical_to_extrinsic`?

nguigs (approved):
This is great!
It seems you've hard coded most of the points for the unit tests, whereas in many other modules random points are used, maybe it is more general and could help spotting more issues? Of course the seed should be set to avoid tol problems

tests/tests_geomstats/test_pullback_metric.py
ninamiolane: To avoid relying on other part of the code, i.e. to really have a unit-test.

tests/tests_geomstats/test_pullback_metric.py
ninamiolane: Same.

geomstats/geometry/pullback_metric.py
ninamiolane: Good point, thanks, I will modify it (in a following PR).

geomstats/geometry/pullback_metric.py
ninamiolane: Good point, thanks!

geomstats/geometry/pullback_metric.py
ninamiolane: @nguigs I couldn't use `SymmetricMatrices.from_vector` because the coo_matrix function relies on scipy that will not accept ArrayBox structure from autograd. I have changed it to use joblib though.

PR #1090, opened 2021-08-14
Add environment.yml for conda

ninamiolane:
This PR:
- adds an `environment.yml` file so that users can easily install the requirements for geomstats into a conda environment,
- adds an explanation into the README about it.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,15 @@
+name: geomstats-3.8
+channels:
+  - conda-forge
+dependencies:
+  - jupyter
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -0,0 +1,15 @@
+name: geomstats-3.8
+channels:
+  - conda-forge
+dependencies:
+  - jupyter
+  - joblib==0.14.1
```
nguigs, line 6: joblib is already in sklearn

PR #1121, opened 2021-08-23
pytest migration

SaitejaUtpala:
this pr moves from nose2 to pytest

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,6 +61,6 @@ jobs:
           pytest --cov-report term --cov=geomstats ${{matrix.test-folder}} 
           
       - name: uploading code coverage [codecov]
-        if: ${{matrix.python-version == 3.7 && matrix.test-folder == 'tests.tests_geomstats'}}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -54,13 +54,13 @@ jobs:
           flake8 --ignore=D,W503,W504 geomstats examples tests
           flake8 geomstats/geometry geomstats/learning 
 
-      - name: unit testing [nose2]
+      - name: unit testing [pytest]
         env: 
           GEOMSTATS_BACKEND : ${{matrix.geomstats-backend}} 
         run: |
-          nose2 --with-coverage ${{matrix.test-folder}}
+          pytest --cov-report term --cov=geomstats ${{matrix.test-folder}}
```
ninamiolane, line 61: I'm confused here: does this mean that the coverage is only reported when running the `matrix.test-folder`, i.e. just tests/tests_geomstats or tests/test_notebooks? 

I know it was already like this in the former build.yml. But should we hardcode tests/test_geomstats as the folder here?

ninamiolane (commented):
Awesome!! 🎉  Thank you so much for doing this, it is exciting that we don't actually need to change the test files.

I have only one question within the build.yml.

PR #1133, opened 2021-08-30
Quotient SRV Metric

alebrigant:
Create class ```QuotientSRVMetric``` induced on the space of shapes = space of curves quotiented by the group of reparametrizations.


geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +618,47 @@ def dist(self, point_a, point_b, **kwargs):
...
+        vec_1 = gs.squeeze(gs.hstack((
+            gs.array([[-1.]]),
+            gs.zeros((1, n_points - 2)),
+            gs.array([[1.]]))))
+        vec_2 = gs.squeeze(gs.hstack((
+            1 / 2 * gs.ones((1, n_points - 2)),
+            gs.array([[1.]]))))
+        vec_3 = gs.squeeze(gs.hstack((
+            gs.array([[1.]]),
+            1 / 2 * gs.ones((1, n_points - 2)))))
```
nguigs, line 646: Why not `vec_3 = gs.array([1.] + [1. / 2] * (n_points - 2))` for readibility?

geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +618,47 @@ def dist(self, point_a, point_b, **kwargs):
...
+
+        Returns
+        -------
+        space_deriv : array-like, shape=[...,n_points, ambient_dim]
+        """
+        n_points = curve.shape[-2]
+
+        vec_1 = gs.squeeze(gs.hstack((
+            gs.array([[-1.]]),
+            gs.zeros((1, n_points - 2)),
```
nguigs, line 639: Did you make sure somewhere that `n_points >=2`?

geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +618,47 @@ def dist(self, point_a, point_b, **kwargs):
 
         return dist
 
+    # TODO (alebrigant): generalize from_vector_to_diagonal_matrix to k
+    # diagonals and modify tensorflow vstack and hstack.
```
nguigs, line 622: What do you mean? `from_vector_to_diagonal_matrix` works for if you input a list of size `n_samples` vectors, i.e. an array of shape `(n_samples,  dim)`, and returns a list of diagonal matrices (shape `(n_samples, dim, dim)`

geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +618,47 @@ def dist(self, point_a, point_b, **kwargs):
...
+    def space_derivative(self, curve):
+        """Compute space derivative of curve using centered differences.
```
nguigs, line 624: Did you try this numpy method: https://numpy.org/doc/stable/reference/generated/numpy.gradient.html
I don't know about the other backends though

geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +618,47 @@ def dist(self, point_a, point_b, **kwargs):
...
+            gs.zeros((1, n_points - 2)),
+            gs.array([[1.]]))))
+        vec_2 = gs.squeeze(gs.hstack((
+            1 / 2 * gs.ones((1, n_points - 2)),
+            gs.array([[1.]]))))
+        vec_3 = gs.squeeze(gs.hstack((
+            gs.array([[1.]]),
+            1 / 2 * gs.ones((1, n_points - 2)))))
+
+        mat_1 = from_vector_to_diagonal_matrix(vec_1)
```
nguigs, line 648: For `mat_1` you can use `gs.array_from_sparse([(0, 0), (n_points - 1,) * 2 ], [-1, 1], [n_points, n_points])`

tests/tests_geomstats/test_discrete_curves.py
```diff
@@ -292,25 +292,25 @@ def test_is_tangent_to_tangent(self):
             tangent_vec, point)
         self.assertTrue(gs.all(result))
 
-    @geomstats.tests.np_autograd_and_torch_only
-    def test_projection_closed_curves(self):
-        """Test that projecting the projection returns the projection
+    # @geomstats.tests.np_autograd_and_torch_only
```
nguigs, line 295: Leftover comments

nguigs (commented):
Great! A few comments to improve the readability of the code!
Did you have a look at the FiberBundle and QuotientMetric classes? I'd be interested to know if it was possible to use it!

geomstats/geometry/discrete_curves.py
alebrigant: I wasn't aware of this syntax ! This was a horrible workaround an error in tensorflow but your suggestion is much better.  Thanks !!

geomstats/geometry/discrete_curves.py
alebrigant: Good point.

geomstats/geometry/discrete_curves.py
alebrigant: This should work, I'll try it out.

geomstats/geometry/discrete_curves.py
alebrigant: Thanks I didn't know we had this function. I'm not sure it really improves readability though in this case.

alebrigant:
> Great! A few comments to improve the readability of the code!
> Did you have a look at the FiberBundle and QuotientMetric classes? I'd be interested to know if it was possible to use it!

Thanks for the comments ! It would definitely make sense to have ```QuotientSRVMetric``` inherit from ```QuotientMetric```, but I'm confused about one thing: the splitting of a tangent vector in vertical and horizontal parts depend on the Riemannian metric that you define on the total space, since the horizontal space is the orthogonal of the vertical space w.r.t. that metric. So shouldn't the ```horizontal_projection``` and others be in ```QuotientMetric``` instead of ```FiberBundle``` ?

nguigs:
> > Did you have a look at the FiberBundle and QuotientMetric classes? I'd be interested to know if it was possible to use it!
> 
> Thanks for the comments ! It would definitely make sense to have `QuotientSRVMetric` inherit from `QuotientMetric`, but I'm confused about one thing: the splitting of a tangent vector in vertical and horizontal parts depend on the Riemannian metric that you define on the total space, since the horizontal space is the orthogonal of the vertical space w.r.t. that metric. So shouldn't the `horizontal_projection` and others be in `QuotientMetric` instead of `FiberBundle` ?

Yes you are right that it depends on the metric. But the `FiberBundle` class is not like a `Manifold`: the metric of the total space is fixed when initializing the class.

geomstats/geometry/discrete_curves.py
```diff
@@ -326,6 +331,47 @@ def square_root_velocity_inverse(self, srv, starting_point):
 
         return curve
 
+    def d_square_root_velocity(self, tangent_vec, curve):
```
ninamiolane, line 334: For consistency with our current namings of differentials, this should be called:
`differential_square_root_velocity`, even if it's a bit long...

See .e.g here:
https://github.com/geomstats/geomstats/blob/14f181ae9d801c729037698e8926587c9f83163b/geomstats/geometry/spd_matrices.py#L133

geomstats/geometry/discrete_curves.py
```diff
@@ -2,8 +2,11 @@
 
 import math
 
+import numpy as np
```
ninamiolane, line 5: Why the numpy import? Do you need anything from the backends?

ninamiolane: In the worst case, let's still import gs. instead of np. but only run the tests with numpy (and/or autograd) backends

geomstats/geometry/discrete_curves.py
ninamiolane: PR on this has been merged! 🎉

geomstats/geometry/discrete_curves.py
```diff
@@ -326,6 +331,47 @@ def square_root_velocity_inverse(self, srv, starting_point):
...
+        d_vec_tangential = gs.einsum(
+            'ij,i->ij', unit_velocity_vec,
+            self.ambient_metric.inner_product(d_vec, unit_velocity_vec))
+        d_srv_vec = d_vec - 1 / 2 * d_vec_tangential
+        d_srv_vec = gs.einsum(
+            'ij,i->ij', d_srv_vec, 1 / velocity_norm**(1 / 2))
+
+        return d_srv_vec
+
+    def inner_product(self, tangent_vec_a, tangent_vec_b, curve):
```
ninamiolane, line 357: Why not calling `srv_inner_product` here?

Additionally, I wonder if there would be something interesting to be done with `PullbackMetric`:
https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/pullback_metric.py

But maybe let's keep this discussion for another time.

ninamiolane:
Awesome discussion on the use of the new QuotientMetric class!

How about the following structure:
- Have `DiscreteCurve` inherit from `FiberBundle`. Within the `DiscreteCurves`'s constructor, the `FiberBundle`'s constructor takes as input:
    - (1) `ambient_metric`, which can be an argument of `DiscreteCurve`'s constructor, defaulting to the `SRVMetric`.
    - (2) `group_action`, which would be the action of the reparameterization.
Note that we can quotient the action of translation, scaling and rotation with the Kendall module, so it would not be included there.
 
@alebrigant , the `split_horizontal_vertical` from your PR could go as a method of `DiscreteCurve` directly, overwriting the methods `horizontal_projection`, and `vertical_projection` from its parent class `FiberBundle` . Your code on horizontal geodesic and path could go there too. If your code only works with the `SRVMetric` (and not the general ElasticMetric), then there could be an `if/else` condition checking for the `self.ambient_metric` argument and returning `NotImplementedError` if it is not `SRVMetric`.

- Your new metric could be called `ReparameterizedDiscreteCurveMetric`, and inherits from `QuotientMetric`, where the `fiber_bundle` attribute of the `QuotientMetric` class is the corresponding `DiscreteCurve`?

Note: How about we use `ReparameterizedDiscreteCurveMetric` instead of `ShapeDiscreteCurveMetric`, and keep the word "shape" for Kendall shapes i.e. for the quotients with scalings, rotations, and translations?

Not for this PR: It would be interesting to see if we can use the structure `PullbackMetric` for these metrics that pullback the L2 metric.
https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/pullback_metric.py

geomstats/geometry/discrete_curves.py
alebrigant: This is different from `srv_inner_product`:  `inner_product` is the actual inner product for the SRV metric, and can be computed for any two tangent vectors, i.e. infinitesimal vector fields along a same curve. The name of `srv_inner_product` is confusing (and in my opinion should be changed), as it simply computes the L2 inner product between the SRV representations of two curves. This is useful for computing distances between curves, but not inner products between tangent vectors. 

I haven't looked at `PullbackMetric` yet, but I will !

alebrigant: Never mind, now I see what you meant and you are right: I can use `srv_inner_product` inside `inner_product`. Thanks !

alebrigant:
> Awesome discussion on the use of the new QuotientMetric class!
> 
> How about the following structure:
> 
>     * Have `DiscreteCurve` inherit from `FiberBundle`. Within the `DiscreteCurves`'s constructor, the `FiberBundle`'s constructor takes as input:
>       
>       * (1) `ambient_metric`, which can be an argument of `DiscreteCurve`'s constructor, defaulting to the `SRVMetric`.
>       * (2) `group_action`, which would be the action of the reparameterization.
>         Note that we can quotient the action of translation, scaling and rotation with the Kendall module, so it would not be included there.
> 
> 
> @alebrigant , the `split_horizontal_vertical` from your PR could go as a method of `DiscreteCurve` directly, overwriting the methods `horizontal_projection`, and `vertical_projection` from its parent class `FiberBundle` . Your code on horizontal geodesic and path could go there too. If your code only works with the `SRVMetric` (and not the general ElasticMetric), then there could be an `if/else` condition checking for the `self.ambient_metric` argument and returning `NotImplementedError` if it is not `SRVMetric`.
> 
>     * Your new metric could be called `ReparameterizedDiscreteCurveMetric`, and inherits from `QuotientMetric`, where the `fiber_bundle` attribute of the `QuotientMetric` class is the corresponding `DiscreteCurve`?
> 
> 
> Note: How about we use `ReparameterizedDiscreteCurveMetric` instead of `ShapeDiscreteCurveMetric`, and keep the word "shape" for Kendall shapes i.e. for the quotients with scalings, rotations, and translations?
> 
> Not for this PR: It would be interesting to see if we can use the structure `PullbackMetric` for these metrics that pullback the L2 metric.
> https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/pullback_metric.py

alebrigant:
@ninamiolane thanks for the comments and for the ideas on the use of the `QuotientMetric` class ! I quite agree with the structure that you're suggesting. I would just suggest calling the new class `UnparameterizedDiscreteCurves` as it would be more coherent with the names usually used in the literature.

ninamiolane:
> I would just suggest calling the new class UnparameterizedDiscreteCurves as it would be more coherent with the names usually used in the literature.

Sounds good! 🎉

geomstats/geometry/discrete_curves.py
```diff
@@ -2,8 +2,11 @@
 
 import math
 
+# import numpy as np
```
ninamiolane, line 5: Rm leftover comment

geomstats/geometry/discrete_curves.py
```diff
@@ -326,6 +331,50 @@ def square_root_velocity_inverse(self, srv, starting_point):
 
         return curve
 
+    def aux_differential_square_root_velocity(self, tangent_vec, curve):
```
ninamiolane, line 334: Thanks for the renaming! I think DeepSource is complaining about the docstring, adding the Parameters and their description here would solve it.

geomstats/geometry/discrete_curves.py
```diff
@@ -326,6 +331,50 @@ def square_root_velocity_inverse(self, srv, starting_point):
...
+            raise AssertionError('The square root velocity inner product '
+                                 'is only implemented for discrete curves '
+                                 'embedded in a Euclidean space.')
+        # n_sampling_points = curve.shape[-2]
+        d_srv_vec_a = self.aux_differential_square_root_velocity(
+            tangent_vec_a, curve)
+        d_srv_vec_b = self.aux_differential_square_root_velocity(
+            tangent_vec_b, curve)
+        inner_prod = self.srv_inner_product(d_srv_vec_a, d_srv_vec_b)
+        # self.l2_metric(n_sampling_points - 1).inner_product(
```
ninamiolane, line 373: Leftover comments

geomstats/geometry/discrete_curves.py
```diff
@@ -572,6 +621,35 @@ def dist(self, point_a, point_b, **kwargs):
...
+        mat_2 = from_vector_to_diagonal_matrix(vec_2, -1)
+        mat_3 = from_vector_to_diagonal_matrix(vec_3, 1)
+        mat_space_deriv = mat_1 - mat_2 + mat_3
+
+        space_deriv = n_points * gs.matmul(mat_space_deriv, curve)
+
+        return space_deriv
+
 
 class ClosedDiscreteCurves(Manifold):
```
ninamiolane, line 654: This should probably be renamed `UnparameterizedClosedDiscreteCurves` then?
Alternatively, we could only have `UnparameterizedDiscreteCurves` and have `self.close_or_open` as an attribute? Maybe this is a discussion for after this PR, and we can just open an issue to think about the namings of these classes afterwards.

tests/tests_geomstats/test_discrete_curves.py
```diff
@@ -344,10 +344,156 @@ def test_srv_norm(self):
...
+        squared_norm = gs.array(sums) / (srvs_ab.shape[-2] + 1)
         expected = gs.sqrt(squared_norm)
         self.assertAllClose(result, expected)
 
         result = result.shape
         expected = [srvs_ab.shape[0]]
         self.assertAllClose(result, expected)
+
+    @geomstats.tests.np_and_autograd_only
+    def test_aux_differntial_square_root_velocity(self):
```
ninamiolane, line 356: NIT/Typo: test_aux_differential

alebrigant:
Actually my bad, there would not be a class for unparameterized curves, only classes for metrics on unparameterized curves. But then I would still argue that vertical and horizontal projections should be methods of the quotient metric, because otherwise it means, as you said, using `if/elif` conditions to check which metric is used (SRV or elastic or any other) to know how to compute the split. But maybe this is something we can discuss at the next meeting !

nguigs:
Sorry for the late reply. Let me clarify my view of the different classes we should have, and please have a look at the quotient metric on [correlation matrices](https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/full_rank_correlation_matrices.py) for an example. Of course this is open for discussion, but the discussion would be on the design of the `FiberBundle` class itself rather than this case of metric on curves.

To implement a quotient space and the corresponding quotient metric, here are the classes that we need:
- A class for the total space
- A class for the Riemannian metric of the total space (called ambient metric)
- A class for the base manifold, if we can actually represent its elements with arrays in a non ambiguous way: this is the case of the correlation matrices, but not of the Kendall shapes, and I don't think it is the case for the curves.
- A `FiberBundle` class. Although it coincides with the total space, it should not replace it, because a FIberBundle works with only one ambient metric! It can however inherit from the total space. This is why `DiscreteCurve` should *not* inherit from `FiberBundle`, but instead a new class should be created specifically for the quotient structure considered.
- A `QuotientMetric` class. It only needs the above arguments to work, and in most cases it does not implement any other method, so it only has the `__init__`, except if we have closed form solutions.

Therefore, in this case we should leave DiscreteCurve as it was and implement new `FiberBundle` and corresponding `QuotientMetric` classes for every quotient that we consider, and there should not be any `if/else`!

I agree that we can discuss whether the methods `horizontal_projection` and `vertical_projection` should be moved to the `QuotientMetric`, but in that case we can actually merge `FiberBundle` into `QuotientMetric`. However, I like that `QuotientMetric` is very simple because all the machinery is in `FiberBundle`, and there is more than just the metric operations in `FiberBundle`, so it would have been an overwhelming `QuotientMetric` class. A `QuotientMetric` is thus like any other metric of geomstats, which is useful for learning algs.

alebrigant:
Thanks @nguigs for these clarifications ! I agree that my suggestion means that `QuotientMetric` class gets more methods, but it wouldn't mean that we could merge `FiberBundle` into `QuotientMetric` because on the contrary, we could define a fiber bundle without the notion of metric, and separately different quotient metrics on it.

In my view, the fiber bundle structure would be particularly interesting in the cases where we don't really know how to compute geodesics on the base space, but the geometry of the total space is known. Therefore, we use horizontal geodesics in the total space to compute geodesics in the base space. In this perspective, I would find the following structure more natural and helpful:
- `FiberBundle` should be seen more as representing the base space than the total space, since the total space is well known without the help of the fiber bundle structure. However it still contains information on both the total space and the base space. The methods `riemannian_submersion`, `lift`, `tangent_riemannian_submersion` definitely belong here, as well as any method fiber-related (check that two elements of the total space belong to the same fiber, i.e. project to the same element of the base space). In other words, methods of this class can be about verticality (independant of the metric) but not horizontality (metric-dependant).
- `QuotientMetric` takes a fiber bundle as argument as well as a metric on the total space of the fiber bundle. I would put all the methods about horizontality there.

With this structure, in the discrete curves example, we could create
- a class `UnparameterizedDiscreteCurves` that inherits from `FiberBundle`, with no notion of metric.
-  classes `QuotientSRVMetric`, `QuotientElasticMetric` etc. that inherit from `QuotientMetric` and that define metrics on the previous class.

alebrigant:
@ninamiolane @nguigs would you agree if I merged this PR and created another one to see how to use `FiberBundle` and `QuotientMetric` here ? I am getting conflicts.

ninamiolane:
Sure, just merged it! @alebrigant

Issue #107, opened 2018-06-22
Implement Homogeneous spaces


ninamiolane:
Create a class `HomogeneousSpace` that inherits from `Manifold` and implements its geometry.
https://en.wikipedia.org/wiki/Homogeneous_space

SwastiKh:
Hi if no one is working on this issue, I'd like to attempt at creating the Homogenous space class. Is there any mathematical background I should go through beforehand?

ninamiolane:
Hi @SwastiKh thank you for your interest! 🎉  This one is a very exciting issue for sure, with a lot of interest coming from the computer vision and machine learning communities.

Since we posted this issue in 2018, its scope has slightly changed. Instead of implementing an abstract class of homogeneous space, we want to focus on implementing two examples of homogeneous spaces: Stiefel manifold and Grasmann manifold... and other *practical* homogeneous spaces too, if you think about one.

As for the mathematical background on each of them, you can have a look at the reference papers cited in:
- our stiefel.py module
https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/stiefel.py
- and our grassman module: https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/grassmannian.py

Then, the initial goal of this issue is to go through the work of the winning participants of the ICLR 2021 computational geometry and topology challenge, and import in geomstats the metrics that they have implemented and that are not present in geomstats:
https://github.com/geomstats/challenge-iclr-2021/blob/main/1st-prize__mihaelanistor__Noise-Invariant-Topological-Features/noise_invariant_topological_features.ipynb

Later on, the goal of this issue is to review the literature on Riemannian geometry of Stiefel and Grassmanians and implement in geomstats the structures that are not yet implemented.

Let me know what you think?

SwastiKh:
Hi @ninamiolane! Thank you for the clarification. I was too thinking about implementing specific examples, mainly positive curvature homogeneous spaces like orthogonal groups or special orthogonal groups. I will look into stiefel and grassman module and work on those, now that I had a concrete starting point. Let me know if I should focus on the initial goal too alongside.

SwastiKh:
[...]

PR #1136, opened 2021-08-31
Implement Homogeneous spaces: complete Stiefel and Grassmann implementations. Issue #107

SwastiKh:
added 6 metrics for stiefel and grassmannian each. Unit tests are incomplete as of now. I could not find examples of the result for testing. Will update the unit tests soon.

### DIFF
--- a/geomstats/geometry/grassmannian.py
+++ b/geomstats/geometry/grassmannian.py
@@ -41,6 +41,7 @@ from geomstats.geometry.general_linear import GeneralLinear
 from geomstats.geometry.matrices import Matrices, MatricesMetric
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 from geomstats.geometry.symmetric_matrices import SymmetricMatrices
+import numpy as np
 
 
 def submersion(point, k):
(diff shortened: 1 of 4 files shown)


SaitejaUtpala:
you need to use gs backend for computations not specific backend (like you use numpy).

ninamiolane:
Woohoo, thanks for looking at it @SwastiKh ! Clean code !

In geomstats, we use the Python class `RiemannianMetric` (from `geometry/riemannian_metric.py`) to implement riemannian metrics. Sorry, I could have pointed that to you before.

This means that, before integrating this code, we need to understand:
- what are the Riemannian metrics that these distance functions define (are they referenced somewhere in the literature?)
- define the corresponding classes inheriting from the RiemannianMetric class.
I let you read on Riemannian metrics, and maybe you could even contact the participants of the challenge (and/or read through their notebook and research paper) so that we get the theoretical background on this?

Thanks, this is exciting!

SwastiKh:
Hi @ninamiolane 

I will make the necessary changes using the Riemann metric and probably contact the contest winners, if required, after reading the research paper and getting the background. 

Thank you for the wonderful support!😃✨

ninamiolane:
@SwastiKh do you have any update on this? Can we help?

ninamiolane:
@SwastiKh any update on this?

PR #1139, opened 2021-09-03
Make Automatic differentiation consistent across backends + fresh tests

ninamiolane:
This PR makes:
- value_and_grad
- custom_gradient
functions consistent across backends, and splits:
- test_backends.py,
- test_autodiff.py (new)
to gather tests related to automatic differentiation.


ninamiolane:
@nguigs I've addressed your comments (cf PR #1134 ), except the move of squared_dist into the class RiemannianMetric because that has given rise to new errors.

I think this PR is gigantic: maybe we should merge this, and leave the move to RiemannianMetric for the next one?

nguigs:
Yes let's merge and I'lll look into moving to RiemannianMetric when I merge the geodesic regression

PR #1142, opened 2021-09-07
Cholesky space

SaitejaUtpala:
This PR implements following things

- [x] vector space of  lower triangular matrices
    - [x] geometry
    - [x] tests
- [x] cholesky space : manifold of lower triangular matrices with positive diagonal elements (as open submanifold of lower triangular matrices)
   - [x] geometry
   - [x] tests 
- [x] Riemannian metric on choleksy space 
   - [x] metric
   - [x] tests


nguigs:
Hi Saiteja!
Great input! In fact the space of lower triangular matrices with positive diag (LT+) is a Lie group, and the vector space of LT matrices is its Lie algebra, so it should be implemented with the additional structure.
What metrics did you have in mind for LT+ ?
It can be embedded with left-invariant metrics, and there is one that coincides with the affine-invariant one on SPD. I have code that checks this.

SaitejaUtpala:
> Hi Saiteja!
> Great input! In fact the space of lower triangular matrices with positive diag (LT+) is a Lie group, and the vector space of LT matrices is its Lie algebra, so it should be implemented with the additional structure.
> What metrics did you have in mind for LT+ ?
> It can be embedded with left-invariant metrics, and there is one that coincides with the affine-invariant one on SPD. I have code that checks this.

Hey @nguigs . yup LT+ is a  matrix lie group. The metric is constructed by imposing different kinds of inner products on strictly lower part and diagonal elements (this  metric doesn't have name).  this is actually bi-invariant metric  

Then this metric is push forwarded to  SPD space, this is called Log Cholesky metric . Log cholesky metric has lots of nice properties and is kind of best of two worlds (Log euclidean, Affine Invariant ).  It has closed  form expression for Frechet mean and  easily computable expressions for parallel trasnport, R-log , R-exp maps . So I don't think this is equivalent to Affine Invariant Metric ?

nguigs:
> Then this metric is push forwarded to SPD space, this is called Log Cholesky metric . Log cholesky metric has lots of nice properties and is kind of best of two worlds (Log euclidean, Affine Invariant ). It has closed form expression for Frechet mean and easily computable expressions for parallel trasnport, R-log , R-exp maps . So, I don't think this is equivalent to Affine Invariant Metric?

Only the left-invariant metric with 2 on the diagonal and 1 on other coefficients is equivalent to AI metric when pushed to SPD. The others are different. This is good to test the implementation

ninamiolane:
Wooohoo, thanks for this! As @nguigs mentioned, we should have `Cholesky` inherits from `LieGroup` too.

On a minor note, maybe check your docstrings, there are periods missing at the end of descriptions here and there, e.g.:
```
    @classmethod
    def differential_gram(cls, tanget_vec, base_point):
        """Compute gram matrix of rows  
        Gram_matrix is mapping from point to point.point^{T}. 
        This is diffeomorphism between cholesky space and spd manifold
```
and at other places.

Otherwise this is awesome, can't wait for the tests.

geomstats/geometry/matrices.py
```diff
@@ -242,30 +441,7 @@ def to_skew_symmetric(cls, mat):
...
-        is_diagonal : array-like, shape=[...,]
-            Boolean evaluating if the matrix is square and diagonal.
-        """
-        is_square = cls.is_square(mat)
-        if not gs.all(is_square):
-            return False
-        diagonal_mat = from_vector_to_diagonal_matrix(cls.diagonal(mat))
-        is_diagonal = gs.all(
-            gs.isclose(mat, diagonal_mat, atol=atol), axis=(-2, -1))
-        return is_diagonal
```
nguigs, line 268: Why does this appear in this PR?

geomstats/geometry/cholesky.py
```diff
@@ -0,0 +1,324 @@
...
+        Returns
+        -------
+        projected: array-like, shape=[..., n, n]
+            SPD matrix.
+        """
+        vec_diag = gs.exp(gs.diagonal(point, axis1 = -2, axis2 = -1))
+        diag = gs.vec_to_diag(vec_diag)
+        strictly_lower_triangular = Matrices.to_lower_triangular(point)
+        projection = diag + strictly_lower_triangular
+        return projection
```
nguigs, line 72: Ideally, the projection should minimize the Frobenius norm. This is tricky as we have an open set here, but one property that must be verified for a projection function p is p(p(x))=x for all x.
This is not the case here, so maybe only do the projection if point does not already belong to LT+, or use epsilon to floor negative values of the diagonal.

nguigs (commented):
Good start!
I think we should discuss the naming though. Firstly, I'm not sure the naming Cholesky space is justified and well known enough, maybe we should stick to something more informative (PositiveLowerTriangular?).
This is even more the case for the metrics. I think the one you are implementing is called Log-Cholesky, and we should be precise because we can implement other ones such as Euclidean-Cholesky, LogEuclidean-Cholesky, and invariant metrics.
@ythanwerdas what do you think?

nguigs:
Maybe the Riemannian metric on SPD could be a PullBack metric @ninamiolane ?

ythanwerdas:
Dear all, let me give you an overview of Riemannian metrics on SPD matrices based on the Cholesky decomposition to help you decide the names.

First of all, Saiteja and Nicolas don't speak about the same Lie group structure. For Nicolas, the internal law is the matrix multiplication. For Saiteja, following Zhenhua Lin, the Lie group structure is a direct product of Lie groups: additive on the strict lower triangular matrix and multiplicative on the diagonal. By the way, it is even a vector space structure. Otherwise said, there is a diffeomorphism from LT^+(n) to the vector space LT(n) which is identity on the off-diagonal part and logarithm on the diagonal part.

I am preparing a paper on left-invariant metrics on the multiplicative Lie group LT^+(n), I will update this comment with the preprint as soon as it is available. I call Lie-Cholesky metrics the pullbacks of these metrics onto SPD matrices.

Zhenhua Lin calls log-Cholesky the metric Saiteja is implementing in this PR. Some people use directly the Euclidean metric on LT^+(n) and the pullback onto SPD matrices has been called "Cholesky metric" (see [chapter 3 of Riemannian Geometric Statistics in Medical Image Analysis](https://hal.inria.fr/hal-02341958/document)). It could be called Euclidean-Cholesky to avoid confusions. Another vector space parameterization is given by the matrix logarithm which sends diffeomorphically LT^+(n) onto LT(n) (see [this paper](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7463054)). This is also geodesically complete of course. The pullback onto SPD matrices could be called log-Euclidean-Cholesky to avoid confusions again.

I don't have opinion on the naming of the space, Cholesky or PositiveLowerTriangular have different advantages.

One last comment if I may, from a modelling point of view, I wouldn't say that log-Cholesky is better than log-Euclidean and affine-invariant since it is not invariant under orthogonal transformations and not even under permutations. It means that changing the order of the axes (representing your signals for example if your SPD matrices are covariance matrices) would change the statistical analysis. This property depends on the application but in many cases, we would like the analysis to be independent under orthogonal transformations (at least).

ninamiolane:
Yes, this metric should inherit from the PullbackMetric, while still overwriting most methods: if we can use closed forms instead of automatic differentiation the results will be better.

I agree with having `LowerTriangularMatrices` and `LowerTriangularPositiveMatrices`, even if it is long: it is more self-explanatory and I am indeed not sure that everybody knows the term `CholeskySpace`.

Thanks @ythanwerdas for the inputs on the classification of metrics for SPD matrices 🙌

geomstats/geometry/cholesky.py
SaitejaUtpala: @nguigs nice catch! I forgot we should we have p(p(x))=x. I will just use take `abs` of the diagonal

geomstats/_backend/numpy/__init__.py
```diff
@@ -381,6 +382,18 @@ def array_from_sparse(indices, data, target_shape):
     return array(coo_matrix((data, list(zip(*indices))), target_shape).todense())
 
 
+def vec_to_diag(vec):
```
ninamiolane, line 385: This function is already present here:
https://github.com/geomstats/geomstats/blob/master/geomstats/algebra_utils.py

As a guideline, let's put algebra utils in this `algebra_utils.py` file, using the `gs.` syntax to write in 4 backends. This avoid re-writing the same function in the 4 backends folders. The `_backend` folder is only here to write **already existing numpy functions** in the other backends, so that we can use the `gs.` syntax following the numpy API. It is not here to put utils from algebra. 

The exceptions are:
- if the implementation of the function cannot be written using the same backends `gs.` functions across the 4 backends,
- if we can get considerable speed-ups by writing the algebra function differently in the 4 backends.

geomstats/_backend/numpy/__init__.py
```diff
@@ -381,6 +382,18 @@ def array_from_sparse(indices, data, target_shape):
...
+    """Convert vector to diagonal matrix"""
+    d = vec.shape[-1]
+    return np.squeeze(vec[..., None, :] * np.eye(d)[None, :, :])
+
+
+def tril_to_vec(x, k=0):
```
ninamiolane, line 391: Ditto: put this in the algebra_utils.py

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -791,6 +807,19 @@ def where(condition, x=None, y=None):
     return tf.where(condition, x, y)
 
 
+def tril_to_vec(x, k=0):
```
ninamiolane, line 810: Can we write this like in Numpy or pytorch using `tril_indices`? If so, this function should also go to algebra_utils.py

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -812,6 +841,11 @@ def tile(x, multiples):
     return tf.tile(x_reshape, multiples)
 
 
+def vec_to_diag(vec):
+    """Convert vec to diagonal matrix"""
+    return tf.linalg.diag(vec)
+
+
 def vec_to_triu(vec):
     """Take vec and forms strictly upper traingular matrix.
```
ninamiolane, line 850: Typo: traingular -> triangular

ninamiolane: Typo: Take vec and form (no s)

geomstats/geometry/lower_triangular_matrices.py
```diff
@@ -0,0 +1,107 @@
...
+        super(LowerTriangularMatrices, self).__init__(
+            dim=int(n * (n + 1) / 2),
+            shape=(n, n),
+            metric=MatricesMetric(n, n),
+            default_point_type="matrix",
+        )
+        self.n = n
+
+    def get_basis(self):
+        """Compute the basis of the vector space of lower triangular"""
```
ninamiolane, line 30: Missing .

geomstats/geometry/lower_triangular_matrices.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Returns
+        -------
+        vec : array-like, shape=[..., n(n+1)/2]
+            Vector.
+        """
+        return gs.tril_to_vec(mat)
+
+    def projection(self, point):
+        """Make a square matrix lower triangular
```
ninamiolane, line 75: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+    def is_strictly_upper_triangular(cls, mat, atol=gs.atol):
+        """Check if a matrix is strictly upper triangular
+
+        Parameters
+        ----------
+        mat : array-like, shape=[..., n, n]
+            Matrix
+        atol : float
+            Absolute tolerance
+            Optional, defualt : backend atol
```
ninamiolane, line 269: Typo: default

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        """
+        is_square = cls.is_square(mat)
+        if not is_square:
+            is_vectorized = gs.ndim(gs.array(mat)) == 3
+            return gs.array([False] * len(mat)) if is_vectorized else False
+        return cls.equal(mat, gs.tril(mat, k=-1), atol)
+
+    @classmethod
+    def is_strictly_upper_triangular(cls, mat, atol=gs.atol):
+        """Check if a matrix is strictly upper triangular
```
ninamiolane, line 261: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        """
+        is_square = cls.is_square(mat)
+        if not is_square:
+            is_vectorized = gs.ndim(gs.array(mat)) == 3
+            return gs.array([False] * len(mat)) if is_vectorized else False
+        return cls.equal(mat, gs.triu(mat), atol)
+
+    @classmethod
+    def is_strictly_lower_triangular(cls, mat, atol=gs.atol):
+        """Check if a matrix is strictly lower triangular
```
ninamiolane, line 238: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        """
+        is_square = cls.is_square(mat)
+        if not is_square:
+            is_vectorized = gs.ndim(gs.array(mat)) == 3
+            return gs.array([False] * len(mat)) if is_vectorized else False
+        return cls.equal(mat, gs.tril(mat), atol)
+
+    @classmethod
+    def is_upper_triangular(cls, mat, atol=gs.atol):
+        """Check if a matrix is upper triangular
```
ninamiolane, line 215: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+
+        Parameters
+        ----------
+        mat : array-like, shape=[..., n, n]
+            Matrix
```
ninamiolane, line 220: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        atol : float
+            Absolute tolerance
+            Optional, default : backend atol
```
ninamiolane, line 223: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+
+        Returns
+        -------
+        is_triu : array-like, shape=[...,]
+            Boolean evaluating if the matrix is upper triangular
```
ninamiolane, line 228: Missing .

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        return cls.equal(mat, gs.triu(mat), atol)
+
+    @classmethod
+    def is_strictly_lower_triangular(cls, mat, atol=gs.atol):
+        """Check if a matrix is strictly lower triangular
+
+        Parameters
+        ----------
+        mat : array-like, shape=[..., n, n]
+            Matrix
```
ninamiolane, line 243: Missing . (I let you check the other ones)

geomstats/geometry/matrices.py
```diff
@@ -163,6 +163,122 @@ def is_square(mat):
...
+        atol : float
+            Absolute tolerance
+            Optional, defualt : backend atol
+
+        Returns
+        ------
+        is_strictly_triu : array-like, shape=[...,]
+            Boolean evaluting if the matirx is strictly upper triangular
```
ninamiolane, line 274: Typo: matrix

ninamiolane: Typo: evaluating

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        mat : array-like, shape=[..., n, n]
+            Matrix to be checked.
+        atol : float
+            Tolerance.
+            Optional, default: backend atol.
+
+        Returns
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean denoting if mat is an SPD matrix.
```
ninamiolane, line 73: Change description, since we do not explicitly have symmetric matrices here

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+
+        Returns
+        -------
+        projected: array-like, shape=[..., n, n]
+            SPD matrix.
+        """
+        return gs.einsum("...ij,...kj->...ik", point, point)
+
+    @staticmethod
+    def differential_gram(tanget_vec, base_point):
```
ninamiolane, line 123: tangent_vec

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        -------
+        differential_gram : array-like, shape=[..., n, n]
+            Differential of the matrix exponential.
+        """
+        mat1 = gs.einsum("...ij,...kj->...ik", tanget_vec, base_point)
+        mat2 = gs.einsum("...ij,...kj->...ik", base_point, tanget_vec)
+        return mat1 + mat2
+
+    @staticmethod
+    def inverse_differential_gram(tanget_vec, base_point):
```
ninamiolane, line 146: tangent_vec

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
+"""The manifold of lower triangular matrices with positive diagonal elements"""
+
+import geomstats.backend as gs
+from geomstats.geometry.base import OpenSet
+from geomstats.geometry.matrices import Matrices
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.geometry.lower_triangular_matrices import LowerTriangularMatrices
+
+
+class PositiveLowerTriangularMatrices(OpenSet):
```
ninamiolane, line 10: In the docstring, indicate that this space is also called the Cholesky space (so that other users who may have heard the term will relate as well)

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        )
+        inverse_differential_gram = Matrices.mul(base_point, aux)
+        return inverse_differential_gram
+
+
+class CholeskyMetric(RiemannianMetric):
+    """Class for the cholesky metric on the cholesky space."""
+
+    def __init__(self, n):
+        """Build the CholeskyMetric
```
ninamiolane, line 176: Init do not have docstrings: the description of the parameters and the reference should go in the docstring of the class.

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        sl_tagnet_vec_a = gs.tril_to_vec(tangent_vec_a, k=-1)
+        sl_tagnet_vec_b = gs.tril_to_vec(tangent_vec_b, k=-1)
+        ip_sl = gs.einsum("...i,...i->...", sl_tagnet_vec_a, sl_tagnet_vec_b)
+        return ip_sl
+
+    @classmethod
+    def inner_product(cls, tangent_vec_a, tangent_vec_b, base_point):
+        """Compute the inner product using only strictly lower triangular elements.
+
+
```
ninamiolane, line 247: Rm one line

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        inv_sqrt_diagonal = gs.power(Matrices.diagonal(base_point), -2)
+        tangent_vec_a_diagonal = Matrices.diagonal(tangent_vec_a)
+        tangent_vec_b_diagonal = Matrices.diagonal(tangent_vec_b)
+        prod = tangent_vec_a_diagonal * tangent_vec_b_diagonal * inv_sqrt_diagonal
+        ip_diagonal = gs.sum(prod, axis=-1)
+        return ip_diagonal
+
+    @staticmethod
+    def strictly_lower_inner_product(tangent_vec_a, tangent_vec_b):
+        """Compute the inner product using only strictly lower triangular elements.
```
ninamiolane, line 224: To be more precise: compute a term of the inner product, since it is not (yet) the inner product here, same for the description of the Returns

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+            Inner-product.
+        """
+        sl_tagnet_vec_a = gs.tril_to_vec(tangent_vec_a, k=-1)
+        sl_tagnet_vec_b = gs.tril_to_vec(tangent_vec_b, k=-1)
+        ip_sl = gs.einsum("...i,...i->...", sl_tagnet_vec_a, sl_tagnet_vec_b)
+        return ip_sl
+
+    @classmethod
+    def inner_product(cls, tangent_vec_a, tangent_vec_b, base_point):
+        """Compute the inner product using only strictly lower triangular elements.
```
ninamiolane, line 245: Ditto

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        diag_inner_product = cls.diag_inner_product(
+            tangent_vec_a, tangent_vec_b, base_point
+        )
+        strictly_lower_inner_product = cls.strictly_lower_inner_product(
+            tangent_vec_a, tangent_vec_b
+        )
+        return diag_inner_product + strictly_lower_inner_product
+
+    def exp(self, tangent_vec, base_point, **kwargs):
```
ninamiolane, line 273: 🎉

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        """Compute the Cholesky exponential map.
+
```
ninamiolane, line 275: Could you put the equation in :math: in the docstring? (I see it is in the paper), while referencing the paper again? Thank you!

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        diag_base_point = Matrices.diagonal(base_point)
+        diag_tangent_vec = Matrices.diagonal(tangent_vec)
+        diag_product_expm = gs.exp(gs.divide(diag_tangent_vec, diag_base_point))
+
+        sl_exp = sl_base_point + sl_tangent_vec
+        diag_exp = gs.vec_to_diag(diag_base_point * diag_product_expm)
+        exp = sl_exp + diag_exp
+        return exp
+
+    def log(self, point, base_point, **kwargs):
```
ninamiolane, line 303: Ditto, could you put the euqation in :math: in the docstring? Ty

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,359 @@
...
+        diag_base_point = Matrices.diagonal(base_point)
+        diag_point = Matrices.diagonal(point)
+        diag_product_logm = gs.log(gs.divide(diag_point, diag_base_point))
+
+        sl_log = sl_point - sl_base_point
+        diag_log = gs.vec_to_diag(diag_base_point * diag_product_logm)
+        log = sl_log + diag_log
+        return log
+
+    def squared_dist(self, point_a, point_b, **kwargs):
```
ninamiolane, line 333: Ditto on the :math: equation

geomstats/geometry/spd_matrices.py
```diff
@@ -412,6 +415,49 @@ def logm(cls, mat):
     from_vector = SymmetricMatrices.__dict__["from_vector"]
     to_vector = SymmetricMatrices.__dict__["to_vector"]
 
+    @classmethod
+    def cholesky_factor(cls, mat):
```
ninamiolane, line 419: Nice!

tests/tests_geomstats/test_lower_triangular_matrices.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertAllClose(result, expected)
+
+    def test_random_point_and_belongs_vectorization(self):
+        """Test of random_point and belongs methods."""
+        points = self.space.random_point(4)
+        result = self.space.belongs(points)
+        expected = gs.array([True] * 4)
+        self.assertAllClose(result, expected)
+
+    def test_to_vec(self):
```
ninamiolane, line 89: name of the unit-test should match name of the function tested: test_to_vector

tests/tests_geomstats/test_lower_triangular_matrices.py
```diff
@@ -0,0 +1,108 @@
...
+            [
+                [[1.0, 0.0, 0.0], [0.6, 7.0, 0.0], [-3.0, 0.0, 8.0]],
+                [[2.0, 0.0, 0.0], [2.6, 7.0, 0.0], [-3.0, 0.0, 28.0]],
+            ]
+        )
+        result = self.space.to_vector(chol_mat)
+        expected = gs.array(
+            [[1.0, 0.6, 7.0, -3.0, 0.0, 8.0], [2.0, 2.6, 7.0, -3.0, 0.0, 28.0]]
+        )
+        self.assertTrue(gs.allclose(result, expected))
```
ninamiolane, line 108: `get_basis` and `projection` are not tested: add unit tests?

tests/tests_geomstats/test_positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,291 @@
+"""Unit tests for Positive lower  triangular matrices"""
```
ninamiolane, line 1: Missing .

tests/tests_geomstats/test_positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,291 @@
...
+        batch_L = gs.array(
+            [
+                [[EULER ** 3, 0.0], [4.0, EULER ** 4]],
+                [[EULER ** 3, 0.0], [7.0, EULER ** 4]],
+            ]
+        )
+        batch_squared_dist_result = self.metric_cholesky.squared_dist(batch_K, batch_L)
+        batch_squared_dist_expected = gs.array([4 + 5, 9 + 5])
+        self.assertAllClose(squared_dist_expected, squared_dist_result)
+        self.assertAllClose(batch_squared_dist_expected, batch_squared_dist_result)
```
ninamiolane, line 291: Nice!! 🎉 thanks for all of these tests

ninamiolane (approved):
Very nice, thank you for this immense PR!! 🎉 

A few comments on docstrings: "." are missing here and there, I didn't flag them all, could you correct this?
Could you add the fact that the LT+ space has a vector space structure, as mentioned by @ythanwerdas, in the docstring? Cf:

"For Saiteja, following Zhenhua Lin, the Lie group structure is a direct product of Lie groups: additive on the strict lower triangular matrix and multiplicative on the diagonal. By the way, it is even a vector space structure. Otherwise said, there is a diffeomorphism from LT^+(n) to the vector space LT(n) which is identity on the off-diagonal part and logarithm on the diagonal part."

Since you added functions in the backends, could you add the corresponding tests in `test_backends.py`? Thank you 🙌 

@nguigs @ythanwerdas which name should we give to the metric: `LogCholeskyMetric` then? or something else? Thanks!

Other comments are inline. Awesome work!

geomstats/geometry/positive_lower_triangular_matrices.py
SaitejaUtpala: nope. It is inner product!. (it satisfies symmetric, non-degenerate, positive bi-linear conditions )

geomstats/geometry/positive_lower_triangular_matrices.py
SaitejaUtpala: nope. It is inner product!. (it satisfies symmetric, non-degenerate, positive bi-linear conditions )

geomstats/geometry/lower_triangular_matrices.py
SaitejaUtpala: done

tests/tests_geomstats/test_lower_triangular_matrices.py
SaitejaUtpala: done

geomstats/_backend/tensorflow/__init__.py
SaitejaUtpala: done

geomstats/_backend/autograd/__init__.py
```diff
@@ -384,12 +385,26 @@ def array_from_sparse(indices, data, target_shape):
...
+
+
 def triu_to_vec(x, k=0):
+    """ """
     n = x.shape[-1]
     rows, cols = triu_indices(n, k=k)
     return x[..., rows, cols]
 
 
+def vec_to_diag(vec):
```
ninamiolane, line 402: How is this different from https://numpy.org/doc/stable/reference/generated/numpy.diag.html ? is it working for minbatches?

geomstats/_backend/numpy/__init__.py
ninamiolane: @SaitejaUtpala what do you think of this?

geomstats/_backend/pytorch/linalg.py
```diff
@@ -53,7 +53,7 @@ def backward(ctx, grad):
 
 
 def cholesky(a):
-    return torch.cholesky(a, upper=False)
+    return torch.linalg.cholesky(a)
```
ninamiolane, line 56: This can go on line 44:
`cholesky = torch.linalg.cholesky`

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -763,10 +763,26 @@ def cumprod(a, axis=None):
     return tf.math.cumprod(a, axis=axis)
 
 
-def tril(m, k=0):
-    if k != 0:
-        raise NotImplementedError("Only k=0 supported so far")
-    return tf.linalg.band_part(m, -1, 0)
+# (sait) there is tf.experimental.tril (we can use it once it moves to stable)
```
ninamiolane, line 766: Use the TODO(sait) syntax: it will automatically create an issue on GitHub, which will be a good reminder for us to look at it.

geomstats/geometry/lower_triangular_matrices.py
```diff
@@ -0,0 +1,104 @@
...
+        super(LowerTriangularMatrices, self).__init__(
+            dim=int(n * (n + 1) / 2),
+            shape=(n, n),
+            metric=MatricesMetric(n, n),
+            default_point_type="matrix",
+        )
+        self.n = n
+
+    def get_basis(self):
+        """Compute the basis of the vector space of lower triangular."""
```
ninamiolane, line 27: Docstring is missing `Returns`

geomstats/geometry/lower_triangular_matrices.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Returns
+        -------
+        vec : array-like, shape=[..., n(n+1)/2]
+            Vector.
+        """
+        return gs.tril_to_vec(mat)
+
+    def projection(self, point):
+        """Make a square matrix lower triangular.
```
ninamiolane, line 72: Maybe add "by zeroing out lower elements."

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
...
+        """
+        vec_diag = gs.abs(Matrices.diagonal(point) - 0.1) + 0.1
+        diag = gs.vec_to_diag(vec_diag)
+        strictly_lower_triangular = Matrices.to_lower_triangular(point)
+        projection = diag + strictly_lower_triangular
+        return projection
+
+    @staticmethod
+    def gram(point):
+        """Compute gram matrix of rows
```
ninamiolane, line 106: Missing "."

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
...
+
+        Gram_matrix is mapping from point to point.point^{T}.
+        This is diffeomorphism between cholesky space and spd manifold
```
ninamiolane, line 109: Missing .

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
...
+        Returns
+        -------
+        projected: array-like, shape=[..., n, n]
+            SPD matrix.
+        """
+        return gs.einsum("...ij,...kj->...ik", point, point)
+
+    @staticmethod
+    def differential_gram(tangent_vec, base_point):
+        """Compute gram matrix of rows
```
ninamiolane, line 125: Change the description, as this is the differential

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
...
+        return mat1 + mat2
+
+    @staticmethod
+    def inverse_differential_gram(tangent_vec, base_point):
+        """Compute inverse differential of gram map
+
+        Parameters
+        ----------
+        tangent_vec : array_like, shape=[..., n, n]
+            tanget vector at gram(base_point)
```
ninamiolane, line 153: tange*n*t vector

tests/tests_geomstats/test_lower_triangular_matrices.py
```diff
@@ -0,0 +1,140 @@
+"""Unit tests for the vector space of symmetric matrices."""
```
ninamiolane, line 1: Change this docstring

tests/tests_geomstats/test_matrices.py
```diff
@@ -227,3 +227,64 @@ def test_diagonal(self):
         result = Matrices.diagonal(mat)
         expected = gs.ones((2, 3))
         self.assertAllClose(result, expected)
+
+    def test_to_diagonal(self):
```
ninamiolane, line 231: Do these functions also work on minibatch ("vectorized")?

tests/tests_geomstats/test_positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,291 @@
...
+        gram_expected = gs.matmul(point, Matrices.transpose(point))
+        self.assertAllClose(gram_expected, gram_result)
+
+        belongs_result = gs.all(SPDMatrices(self.n).belongs(gram))
+        belongs_expected = True
+        self.assertAllClose(belongs_expected, belongs_result)
+
+    def test_differential_gram(self):
+        """Test differential of gram for single point"""
+        L = gs.array([[1.0, 0.0], [2.0, 1.0]])
```
ninamiolane, line 124: Avoid using uppercase letters for variables, these are reserved for classes, see PEP8:
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

Also, let's avoid using single letters as variable names: it makes "find-and-replace" very complicated during refactoring. Thanks!

geomstats/geometry/positive_lower_triangular_matrices.py
ninamiolane: Adapt the docstring

geomstats/geometry/positive_lower_triangular_matrices.py
ninamiolane: Right, it is an inner-product, but not *the* inner-product on the space (which comes on line 246). Looks good like this.

ninamiolane (commented):
Very nice, I think we will be able to merge this great PR today 🎉  High-level comments:

- I have important comments/questions on the structure of the algebraic tools in the backends, see inlines.
- Some tests do not include tests on minibatch, aka `_vectorization` tests
- DeepSource is failing (there's not much to do, though)
- Can you make a pass on all the docstrings? 
  - Some docstrings are not right (I believed you copy pasted something but then didn't edit it)
  - Some words have uppercase in the docstrings:
    - cholesky -> Cholesky
    - spd -> SPD
  - Some "." are missing here and there.

Thank you very much for this!

geomstats/_backend/numpy/__init__.py
SaitejaUtpala: @ninamiolane  both tensorflow(https://www.tensorflow.org/api_docs/python/tf/linalg/diag) and pytorch (https://pytorch.org/docs/1.8.1/generated/torch.diag_embed.html?highlight=diag#torch.diag_embed) has inbuilt support for this and only  numpy doesn't have this thing

SaitejaUtpala: we can write two versions in numpy which uses assignment operation and which doesn't

SaitejaUtpala: so I think this should be in backend itself

SaitejaUtpala: and this is so common in ML. I think numpy may natively support this in future

SaitejaUtpala: I will remove from_vector_to_diagonal_matrix from algebra utils in next PR

ninamiolane: Ok, agreed, this function should be in the backends. You say that numpy does not have this but how about this:
https://numpy.org/doc/stable/reference/generated/numpy.diag.html
?

SaitejaUtpala: because on 2d array it doesn't return 3d array. (which is what we want)
`If v is a 2-D array, return a copy of its k-th diagonal. If v is a 1-D array, return a 2-D array with v on the k-th diagonal`.

ninamiolane: Goooot it, thank you for the explanation.

geomstats/_backend/pytorch/__init__.py
```diff
@@ -780,6 +777,18 @@ def vectorize(x, pyfunc, multiple_args=False, **kwargs):
     return stack(list(map(pyfunc, x)))
 
 
+def vec_to_diag(vec):
```
ninamiolane, line 780: Why not: https://pytorch.org/docs/1.8.1/generated/torch.diag_embed.html?highlight=diag#torch.diag_embed ?

In this case, this would go on Line 44 as:
vec_to_diag = torch.diag_embed

geomstats/_backend/pytorch/linalg.py
ninamiolane: (since the function is exactly the same, with same parameters)

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -817,8 +846,13 @@ def tile(x, multiples):
     return tf.tile(x_reshape, multiples)
 
 
+def vec_to_diag(vec):
```
ninamiolane, line 849: Put on top of the file as, see:
https://github.com/geomstats/geomstats/blob/1b2c2cc47c5bccdfb0236adf8246b942f934d4e8/geomstats/_backend/tensorflow/__init__.py#L74

geomstats/geometry/matrices.py
```diff
@@ -243,28 +438,23 @@ def to_skew_symmetric(cls, mat):
...
-    def is_diagonal(cls, mat, atol=gs.atol):
-        """Check if a matrix is square and diagonal.
+    def to_lower_triangular_diagonal_scaled(cls, mat, K=2.0):
+        """Make a matrix lower triangular, by zeroing out upper elements
+        and also diagonal is divided by factor K.
 
         Parameters
         ----------
-        mat : array-like, shape=[..., n, n]
+        mat : array-like, shape = [..., n, n]
```
ninamiolane, line 447: No space around = in our docstrings.

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
+"""The manifold of lower triangular matrices with positive diagonal elements"""
```
ninamiolane, line 1: Missing . at the end

geomstats/geometry/positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,361 @@
...
+            Boolean denoting if mat belongs to cholesky space.
+        """
+        is_lower_triangular = self.ambient_space.belongs(mat, atol)
+        diagonal = Matrices.diagonal(mat)
+        is_positive = gs.all(diagonal > 0, axis=-1)
+        belongs = gs.logical_and(is_lower_triangular, is_positive)
+        return belongs
+
+    def projection(self, point):
+        """Project a matrix to the cholesksy space.
```
ninamiolane, line 83: Cholesky

geomstats/geometry/positive_lower_triangular_matrices.py
ninamiolane: (the copy-pasting has had a problem)

tests/tests_geomstats/test_positive_lower_triangular_matrices.py
ninamiolane: and double space between lower and triangular

tests/tests_geomstats/test_positive_lower_triangular_matrices.py
```diff
@@ -0,0 +1,291 @@
...
+
+import math
+import warnings
+
+import tests.helper as helper
+
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.positive_lower_triangular_matrices import (
+    PositiveLowerTriangularMatrices,
```
ninamiolane, line 12: Are you using pre-commit and isort? CholeskyMetric should come before PositiveLower..

ninamiolane (commented):
Almost there! A few cosmetic comments to address (e.g. see my previous code review comments on the docstrings' cleanup).

Will be merged very soon! 🎉

ninamiolane (changes requested):
Moving to "request changes" to clean up docstrings and address other minor comments. Excellent work, looking forward to merge.

SaitejaUtpala:
@ninamiolane ready to merge. deep source is complaining about test files. that is fine though.

geomstats/_backend/tensorflow/__init__.py
ninamiolane: @SaitejaUtpala

geomstats/geometry/lower_triangular_matrices.py
ninamiolane: @SaitejaUtpala

geomstats/geometry/matrices.py
ninamiolane: @SaitejaUtpala

geomstats/geometry/positive_lower_triangular_matrices.py
ninamiolane: @SaitejaUtpala

ninamiolane (changes requested):
@SaitejaUtpala there are still some comments to be addressed + can you look at how this integrates with the PR #1232 merged two days ago? Someone needed triu, and we didn't want to wait until the Cholesky PR is merged thus it got merged. Thank you very much!

tests/tests_geomstats/test_matrices.py
SaitejaUtpala: there are vectorized tests!

ninamiolane (approved):
Fantastic, thank you very much!

PR #1143, opened 2021-09-07
faster spd checking

SaitejaUtpala:
faster SPD checking by cholesky decomposition
- [x] numpy
- [x] autograd
- [x] tensorflow
- [x] pytorch
- [x] tests

some simple benchmarks on numpy backend.
|size |  geomstats time| new time|
|------| -------------------- | -----  |  
| (10, 2000, 2000) | 21 sec | 3sec |
| (100 , 1000, 1000) | 25 sec | 4 sec|
|  (1000, 200, 200)  | 5 sec | ~0 sec |


SaitejaUtpala:
@SaitejaUtpala RTM

ninamiolane (changes requested):
Very nice! Smart to have though about this, thank you! 🎉 

Did you check that this was leading to a speed-up in a 4 backends? I trust you with the benchmark.

Just one comment: would it make more sense to have the functions:
- _is_single_matrix_pd
- is_pd
*within* the geomstats core code base, in the `matrices.py` module, and only put the `gs.linalg.cholesky` function in the backends?

This way:
- we only put in the `_backend` folder the mapping between the functionalities from torch, tf, autograd and np,
- the logic that we use to perform computations is thus in the codebase, more visible
- we avoid copy-pasting code in the 3 backends as much as possible.

What do you think?

 Note: if you do put these functions in the core codebase, then the docstrings will need to be adapted to respect our conventions ;)

SaitejaUtpala:
@ninamiolane  I think we should keep _is_single_matrix_pd in the backends and move is_pd to the matrices module.. because currently I am using  torch.linalg.cholesky, but there is https://pytorch.org/docs/stable/generated/torch.linalg.cholesky_ex.html (still in experimental stage). but it can be even faster than torch.linalg.cholesky for pd checking ( quoting `This makes this function a faster way to check if a matrix is positive-definite, and it provides an opportunity to handle decomposition errors more gracefully or performantly than torch.linalg.cholesky() does.`)

geomstats/_backend/autograd/linalg.py
```diff
@@ -100,3 +101,18 @@ def qr(*args, **kwargs):
     return np.vectorize(
         np.linalg.qr, signature="(n,m)->(n,k),(k,m)", excluded=["mode"]
     )(*args, **kwargs)
+
+
+def is_single_matrix_pd(mat):
+    """Check if a two dimensional square matrix is
+    positive definite
```
ninamiolane, line 108: Missing .

geomstats/_backend/numpy/linalg.py
```diff
@@ -71,3 +71,18 @@ def qr(*args, **kwargs):
     return np.vectorize(
         np.linalg.qr, signature="(n,m)->(n,k),(k,m)", excluded=["mode"]
     )(*args, **kwargs)
+
+
+def is_single_matrix_pd(mat):
+    """Check if a two dimensional square matrix is
+    positive definite
```
ninamiolane, line 78: Missing .

geomstats/geometry/matrices.py
```diff
@@ -186,6 +186,47 @@ def is_symmetric(cls, mat, atol=gs.atol):
...
+        ----------
+        mat : array-like, shape=[..., n, n]
+            Matrix.
+        atol : float
+            Absolute tolerance.
+            Optional, default: backend atol.
+
+        Returns
+        -------
+        is_pd : array-like, shape=[...,]
```
ninamiolane, line 224: NIT: is_pd -> is_spd

tests/tests_geomstats/test_backends.py
```diff
@@ -941,6 +941,23 @@ def test_prod(self):
         expected = gs.cumprod(vec)[-1]
         self.assertAllClose(result, expected)
 
+    def test_is_singel_matrix_pd(self):
```
ninamiolane, line 944: NIT: is_single

ninamiolane (approved):
There are a few typos here and there: let's correct them and we can merge! great work!

ninamiolane (approved):
Great, thank you!

PR #1144, opened 2021-09-09
From vector to lower or upper diagonal

alebrigant:
This PR adds a `num_diag` argument to `from_vector_to_diagonal_matrix` to transform a vector into a lower diagonal (`num_diag < 0`) or upper diagonal (`num_diag > 0`), just like [np.diag](https://numpy.org/doc/stable/reference/generated/numpy.diag.html). When `num_diag=0` the function is the same as before.

### DIFF
--- a/geomstats/algebra_utils.py
+++ b/geomstats/algebra_utils.py
@@ -92,11 +92,12 @@ def from_vector_to_diagonal_matrix(vector, num_diag=0):
         3-dimensional array where the `i`-th n-by-n array `diagonals[i, :, :]`
         is a diagonal matrix containing the `i`-th row of `vector`.
     """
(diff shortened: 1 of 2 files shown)


ninamiolane (approved):
Nice, thanks! LGTM as soon as the tests are also passing in PyTorch

alebrigant:
> Nice, thanks! LGTM as soon as the tests are also passing in PyTorch

The errors in PyTorch come from the static method `mul` of `Matrices`, I don't see the link with this PR. Is it possible that it comes from something else ?

ninamiolane:
Ah yes that's weird... Ok let's merge this and open a new fix PR then? Could be a change of version in pytorch...?

PR #1148, opened 2021-09-10
Levina Bickel Estimator for intrinsic dimension

SaitejaUtpala:
This pr implements Levina Bickel Estimator for intrinsic dimension of embedded manifolds. example, tests, doc strings yet to be added


ninamiolane (commented):
Nice, I didn't know about this! Maybe a few additional line of description (possibly using r""" syntax for docstrings with mathematical formulas) could help.

But I understand this is work in progress so I'll come back to it!

SaitejaUtpala:
@ninamiolane RTM

geomstats/learning/intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,100 @@
...
+        self : object
+            Returns self.
+        """
+        pairwise_sub = gs.expand_dims(X, axis=0) - gs.expand_dims(X, axis=1)
+        pairwise_dist = gs.linalg.norm(pairwise_sub, axis=-1)
+        sorted_dist = gs.sort(pairwise_dist, axis=-1)[:, 1:]
+        self.log_sorted_dist = gs.log(sorted_dist)
+        return self
+
+    def T(self, j, x):
```
ninamiolane, line 74: could this be called `nearest_neighbor` instead? would be more intuitive + different from the usual T used for transpose in numpy.

geomstats/learning/intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,100 @@
+"""Intrinsic Dimension Estimation."""
+
+import geomstats.backend as gs
+
+
+class LevinaBickelEstimator:
```
ninamiolane, line 6: Excellent 🎉 !!

geomstats/learning/intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,100 @@
...
+        self.min_neighbors = min_neighbors
+        self.max_neighbors = max_neighbors
+        self.log_sorted_dist = None
+
+    def fit(self, X):
+        """Calculate distance matrix
+
+        Parameters
+        ----------
+        X : array-like
```
ninamiolane, line 60: What is the shape here, probably [n_samples, dim]? Do we want to make it work for shapes such as [n_samples, n, m], i.e. for the manifolds of matrices or discrete landmarks/curves that we have?

geomstats/learning/intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,100 @@
...
+            index of the point in the data
+
+        Returns
+        -------
+        X : array like, shape=[N, dim]
+        """
+        return gs.exp(self.log_sorted_dist[x - 1, j - 1])
+
+    def predict(self):
+        """Predict intrinsic dimension."""
```
ninamiolane, line 91: Docstring missing Returns

tests/tests_geomstats/test_intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,29 @@
...
+        expected = (500, 499)
+        self.assertAllClose(result, expected)
+
+    def test_predict(self):
+        """Test the predict method."""
+        X, _ = datasets.make_swiss_roll(n_samples=2000)
+        X = gs.array(X)
+        lbe = LevinaBickelEstimator(min_neighbors=10, max_neighbors=200)
+        int_dim = lbe.fit(X).predict()
+        result = (1.9 <= int_dim <= 2.1)
```
ninamiolane, line 27: why not `result = (int_dim == 2)`?

NIT: Could also be written as:
```
result = lbe.fit(X).predict()
expected = 2
```

tests/tests_geomstats/test_intrinsic_dimension_estimation.py
```diff
@@ -0,0 +1,29 @@
...
+
+class TestIntrinsicDimensionEstimator(geomstats.tests.TestCase):
+    """Test of IntrinsicDimensionEstimators."""
+
+    def test_fit(self):
+        """Test the fit method"""
+        lbe = LevinaBickelEstimator()
+        X = gs.random.rand(500, 3)
+        dist_mat = lbe.fit(X).log_sorted_dist
+        result = (dist_mat.shape[0], dist_mat.shape[1])
```
ninamiolane, line 17: `result = dist_mat.shape` should work?

geomstats/learning/intrinsic_dimension_estimation.py
ninamiolane: I think `IntrinsicDimensionEstimator` would be more intuitive: instead of giving it a name, with tell *what* it is. 

NIT: The file could also be renamed to `intrinsic_dimension_estimator.py`, for consistency.

Any reason why this does not inherit from `BaseEstimator`? Maybe inheriting from BaseEstimator restricts the gs.arrays that we can use (i.e. since sklearn mostly works in np.arrays, maybe we can't use pytorch arrays there?)

ninamiolane (approved):
Looking good, thank you for the addition! 🎉  I made small comments.

How do we position this code vs `scikit-dimension` though?
https://scikit-dimension.readthedocs.io/en/latest/

We don't want to re-code something that exists. Since this works in all 4 backends, this could be 1 advantage?

Otherwise, maybe we should link to their package in our docstring and explain the difference?

geomstats/learning/intrinsic_dimension_estimation.py
SaitejaUtpala: it would be intutive. but they are many

tests/tests_geomstats/test_intrinsic_dimension_estimation.py
SaitejaUtpala: nope. this is approximate estimator. It doesn't necessarily give integer.

PR #1151, opened 2021-09-14
Add unit tests for geodesic regression on SE2

ninamiolane:
This PR adds unit tests for geodesic regression on SE2, with the backends tf and autograd (torch is missing the implementation of the matrix expm?).

### DIFF
--- a/geomstats/_backend/numpy/autodiff.py
+++ b/geomstats/_backend/numpy/autodiff.py
@@ -18,7 +18,7 @@ def detach(x):
     return x
 
 
-def value_and_grad(func):
+def value_and_grad(*args, **kwargs):
     """Return an error when using automatic differentiation with numpy."""
     raise RuntimeError(
         "Automatic differentiation is not supported with numpy backend. "
--- a/geomstats/learning/geodesic_regression.py
+++ b/geomstats/learning/geodesic_regression.py
@@ -72,7 +72,8 @@ class GeodesicRegression(BaseEstimator):
 
     def __init__(
             self, space, metric=None, center_X=True, method='extrinsic',
-            max_iter=100, learning_rate=.1, tol=1e-5, verbose=False):
+            max_iter=100, learning_rate=.1, tol=1e-5, verbose=False,
+            initialization='random'):
         if metric is None:
             metric = space.metric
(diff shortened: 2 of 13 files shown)


nguigs:
> This PR adds unit tests for geodesic regression on SE2, with the backends tf and autograd (torch is missing the implementation of the matrix expm?).

PyTorch does have expm but doesn't have logm. Apparently the problem comes from the implementation on GPU, but we could use [this](https://github.com/pytorch/pytorch/issues/9983#issuecomment-891777620) workaround to have a differentiable version of logm in pytorch?

geomstats/learning/geodesic_regression.py
```diff
@@ -238,6 +257,53 @@ def _fit_extrinsic(self, X, y, weights=None, compute_training_score=False):
 
         return self
 
+    def initialize_parameters(self, y):
```
ninamiolane, line 260: Nice!! 🎉

geomstats/_backend/numpy/autodiff.py
```diff
@@ -18,7 +18,7 @@ def detach(x):
     return x
 
 
-def value_and_grad(func):
+def value_and_grad(*args, **kwargs):
```
ninamiolane, line 21: `value_and_grad(func, to_numpy=False)` could work too and would more closely follow the API of the others.

tests/tests_geomstats/test_examples.py
```diff
@@ -68,6 +69,11 @@ def test_empirical_frechet_mean_uncertainty_sn():
     def test_geodesic_regression_hypersphere():
         geodesic_regression_hypersphere.main()
 
+    @staticmethod
```
ninamiolane, line 72: Nice!!

ninamiolane (commented):
Wohoooo! It's happening, congrats on making it work 🎉 

I let you merge when you feel like it. (Maybe keep the grassmanian regression for the next PR?)

geomstats/_backend/numpy/autodiff.py
nguigs: The point is to get the wrong backend error instead of a possible error due to the arguments

PR #1171, opened 2021-09-28
Precommit hooks

SaitejaUtpala:
precommit hooks with black, blacken-docs, isort etc


ninamiolane (approved):
Excellent, thank you!

PR #1003, opened 2021-05-13
Unstall the Forever stalled MDM PR

SaitejaUtpala:
This is an (optimistic )attempt to un-stall the #962 by breaking into series of smaller PRs. As first this just includes mdm algorithm and corresponding test.

PR #962, opened 2021-03-31
NEW: Add Riemmanian minimum distance to mean (MDM) classification for SPD matrices

ninamiolane:
This PR follows the stalled PR #623

PR #1188, opened 2021-11-05
Improve MDM

qbarthelemy:
This PR aims to:
- add MDM in the geomstats documentation, in a new Classification section;
- correct shape of labels in MDM: to be consistent with sklearn classifiers, labels must be a vector of shape (n_samples) instead of (n_samples, n_classes); so `split_data_in_classes` function is deleted;
- distinguish class index from class label;
- add `score` function to MDM;
- correct and complete MDM tests;
- complete example on EMG classification, with Euclidean and Riemannian MDM.

Help needed: in EMG example, Euclidean MDM output is NaN...

Suggestion: the old PR #962 could be closed, because MDM has been added in #1003.


ninamiolane (approved):
Excellent, thank you so much for having a pass at this! 🙌 

The tests are failing in tensorflow: can you see if you can address these? Otherwise, it is fine to skip the corresponding test in TF.

Could you also raise an issue about the Euclidean MDM problem? Thanks!

qbarthelemy:
> Could you also raise an issue about the Euclidean MDM problem?

I think that obtaining NaN values with SPDMetricEuclidean is partially intended, because geomstats raises this error:
`RuntimeError: Automatic differentiation is not supported with numpy backend. Use autograd, pytorch or tensorflow backend instead.`
This said, it remains surprising that advanced metrics as SPDMetricAffine and SPDMetricLogEuclidean are supported by np backend, but not SPDMetricEuclidean.

nguigs:
> This said, it remains surprising that advanced metrics as SPDMetricAffine and SPDMetricLogEuclidean are supported by np backend, but not SPDMetricEuclidean.

Hi @qbarthelemy, thank you for contributing!!
SPDMetricAffine and SPDMetricLogEuclidean are both implemented with close form solutions so there are no problems supporting them with numpy, actually it's even the most precise.
You get the `RuntimeError: Automatic differentiation ` error because exp/log maps are not implemented for SPDMetricEuclidean, so more general methods are called. In fact SPDMetricEuclidean should probably inherit MatricesMetric to implement exp and log as addition/substraction. This could be an issue to raise!

qbarthelemy:
Ok, thanks for the explanation! I have opened the PR #1192 to add log and exp maps for SPDMetricEuclidean.

PR #1191, opened 2021-11-09
Committing new files for the psd matrices of rank k

annacalissano:


geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        vector_sym = Matrices(self.n, self.n).to_symmetric(vector)
+
+        _, r = gs.linalg.eigh(base_point)
+        r_ort = r[..., :, self.n - self.rank: self.n]
+        r_ort_t = Matrices.transpose(r_ort)
+
+        candidates = gs.matmul(
+            gs.matmul(gs.matmul(r_ort, r_ort_t), vector_sym), gs.matmul(r_ort, r_ort_t)
+        )
```
nguigs, line 142: First you can use `Matrices.mul(a,b,c,d...)` instead of repeating the call to `matmut`. Than you can assigne `gs.matmul(r_ort, r_ort_t)` to a variable to avoid computing it twice.

geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -0,0 +1,209 @@
...
+
+        result = gs.where(gs.sum(gs.isclose(candidates, 0.0, gs.atol),
+                                 axis=(-2, -1)) < (self.n * self.n), False, True)
```
nguigs, line 145: This could be simplified to `gs.all(gs.isclose(candidates, 0., gs.atol), axis=(-2, -1))`

geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -0,0 +1,209 @@
...
+            Projection of the tangent vector at base_point.
+        """
+        vector_sym = Matrices(self.n, self.n).to_symmetric(vector)
+        _, r = gs.linalg.eigh(base_point)
+        r_ort = r[..., :, self.n - self.rank: self.n]
+        r_ort_t = Matrices.transpose(r_ort)
+
+        return (vector_sym - gs.matmul(
+                gs.matmul(gs.matmul(r_ort, r_ort_t), vector_sym),
+                gs.matmul(r_ort, r_ort_t)))
```
nguigs, line 172: same comment as above for `Matrices.mul`

tests/tests_geomstats/test_rank_k_psd_matrices.py
```diff
@@ -0,0 +1,85 @@
...
+        psd_n_k = self.space
+        mat_not_psd_n_k = gs.array(
+            [[0.8369314, -0.7342977, 1.0402943],
+             [0.04035992, -0.7218659, 1.0794858],
+             [0.9032698, -0.73601735, -0.36105633]]
+        )
+        mat_psd_n_k = gs.array([[1.0, 1.0, 0], [1.0, 4.0, 0], [0, 0, 0]])
+        result = psd_n_k.belongs(mat_not_psd_n_k)
+        expected = False
+        self.assertFalse(result, expected)
```
nguigs, line 38: `self.assertFalse(result)` is sufficient

tests/tests_geomstats/test_rank_k_psd_matrices.py
```diff
@@ -0,0 +1,85 @@
...
+
+        result = psd_n_k.belongs(mat_psd_n_k)
+        expected = True
+        self.assertTrue(result, expected)
```
nguigs, line 42: `self.assertTrue(result)` is sufficient

nguigs (commented):
Great contribution! I made a few comments to simplify the code a bit, which will make it understandable by more people (hopefully).

geomstats/geometry/rank_k_psd_matrices.py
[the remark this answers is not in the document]
annacalissano: gs.array([gs.allclose(c, 0.0, gs.atol) for c in candidates])

but it is not working for all input dimensions. The problem is that I need to check it for each element of the gs.array separately and gs.allclose doesn't seem to have an axis option right?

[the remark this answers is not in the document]
nguigs: Sorry I meant to delete this comment, but I suggested something simpler below

nguigs: This one should work

geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Returns
+        -------
+        projected: array-like, shape=[..., n, n]
+            PSD matrix rank k.
+        """
+
+        sym = Matrices.to_symmetric(point)
+        _, s, v = gs.linalg.svd(sym)
+        h = gs.matmul(Matrices.transpose(v), s[..., None] * v)
```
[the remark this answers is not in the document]
annacalissano: I agree with you,  it seems like a redundant step to turn the matrix into a symmetric one (as we already do that in the line before). It is a step in the implementation in matlab take from this reference: https://www.sciencedirect.com/science/article/pii/0024379588902236
Let's discuss it in December because the function works but I am not sure if it is a smart solution and I haven't had much time to check for other smarter option.

PR #1192, opened 2021-11-09
Add exp and log maps for SPD Euclidean

qbarthelemy:
This PR aims to:
- add log and exp maps for `SPDMetricEuclidean`;
- complete tests for spd_matrices with Euclidean log and exp maps;
- complete tests for MDM, adding Euclidean and LogEuclidean metrics;
- update example on EMG classification, to display values for Euclidean MDM.


qbarthelemy:
1. To avoid multiple copies of quasi-similar tests, I tried to use fixtures to run the same test with different metrics.
However,
```
@pytest.mark.parametrize("metric", [SPDMetricAffine, SPDMetricLogEuclidean, SPDMetricEuclidean])
class MyTestClass(geomstats.tests.TestCase):
    def my_test(self, metric):
        my test...
```
raises an error in `unittest`.
So, I have decided to apply a loop inside each test: I don't know if it is compliant with geomstats practices.

2. In EMG example, I have renamed DNN into MLP, because MLP with one hidden layer is not really considered as a deep neural network in the deep learning community.

geomstats/geometry/spd_matrices.py
```diff
@@ -885,6 +885,47 @@ def exp_domain(tangent_vec, base_point):
 
         return domain
 
+    def exp(self, tangent_vec, base_point, **kwargs):
```
ninamiolane, line 888: Does this work for power-euclidean metrics for `self.power_euclidean != 1`?

ninamiolane (approved):
Great PR, thank you very much!

1. Definitely, we want to refactor our (monstrous) unit-tests and use pytest fixtures. @SaitejaUtpala is planning to work on it, but any help would be appreciated here if you have ideas to simplify the CI! 🙌  
2. DNN -> MLP: Definitely, good call, thank you.

I just left one question about the exp and log map for the power-euclidean metrics with power != 1. See Table 2 here: https://arxiv.org/pdf/2111.02990.pdf

qbarthelemy:
> I just left one question about the exp and log map for the power-euclidean metrics with power != 1. See Table 2 here: https://arxiv.org/pdf/2111.02990.pdf

Yes, you are right: current code is only for `power_euclidean = 1`.
However, I don't see how to code with geomstats functions the generalized formula of Table 2 applied to the particular case of exp and log maps of power Euclidean metrics.
Maybe @ythanwerdas or @xpennec could help on this point.

geomstats/geometry/spd_matrices.py
```diff
@@ -885,6 +885,59 @@ def exp_domain(tangent_vec, base_point):
...
+        -------
+        exp : array-like, shape=[..., n, n]
+            Euclidean exponential.
+        """
+        power_euclidean = self.power_euclidean
+
+        if power_euclidean == 1:
+            exp = tangent_vec + base_point
+        else:
+            exp = 1  # TODO
```
ninamiolane, line 912: Maybe add "raise NotImplementedError" here?

ninamiolane (commented):
Looking good! @ythanwerdas if you have some inputs on the exp and logs for power !=1 that would be great.

Otherwise, @qbarthelemy maybe just add a "not implemented error" there and we can merge like this for now.

Thanks!

ythanwerdas:
Dear all, sorry for replying that late. The exponential map for the power-Euclidean metric with power p is $Exp_S(X)=(S^p + d_S pow_p(X))^{1/p}$ and the logarithm map is $Log_S(T)=(d_S pow_p)^{-1}(T^p - S^p)$ where $S,T$ are SPD matrices and $X$ is a vector at S, i.e. a symmetric matrix. The differential of pow_p : S -> S^p is already implented.

The reason why I stopped implementing the power-Euclidean metrics (and more generally all metrics obtained by pullback by a map f) is that computing with them amounts to compute with the original metric (here Euclidean) on the deformed dataset [S_1^p, ..., S_N^p]. So in theory, we could avoid using these implementations. However, I agree that it is also good to have them ready to use in the classical way.

qbarthelemy:
Thanks @ythanwerdas! I will implement these formula.

qbarthelemy:
@ythanwerdas , can you check that I have correctly _geomstatified_ your equations?

Exp_S(X)=(S^p + d_S pow_p(X))^{1/p} :
```
exp = SymmetricMatrices.powerm(
    SymmetricMatrices.powerm(S,p) + SPDMatrices.differential_power(p, X, S),
    1/p
)
```
and Log_S(T)=(d_S pow_p)^{-1}(T^p - S^p):
```
log = SPDMatrices.inverse_differential_power(p, SymmetricMatrices.powerm(T, p) - SymmetricMatrices.powerm(S, p), S)
```

ythanwerdas:
That's perfect! Thank you @qbarthelemy!

PR #1193, opened 2021-11-15
Add predict_proba to MDM

qbarthelemy:
This PR aims to:
- add the function `predict_proba` to MDM classifier;
- complete and update tests for MDM.

### DIFF
--- a/geomstats/learning/mdm.py
+++ b/geomstats/learning/mdm.py
@@ -1,5 +1,6 @@
 """The MDM classifier on manifolds."""
 
+from scipy.special import softmax
 from sklearn.metrics import accuracy_score
 import geomstats.backend as gs
 from geomstats.learning.frechet_mean import FrechetMean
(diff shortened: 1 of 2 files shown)


geomstats/learning/mdm.py
```diff
@@ -94,6 +95,31 @@ def predict(self, X):
...
+
+        Returns
+        -------
+        probas : array-like, shape=[n_samples, n_classes]
+            Probability of the sample for each class in the model.
+        """
+        n_samples = X.shape[0]
+        probas = []
+        for i in range(n_samples):
+            dists = self.riemannian_metric.dist(X[i], self.mean_estimates_)
```
ninamiolane, line 119: From an efficiency perspective, maybe it would make more sense to use `riemannian_metric.squared_dist` here. Most often, the Riemannian square distance is computed using the inner-product of the Riemannian Log with itself, and then the Riemannian distance is computed by taking the square of this:
https://github.com/geomstats/geomstats/blob/efe8f0215df802f4dd2816fcac42bd81b7a900d7/geomstats/geometry/riemannian_metric.py#L200


If we directly use the .squared_dist here, we spare one square-root and one square in the default case, which could be interesting.

ninamiolane (approved):
LGTM as long as the TF tests pass too, thank you 🎉 !

Edit: The failed test seemed to have been a network error so let me just re-run the unit-tests :)

geomstats/learning/mdm.py
qbarthelemy: Yes, you are right! I will correct this point.

PR #1195, opened 2021-11-30
add extrinsic to spherical and its tangent

nguigs:
Add functions to work with spherical coordinates

### DIFF
--- a/geomstats/geometry/hypersphere.py
+++ b/geomstats/geometry/hypersphere.py
@@ -207,7 +207,7 @@ class _Hypersphere(EmbeddedManifold):
         theta = base_point_spherical[..., 0]
         phi = base_point_spherical[..., 1]
 
-        theta_safe = gs.where(gs.abs(theta < gs.atol), gs.atol, theta)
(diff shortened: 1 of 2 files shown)


geomstats/geometry/hypersphere.py
```diff
@@ -173,6 +174,71 @@ def tangent_spherical_to_extrinsic(
 
         return tangent_vec_extrinsic
 
+    def extrinsic_to_spherical(self, point_extrinsic):
```
ninamiolane, line 177: Add a docstring?

It would be nice to add there which point is considered the origin of the spherical coordinates.

ninamiolane (approved):
Good as long as the docstring is added, with a specification of what 3D point is chosen as the origin of the theta and phi.

geomstats/geometry/hypersphere.py
nguigs: Sorry forgot! Weird that DeepSource didn't detect it though

ninamiolane: Yeah....

nguigs: And now adding docstrings fails codecov?

Issue #1200, opened 2021-12-04
SPD.exp() and cuda 11.4


PLabarbarie:
Hello, i have an issue using the function exp from SPDMetricAffine. I use this function in a riemannian optimization algorithm which can be sumerized as followed

C_t = a certain matrix in SPD.

for i in range(nb_ite):
        nabla_R = ....
        C_t = SPD.exp(nabla_R, C_t)

I have launched this algorithm on multiple servers without any problem but yesterday one appeared on a new server. The function exp() was extremely long. Normally it took less than 1 second and now 40 seconds. 

This is my settings :

CUDA 11.4
torch '1.10.0+cu113'
torchvision>=0.7.0
tqdm>=4.44.0
scipy>=1.5.0
statsmodels>=0.11.1
joblib>=0.14.1
matplotlib>=3.3.4
numpy>=1.18.1
scikit-learn>=0.22.1
autograd>=1.3
geomstats>=2.3.1 

I just remarked that my other servers were on CUDA 10.2.

Can you help me please ?

Best regards.

Pol

PLabarbarie:
UserWarning: torch.symeig is deprecated in favor of torch.linalg.eigh and will be removed in a future PyTorch release.
The default behavior has changed from using the upper triangular portion of the matrix by default to using the lower triangular portion.
L, _ = torch.symeig(A, upper=upper)
should be replaced with
L = torch.linalg.eigvalsh(A, UPLO='U' if upper else 'L')
and
L, V = torch.symeig(A, eigenvectors=True)
should be replaced with
L, V = torch.linalg.eigh(A, UPLO='U' if upper else 'L') (Triggered internally at  ../aten/src/ATen/native/BatchLinearAlgebra.cpp:2499.)
  torch.symeig(C)

Okay I noticed that it's caused by symeig() function. I don't know why it works on other servers but not on this one.

If you have any idea to skirt this problem ?

Pol

captain-pool:
@PLabarbarie can you share a small test script that demonstrates this problem?

PLabarbarie:
Hello, like a said before i tried to understand whats causing the slowness of my main script. Firstly with multiple print and time print i remarked that it was the function exp of SPD. After that, i created a little script (with the same conditions of my main script) to see if it was really the function exp and it was. When i saw the message mention above 

UserWarning: torch.symeig is deprecated in favor of torch.linalg.eigh and will be removed in a future PyTorch release.
[...]

PR #1201, opened 2021-12-06
updating symeig with eigh

captain-pool:
Replacing `symeig` implementation of `eigh` with `torch.linalg.eigh`. (Ref: https://github.com/geomstats/geomstats/issues/1200)


ninamiolane (approved):
Excellent, thank you!

Issue #1187, opened 2021-11-05
parallel transport documentation


SaitejaUtpala:
currently parallel transport takes 

* tangent_vec_a (tangent vector at base point to be transported.)
* tangent_vec_b (tangent vector at base point, along which the parallel transport is computed)
* base_point ( point on the hypersphere)

but if someone has 
* point_a (point from which tangent vector is to transported)
* tangent vector (tangent vector at point_a)
* point_b (a point to which tangent vector at point_a is to be transported)

one has to use (tangent_vector, Log_{point_a} point_b , base_point). it's better to properly document this so that someone new can use it.

ninamiolane:
Agreed! Additionally, tangent_vec_a and tangent_vec_b could be renamed to:
- tangent_vec
- tangent_vec_transport_direction
or something like this, because I am always confused which one is tangent vec a and which one is tangent vec b.

nguigs:
Yes, we could also add the option to give an `end_point` instead of `tangent_vec_b`, but we would need to change the signature of the method

PR #1202, opened 2021-12-07
Address issue about parallel transport API

nguigs:
Start addressing #1187


examples/geodesic_regression_hypersphere.py
```diff
@@ -73,8 +73,8 @@ def main():
     tangent_vec_of_transport = METRIC.log(intercept, base_point=intercept_hat)
     transported_coef_hat = METRIC.parallel_transport(
         tangent_vec_a=tangent_vec_to_transport,
```
ninamiolane, line 75: Should we rename tangent_vec_a and tangent_vec_b : they play different roles, and I always have to go back to the file implementing the parallel transport to remember which one is which one.

How about: `tangent_vec_to_transport` and `tangent_vec`?

geomstats/geometry/connection.py
```diff
@@ -627,22 +627,29 @@ def path(t):
...
-        tangent_vec_b)`.
+        along the geodesic between two points `base_point` and `end_point`
+        or alternatively defined by :math:`t\mapsto exp_(base_point)(
+        t*tangent_vec_b)`.
 
         Parameters
         ----------
         tangent_vec_a : array-like, shape=[..., {dim, [n, m]}]
             Tangent vector at base point to be transported.
+        base_point : array-like, shape=[..., {dim, [n, m]}]
```
ninamiolane, line 644: how about `initial_point` to be consistent with the geodesic's API?

geomstats/geometry/symmetric_matrices.py
```diff
@@ -228,9 +228,13 @@ def apply_func_to_eigvals(mat, function, check_positive=False):
         """
         eigvals, eigvecs = gs.linalg.eigh(mat)
         if check_positive and gs.any(gs.cast(eigvals, gs.float32) < 0.0):
-            logging.warning(
-                "Negative eigenvalue encountered in" " {}".format(function.__name__)
-            )
+            try:
```
ninamiolane, line 231: what is this? and does this have to be coded with a try/except? why not if and raise Error?

tests/helper.py
```diff
@@ -96,35 +96,39 @@ def is_isometry(tan_a, trans_a, endpoint):
     tan_vec_b = space.to_tangent(gs.random.rand(*shape) / 5, base_point)
     end_point = metric.exp(tan_vec_b, base_point)
 
-    transported = metric.parallel_transport(tan_vec_a, tan_vec_b, base_point)
+    transported = metric.parallel_transport(tan_vec_a, base_point, tan_vec_b)
+    result = is_isometry(tan_vec_a, transported, end_point)
+    results.append(gs.all(result))
+
+    transported = metric.parallel_transport(tan_vec_a, base_point, end_point=end_point)
```
ninamiolane, line 103: do you have a test for vectorization when called with base_point/end_point?

ninamiolane: Could we also add a test that checks that for the same geodesic (of initial_point, end_point and initial_tangent_vec_) both versions of the parallel transport give the same result?

tests/tests_geomstats/test_pre_shape.py
```diff
@@ -766,3 +766,9 @@ def test_parallel_transport(self):
         is_horizontal = space.is_horizontal(transported, end_point)
         self.assertTrue(gs.all(is_tangent))
         self.assertTrue(gs.all(is_horizontal))
+
+        transported = metric.parallel_transport(
+            tan_a[0], point, end_point=end_point[0], n_steps=150, step="rk4"
```
ninamiolane, line 771: could we check that both versions of the parallel transport give the same result?

ninamiolane (approved):
LGTM! Two main comments:
- the API of the parallel transport could benefit matching the API of the geodesic: initial_point, end_point, initial_tangent_vec, tangent_vec_to_transport?
- could you add slightly more tests, i.e. checking that both parallel transports give the same transported vector?

Otherwise good to merge! 🎉

examples/geodesic_regression_hypersphere.py
nguigs: what about `tangent_vec` and `initial_direction`?

geomstats/geometry/connection.py
nguigs: Ok

geomstats/geometry/symmetric_matrices.py
nguigs: I got an error from this line when function is a list of functions (in powerm).

tests/helper.py
nguigs: This is with multiple inputs for all args, I will try with only one `end_point`.

tests/tests_geomstats/test_pre_shape.py
nguigs: What do you mean both versions? Do mean ladder methods? This check is done for the hypersphere.

ninamiolane: Sorry I wasn't clear. i mean a version with initial_point and end_point, and a version with initial_point and initial_tangent_vec.

examples/geodesic_regression_hypersphere.py
ninamiolane: or tangent_vec and initial_tangent_vec? which ever you prefer!

Issue #636, opened 2020-04-27
Use gs.assignment to clean code + increase torch & tf coverages


ninamiolane:
The new `gs.assignment` function (thanks @pchauchat !) allows us:
- to clean code such as:
https://github.com/geomstats/geomstats/blob/1e5f7cea94a7884b798f07723285b04592ac09ab/geomstats/geometry/special_orthogonal.py#L183
- to translate functions/methods in order for them to run with tf and torch.

We should use it in the whole codebase.

nguigs:
What about writing a `divide` function for tf with similar behavior as np and pytorch (maybe using `gs.assignement`)? It is similar to `where`, but unlike `where` it doesn't compute the division where the condition is false. As in most cases in the code, a mask is used to avoid dividing by 0, refactorizing with `divide` will probably be cleaner than with `gs.assignment` everywhere.

PR #1204, opened 2021-12-08
Implement Special Linear (nguigs-hiring)

lpereira95:
Here it goes my implementation of special linear along with corresponding tests.

Looking forward for the feedback.


geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+        Returns
+        -------
+        belongs : array-like, shape=[...,]
+            Boolean evaluating if point belongs to the manifold.
+        """
+        has_right_shape = self.lie_algebra.ambient_space.belongs(point)
+
+        if gs.all(has_right_shape):
+            # ???: det can be negative?
+            return gs.isclose(gs.abs(gs.linalg.det(point)), 1.0, atol=atol)
```
nguigs, line 52: No this space is restricted to positive determinant

geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+        point : array-like, shape=[..., n, n]
+            Point in embedding manifold.
+
+        Returns
+        -------
+        projected : array-like, shape=[..., n, n]
+            Projected point.
+        """
+        scale_coeff = gs.power(gs.abs(gs.linalg.det(point)),
+                               1.0 / self.n)
```
nguigs, line 73: `algebra_utils.flip_determinant` can be used to make sure det > 0

geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+        n_non_diag_half = (self.dim - n + 1) // 2
+
+        self.basis = gs.zeros((self.dim, n, n))
+
+        self.basis[dims[:n_non_diag_half], rows, cols] = 1.
+        self.basis[dims[n_non_diag_half:-n + 1], cols, rows] = 1.
+
+        self.basis[dims[-n + 1:], 0, 0] = -1
+        self.basis[dims[-n + 1:], diag_rows[1:], diag_cols[1:]] = 1
+
```
nguigs, line 138: Assignment won't work in tensorflow...

geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+        Integer representing the shape of the matrices: n x n.
+    """
+
+    def __init__(self, n):
+        super(SpecialLinearLieAlgebra, self).__init__(
+            dim=n**2 - 1,
+            n=n,
+        )
+        # ???: not sure it is in the right place
+        self.ambient_space = GeneralLinear(n)
```
nguigs, line 122: Should be `Matrices(n)`

geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+        point: array-like, shape=[..., n, n]
+            Point.
+
+        Returns
+        -------
+        point: array-like, shape=[..., n, n]
+            Projected point.
+        """
+        projected_point = gs.copy(point)
+        projected_point[..., 0, 0] = projected_point[..., 0, 0] - gs.trace(
```
nguigs, line 208: Assignment -> you can use the first element of the basis: `p -= trace * e0`

nguigs (changes requested):
Great first contribution! Please avoid assignment to make it work with Tensorflow

geomstats/geometry/special_linear.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Returns
+        -------
+        projected : array-like, shape=[..., n, n]
+            Projected point.
+        """
+        scale_coeff = gs.power(gs.abs(gs.linalg.det(point)),
+                               1.0 / self.n)
+        aux_shape = [1] if len(gs.shape(point)) == 2 else [1, 1]
+        return gs.divide(point, gs.reshape(scale_coeff, (-1, *aux_shape)))
```
nguigs, line 75: I didn't know divide worked in that case (multiple arrays). There is an issue related to divide that you could look at: #636

geomstats/geometry/special_linear.py
lpereira95: return gs.isclose(gs.linalg.det(point), 1.0, atol=atol)

geomstats/geometry/special_linear.py
lpereira95: I guess this solves the problem (and avoids use of divide):

```
        proj_point = utils.flip_determinant(point, gs.linalg.det(point))
        scale_coeff = gs.power(gs.linalg.det(proj_point), 1. / self.n)
        return gs.einsum('...ij,...->...ij', proj_point, 1. / scale_coeff)
```

geomstats/geometry/special_linear.py
lpereira95: At the end we can avoid divide and use `gs.einsum` as in other parts of the package.

geomstats/geometry/special_linear.py
lpereira95: this solves the issue (after `Matrices.basis` implementation `gs.array_from_sparse` can be replaced):

```
trace = gs.trace(point, axis1=-2, axis2=-1)
# TODO: use ambient_matrix basis instead
sps_matrix = gs.array_from_sparse([(0, 0)], [1.0], (self.n, self.n))

return gs.copy(point) - gs.einsum('...,...ij->...ij', trace, sps_matrix)
```

lpereira95:
I will close this now and open a new one, since it was open too long ago and tons of commits were done in the meantime.

PR #1215, opened 2021-12-13
Rm deprecated code in lie_groups.py about loss and grad

ninamiolane:
The functions loss and grad are not used anymore: this PR removes them.


ssubra25:
Hey! I notice that the loss and grad function has been removed for Reimannian metric from the riemannian_metric.py and ;ie_group.py file so can you please assist me and tell me how can I compute the grad and loss for reimannian metric? Thanks

nguigs:
Hi @ssubra25 what is the expression of the loss you want to compute? You can uses the method `squared_distance` in any `RiemannianMetric` class and it will be differentiable.

Issue #1197, opened 2021-12-02
Issues with Codecov and DeepSource


ninamiolane:
It seems that:
- DeepSource does not react when docstrings are missing
- Too many docstring make codecov fail (because it's trying to find unit-tests for those docstrings?)

We need to clean this.

ninamiolane:
cf PR #1195 . @SaitejaUtpala any idea there?

ninamiolane:
It seems that this was a one time issue with DeepSource. New tests showed that DeepSource and Codecov behave as expected.

PR #1216, opened 2021-12-13
Address error in CI and documentation

ninamiolane:
This PR addresses issue #1197 .


ninamiolane:
It seems that this is not an error, DeepSource behaved weirdly once.

PR #1220, opened 2021-12-13
Example of geodesic regression on the Grassmann manifold

nguigs:


ninamiolane:
I think it tries to differentiate with respect to the last argument which is of type GrassmannMetric and thus it fails. This error was the reason why we did the hack with private_squared_metric, etc...

Maybe trying to disentangle the GrassmannMetric argument from the function to be differentiated, i.e. having a new function that only depends on the tensor-arguments, since the metric will not change, could help?

Or just avoiding to pass the kwargs as args here:
https://github.com/geomstats/geomstats/blob/f728919217fda31706ae899560c9f32358219eed/geomstats/_backend/pytorch/autodiff.py#L50
and passing them as kwargs instead.

PR #1223, opened 2021-12-14
Add a subcase to torch solve_sylvester

emaignant:
Add the subcase to the resolution of AX+XA=Q when

    Q is a skew matrix
    A is a positive symmetric matrix of rank m-1 (where m is the size of A)


ninamiolane (approved):
Very nice! Usually we should add a unit-test in the backend. I will let it go for this one, since backends' tests will be refactored soon, and I know you have a lot to do with the visualization.

Issue #1118, opened 2021-08-20
Fix parallel transport for TF and Torch


ninamiolane:
The parallel transport is:
- very slow in tensorflow
- inaccurate in pytorch
See, for example, the tests of `test_pullback_metric.py` that are failing in pytorch and tensorflow.

We should look into this.

nguigs:
Which parallel transport are you referring to? The approximation with ladders?

ninamiolane:
These two tests:

https://github.com/geomstats/geomstats/blob/1cdf98d2b3692fc0f52a5ba93c9bed66c9200d44/tests/tests_geomstats/test_pullback_metric.py#L217

https://github.com/geomstats/geomstats/blob/1cdf98d2b3692fc0f52a5ba93c9bed66c9200d44/tests/tests_geomstats/test_pullback_metric.py#L277

and yes the ladder for the second one.

nguigs:
This will be easier to address once #1224 is merged

PR #1224, opened 2021-12-14
Optimize geodesic computation when no closed form exists

alebrigant:
The goal of this PR is to modify the way geodesics are computed in the `connection` class.

These are changes to be made:

1. In classes where `exp` is implemented in closed form, `exp` is renamed into `exp_closed_form`.

2. In `connection`, `exp` is redefined to be `exp_closed_form` if this attribute exists, otherwise, it calls a function `_geodesic_ivp` that integrates the geodesic equation given initial values.

3. In `connection`, a function `_geodesic_bvp` integrates the geodesic equation given boundary values. There are two available methods: geodesic shooting (existing) and scipy's solve_bvp (new).

4. In `connection`, `geodesic` is defined using `exp` or `log` if closed forms are implemented, and `_geodesic_ivp` or `_geodesic_bvp` otherwise.

### DIFF
--- a/geomstats/geometry/special_euclidean.py
+++ b/geomstats/geometry/special_euclidean.py
@@ -1048,8 +1048,8 @@ class SpecialEuclideanMatrixCannonicalLeftMetric(_InvariantMetricMatrix):
         """
         return Matrices.frobenius_product(tangent_vec_a, tangent_vec_b)
 
-    def exp_closed_form(self, tangent_vec, base_point=None, n_steps=10, method="rk4",
-                        **kwargs):
+    def exp(self, tangent_vec, base_point=None, n_steps=10, method="rk4",
+            **kwargs):
         """Exponential map associated to the cannonical metric.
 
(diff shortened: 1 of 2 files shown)


ninamiolane (commented):
Very nice, thanks! I can review the code when the tests pass. Just one comment: I think we should have `_exp_closed_form` as a private function, and still have the manifolds using `exp`. 

The general user will look for an `exp` function, and might not care about how that exp has been computed, thus the user-facing API should have `exp`. What do you think?

ninamiolane:
@alebrigant @nguigs what should we do about this?

@nguigs you mentioned you wanted this PR to solve issue #1118 , but I think this refactor was encountering a API problem in the reorganization of the parent class/child class parameter names.

@alebrigant should we close this PR? Do you think you will give it a try in the future, and if so when?

Sorry if we talked about this at the end of the hackathon and I forgot :)

alebrigant:
@ninamiolane sorry I had forgotten about this PR ! I think the best thing to do is probably to close it. I can give it another try at the next hackathon in February.

Issue #1218, opened 2021-12-13
In Hypersphere, SOn and Stiefel, Log of antipodal points should throw an error


ninamiolane:
Taking the log of two antipodal points should throw an error.

However, the manifolds mentioned above try to compute something and return either:
- 0,
- a complex number.

sdoygb:
I'll take this one. The fix: in `Hypersphere.log`, `SOn.log` and `Stiefel.log`, detect when the base point is antipodal to the target point (geodesic distance = π, i.e. exp-map degenerates / log is multi-valued) and raise an explicit error instead of returning 0 or a complex number.

Plan:
- add an `_is_antipodal` check in the three `log` implementations (with vectorized batch support)
- raise `ValueError` (or the project's preferred exception) with a clear message
- add unit tests covering the antipodal case for each manifold, plus the near-antipodal boundary

PR #1229, opened 2021-12-15
Test if rotation matrices are antipodals before trying to compute the Log

ninamiolane:
This starts to address issue #1218  for the case of Lie group SpecialOrthogonal(3).

This PR:
- introduces the function `are_antipodals`, with its tests
- uses `are_antipodals` tests before trying to take the group log or the Riemannian log on SpecialOrthogonal (Matrices)
- adds the function `matrix_from_rotation_vector` and `rotation_vector_from_matrix` to `_SpecialOrthogonalMatrices`. Previouslly, it was only in SpecialOrthogonal2/3Vectors, which was confusing: I had to go back into the source code to understand where this function was.


ninamiolane:
About the failing tests:
- I am working on cleaning the Docs check in PR #1262 .
- DeepSource errors are not from this PR. @nguigs do you remember why we have `metric` and `base_point` parameters in these `regularize` functions? They are not used at all, should we remove them altogether?
Screen Shot 2021-12-29 at 10 47 37 AM

geomstats/geometry/special_orthogonal.py
```diff
@@ -163,6 +163,120 @@ def vector_from_skew_matrix(self, skew_mat):
...
+
+        Returns
+        -------
+        _ : array-like, shape=[...,]
+            Array of booleans determining if each pair of rotation
+            matrices corresponds to a pair of antipodal rotation
+            matrices.
+        """
+        sq_rot_mat1 = gs.matmul(rotation_mat1, rotation_mat1)
+        sq_rot_mat2 = gs.matmul(rotation_mat2, rotation_mat2)
```
nguigs, line 237: Would it not be faster to compare `rotation_mat1` with `rotation_mat2` and `-rotation_mat2`?

nguigs (commented):
Just a comment on docstrings and efficiency

nguigs:
> * DeepSource errors are not from this PR. @nguigs do you remember why we have `metric` and `base_point` parameters in these `regularize` functions? They are not used at all, should we remove them altogether?

No idea, I think it is because the norm of the vector can change if we change the metric, but the vector representations are not used with any other metrics... I agree we should remove them

geomstats/geometry/special_orthogonal.py
ninamiolane: I'm not sure this would work? This is a sketch of what I had in mind for testing antipodality:

IMG_3466

ninamiolane:
Thanks! I've edited the docstring and removed the unnecessary parameters for the regularize functions. Let me know if you agree with the antipodal tests? Then I'll merge.

ninamiolane:
Just undid the modifications on the regularize_tangent_vec functions, it seems that we are using the metric somewhere. I will raise an issue instead.

PR #1232, opened 2021-12-15
Add triu and test

mortenapedersen:
Adds a function that generates the upper triangular matrix from a given matrix. Including a test.


ninamiolane (changes requested):
Looking great! Could you add it in autograd as well? thanks!

ninamiolane (approved):
Great, thanks!

PR #1235, opened 2021-12-16
add example of computations of geodesics of MPE metric on SPD matrices

nguigs:
for @mortenapedersen

### DIFF
--- a/geomstats/geometry/hamiltonian_geodesics.py
+++ b/geomstats/geometry/hamiltonian_geodesics.py
@@ -1,0 +1,241 @@
+from time import time
+
+from autograd import elementwise_grad, value_and_grad
+from autograd.scipy.integrate import odeint
+from scipy.integrate import solve_bvp
+from scipy.optimize import minimize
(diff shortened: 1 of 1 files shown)


geomstats/geometry/hamiltonian_geodesics.py
```diff
@@ -0,0 +1,241 @@
...
+            return gs.array([H_p, -H_q])
+
+        return vector
+
+    @classmethod
+    def symp_euler(cls, hamiltonian, step_size):
+        def step(state):
+            position, momentum = state
+            dq, _ = cls.symp_grad(hamiltonian)(state)
+            y = gs.array([position + dq, momentum])
```
mortenapedersen, line 143: This line should be:

y = gs.array([position + step_size * dq, momentum])

(that fixed my wrong results when integrating the Hamiltonian eq.'s in the SubRiemannianMetric class)

nguigs:
This was used by @mortenapedersen in #1237 and related PRs and is not meant to be merged.

PR #1236, opened 2021-12-16
Xpennec stiefel tests

xpennec:
Stiefel logarithm raises an error if point and base_point are not in the same sheet for p=n.

### DIFF
--- a/tests/tests_geomstats/test_stiefel.py
+++ b/tests/tests_geomstats/test_stiefel.py
@@ -85,14 +85,15 @@ class TestStiefel(geomstats.tests.TestCase):
 
         self.assertAllClose(gs.shape(result), (self.n, self.p))
 
-    def test_log_two_sheets_error(self):
+    @staticmethod
+    def test_log_two_sheets_error():
         stiefel = Stiefel(n=3, p=3)
         base_point = stiefel.random_point()
         det_base = gs.linalg.det(base_point)
         point = stiefel.random_point()
         det_point = gs.linalg.det(point)
         if gs.all(det_base * det_point > 0.):
-            point = -point
+            point *= -1.0
 
         with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


tests/tests_geomstats/test_stiefel.py
```diff
@@ -81,6 +85,18 @@ def test_random_uniform_shape(self):
 
         self.assertAllClose(gs.shape(result), (self.n, self.p))
 
+    @geomstats.tests.np_only
+    def test_log_two_sheets_error(self):
+        base_point = Stiefel(n=3, p=3).random_point()
```
ninamiolane, line 90: stiefel = Stiefel(n=3, p=3) can be instantiated only once

tests/tests_geomstats/test_stiefel.py
```diff
@@ -81,6 +85,18 @@ def test_random_uniform_shape(self):
...
+        det_base = gs.linalg.det(base_point)
+        point = Stiefel(n=3, p=3).random_point()
+        det_point = gs.linalg.det(point)
+        if gs.all(det_base * det_point > 0.):
+            point[:, 1] *= -1.0
```
ninamiolane, line 95: point = - point

tests/tests_geomstats/test_stiefel.py
```diff
@@ -81,6 +85,18 @@ def test_random_uniform_shape(self):
 
         self.assertAllClose(gs.shape(result), (self.n, self.p))
 
+    @geomstats.tests.np_only
```
ninamiolane, line 88: We can remove this decorator, after the removal of the assignment it should work in three backends

ninamiolane (changes requested):

ninamiolane (approved):
Great, thank you!

PR #1249, opened 2021-12-17
Introduce n-fold manifolds

nguigs:
For the direct product of a given manifold repeated n times.

### DIFF
--- a/geomstats/geometry/base.py
+++ b/geomstats/geometry/base.py
@@ -26,7 +26,7 @@ class VectorSpace(Manifold, abc.ABC):
         if "dim" not in kwargs.keys():
             kwargs["dim"] = int(gs.prod(gs.array(shape)))
         super(VectorSpace, self).__init__(
-            default_point_type=default_point_type, **kwargs
+            shape=shape, default_point_type=default_point_type, **kwargs
         )
         self.shape = shape
 
@@ -167,6 +167,8 @@ class LevelSet(Manifold, abc.ABC):
         default_coords_type="intrinsic",
         **kwargs
     ):
+        if "shape" not in kwargs:
+            kwargs["shape"] = embedding_space.shape
         super(LevelSet, self).__init__(
             dim=dim,
             default_point_type=embedding_space.default_point_type,
@@ -318,6 +320,8 @@ class OpenSet(Manifold, abc.ABC):
     def __init__(self, dim, ambient_space, **kwargs):
         if "default_point_type" not in kwargs:
             kwargs["default_point_type"] = ambient_space.default_point_type
+        if "shape" not in kwargs:
+            kwargs["shape"] = ambient_space.shape
         super().__init__(dim=dim, **kwargs)
         self.ambient_space = ambient_space
 
--- a/geomstats/geometry/base.py
+++ b/geomstats/geometry/base.py
@@ -8,6 +8,8 @@ import abc
 import geomstats.backend as gs
 from geomstats.geometry.manifold import Manifold
 
+POINT_TYPES = {1: "vector", 2: "matrix"}
+
 
 class VectorSpace(Manifold, abc.ABC):
     """Abstract class for vector spaces.
@@ -22,12 +24,10 @@ class VectorSpace(Manifold, abc.ABC):
         Optional, default: 'vector'.
     """
 
-    def __init__(self, shape, default_point_type="vector", **kwargs):
+    def __init__(self, shape, **kwargs):
         if "dim" not in kwargs.keys():
             kwargs["dim"] = int(gs.prod(gs.array(shape)))
-        super(VectorSpace, self).__init__(
-            shape=shape, default_point_type=default_point_type, **kwargs
-        )
+        super(VectorSpace, self).__init__(shape=shape, **kwargs)
         self.shape = shape
 
     def belongs(self, point, atol=gs.atol):
@@ -47,13 +47,8 @@ class VectorSpace(Manifold, abc.ABC):
         belongs : array-like, shape=[...,]
             Boolean evaluating if point belongs to the space.
         """
-        if self.default_point_type == "vector":
-            point_shape = point.shape[-1:]
-            minimal_ndim = 1
-        else:
-            point_shape = point.shape[-2:]
-            minimal_ndim = 2
-        belongs = point_shape == self.shape
+        minimal_ndim = len(self.shape)
+        belongs = point.shape[-minimal_ndim:] == self.shape
         if point.ndim == minimal_ndim:
             return belongs
         return gs.tile(gs.array([belongs]), [point.shape[0]])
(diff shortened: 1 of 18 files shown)


geomstats/geometry/base.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, shape, default_point_type="vector", **kwargs):
         if "dim" not in kwargs.keys():
             kwargs["dim"] = int(gs.prod(gs.array(shape)))
         super(VectorSpace, self).__init__(
-            default_point_type=default_point_type, **kwargs
+            shape=shape, default_point_type=default_point_type, **kwargs
```
ninamiolane, line 26: 🎉

geomstats/geometry/product_manifold.py
```diff
@@ -311,3 +312,263 @@ def is_tangent(self, vector, base_point, atol=gs.atol):
...
+
+    Define a manifold as the product manifold of n copies of a given base manifold M.
+
+    Parameters
+    ----------
+    base_manifold : Manifold
+        Base manifold.
+    n_copies : int
+        Number of replication of the base manifold.
+    """
```
ninamiolane, line 328: Maybe add the other parameters too?

geomstats/geometry/product_manifold.py
```diff
@@ -311,3 +312,263 @@ def is_tangent(self, vector, base_point, atol=gs.atol):
...
+        Returns
+        -------
+        samples : array-like, shape=[..., n_copies, base_shape]
+            Points sampled on the product manifold.
+        """
+        sample = self.base_manifold.random_point(n_samples * self.n_copies, bound)
+        reshaped = gs.reshape(sample, (n_samples, self.n_copies) + self.base_shape)
+        return gs.squeeze(reshaped)
+
+    def projection(self, point):
```
ninamiolane, line 445: Missing docstring (but DS got it)

geomstats/geometry/product_manifold.py
```diff
@@ -311,3 +312,263 @@ def is_tangent(self, vector, base_point, atol=gs.atol):
...
+    ----------
+    base_metric : RiemannianMetric
+        Base metric.
+    n_copies : int
+        Number of replication of the base metric.
+    base_shape : tuple of int
+        Shape of one element of the underlying base manifold
+    """
+
+    def __init__(self, base_metric, n_copies, base_shape):
```
ninamiolane, line 471: Should we do:
self.base_shape = base_metric.shape

and have each metric have a shape attribute too?

It would be more symmetric between the Metric and Manifold classes.

geomstats/geometry/product_manifold.py
```diff
@@ -311,3 +312,263 @@ def is_tangent(self, vector, base_point, atol=gs.atol):
...
+        inner_each = self.base_metric.inner_product(vector_a, vector_b, point_)
+        reshaped = gs.reshape(inner_each, (-1, self.n_copies))
+        return gs.squeeze(gs.sum(reshaped, axis=-1))
+
+    def exp(self, tangent_vec, base_point, **kwargs):
+        """Compute the Riemannian exponential of a tangent vector.
+
+        Parameters
+        ----------
+        tangent_vec : array-like, shape=[..., n_copies, base_shape]
```
ninamiolane, line 535: technically, it would be *base_shape, but maybe we should keep it this way otherwise it's going to be hard to read.

ninamiolane (changes requested):
Great! Should we have default_point_type be defined only in the base Manifold, through point_types = {2: "vector", 3: "matrix"} and shape.ndim?

Tests and DS should pass.

geomstats/geometry/base.py
nguigs: Good idea!

nguigs:
The remaining DeepSource issues are not related to this PR @ninamiolane

geomstats/geometry/product_manifold.py
```diff
@@ -7,12 +7,13 @@
...
 from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.product_riemannian_metric import ProductRiemannianMetric
+from geomstats.geometry.riemannian_metric import RiemannianMetric
 
 
 class ProductManifold(Manifold):
     """Class for a product of manifolds M_1 x ... x M_n.
 
-    In contrast to the class Landmarks or DiscretizedCruves,
+    In contrast to the classes NFoldManifolds, Landmarks, or DiscretizedCruves,
```
ninamiolane, line 16: NIT: NFoldManifold (no s) and DiscretizedCurves (not Cruves)

geomstats/geometry/heisenberg.py
```diff
@@ -1,12 +1,13 @@
...
 from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.lie_group import LieGroup
 from geomstats.geometry.symmetric_matrices import SymmetricMatrices
 
 
-class HeisenbergVectors(LieGroup):
+class HeisenbergVectors(LieGroup, VectorSpace):
     """Class for the 3D Heisenberg group in the vector representation.
 
     The 3D Heisenberg group represented as R^3. It is a step-2 Carnot Lie
```
ninamiolane, line 13: NIT: Actually, the parameters of the init should be added, instead of "No parameters" in the class' docstring

geomstats/geometry/product_manifold.py
```diff
@@ -311,3 +318,291 @@ def is_tangent(self, vector, base_point, atol=gs.atol):
 
         is_tangent = gs.all(is_tangent, axis=-1)
         return is_tangent
+
+
+class NFoldManifold(Manifold):
```
ninamiolane, line 323: Maybe it should have its own `nfold_manifold.py` file, for visibility for new users?

ninamiolane: It does not inherit from ProductManifold bcs the machinery is different?

ninamiolane (approved):
great, thanks! I have just put minor comments in the PR directly.

geomstats/geometry/heisenberg.py
nguigs: But there are no parameters for the init either!

geomstats/geometry/product_manifold.py
nguigs: Yes ProductManifold won't work for matrix point types, so I started from scratch to avoid this issue.
I think a user might be looking for a product manifold before realizing they can use the nfold so it's good to keep it here. The name of the file won't matter though if we get to do imports from the geometry module itself. Did you try?

ninamiolane: Makes sense.

I feel that exposing a module `nfold_manifold.py` makes it easier for the user going through the files (on GitHub) to see that we have it. But that may be personal, and it does not matter too much.

Yes I tried doing the imports, it was working first but then I ran into circular imports, and stopped: I will not have the bandwidth to do it this quarter. There is a closed PR on this, if you want to have a look. I wonder if having one class per .py could help disentangle the imports?

ninamiolane (approved):

PR #1251, opened 2021-12-17
Parametrization - 1

SaitejaUtpala:
clean tests. will clean one test file at time


ninamiolane (commented):
Excellent, thank you so much! We had been talking about splitting the tests for such a long time (@xpennec @nguigs you might be interested in this!).

It looks perfect, very readable 🙌 . The only downsides I could see with this design are:
- within the `*_data` functions, it is hard to known what the data actually represent (are they tangent vectors, symmetric matrices, or the expected results?)
  - ---> Could a `dict` structure help with this? You use a combo `string` + `list`, thus why not a `dict`? If a `dict` is not possible (because of pytest's API?), maybe naming variables within the `*_data` functions could already be helpful.
-  The `_data` values that are input into the tests are far from the actual test in the `*_test.py` file: the user will need to scroll up and down... I'm not sure this can be avoided, and I think this is completely fine.

Many thanks again! 🎉 🎄

SaitejaUtpala:
> Excellent, thank you so much! We had been talking about splitting the tests for such a long time (@xpennec @nguigs you might be interested in this!).
> 
> It looks perfect, very readable raised_hands . The only downsides I could see with this design are:
> 
>     * within the `*_data` functions, it is hard to known what the data actually represent (are they tangent vectors, symmetric matrices, or the expected results?)
>       
>       * ---> Could a `dict` structure help with this? You use a combo `string` + `list`, thus why not a `dict`? If a `dict` is not possible (because of pytest's API?), maybe naming variables within the `*_data` functions could already be helpful.
> 
>     * The `_data` values that are input into the tests are far from the actual test in the `*_test.py` file: the user will need to scroll up and down... I'm not sure this can be avoided, and I think this is completely fine.
> 
> 
> Many thanks again! tada christmas_tree

yaa sure. I have made only smoke tests dictionary and peeled values while generating tests. regarding scrolling, usually with recent rewrite, there will be little to zero other computation in the files. So there should not be that much scrolling

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    smoke: Basic testing.
+    rt: tests on large number of randomly generated test data
```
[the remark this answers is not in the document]
ninamiolane: Maybe:
random: Tests on large number of randomly generated test data.
(rt --> random to be consistent with the naming).

tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
```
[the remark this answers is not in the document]
ninamiolane: Could this go in geomstats/tests.py ? I feel geomstats/tests.py has some configuration wrt tests already, maybe it could be nice to have everything at the same place?

tests/tests_geomstats/test_spd_matrices.py
```diff
@@ -4,6 +4,9 @@
 import math
 import warnings
 
+import pytest
+from numpy.linalg import cholesky
```
ninamiolane, line 8: No numpy

tests/tests_geomstats/test_spd_matrices.py
```diff
@@ -19,192 +22,156 @@
...
 
-class TestSPDMatrices(geomstats.tests.TestCase):
-    """Test of SPDMatrices methods."""
-
-    def setup_method(self):
-        """Set up the test."""
-        warnings.simplefilter("ignore", category=ImportWarning)
+def belongs_data():
+    smoke_data = [
+        dict(n=2, mat=[[3.0, -1.0], [-1.0, 3.0]], expected=True),
```
ninamiolane, line 32: Any reason why you use the `dict` syntax as opposed to `{}` with keys as strings? (no need to change everything, I am just curious).

ninamiolane (commented):
Excellent, I think this is a great design. Congrats! 

I have left comments inline directly (see also the comments from my former review.), but at a high-level:
- tests/conftest.py and tests.py could be merged.
- docstrings should be added
- tests need to pass.

Excited to see this coming along 💯

ninamiolane (commented):
Also, some tests are missing:
- random_point_and_belongs
- the _vectorizations
Is it on purpose?

tests/tests_geomstats/test_spd_matrices.py
```diff
@@ -19,192 +22,156 @@
...
+                [1.0, 1 / 3, 1 / 3],
+                [1 / 3, 0.125, 0.125],
+                [1 / 3, 0.125, 0.125],
+            ],
+        )
+    ]
+    return generate_tests(smoke_data)
+
+
+def inverse_differential_power_data(power, tangent_vec, base_point, expected):
```
ninamiolane, line 102: Should this function take input args? They are not used + upon calling the function, no args is passed as input in the test files.

tests/tests_geomstats/test_symmetric_matrices.py
```diff
@@ -1,130 +1,165 @@
...
-    def test_projection_and_belongs(self):
-        shape = (2, self.n, self.n)
-        result = helper.test_projection_and_belongs(self.space, shape)
-        for res in result:
-            self.assertTrue(res)
-
-    def test_random_and_belongs(self):
-        mat = self.space.random_point()
-        result = self.space.belongs(mat)
+    class TestingData(TestData):
```
ninamiolane, line 17: NIT: TestingData(TestData) looks a bit redundant and weird: 

Maybe `SymmetricMatricesData(TestData)` ?

PR #1252, opened 2021-12-17
cleaning github actions

SaitejaUtpala:
separate linting and testing in github actions


ninamiolane:
Yes, that's a great idea! 🙌

Issue #1194, opened 2021-11-29
Support for general function/Hilbert spaces


kiranvad:
Hello everyone,
Really enjoy this package both for understanding and computing with differential geometry and manifolds.
This is a follow-up discussion with @ninamiolane on Twitter about the prospect of having general function spaces or Hilbert spaces as manifolds along with the discrete curves.
Lots of experimental data collected in material science is in the form of functions (optical spectroscopy wavelength vs intensity). While this can in principle is considered as discrete curves in R2, the resulting metric would be suboptimal.
We recently started using functional spaces and their geometry to compute statistics and distances for this type of data and were wondering if generalizing them would be of interest to geomstats team.
I have a very crude implementation [here](https://github.com/kiranvad/geomstats/blob/948b6fb4650ff3e62ee63aaad94f5926f84a3b3d/geomstats/geometry/functions.py) for a few of the spaces such as the L2, infinite-dimensional Hypersphere, SRSF (which I think is very closely related to SRVF implemented herein).
Would love to chat more about this if there's sufficient interest either here or over email.

Thank you!
Kiran Vaddi

ninamiolane:
Hi @kiranvad thank you very much for reaching out! Good to have you here.

Indeed, geomstats implementation only focuses on [discretized curves](https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/discrete_curves.py) with a fixed number of sampling points. [See a notebook on this here](https://github.com/geomstats/geomstats/blob/master/notebooks/08_shape_analysis.ipynb).

Your implementation would generalize this?

Additionally, I see that you are writing methods related to information geometry: fantastic! Do you think these methods could be integrated into our model [information_geometry](https://github.com/geomstats/geomstats/tree/master/geomstats/information_geometry)? S[ee a notebook on this here](https://github.com/geomstats/geomstats/blob/master/notebooks/06_information_geometry.ipynb).

To summarize, there is definitely a huge interest on our side to integrate your methods here! I would be happy to assist you in the integration of this code, and I believe that @alebrigant and @nguigs could be interested too.

kiranvad:
That sounds great! 
The implementations I linked are pretty rudimentary but yes they are somewhat generalizations of discretized curves (cf.  [this book](https://link.springer.com/book/10.1007/978-1-4939-4020-2)) and I would love to work with all of you to improve them and learn more.
I will try to put together a pull request for these and maybe we can go from there?

nguigs:
[...]

PR #1257, opened 2021-12-18
Function spaces implementation : starting with infinite dimensional spheres

kiranvad:
Hello,
This is a follow-up for #1194 (apologies for the extreme delay on this)
This PR:
* Introduces function spaces module with the implementation of infinite-dimensional spheres with L2 metrics that are mentioned in Chapter 4.3.2 in the book [Functional and Shape Data Analysis](https://link.springer.com/book/10.1007/978-1-4939-4020-2).
* This class will later be used to aid the implementation of general probability measures under the Fisher-Rao metric and SRSF for functions to be added later.


ninamiolane:
Excellent @kiranvad , thank you very much! We can look at the code once you request our reviews. In the meantime, it seems that some GitHub actions are failing: could you have a look? Thanks!

kiranvad:
Hi @ninamiolane, this PR should be ready for review now. Looking forward to your feedback!

ninamiolane:
Looking great @kiranvad , thank you for this great addition ! Some "Linting and Testing" workflow are still failing though, could you have a look (click on "Details" to see what is the bug) before we get a human code reviewer on this? Many thanks! 🙌 

Don't worry about the "Docs check and Deployment" workflow though, you do not need to fix this one.

Also, very very minor but the formatting of the docstrings is not exact (just missing blank lines), see here:
https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#the-anatomy-of-a-docstring

(Respecting this exact docstring format helps generating the doc website automatically).

kiranvad:
Hi @ninamiolane , thanks for the feedback. 
I have a very quick question regarding the testing workflows. It appears that test functions attributes from the `setup_method` are not being accessible from other methods in the file/module (that's the error and I can reproduce it on my local machine for other functions in the tests folder. I tried this with `test_euclidean`). 
On my local machine, I can fix this by renaming `setup_method` to the usual `setUp`.
Running `nose2 --verbose tests.tests_geomstats.test_functions` then works perfectly fine for me.
I am not familiar with fixing these issues so any help would be greatly appreciated.

ninamiolane:
@kiranvad sorry, this is our bad! Our contributing guidelines were still indicating to use `nose2` as our testing suite, whereas we actually moved to `pytest` now. I am updating the contributing guidelines in this PR #1262 . So sorry, I hope you did not loose too much time on this. 

`setUp` was used by `nose2`,  `setup_method` is used by `pytest`. 

Could you try again with `pytest` and `setup_method`? I didn't see anything wrong with your code otherwise.

tests/tests_geomstats/test_functions.py
```diff
@@ -0,0 +1,44 @@
...
+import numpy as np
+
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.functions import SinfSpace
+
+
+class TestFunctionsSinf(geomstats.tests.TestCase):
+    def setup_method(self):
+        self.domain = gs.linspace(-math.pi, math.pi)
```
ninamiolane, line 15: @kiranvad I believe replacing this by:
`gs.linspace(-gs.pi, gs.pi, num-50)` would solve the remaining issues.

kiranvad: Thanks for the suggestion @ninamiolane 
I fixed this but the tests still fail with a pytorch backend because `gs.vectorize` thrown an error. I suspect this is because I make use of the `signature` parameter that exists in NumPy but not in PyTorch. I am not sure how to fix this, so for now I made the tests only for NumPy and autograd.

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+import math
+
+import numpy as np
+
+import geomstats.backend as gs
+from geomstats.geometry.manifold import Manifold
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+
+
+class SinfSpaceMetric(RiemannianMetric):
```
nguigs, line 12: Could we use a more explicit class name? Sinf refers to the notation and does not read well if you don't know what it stands for.

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+    r"""A Riemannian metric on the S_{\inf} space
+
+    Parameters:
+    -------
+    domain_samples : array of shape (n_samples, )
+        grid points on the domain
```
nguigs, line 18: Add a capital letter and a full stop.

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+            Optional, default: None.
+
+        Returns
+        -------
+        inner_product : array-like, shape=[...,]
+            Inner-product.
+        """
+
+        def _f(v1, v2):
+            return gs.cast(np.trapz(v1 * v2, x=self.x), gs.float32)
```
nguigs, line 49: The `trapz` method could be added to the backend, see the equivalent in [torch](https://pytorch.org/docs/stable/generated/torch.trapezoid.html) and [tensorflow](https://www.tensorflow.org/probability/api_docs/python/tfp/math/trapz).

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+
+        inner_prod = gs.vectorize(
```
nguigs, line 51: There is an `axis` parameter in the `trapz` that should allow to vectorize your computations without using the `gs.vectorize` method. That could help making the code work with other backends.

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+        -------
+        exp : array-like, shape=[..., n_evals]
+            Point on the hypersphere equal to the Riemannian exponential
+            of tangent_vec at the base point.
+        """
+
+        def _f(v, p):
+            norm_v = self.norm(v)
+            t1 = gs.cos(norm_v) * p
+            t2 = (gs.sin(norm_v) / norm_v) * p
```
nguigs, line 81: All these operations are already vectorized or can be by using einsum, please have a look at the finite dim equivalent. Also the use of taylor_exp_even_func make sure we stay precise around 0.
https://github.com/geomstats/geomstats/blob/d89ee0a4eb8cd178a5de5bccc095fda52d9c0732/geomstats/geometry/hypersphere.py#L743-L747

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+        Returns
+        -------
+        log : array-like, shape=[..., n_evals]
+            Tangent vector at the base point equal to the Riemannian logarithm
+            of point at the base point.
+        """
+
+        def _f(p0, p1):
+            theta = gs.arccos(self.inner_product(p1, p0, base_point=None))
+            return (p1 - p0 * gs.cos(theta)) * (theta / gs.sin(theta))
```
nguigs, line 113: Same comment as above.

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+        ----------
+        point : array-like, shape=[..., dim]
+            discrete evaluation of a function.
+        Returns
+        -------
+        projected_point : array-like, shape=[..., dim]
+            Point projected to the hypersphere.
+        """
+        norm_p = self.metric.norm(point)
+        return point / norm_p
```
nguigs, line 161: Does this work for multiple samples?

geomstats/geometry/functions.py
```diff
@@ -0,0 +1,245 @@
...
+        base_point : array-like, shape=[..., dim]
+            Point on the manifold.
+
+        Returns
+        -------
+        tangent_vec : array-like, shape=[..., dim]
+            Tangent vector at base point.
+        """
+        inner_product = self.metric.inner_product(vector, base_point)
+        tangent_vec = vector - inner_product
```
nguigs, line 220: Is this right? inner_product is a scalar, shouldn't it be multiplied by a base_point?

tests/tests_geomstats/test_functions.py
```diff
@@ -0,0 +1,46 @@
...
+        self.manifold = SinfSpace(self.domain)
+
+    @geomstats.tests.np_and_autograd_only
+    def test_manifold(self):
+        proj_f = self.manifold.projection(self.f)
+        result = self.manifold.belongs(proj_f)
+        self.assertTrue(result, "Expected True but got %s" % result)
+
+    @geomstats.tests.np_and_autograd_only
+    def test_metric(self):
```
nguigs, line 24: Could you add lower level tests (test_inner_product, test_exp, test_log, test_log_and_exp, etc?)

tests/tests_geomstats/test_functions.py
```diff
@@ -0,0 +1,46 @@
...
+        num_funcs = 10
+        point_a = self.funcs(1)
+        point_b = self.funcs(2)
```
nguigs, line 27: Could you also test with multiple inputs?

nguigs (changes requested):
Good input, thanks! 
- I made some comments that will allow to generalize the code to other backends.
- The code does not seem vectorized everywhere, have a look at the Hypersphere class to see how to do.
- In fact this manifold and its metric are very close to the Hypersphere ones, could we reuse code from there instead of rewriting it? I guess the only substantial change is the computation of the inner_product.

ninamiolane:
@kiranvad any update on this?

kiranvad:
Hi @ninamiolane and @nguigs 
Thanks for the feedback. Still working through them. Got back from some time away from Seattle recently. Will try to push changes in a couple of weeks or so.

ninamiolane:
@kiranvad excellent, keep us in touch, we're looking forward to this.

geomstats/geometry/functions.py
kiranvad: Thanks, I am calling this a `HilbertSphere` in the updated code

geomstats/geometry/functions.py
kiranvad: Great, this is really helpful. I tried to use as much code as possible but it was not straightforward to define the new metric using the `HypersphereMetric`. The main challenge is we need the embedding metric to be  `L2Metric` on functions. I did not want to define one separately but if the existing `L2Metric` can be updated (somehow), I think I can write the new metric mostly as another sub-class of Hypersphere metric.

geomstats/geometry/functions.py
kiranvad: I think the updated one should work! thanks for pointing this out.

geomstats/geometry/functions.py
kiranvad: Yeah, that was wrong. I updated this code to do this properly. Thanks for pointing this out again.

kiranvad:
Update the code for now. Will modify the tests accordingly.

kiranvad:
For some reason, I am not able to push changes to this branch. Suspect the error is the [following](https://github.com/psf/black/issues/2964). Will follow up on this.

ninamiolane:
@kiranvad sounds good, let us know if we can help? feel free to join us on slack (link on geomstats readme) if you have any question.

@lpereira95 : would you know what could be happening with this PR?

kiranvad:
Hello @ninamiolane and @nguigs 
An update on this PR. The latest commits should:
1. Incorporate and address changes requested by @nguigs 
2. All the tests related to `geometry/functions.py` pass but I need to implement a backend function for [_trapz](https://github.com/kiranvad/geomstats/blob/0fef30d0ee4f3254b471eb560fadd0f6c9e18177/geomstats/geometry/functions.py#L13) to get this work with PyTorch and TensorFlow. I couldn't find any resources on how to do this (I tried simply adding it to the folders in `_backend` but that seems to break the code). Any suggestions?

ninamiolane:
Hi @kiranvad , thank you for the update! I see that [pytorch has trapz](https://pytorch.org/docs/stable/generated/torch.trapz.html), [tf also](https://www.tensorflow.org/probability/api_docs/python/tfp/math/trapz) and autograd probably has it. 

Thus, the strategy you mention (putting it in `_backend/pytorch/__init__.py` and likewise for the other backends) should work. What broke the code when you tried?

@lpereira95 any chance you could help @kiranvad with the backends? 🙌  Thank you!

tests/tests_geomstats/test_functions.py
kiranvad: Done!

tests/tests_geomstats/test_functions.py
kiranvad: Done!

lpereira95:
Sure, I'll add them to the backend this afternoon, so @kiranvad can assume they're available as `gs.trapz`.

kiranvad:
Thanks, @lpereira95, I modified the code to use `gs.trapz` using #1522 and increased the testing to `pytorch` backend as well.

kiranvad:
Hello @ninamiolane,
I think this PR is ready for another review. The test fails are not related to the contents this PR adds. 
Let me know if you'd like to see more changes or if it's good to go.

ninamiolane (approved):
Looking good, many thanks!!

PR #1258, opened 2021-12-20
fix docs

SaitejaUtpala:
geomstats.ai doesn't contain updated docs.


ninamiolane (approved):
Excellent, thank you so much!

PR #1259, opened 2021-12-20
Fix docs

SaitejaUtpala:
* run sphinx build everytime and fill docs folder (or complain if there is an error)
* commit changes 
* deploy to geomstats.ai on merge


ninamiolane:
See PR #1258 and PR #1259

PR #1272, opened 2022-01-10
doc string auto generation template and checker

SaitejaUtpala:
This PR adds

- automatic generation of doc string template
- check doc string convention with flake8-docstrings


ninamiolane:
Fantastic, thanks! Linting is failing: maybe we could get rid of flake8, and leave the linting to DeepSource only? So that we do not have to recalibrate flake8 each time?

SaitejaUtpala:
> Fantastic, thanks! Linting is failing: maybe we could get rid of flake8, and leave the linting to DeepSource only? So that we do not have to recalibrate flake8 each time?

It's not because of linting in code. But linting in doc strings, like first line doesn't contain period and there is not doc string. I have already made it part of pre-commit hooks. commits will actually fail if flake8-docstrings is installed. so ppl will correct things locally instead of waiting for deep-source to complain

ninamiolane:
Makes sense, good point! 💯 Let's correct the Linting failing Gihub Action then.

PR #1273, opened 2022-01-11
Riemannian Mean Shift algorithm

shubhamtalbar96:
Added Riemannian Mean Shift algorithm implementation for manifold. The following learning method assumes a numpy backend and flat kernel method.


geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+        return np.array(out).reshape((n_A, n_B))
+
+    def fit(
+        self,
+        points,
+        n_centers,
+        n_jobs=1,
+        max_iter=100,
+        init_centers="from_points",
+        kernel="flat",
```
nguigs, line 71: Following scikit learn's API, these parameters should be attributes of the object and passed to init.
`points` should be named `X`, and a dummy `y=None` argument needs to be added for consistency with other estimators.

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+        self.centers = centers
+
+    def predict(self, points):
+        """
+        Predict the closest cluster each point in 'points' belongs to
+        """
+
+        if self.centers is None:
+            raise Exception("Not fitted")
+        else:
```
nguigs, line 139: else is not necessary here

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+            out = []
+            for i in range(points.shape[0]):
+                j_closest_center = self.metric.closest_neighbor_index(
+                    points[i, :], self.centers
+                )
+                out.append(self.centers[j_closest_center, :])
+
+            return np.array(out)
```
nguigs, line 147: Use gs instead of np

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+        Fit centers in all the input points to find 'n_centers' number of clusters.
+        """
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_mean(points, weights):
+            return self.mean.fit(points, weights=weights).estimate_
+
+        # 'from_points' is the default initialization for centers used
+        if init_centers == "from_points":
```
nguigs, line 84: Different options should be listed in the docstring

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+            out = pool(
+                pickable_mean(points_to_average[j], nonzero_weights[j])
+                for j in range(centers.shape[0])
+            )
+
+            new_centers = np.array(out)
+
+            displacements = [
+                self.metric.dist(centers[j], new_centers[j]) for j in range(n_centers)
+            ]
```
nguigs, line 124: The dist function works with lists of inputs. Are `centers[j]` and `new_centers[j]` single points? In this case you can use `self.metric.dist(centers, new_centers)` directly

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,147 @@
...
+            dists = self.__intersets_distances(
+                centers, points, n_jobs=n_jobs, **joblib_kwargs
+            )
+
+            # assuming the use of 'flat' kernel
+            if kernel == "flat":
+                weights = np.ones_like(dists)
+
+            weights[dists > self.bandwidth] = 0.0
+            weights = weights / weights.sum(axis=1, keepdims=1)
```
nguigs, line 101: Use gs.sum instead

nguigs (commented):
Great addition to the package, thanks! A few remarks on code, conventions and docstrings.
Please use the gs backend (`import geomstats.backend as gs`) instead of `np` and it should be straightforward to have this work with the other backends.

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Thank you Nicolas for your comments. I have raised a new PR with the changes that you have suggested. Please review them at your convenience and let me know of more changes or improvements.

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -1,147 +1,161 @@
...
-            The Geomstats Riemannian metric associated to the space used.
-        bandwidth : size of neighbourhood around each center
-            All points in 'bandwidth' size around center are considered for
-            calculating new mean centers.
-        tot : float, stopping condition
-            Computation of subsequent mean centers is stopped when the distance
-            between them is less than 'tot'
-        mean : mean method used to calculate the new center of a cluster
-        """
+    """
```
nguigs, line 14: Title needs to be on the first line, right after the quotes """, and there needs to be a blank line between the title and the description.

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -1,147 +1,161 @@
...
+    density function given discrete data sampled from that function. It is
+    an iterative method for finding the centers of a collection of clusters.
+
+    Following implementation assumes a flat kernel method.
+
+    Parameters
+    ----------
+    manifold : object of class RiemannianManifold
+        Geomstats Riemannian manifold on which the
+        Riemannian Mean Shift algorithm is to applied.
```
nguigs, line 26: TYPO: to apply or is to be applied

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -1,147 +1,161 @@
...
+        Geomstats Riemannian metric associated to the space used.
+    bandwidth : size of neighbourhood around each center
+        All points in 'bandwidth' size around center are considered for
+        calculating new mean centers.
+    tot : float, stopping condition
+        Computation of subsequent mean centers is stopped when the distance
+        between them is less than 'tot'
+    mean : Mean method used to calculate the new center of a cluster
+    n_centers : Total initial centers
+    n_jobs : Number of parallel threads to be initiated for parallel computation
```
nguigs, line 37:
```suggestion
    mean : str
        Method used to calculate the new center of a cluster.
    n_centers : int
        Number of centers.
        Optional, default : 1.
    n_jobs : int
        Number of parallel threads to be initiated for parallel 
        Optional, default : 1.
```

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -1,147 +1,161 @@
...
-        n_jobs=1,
-        max_iter=100,
-        init_centers="from_points",
-        kernel="flat",
-        **joblib_kwargs
-    ):
-        """
-        Fit centers in all the input points to find 'n_centers' number of clusters.
-        """
+    def fit(self, X):
```
nguigs, line 92: For compatibility with pipelines
```suggestion
    def fit(self, X, y=None):
```

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -1,147 +1,161 @@
...
-                    points[i, :], self.centers
-                )
-                out.append(self.centers[j_closest_center, :])
-
-            return np.array(out)
+
+        out = []
+        for i in range(points.shape[0]):
+            j_closest_center = self.metric.closest_neighbor_index(
+                points[i, :], self.centers
```
nguigs, line 157: use `points[i]` as points may be matrices.

nguigs (commented):
Thanks, good corrections! Could you maybe add a test with two clusters?

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+    def fit(self, X):
+        """Fit centers in all the input points to find 'n_centers' number of clusters"""
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_mean(points, weights):
+            """Frechet Mean of all points weighted by weights"""
+            return self.mean.fit(points, weights=weights).estimate_
+
+        # 'from_points' is the default initialization for centers used
```
ninamiolane, line 101: No left-over comments

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        if self.init_centers == "from_points":
+            n_points = X.shape[0]
+            centers = X[gs.random.randint(n_points, size=self.n_centers), :]
+        if self.init_centers == "random_uniform":
+            n_centers = centers.shape[0]
+            centers = self.manifold.random_uniform(n_samples=n_centers)
+
+        print(type(self.max_iter))
```
ninamiolane, line 109: remove left-over prints

geomstats/learning/riemannian_mean_shift.py
ninamiolane: The comments about the code can go in the docstring of `pickable_mean`

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        @joblib.wrap_non_picklable_objects
+        def pickable_mean(points, weights):
+            """Frechet Mean of all points weighted by weights"""
+            return self.mean.fit(points, weights=weights).estimate_
+
+        # 'from_points' is the default initialization for centers used
+        if self.init_centers == "from_points":
+            n_points = X.shape[0]
+            centers = X[gs.random.randint(n_points, size=self.n_centers), :]
+        if self.init_centers == "random_uniform":
```
ninamiolane, line 105: if -> elif

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        self.tol = tol
+        self.mean = FrechetMean(self.metric, **frechet_mean_kwargs)
+        self.n_centers = n_centers
+        self.n_jobs = n_jobs
+        self.max_iter = max_iter
+        self.init_centers = init_centers
+        self.kernel = kernel
+        self.centers = None
+
+        print(type(max_iter))
```
ninamiolane, line 70: Rm leftover prints

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        print(max_iter)
+
+    def __intersets_distances(self, points_A, points_B, **joblib_kwargs):
```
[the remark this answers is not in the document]
ninamiolane: points_A -> points_a

ninamiolane, line 73: This function is great! Can't believe we did not have it (@nguigs ). Could rename it to `dist_intersets` and put it on the Line 389 on this file?
https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/riemannian_metric.py#L389

It would come just after:
- dist_broadcast
- dist_pairwise
which are doing similar computations (but not quite).

It would also become a public function, which means that a docstring will be required (with  parameters' description).

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        n_A, n_B = points_A.shape[0], points_B.shape[0]
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_dist(x, y):
+            """Riemannian metric between points x & y"""
+            return self.metric.dist(x, y)
+
+        pool = joblib.Parallel(n_jobs=self.n_jobs, **joblib_kwargs)
+        out = pool(
```
ninamiolane, line 84: You can get rid of the indices i,j here, it will improve readability.

This could be:
`pickable_dist(one_point_a, one_point_b) for one_point_a in points_a for one_point_b in points_b`

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+            for j in range(self.n_centers):
+                indexes = gs.where(weights[j] > 0)
+                nonzero_weights += [
+                    weights[j][indexes],
+                ]
+                points_to_average += [
+                    X[indexes],
+                ]
+
+            # compute Frechet means in parallel
```
ninamiolane, line 132: Excellent!! 🎉  please juste remove the comment: comments are hard to maintain: very quickly, they become deprecated because the person who will change a piece of code will forget to change the comment. Instead, put all the information needed in the docstrings of the corresponding functions (this comment could go in pickable_mean for example); and if needed, further break the code into more auxiliary functions (I don't think it is needed here).

The advantage of using docstring to "comment" the code, instead of using the actual comments, is that the explanations show on the documentation website: geomstats.ai. Thanks!

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        def pickable_mean(points, weights):
+            """Frechet Mean of all points weighted by weights"""
+            return self.mean.fit(points, weights=weights).estimate_
+
+        # 'from_points' is the default initialization for centers used
+        if self.init_centers == "from_points":
+            n_points = X.shape[0]
+            centers = X[gs.random.randint(n_points, size=self.n_centers), :]
+        if self.init_centers == "random_uniform":
+            n_centers = centers.shape[0]
```
ninamiolane, line 106: You probably don't need this, you can use `self.n_centers` directly, correct?

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+            # compute Frechet means in parallel
+            pool = joblib.Parallel(n_jobs=self.n_jobs)
+            out = pool(
+                pickable_mean(points_to_average[j], nonzero_weights[j])
+                for j in range(self.n_centers)
+            )
+
+            new_centers = gs.array(out)
+
+            displacements = [self.metric.dist(centers[j], new_centers[j])]
```
ninamiolane, line 141: Should this be inside a for loop that depends on `j`?

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+            centers = new_centers
+
+            if (gs.array(displacements) < self.tol).all():
```
ninamiolane, line 144: Yes, you might not need the list (see comment above) especially if you are re-casting to array then. Try removing the list `[]` on line 141 and the `gs.array` here

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+                break
+
+        self.centers = centers
+
+    def predict(self, points):
+        """Predict the closest cluster each point in 'points' belongs to"""
```
ninamiolane, line 150: Need to follow docstring conventions (see other comments)

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,161 @@
...
+        if self.centers is None:
+            raise Exception("Not fitted")
+
+        out = []
+        for i in range(points.shape[0]):
```
ninamiolane, line 155: You can use `for point in points` works in python and is shorter and easier to read.

geomstats/learning/riemannian_mean_shift.py
ninamiolane: ^This would become `point` with the syntax for point in points

tests/tests_geomstats/test_riemannian_mean_shift.py
```diff
@@ -0,0 +1,40 @@
+""" Unit tests for Riemannian Mean Shift method """
```
ninamiolane, line 1: 🎉 

Missing "." at the end of the module's docstring. 

@SaitejaUtpala do you know why DeepSource is not picking up on these docstrings' conventions?

tests/tests_geomstats/test_riemannian_mean_shift.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
```
ninamiolane, line 3: Remove this line and instead:
`import geomstats.backend as gs`

and on line 16, use:
`gs.random.seed(10)`

tests/tests_geomstats/test_riemannian_mean_shift.py
```diff
@@ -0,0 +1,40 @@
...
+
+import geomstats.tests
+from geomstats.geometry.hypersphere import Hypersphere, HypersphereMetric
+from geomstats.learning.frechet_mean import FrechetMean
+from geomstats.learning.riemannian_mean_shift import RiemannianMeanShift as RMS
+
+
+@geomstats.tests.np_and_autograd_only
```
ninamiolane, line 11: Do you encounter an error when removing this decorator, and if so which one? Thanks!

ninamiolane (changes requested):
Great work, and very clean code, thanks! 🎉  I put a lot of comments, but they are all very minor. Look at the docstrings' conventions still, as many docstrings are missing the parameters description (which we need to generate the documentation website: https://geomstats.github.io/learning.html#module-geomstats.learning.frechet_mean).

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Noted, updated docstring.

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Addessed

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Am confused about this comment. What should I assign to y, since we are trying to fit the centers in "self.centers".

ninamiolane: It is for API compatibility, you can just add `, y=None` after the `X`. It will not be used.

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Agreed.

geomstats/learning/riemannian_mean_shift.py
shubhamtalbar96: Okay noted.

tests/tests_geomstats/test_riemannian_mean_shift.py
```diff
@@ -0,0 +1,40 @@
...
+class TestRiemannianMeanShift(geomstats.tests.TestCase):
+    _multiprocess_can_split_ = True
+
+    def test_hypersphere_riemannian_mean_shift_fit(self):
```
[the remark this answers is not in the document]
shubhamtalbar96: Sure.

tests/tests_geomstats/test_riemannian_mean_shift.py
shubhamtalbar96: No, I just removed the decorator and the test case works fine.

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,177 @@
...
+        self.tol = tol
+        self.mean = FrechetMean(self.metric, **frechet_mean_kwargs)
+        self.n_centers = n_centers
+        self.n_jobs = n_jobs
+        self.max_iter = max_iter
+        self.init_centers = init_centers
+        self.kernel = kernel
+        self.centers = None
+
+    def dist_intersets(self, points_a, points_b, **joblib_kwargs):
```
ninamiolane, line 79: Parameter description is missing in the docstring.

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,177 @@
...
+        pool = joblib.Parallel(n_jobs=self.n_jobs, **joblib_kwargs)
+        out = pool(
+            pickable_dist(point_a, point_b)
+            for point_a in points_a
+            for point_b in points_b
+        )
+
+        return gs.array(out).reshape((n_a, n_b))
+
+    def fit(self, x, y=None):
```
ninamiolane, line 98: Docstring does not respect conventions: add one liner + parameters' description

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,177 @@
...
+        self.kernel = kernel
+        self.centers = None
+
+    def dist_intersets(self, points_a, points_b, **joblib_kwargs):
+        """Parallel computation of distances between two sets of points."""
+        n_a, n_b = points_a.shape[0], points_b.shape[0]
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_dist(x, y):
```
ninamiolane, line 85: docstring

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,177 @@
...
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_mean(points, weights):
+            """Frechet Mean of all points weighted by weights.
+
+            Parameters
+            ----------
+            points : array-like,
+                     Clusters of points.
```
ninamiolane, line 112: Docstring conventions:
- Indentation should be +4 for parameter's description
- Include the shape of the array after the `,` (see examples of docstrings in geomstats)

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,177 @@
...
+        if self.init_centers == "from_points":
+            n_points = x.shape[0]
+            centers = x[gs.random.randint(n_points, size=self.n_centers), :]
+        elif self.init_centers == "random_uniform":
+            centers = self.manifold.random_uniform(n_samples=self.n_centers)
+
+        for _ in range(self.max_iter):
+            dists = self.dist_intersets(centers, x)
+
+            # assuming the use of 'flat' kernel
```
ninamiolane, line 127: Rm comment

ninamiolane (changes requested):
Very nice!! Good job on getting these firsts tests to pass. Something is failing the GitHub Action's tests on tf and torch: I let you have a look at this?

Also, see my comments on docstrings. (they are important so that our doc website can be generated automatically geomstats.ai).

Thanks! 🙌

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,198 @@
...
+        points_b : array-like, shape=[..., n_features]
+            Clusters of points.
+        """
+        n_a, n_b = points_a.shape[0], points_b.shape[0]
+
+        @joblib.delayed
+        @joblib.wrap_non_picklable_objects
+        def pickable_dist(x, y):
+            """
+            Riemannian metric between points x & y.
```
ninamiolane, line 96: This is actually `Riemannian distance` between points x, and y. The distance comes from the Riemannian metric, but they are two different concepts.

Also: the one-liner docstring should come after the """, like:
```
"""Compute Riemannian distances between x and y.

Parameters
------------
...
```

geomstats/learning/riemannian_mean_shift.py
```diff
@@ -0,0 +1,198 @@
...
+        out = pool(
+            pickable_dist(point_a, point_b)
+            for point_a in points_a
+            for point_b in points_b
+        )
+
+        return gs.array(out).reshape((n_a, n_b))
+
+    def fit(self, x, y=None):
+        """
```
ninamiolane, line 117: Follow:
```
"""Fit centers in all the input points.

Parameters
------------
etc...
```

geomstats/learning/riemannian_mean_shift.py
ninamiolane: @shubhamtalbar96 removing the list didn't work?

tests/tests_geomstats/test_riemannian_mean_shift.py
```diff
@@ -0,0 +1,109 @@
...
+import geomstats.backend as gs
+import geomstats.tests
+from geomstats.geometry.hypersphere import Hypersphere, HypersphereMetric
+from geomstats.geometry.special_orthogonal import SpecialOrthogonal
+from geomstats.learning.frechet_mean import FrechetMean
+from geomstats.learning.riemannian_mean_shift import (
+    RiemannianMeanShift as riemannian_mean_shift,
+)
+
+os.environ["GEOMSTATS_BACKEND"] = "pytorch"
```
ninamiolane, line 14: You should remove this line, because this file should be able to run for all of the backends: we don't want a specific backend to get hardcoded here.

ninamiolane (changes requested):
Looking good! A few comments on docstrings (still, haha), and one on the backend. Let me see why the GitHub Action Tests are failing...

ninamiolane:
Let's try removing the `os.environ` and see if more tests can pass.

We might have some tests failure remaining in fiber_bundle.py after this, not exactly sure why this is happening (@SaitejaUtpala @nguigs if you have any idea let me know? It also happened in this PR: #1289 )

nguigs:
> We might have some tests failure remaining in fiber_bundle.py after this, not exactly sure why this is happening (@SaitejaUtpala @nguigs if you have any idea let me know? It also happened in this PR: #1289 )

I don't understand what's triggering the error in fiber_bundle! Maybe a NotImplementedError raised somewhere and included in the exception of the horizontal projection. Was any added recently in master? I cannot reproduce the errors locally

ninamiolane:
@nguigs I really don't know what addition to master could have triggered it. Could it be an update in python/numpy/torch/tensorflow (since we are doing `>= x.x.x` dependencies now)?

geomstats/learning/riemannian_mean_shift.py
ninamiolane: Incorrect indentation, see above

ninamiolane (approved):
Looking good! Please correct the last 3 docstring errors (1 in my comments here, and 2 on the Linting GitHub actions) and then we can merge 🎉 

The errors related to the fiber_bundle are not yours.

ninamiolane:
Merged via #1298 (which was merged into master, but not recorded in the PR pages for some reasons, thus I had to close it too).

PR #1282, opened 2022-01-13
New visualisation module

emaignant:
Notebooks should run fin now. Next step consists in updating the tests using visualisation so they run fine too.


lpereira95:
@ninamiolane, I've updated the changes from @emaignant. In this version, the visualization file was broken into a subpackage without breaking backwards compatibility (tests and notebooks work seamlessly). If you want to merge, it will not mess up with anything your students may be doing. This is just the first step towards a better visualization module.

@emaignant, I reverted the changes in the notebooks, because I think it is premature to change things there (since we are still designing the API - which mean we would have to change everything again). Everything else was kept (I've just cleaned the imports). Thanks for the work put in moving things around (it can be very tricky!)!

ninamiolane (approved):
Excellent! thank you so much! @lpereira95 I let you merge when the linting tests pass.

lpereira95:
@ninamiolane, I'll merge this PR even though deep source is "complaining". In fact, this PR only moves things around (and solves some small issues). The API is kept exactly the same. The code has barely changed. Nevertheless, I think it is important to merge because this new structure is closer to what we've envisioned for the new API (which I'll start already today).

Also, I think it is nice to have this new structure already available in our today's release.

PR #1285, opened 2022-01-15
Adding option to obtain backend module

captain-pool:
With this patch, an user can directly get the module object of the current backend. Should facilitate in writing more backend independent code.

An user can get the module object via `geomstats.backend.module`. The actual backend name can also be obtained as `geomstats.backend.module.__name__`.

CC: @ninamiolane

### DIFF
--- a/geomstats/_backend/__init__.py
+++ b/geomstats/_backend/__init__.py
@@ -86,7 +86,7 @@ BACKEND_ATTRIBUTES = {
         "mean",
         "meshgrid",
         "mod",
-        "module",
+        "module_name",
         "ndim",
         "one_hot",
         "ones",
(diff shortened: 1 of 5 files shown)


ninamiolane (approved):
LGTM! Thanks for this patch @captain-pool 🎉  @SaitejaUtpala what do you think?

SaitejaUtpala:
Hi @captain-pool. Looks good, can you just add simple unit test case?

nguigs:
Hi,
Can you explain what would be the usecase of accessing the module? I am just afraid this will introduce code that is not compatible with all backends, because contributors will use `geomstats.backend.module.<method>` instead of adding `method` to each backend in a compatible way.

captain-pool:
Makes sense. The reason I wanted to add something like this is to provide a read-only method to gather information about the environment being currently used. I think, instead of returning the module, just returning the name of the backend module would be enough. What do you think?

lpereira95:
An easier solution to get backend name is to add `module.__name__ = _BACKEND` in l265 of `geomstats._backend,__init__.py` and then do

```

print(gs.__name__)
```

I'll add this soon.

PR #1301, opened 2022-01-24
Add more conditions to docs.yml

SaitejaUtpala:
try something diff because it's not working


SaitejaUtpala:
Professor @ninamiolane, it will work now

PR #1303, opened 2022-01-27
Removing duplicate centroids from the predicted centers

shubhamtalbar96:
Since multiple centers could converge to the same points in Mean Shift it makes sense to get rid of the duplicate centroids converged.

### DIFF
--- a/geomstats/learning/riemannian_mean_shift.py
+++ b/geomstats/learning/riemannian_mean_shift.py
@@ -174,7 +174,7 @@ class RiemannianMeanShift(ClusterMixin, BaseEstimator):
             if (gs.array(displacements) < self.tol).all():
                 break
 
-        self.centers = centers
+        self.centers = gs.unique(centers, axis=0)
 
     def predict(self, points):
         """Predict the closest cluster each point in 'points' belongs to.


ninamiolane:
Screen Shot 2022-01-27 at 7 49 14 AM

Unfortunately, numpy and torch's APIs seem to differ in their use of `unique`: the parameter `axis` from numpy is called `dim` in pytorch. Thus, we need to addresss this in our backend.

- Remove this line: https://github.com/geomstats/geomstats/blob/0eed9ad02291fa3a87230bbed024ba038b8258bc/geomstats/_backend/pytorch/__init__.py#L51
- Redefine the `unique` function to map `axis` to `dim` a bit like here:
https://github.com/geomstats/geomstats/blob/0eed9ad02291fa3a87230bbed024ba038b8258bc/geomstats/_backend/pytorch/__init__.py#L207


Thanks!

ninamiolane:
What was the decision after Friday's lab meeting's discussion: should we keep this, or rather let the centers as they are?

shubhamtalbar96:
I think we should keep it as it is for now. Since, when we apply a unique method we also need to take care of floating point numbers as @SaitejaUtpala suggested.

ninamiolane:
Sounds good, thank you!

PR #1307, opened 2022-01-28
First slim implementation of Wald space 

JonasLueg:
The main class is WaldSpace in waldspace.py, and this uses the classes defined in trees.py.
Additionally, there is an example and code for vizualisation.


geomstats/geometry/trees.py
```diff
@@ -0,0 +1,469 @@
...
+        empty intersection with ``part2``.
+    part2 :
+        The second part of the split, an iterable that is a subset of
+        :math:`\{0,\dots,n-1\}`. It may be empty for technical reasons, but must have
+        empty intersection with ``part1``.
+    """
+
+    def __init__(self, n, part1, part2):
+        # sort both parts and convert them into tuples
+        part1, part2 = tuple(np.sort(list(part1))), tuple(np.sort(list(part2)))
```
nguigs, line 41: Can you use `gs.sort` instead of np?

nguigs (commented):
Great start!

geomstats/geometry/trees.py
JonasLueg: done.

geomstats/geometry/trees.py
```diff
@@ -0,0 +1,473 @@
...
+    """
+
+    def __init__(self, n, partition, split_sets):
+        """ Initialize a structure with partition and sets of splits of components. """
+        self._n = n
+        # make some assertions about the given parameters.
+        # same number of components in both partition and split_sets
+        assert len(split_sets) == len(partition)
+        # all parts of partition give the whole set of labels.
+        assert set.union(*[set(part) for part in partition]) == set(range(n))
```
nguigs, line 187: We usually raise value errors instead of using assert statement.

nguigs (commented):
A few comments on style

geomstats/geometry/trees.py
```diff
@@ -0,0 +1,473 @@
...
+        return self.part1 if set(self.part1) & set(other.part1) and set(
+            self.part1) & set(other.part2) else self.part2
+
+    def point_away_split(self, other):
+        """ Gives back the part of this split that is directed away from other split."""
+        return self.part2 if set(self.part1) & set(other.part1) and set(
+            self.part1) & set(other.part2) else self.part1
+
+    def compatible_with(self, other):
+        """ Checks whether this split is compatible with another split. """
```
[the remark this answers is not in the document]
JonasLueg: Two splits are compatible if at least one intersection is empty, I've added it into the description.

geomstats/geometry/trees.py
JonasLueg: I've replaced the assert statements with if-clauses and ValueErrors

geomstats/geometry/trees.py
```diff
@@ -0,0 +1,473 @@
...
+
+    The parameters ``part1`` and ``part2`` are assigned to the attributes ``self.part1``
+    and ``self.part2`` sorted, and the one that contains the smallest minimal value is
+    assigned to ``self.part1``, to obtain consistency.
+
+    Parameters
+    ----------
+    n : int
+        The number of labels, the set of labels will be :math:`\{0,\dots,n-1\}`.
+    part1 :
```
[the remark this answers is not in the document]
JonasLueg: I've added more type infos, what should I do if a method is returned? I've put 'callable' for now as a placeholder. I could do something like abc.Callable.
Also, what if I have a list of list of int, should I do list[list[int]]? I've put list of list of int for now.

PR #1315, opened 2022-02-03
cleaning product manifold

SaitejaUtpala:


tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
+            ]
+            return self.generate_tests(smoke_data)
+
+        def inner_product_data(self):
+            smoke_data = []
+            if not tf_backend():
+                euclidean = Euclidean(3)
+                minkowski = Minkowski(3)
+                space = ProductManifold(
+                    manifolds=[euclidean, minkowski], default_point_type="matrix"
```
ninamiolane, line 84: - There was a test with default_point_point="vector" for this one too: where did it go?

tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
+            n_samples = [1, 5, 10]
+            smoke_data = [
+                dict(manifold=manifold, n_samples=n_samples)
+                for (manifold, n_samples) in itertools.product(
+                    [manifolds[0]], n_samples
+                )
+            ]
+            return self.generate_tests(smoke_data)
+
+        def inner_product_data(self):
```
ninamiolane, line 78: This should be `inner_product_matrix_data`: we are checking that the inner product matrix is correct, e.g. it is gs.eye(6).

tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
-        result = self.space_vector.regularize(expected)
-        self.assertAllClose(result, expected)
-
-    def test_regularize_matrix(self):
-        expected = self.space_matrix.random_point(5)
-        result = self.space_matrix.regularize(expected)
-        self.assertAllClose(result, expected)
-
-    @geomstats.tests.np_autograd_and_torch_only
-    def test_inner_product_matrix(self):
```
ninamiolane, line 140: Where did this go? maybe we can rename it `test_log_inner_product_squared_dist`?

tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
+                expected = gs.eye(6)
+                expected[3, 3] = -1
+                smoke_data = [dict(manifold=space, point=point, expected=expected)]
+            return self.generate_tests(smoke_data)
+
+        def to_tangent_is_tangent_data(self):
+            smoke_data = [dict(manifold=space_matrix), dict(manifold=space_vector)]
+            return self.generate_tests(smoke_data)
+
+        def projection_and_belongs_vector_data(self):
```
ninamiolane, line 96: This should just be called: `projection_and_belongs`:
- the `_vector` was coming from the fact that the function was only tested on the default point type vector,
- but now, we are testing it on both.

tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
-        result = helper.test_projection_and_belongs(
-            self.product, shape=(3, 2, 3, 3), atol=1e-4
+    def test_projection_and_belongs_vector(self, manifold, shape):
+        result = helper.test_projection_and_belongs(manifold, shape, atol=gs.atol * 100)
+        self.assertAllClose(gs.all(result), gs.array(True))
+
+
+class TestNFoldManifold(TestCase, metaclass=Parametrizer):
+    class TestDataNFoldManifold(TestData):
+        def random_and_belongs_data(self):
```
ninamiolane, line 163: The case where the data does not belong to the NFold manifold is missing.

ninamiolane (changes requested):
Thanks! I put some comments. Any reason why you are not using the RiemannianMetricParametrizer?

tests/tests_geomstats/test_product_manifold.py
```diff
@@ -9,277 +11,286 @@
...
+                dict(manifold=manifold, n_samples=n_samples)
+                for (manifold, n_samples) in itertools.product(
+                    [manifolds[0]], n_samples
+                )
+            ]
+            return self.generate_tests(smoke_data)
+
+        def inner_product_data(self):
+            smoke_data = []
+            if not tf_backend():
```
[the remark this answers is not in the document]
SaitejaUtpala: test part already has decorator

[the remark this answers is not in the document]
ninamiolane: why do we need a decorator inside the `_data` function then?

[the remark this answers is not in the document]
SaitejaUtpala: because, data function is called regardless

PR #1335, opened 2022-02-17
refactor skew symmetric, PLT, LT, PSD(k), Full Rank Matrices, Matrices.

SaitejaUtpala:
refactor skew symmetric, PLT, LT, PSD(k), Full Rank Matrices, Matrices.


tests/data_generation.py
```diff
@@ -13,159 +12,6 @@ def better_squeeze(array):
     return array
 
 
-class TemporaryTestData:
```
ninamiolane, line 16: We do not need this anymore, no backward compatibility?

tests/data_generation.py
```diff
@@ -457,8 +303,8 @@ def _basis_cardinality_data(self, space_args_list):
         return self.generate_tests([], random_data)
 
 
-class LieAlgebraTestData(VectorSpaceTestData):
-    def basis_representation_matrix_representation_composition_data(
+class MatrixLieAlgebraTestData(VectorSpaceTestData):
```
ninamiolane, line 306: Good catch! 🎉

tests/data_generation.py
```diff
@@ -743,36 +589,31 @@ def _log_exp_composition_data(
         self,
         connection_args_list,
         space_list,
-        shape_list,
         n_samples_list,
         rtol=gs.rtol,
         atol=gs.atol,
     ):
-        """Generate data to check that exponential and logarithm are inverse.
+        """Generate data to check that logarithm and exponential are inverse.
```
ninamiolane, line 596: Good catch also!

ninamiolane (approved):
Fantastic, thank you! (And thanks for correcting my mistakes in the docstrings too 🙌 )

PR #1345, opened 2022-02-21
add Matrix basis and General linear algebra

### LINKED ISSUE #1336: Matrices is missing basis
Matrices is missing basis

nguigs:
The `basis` is turned into a property of all vector spaces. It is only created and stored if needed, to avoid memory overload.

closes #1336

### DIFF
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -3,6 +3,7 @@
 import geomstats.algebra_utils as utils
 import geomstats.backend as gs
 from geomstats.geometry.base import OpenSet
+from geomstats.geometry.lie_algebra import MatrixLieAlgebra
 from geomstats.geometry.lie_group import MatrixLieGroup
 from geomstats.geometry.matrices import Matrices
 
@@ -27,7 +28,10 @@ class GeneralLinear(MatrixLieGroup, OpenSet):
         if "dim" not in kwargs.keys():
             kwargs["dim"] = n**2
         super(GeneralLinear, self).__init__(
-            ambient_space=Matrices(n, n), n=n, lie_algebra=Matrices(n, n), **kwargs
+            ambient_space=Matrices(n, n),
+            n=n,
+            lie_algebra=GeneralLinearLieAlgebra(n),
+            **kwargs
         )
         self.positive_det = positive_det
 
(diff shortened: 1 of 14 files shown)


SaitejaUtpala:
@nguigs. Can you add skip_test_basis_belongs, skip_test_basis_cardinality that I have to set to True

tests/tests_geomstats/test_matrices.py
```diff
@@ -630,6 +634,17 @@ def test_to_matrix_type_is_matrix_type(self, m, n, matrix_type, mat):
...
+    def test_basis(self, m, n):
+        expected = gs.array(
+            [
+                gs.array_from_sparse([(i, j)], [1], (m, n))
+                for i in range(m)
+                for j in range(n)
+            ]
+        )
+        result = Matrices(m, n).basis
+        self.assertAllClose(result, expected)
```
SaitejaUtpala, line 646: expected should be parameter. so test_basis(self, m, n, expected)

geomstats/geometry/general_linear.py
```diff
@@ -159,3 +165,57 @@ def path(time):
             return cls.exp(vecs, base_point)
 
         return path
+
+
+class GeneralLinearLieAlgebra(MatrixLieAlgebra):
```
ninamiolane, line 170: This is the Lie algebra of the general linear group, thus the space of matrices. 
Why isn't `Matrices(MatrixLieAlgebra)` enough, i.e. why not adding the functions into `Matrices` directly?

ninamiolane: We usually do not name classes in terms of whose Lie algebra they are, but in terms of what they are, thus GeneralLinearLieAlgebra feels a bit weird there?

ninamiolane (commented):
Some questions about GeneralLinearLieAlgebra

geomstats/geometry/general_linear.py
ninamiolane: I see... But then:
- we have two spaces that are the same but have different names... (Matrices and GeneralLinearAlgebra)
- the fact that a space is the linear algebra of another is usually contained in the attribute: space.lie_algebra, not in the name. 
  - I did not see SpecialLinearLieAlgebra yet: is it in @lpereira95 's PR? I would argue against naming the manifolds this way.
  - Otherwise, we would have two logics to name Lie algebras: either by the manifold they actually represent (skewsym matrices for example) or by the space of whom they are a Lie algebra. Which can become confusing.

Is there another way we could make this work with the inheritances? :/ How about SquareMatrices(MatrixLieAlgebra)?

ninamiolane (approved):
Excellent, thank you very much for this it will improve performances indeed 🙌

PR #1358, opened 2022-02-24
Make VectorSpace basis property more consistent

lpereira95:
It will initialize `_basis`, instead of `basis` in VectorSpace instantiation.


nguigs (approved):
Great thanks!

PR #1360, opened 2022-02-24
Solve mutability issue when projecting point

lpereira95:
Avoid returning as projected point the input object (to avoid mutability issues, which are very often tricky to spot).

(I've quickly looked to all the projection functions and have only found this. Probably there's more behavior like this)


ninamiolane (commented):
Thanks @lpereira95 ! I've tried to re-run the tests many times (in case the tests were failing just by "chance"), but each time it's tensorflow failing. Do you know what is going on?

lpereira95:
I don't know what's going on, but I'll try to identify it.

ninamiolane:
@lpereira95 any update on this?

lpereira95:
This will be solved in a new PR.

PR #1361, opened 2022-02-24
speedup parallel transport on SPD manifold

sylvchev:
Trying to use Bures-Wasserstein metric, we realize that is possible to avoid inverting a matrix in the parallel transport for Affine Invariant metric by using a rewriting of the squareroot of a product: sqrtm(AB^-1) being rewrite as B^1/2(B^-1/2 A B^-1/2)B^-1/2 instead.

Tests on random points shows a speedup of 1.3x for small dimension (5) up to 3x for larger dimension (100-150) with the different backend. The code results in the same results, up to the machine precision.


ninamiolane (approved):
Great, thanks!

PR #1362, opened 2022-02-24
Add the Bures-Wasserstein metric for fixed rank PSD matrices

nguigs:
According to the [paper](https://epubs.siam.org/doi/10.1137/18M1231389) by Massart and Absil.
The metric is obtained as the quotient metric of the Frobenius metric on the space of full rank matrices by the orthogonal group.

@annacalissano looking forward to your tests :)


sylvchev:
Thanks for the PR!

nguigs:
@SaitejaUtpala I started a new FiberBundle parametrizer here.

nguigs:
@SaitejaUtpala both are used to read function composition ([see wiki](https://en.wikipedia.org/wiki/Function_composition)), and I think g after f is less confusing because it corresponds to the order of g \circ f unlike f then g.

ninamiolane:
Thanks for this PR!

I don't have a strong preference for `then` versus `after`. 

However, the current implementation of the tests is using `then`. If you want to change to `after`, then it should be changed everywhere in the tests.

ninamiolane (approved):
Great, thanks!

PR #1364, opened 2022-02-24
Refactor Special Orthogonal Group

SaitejaUtpala:


SaitejaUtpala:
@ninamiolane  kindly ignore codecov floating errors

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
+                            point=elements[angle_type],
+                            base_point=angle_type_base,
+                        )
+                    ]
+            return self.generate_tests(smoke_data)
+
+        def group_exp_then_log_with_angles_close_to_pi_data(self):
+            return self.group_log_then_exp_with_angles_close_to_pi_data()
+
+        def left_jacobian_vectorization_data(self):
```
nguigs, line 809: no data generated here

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
+        result = group.compose(group.identity, point)
+        expected = group.regularize(point)
+        inv_expected = -expected
+        self.assertTrue(
+            gs.allclose(result, expected) or gs.allclose(result, inv_expected)
         )
-        result = group.matrix_from_rotation_vector(rot_vec_3)
+
+    @geomstats.tests.np_autograd_and_tf_only
+    def test_regularize_extreme_cases(self, elements_all):
```
nguigs, line 1046: Shouldn't the loop for different inputs be different elements of smoke_data, instead of having assertAllClose inside the loop?

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
+    @geomstats.tests.np_autograd_and_tf_only
+    def test_distance_broadcast(self, n):
+        group = SpecialOrthogonal(n=n)
+        point = group.random_point(5)
+        result = group.bi_invariant_metric.dist_broadcast(point[:3], point)
+        expected = []
+        for a in point[:3]:
+            expected.append(group.bi_invariant_metric.dist(a, point))
+        expected = gs.stack(expected)
+        self.assertAllClose(result, expected)
```
nguigs, line 1374: shouldn't expected be passed as argument?

nguigs (commented):
This is huge work, thanks. A few comments, mainly:
- we should rename tests that test the composition of two functions to make the order more clear (f after g)
- Some tests seem not to have been refactored, i.e. the data is generated in the test instead of the data method

SaitejaUtpala:
@nguigs data not being refactored is primarily because of two reasons

- mostly tests where tests and data is not split is because these tests shapes. and for shapes we don't need to pass that tangent_vec etc.
- I was planning to remove  these things because I thought of adding shape checker.  @ninamiolane  made a great point that it will decrease the codecov and protection of codebase. so I added as is (without decoupling data and test)

tests/tests_geomstats/test_special_orthogonal.py
SaitejaUtpala: I don't think so. This is not some testing function in isolation. It is testing whether function output on vectorized inputs is stack of output of function of single-batch inputs. So no need for expected

tests/tests_geomstats/test_special_orthogonal.py
SaitejaUtpala: yup not needed. it's checking for shape

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
-    def test_to_tangent(self):
-        theta = 1.0
-        vec_1 = gs.array([[0.0, -theta], [theta, 0.0]])
-        result = self.group.to_tangent(vec_1)
-        expected = vec_1
+    def test_compose_shape(self, n, point_type, n_samples):
+        group = self.space(n, point_type=point_type)
+        n_points_a = group.random_uniform(n_samples=n_samples)
+        n_points_b = group.random_uniform(n_samples=n_samples)
+        one_point = group.random_uniform(n_samples=1)
```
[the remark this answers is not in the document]
SaitejaUtpala: yup not needed. it's checking for shape

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
+
+        def exp_log_composition_data(self):
+            return self._exp_log_composition_data(
+                SpecialOrthogonal,
+                self.space_args_list,
+                self.shape_list,
+                self.n_samples_list,
+                amplitude=100.0,
+                rtol=gs.rtol * 10000,
+                atol=gs.atol * 100000,
```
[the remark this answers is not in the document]
SaitejaUtpala: I didn't fine tune. It failed once for 10. So I increased to 100 amplitude

tests/tests_geomstats/test_special_orthogonal.py
```diff
@@ -1,289 +1,1414 @@
...
+            ]
+            return self.generate_tests(smoke_data)
+
+        def distance_broadcast_shape_data(self):
+            n_list = [2, 3]
+            n_samples_list = random.sample(range(1, 20), 2)
+            smoke_data = [
+                dict(n=n, n_samples=n_samples)
+                for n, n_samples in zip(n_list, n_samples_list)
+            ]
```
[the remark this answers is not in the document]
SaitejaUtpala: yup not needed. it's checking for shape

PR #1365, opened 2022-02-25
Add special linear

lpereira95:
Tests follow @SaitejaUtpala design. There's a lot of repetition between the tests of `SpecialLinear` and `GeneralLinear` that probably should be removed in the future.

### DIFF
--- a/geomstats/geometry/special_linear.py
+++ b/geomstats/geometry/special_linear.py
@@ -1,0 +1,227 @@
+"""Class for the group of special linear matrices."""
+
+import geomstats.algebra_utils as utils
+import geomstats.backend as gs
+from geomstats.geometry.invariant_metric import InvariantMetric
+from geomstats.geometry.lie_group import MatrixLieGroup
+from geomstats.geometry.lie_algebra import MatrixLieAlgebra
+from geomstats.geometry.matrices import Matrices
+
+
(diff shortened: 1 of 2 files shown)


ninamiolane:
Nice! 🙌  thanks for re-opening this PR with only the SL files. A few comments:
- Could you make sure that the tests pass before moving the PR from "draft" to "ready for review"?
- if some tests are failing because of the randomized tests, could you let @SaitejaUtpala know?
- To avoid repeated code (coming from GL), we could use the LevelSet structure, as here:
https://github.com/geomstats/geomstats/blob/02fa5e24cad78611353f3b4e87d3f7cba6b339bc/geomstats/geometry/special_orthogonal.py#L31
basically encoding the fact that SL is a subgroup within GL.

I let @nguigs advice if he has a better idea.

lpereira95:
Thanks for the feedback @ninamiolane!

In this case the tests are failing due to incompatibilities with tensorflow (I don't have it installed locally, which makes it a little hard to debug before the pull request - I'm using a mac M1 right now, which makes almost everything very trick to install).

I'll address the LevelStructure design soon (I completely agree with avoiding any kind of code duplication).

ninamiolane:
@lpereira95 any update on this?

PR #1366, opened 2022-02-25
Reformat test_dirichlet

alebrigant:


SaitejaUtpala:
@alebrigant . Nice refactoring. let me know if there is any help in the refactoring

alebrigant:
Thanks @SaitejaUtpala, I'll let you know if I need any help ! I will find some time this week to finish this.

alebrigant:
@SaitejaUtpala I do have one question: how does one skip a test with certain backends ? Before, this was possible using decorators, e.g. `@geomstats.tests.np_and_autograd_only`.

SaitejaUtpala:
@alebrigant  if it is for test function that is residing in the class @geomstats.tests.np_and_autograd_only still works. But if it's one of implicit test functions that gets added automatically though metaclass. currently we can't do that, I will be adding this as part of 2nd iteration.

alebrigant:
@ninamiolane @SaitejaUtpala I don't know how to deal with the random tests. As I see it, there are 2 options: 
Option 1: generate the random data in the data generator. For example
```
        def sample_belongs_data(self):
            random_data = [
                dict(
                    dim=2,
                    point=self.space(2).random_point(3),
                    n_samples=4,
                    expected=gs.ones((3, 4)),
                ),
                dict(
                    dim=3,
                    point=self.space(3).random_point(1),
                    n_samples=2,
                    expected=gs.ones(2),
                ),
                dict(
                    dim=4,
                    point=self.space(4).random_point(2),
                    n_samples=3,
                    expected=gs.ones((2, 3)),
                ),
            ]
```
Option 2: only put some parameters such as `n_points` and `dim` in the data generator and generate some random points in the test. For example:
```
        def sample_belongs_data(self):
            random_data = [
                dict(dim=n, n_points=n_points, n_samples=n_samples)
                for n in self.n_list
                for n_points in self.n_points_list
                for n_samples in self.n_samples_list
            ]
```
The second option is much more concise and practical I would say, but then the data defined in `random_data` are not exactly random, they are deterministic parameters used to generate random data in the tests. What do you think ?

ninamiolane:
Thanks for your feedback! 

I will let @SaitejaUtpala answer, but I think one advantage of generating the data in the *_data function allows us to take advantage of the hierarchy of TestData classes: there might be similar ways of generating random data for the TestConnectionData and TestRiemannianMetricData classes.

But I have no strong preference between the two designs you suggest.

@nguigs what do you think?

SaitejaUtpala:
@alebrigant @ninamiolane. great question! sorry for the delay. I thought about this a bit in start of test refactoring. Answer  it depends on the function being tested. Let me explain.  This is protocol I came up with (may not be ideal) . 
- If the test function is about individual function being tested. almost all the inputs to function should be passed as data to the test_function. Ex: for `exp` tangent_vec and base_point should generated in data function. for `random_point` n_samples should be generated in data function. Now why we should ideally generate all inputs.  because later we can log the inputs for which functions is not working as expected. @ninamiolane has this interesting idea of running diagnostics on data points on which functions that give erroneous outputs. Ex. `exp/log` not working properly

- Now, Based on how inputs  to function are being generated we classify data (inputs to test_function) into smoke_data (which is deterministic data) and random_data (which is randomized data). Now we can follow other classification of data based on whether  output of functions is deterministic, this seem more preferrable and gives justification to names deterministic and randomized data, because classification is supposed to make tests reproducible. But here is the problem with this classification, the function being tested could be inherently random , i.e., for same input it could give different output (so no more reproducibility, some hacks would be fixing seed, but seed is not explicit input to it). ex `random_point` . Input to this function `n_sample`. even if input is deterministic (n_samples are fixed) output is not.  That is the reason I thought first one is better. 

takeway is following 

- input to function being tested should be generated in data_functions.
- based on how inputs are being generated, it should  either go into smoke_data and randomized_data

alebrigant:
Thanks @ninamiolane and @SaitejaUtpala for your answers. These guidelines are fine with me, and I think the option of generating the random data in the *_data functions makes things easier to read.

I have finished the reformatting. Some comments:
- Among the test functions that get added automatically through the metaclass, many are skipped because the functions that are tested cannot be used with all the backends. 
- `test_exp_shape` is skipped because as I understand it, it tests several base points for one initial tangent vector, which is a configuration that is excluded in `DirichletMetric`. I thought it was common to all the Riemannian metrics, but maybe it isn't ? I'm not sure it makes much sense because the one vector given as argument would actually represent the coordinates of different vectors in different tangent spaces.
- The tests that are failing seem to come from other files.

alebrigant:
Do you have a solution that skips a test in both pytorch and tf backends ? Unfortunately this is what I need in most cases.

SaitejaUtpala:
> @SaitejaUtpala do you have a solution that skips a test in both pytorch and tf backends ? Unfortunately this is what I need in most cases.

@alebrigant . Yup. `tf_backend() or pytorch_backend()` :)

SaitejaUtpala:
@alebrigant @ninamiolane .  If possible can you give us some feedback about overall design,since you have actually used design.  Here are questions

- How do you find new design. things that you liked and didn' like and would like
- How long did you take to refactor this file( roughly in terms of hours you allocated )?

alebrigant:
I really like the new design, it is definitely an improvement over the last one. I like the fact that there are some standard tests that are automatically added from the metaclass, and once you understand how everything works, it should speed things up. It's hard to say how much time I spent re-writing this file, maybe 7-10 hours in total ? It took me a while to understand the structure. Here are the critics I have:
- I find that the docstrings are not detailed enough in `data_generation.y` and I had to look at the detail of the code to understand what was done. For example, in `_exp_shape_data`, there is an argument `n_samples_list` = list of number if random data to generate. But it's not said if it's the number of base points, or the number of tangent vectors. Which are the options that are tested ? One base point and several tangent vectors ? Several base points and several tangent vectors ? Also, the argument `shape_list` = list of shapes for random data to generate could be clarified: it does not depend on a number of points, it corresponds to the shape of a given point in the manifold. So I think it would be great to have more details in the docstrings in general, especially to help new users.
- In the standard tests that come from the metaclass, I didn't find a way to pass certain arguments to the functions that are tested, that are specific to the manifold that is tested. For example, the exp and log functions of the `DirichetMetric` take an argument `n_steps` (there is no closed form and the values are computed using iterative schemes) and I need this argument to be high for the `exp_log_composition_test` to pass. Is there a way to do that apart from rewriting the test ?

This is all I can think of right now, but if I think of something else I'll get back to you.

SaitejaUtpala:
@alebrigant @ninamiolane Thanks for feedback. I completely agree with both of the things!.  yeah we didn't completely finalize the design. Plus I didn't know about any other manifolds except SPD matrices, Positive Lower Triangular Matrices, Lower triangular matrices before starting this refactoring. I have incorporated things back into design that while refactoring each file. also, we saved some things for iteration 2.0 because if we change any thing, we have to change lot of other test files. I have noted it both of the things you said. please let me know anything that would like to have

alebrigant:
@ninamiolane @SaitejaUtpala Don't know what to do about the last errors... Do you see other modifications to be done ?

SaitejaUtpala:
@alebrigant yup. it's version mismatch error. can you pull latest from upstream.

SaitejaUtpala:
@alebrigant  Hi!. you can wait for a week after which design will be more stable. I will be working on making it easier to use and also things you have mentioned about passing n_steps to log_exp.

alebrigant:
> @alebrigant Hi!. you can wait for a week after which design will be more stable. I will be working on making it easier to use and also things you have mentioned about passing n_steps to log_exp.

Hi @SaitejaUtpala, I am coming back to this PR. I saw that you made some refactoring

tests/tests_geomstats/test_dirichlet.py
```diff
@@ -1,432 +1,647 @@
...
+from tests.conftest import Parametrizer
+from tests.data_generation import _OpenSetTestData, _RiemannianMetricTestData
+from tests.geometry_test_cases import OpenSetTestCase, RiemannianMetricTestCase
+
+
+class TestDirichlet(OpenSetTestCase, metaclass=Parametrizer):
+    space = DirichletDistributions
+
+    class TestDataDirichlet(_OpenSetTestData):
+        space = DirichletDistributions
```
SaitejaUtpala, line 23: TestDataDirichlet -> DirichletTestData

tests/tests_geomstats/test_dirichlet.py
```diff
@@ -1,432 +1,647 @@
...
+        pdf = self.space(dim).point_to_pdf(point)
+        alpha = gs.ones(dim)
+        samples = self.space(dim).sample(alpha, n_samples)
         result = pdf(samples)
-        expected = [dirichlet.pdf(x, point) for x in samples]
+        pdf = []
+        for i in range(n_points):
+            pdf.append(gs.array([dirichlet.pdf(x, point[i, :]) for x in samples]))
+        expected = gs.squeeze(gs.stack(pdf, axis=0))
         self.assertAllClose(result, expected)
```
SaitejaUtpala, line 165: generate `expected` in the corresponding test_data function

tests/tests_geomstats/test_dirichlet.py
```diff
@@ -1,432 +1,647 @@
...
+    skip_test_exp_then_log = True
+    skip_test_parallel_transport_ivp_is_isometry = True
+    skip_test_parallel_transport_bvp_is_isometry = True
+    skip_test_geodesic_ivp_belongs = True
+    skip_test_geodesic_bvp_belongs = True
+    skip_test_exp_geodesic_ivp = True
+    skip_test_exp_ladder_parallel_transport = True
+
+    class TestDataDirichletMetric(_RiemannianMetricTestData):
+        space = DirichletDistributions
```
SaitejaUtpala, line 194: TestDataDirichletMetric - > DirichletMetricTestData

tests/tests_geomstats/test_dirichlet.py
```diff
@@ -1,432 +1,647 @@
...
         vector = gs.stack(
-            [polygamma_a - polygamma_ab, -polygamma_ab, polygamma_b - polygamma_ab],
+            [
+                gs.polygamma(1, param_a) - gs.polygamma(1, param_a + param_b),
+                -gs.polygamma(1, param_a + param_b),
+                gs.polygamma(1, param_b) - gs.polygamma(1, param_a + param_b),
+            ],
             axis=-1,
         )
         expected = SymmetricMatrices.from_vector(vector)
```
SaitejaUtpala, line 534: move expected to corresponding _test_data function

SaitejaUtpala (commented):
Looks great! I have made some minor comments.

SaitejaUtpala (approved):

PR #1372, opened 2022-02-26
refactor landmarks

SaitejaUtpala:


tests/parametrizers.py
```diff
@@ -475,6 +475,8 @@ def test_log_shape(self, connection_args, point, base_point, expected):
             connection = self.connection(*connection_args)
             log = connection.log(gs.array(point), gs.array(base_point))
             result = gs.shape(log)
+            print("expected", expected)
```
ninamiolane, line 478: Remove left over prints

tests/tests_geomstats/test_landmarks.py
```diff
@@ -1,164 +1,214 @@
...
-        )
-        log_ab = tangent_vec
-        result = self.l2_metric_s2.squared_norm(
-            vector=log_ab, base_point=self.landmarks_a
-        )
-        expected = self.l2_metric_s2.dist(self.landmarks_a, self.landmarks_b) ** 2
-
-        self.assertAllClose(result, expected)
-
-    def test_l2_metric_log_and_exp(self):
```
ninamiolane, line 74: This test was passing, but the new test is skipped with:
`skip_test_exp_log_composition = True`
 
Please add this test back. (We do not want decrease in protection of the codebase).

tests/tests_geomstats/test_landmarks.py
```diff
@@ -1,164 +1,214 @@
...
-        landmarks_ab = self.l2_metric_s2.geodesic(self.landmarks_a, self.landmarks_b)
-        landmarks_bc = self.l2_metric_s2.geodesic(self.landmarks_b, self.landmarks_c)
-        landmarks_ab = landmarks_ab(self.times)
-        landmarks_bc = landmarks_bc(self.times)
-
-        result = self.l2_metric_s2.dist(landmarks_ab, landmarks_bc)
-        self.assertAllClose(gs.shape(result), (n_samples,))
-
-    @geomstats.tests.np_autograd_and_tf_only
-    def test_l2_metric_exp_vectorization(self):
```
ninamiolane, line 118: This test had an exp whose shape was correct, but exp_shape is skipped with
```
skip_test_exp_shape = True
```

Please add this test back, since we do not want to decrease protection.

tests/tests_geomstats/test_landmarks.py
```diff
@@ -1,164 +1,214 @@
...
-            point=landmarks_bc, base_point=landmarks_ab
-        )
-
-        result = self.l2_metric_s2.exp(
-            tangent_vec=tangent_vecs, base_point=landmarks_ab
-        )
-        self.assertAllClose(gs.shape(result), gs.shape(landmarks_ab))
-
-    @geomstats.tests.np_autograd_and_tf_only
-    def test_l2_metric_log_vectorization(self):
```
ninamiolane, line 135: Ditto: we want to keep testing the shape of log, since it was passing here

tests/tests_geomstats/test_landmarks.py
```diff
@@ -1,164 +1,214 @@
...
-
-        tangent_vecs = self.l2_metric_s2.log(
-            point=landmarks_bc, base_point=landmarks_ab
-        )
-
-        result = tangent_vecs
-        self.assertAllClose(gs.shape(result), gs.shape(landmarks_ab))
-
-    @geomstats.tests.np_autograd_and_tf_only
-    def test_l2_metric_geodesic(self):
```
ninamiolane, line 150: Please add this test.

ninamiolane (changes requested):
Nice, thanks! Some tests should stay, as the new tests in landmarks make the codebase loose some protectino.

tests/tests_geomstats/test_landmarks.py
SaitejaUtpala: @ninamiolane  this is not testing anything new. It is basically testing if output on points is stack of output of points.

SaitejaUtpala: @ninamiolane  But sure I will add it. You are correct and I was wrong,  it will decrease protection of Codebase

tests/tests_geomstats/test_landmarks.py
SaitejaUtpala: #1386 #1385 is the reason for this error. unfortunately I cannot skip this. instead I will add the test back.

tests/tests_geomstats/test_landmarks.py
SaitejaUtpala: @ninamiolane  https://github.com/geomstats/geomstats/issues/1386 https://github.com/geomstats/geomstats/issues/1385 is the reason for this error. unfortunately I cannot skip this. instead I will add the test back. apologies

ninamiolane:
This PR does not pass the tests and should not be marked for review.

SaitejaUtpala:
> This PR does not pass the tests and should not be marked for review.

Sorry, Originally tests were passing. then pushed something and precision errors were there.

PR #1374, opened 2022-02-27
refactor invariantt metric

SaitejaUtpala:


ninamiolane:
This PR does not pass the deep source test. Add a comment on why this can be ignored in this PR, or pass the tests before requesting a review.

SaitejaUtpala:
> This PR does not pass the deep source test. Add a comment on why this can be ignored in this PR, or pass the tests before requesting a review.

@ninamiolane  sorry. It's self issue that we cannot do anything right now.

ninamiolane (changes requested):
Where is the codecoverage decrease coming from?

I also see many tests skipped: @nguigs what do you think, should these tests be skipped?

SaitejaUtpala:
@ninamiolane yup that is reason it is marked as draft. some tests were failing. I wanted to see if removing that would make difference.

SaitejaUtpala:
@ninamiolane ignore deep source and -0.05%, are skipped due to some constant failing. Thanks

tests/tests_geomstats/test_invariant_metric.py
```diff
@@ -1,755 +1,771 @@
...
+from tests.geometry_test_cases import RiemannianMetricTestCase
+
+
+class TestInvariantMetric(RiemannianMetricTestCase, metaclass=Parametrizer):
+    metric = connection = InvariantMetric
+    skip_test_parallel_transport_ivp_is_isometry = True
+    skip_test_parallel_transport_bvp_is_isometry = True
+    skip_test_exp_geodesic_ivp = True
+    skip_test_exp_shape = np_backend()
+    # skip_test_exp_belongs = True
```
ninamiolane, line 25: leftover comment, which was also caught by deepsource.

ninamiolane (approved):
I will merge, but can you remove the leftover comment in the next PR? Thanks for this huge work!

PR #1388, opened 2022-03-03
Add injectivity radius of all metrics

nguigs:
After discussions with @SaitejaUtpala and @ythanwerdas, we add a method to return the injectivity radius at a given base point.

### DIFF
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -664,8 +664,9 @@ class Connection(ABC):
     def injectivity_radius(self, base_point):
         """Compute the radius of the injectivity domain.
 
-        This is the largest radius r for which the exponential map is a
-        injective on the open ball of radius r and centered at the base point.
+        This is is the supremum of radii r for which the exponential map is a
+        diffeomorphism from the open ball of radius r centered at the base point onto
+        its image.
 
         Parameters
         ----------
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -535,7 +535,7 @@ class Connection(ABC):
 
         References
         ----------
-        .. [Pennec] Pennec, xavier. Curvature effects on the empirical mean in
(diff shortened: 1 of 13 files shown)


geomstats/geometry/connection.py
```diff
@@ -660,3 +660,21 @@ def parallel_transport(
             "The closed-form solution of parallel transport is not known, "
             "use the ladder_parallel_transport instead."
         )
+
+    def injectivity_radius(self, base_point):
+        """Compute the radius of the injectivity domain.
+
+        This is the largest radius r for which the exponential map is a
+        injective on the open ball of radius r and centered at the base point.
+
```
ythanwerdas, line 669: I think the original definition of the injectivity radius is the supremum of radiuses r for which the exponential map is a diffeomorphism from the open ball of radius r centered at the base point onto its image. I don't know if it is equivalent to "largest radius for which the exponential map is injective".

geomstats/geometry/spd_matrices.py
```diff
@@ -861,6 +879,24 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+        base_point : array-like, shape=[..., n, n]
+            Point on the manifold.
+
+        Returns
+        -------
+        radius : float
+            Injectivity radius.
+        """
+        eigen_values = gs.linalg.eigvalsh(base_point)
+        return eigen_values[..., 0]
```
ythanwerdas, line 898: Sorry, I probably told you something wrong: it is the square-root of the smallest eigenvalue. (It is the smallest eigenvalue for the Euclidean metric.)

ythanwerdas (commented):
Great improvement Nicolas, thank you! Do you plan to adapt the tests log(exp(v))=v by restricting v to the ball of injectivity radius as we previously discussed?

ninamiolane (approved):
Great, thanks!

PR #1403, opened 2022-03-10
Master precision fixes

SaitejaUtpala:


ninamiolane (changes requested):
We need to have geomstats working in Python 3.8 and 3.9, less and less people are using Python 3.7 now. What is the issue with more recent versions of Python?

@nguigs many tests are now skipped in quotient_metric and pre_shape: did you have problems with these tests before?

SaitejaUtpala:
@ninamiolane  yup. 

> We need to have geomstats working in Python 3.8 and 3.9, less and less people are using Python 3.7 now. What is the issue with more recent versions of Python?
> 
> @nguigs many tests are now skipped in quotient_metric and pre_shape: did you have problems with these tests before?

@ninamiolane 
there is this wierd error that kept popping up (you can look here https://github.com/geomstats/geomstats/runs/5500215908?check_suite_focus=true on master). pytorch couldn't find `complex32` data type. So I tried couple of fixes and reducing python version to 3.7 and setting pytorch version to 1.9.1 and it worked. Now I reverted to 3.8.

Regarding the tests failing for preshape #1405.  last time random seeds were fine tuned.  Now we are not doing that. That is the reason it keeps failing for pytorch and tf backends.

PR #1410, opened 2022-03-14
Single parametrizer but mutliple geometry parent testing classes

SaitejaUtpala:
Following PR implements

TestCases are geometry specific and Parametrizer is single. This solves two problems

- If particular geometry class inherits two core classes. Before couldn't add implicit tests from both because for a class there is only one parametrizer. There is no way to do it with older design
-  Also if we add some additional tests and we have keep specifically adding to attribute. This is no longer needed.
- Also skipped tests are skipped instead of being deleted
- Skipped tests data is not generated instead empty list is passed. This should speedup tests a bit


ninamiolane (approved):
Great, thanks! Is there any way the new method in Parameterizer could be made more readable, e.g. by:
- using this style of string formatting: f"testing_data object doesn't have {data_fn_str} function for pairing with {attr_name}".
- being broken down in smaller auxiliary functions that would have explanatory names and docstrings?

tests/conftest.py
```diff
@@ -94,6 +96,119 @@ def pytorch_error_msg(a, b, rtol, atol):
...
+    References
+    ----------
+    More on pytest's parametrizers can be found here:
+    https://docs.pytest.org/en/6.2.x/parametrize.html
+    """
+
+    def __new__(cls, name, bases, attrs):
+        """Decorate the test class' methods with pytest."""
+        for base in bases:
+            test_fn_list = [fn for fn in dir(base) if fn.startswith("test") is True]
```
ninamiolane, line 174: "is True" is not needed.

SaitejaUtpala:
@ninamiolane sure definitely. I have noted it down. I will raise another PR.

PR #1414, opened 2022-03-17
fix vectorization of geodesic in landmarks

nguigs:


ninamiolane (approved):
LGTM but the tests are not triggered for some reason.

PR #1417, opened 2022-03-17
log_exp to log_then_exp

SaitejaUtpala:
after discussion @nguigs , order is more explicit for composition. But basis_representation_then_matrix_representation_composition is crossing 89 characters(linting wouldn't pass). @ninamiolane  @nguigs  Let me know what would be ideal here


ninamiolane:
I let @nguigs decide, I have no preference except that we should have a passing Linting GitHub action.

nguigs:
Maybe we don't need the word `composition` as it is clear that two methods are applied?

PR #1418, opened 2022-03-18
Stratified spaces

annacalissano:
Pulling:

stratified_space abstract structure
spider example of a stratified space
test for the spider (built it the old way)

Note1: it works only with numpy backend
Note2: wrapper and other solutions (e.g. to_arrays function, test) has been discussed yesterday with Nicolas.
Note3: names are still draft


ythanwerdas (commented):
Great job on this new class for stratified spaces! My main concern is on the geodesic method, regarding consistency with the geodesic method on manifolds.

PR #1430, opened 2022-03-26
Add wgpr support for tensor of dim > 1, and default priors

arthurPignet:
Add wgpr support for tensor of dim > 1, and 2 default priors

Sickit-learn can only handle vectors as target, while some objects in geomstats are matrices, like in the SDP cone. 
In this new version, the target are vectorized before being passed to the sklearn regressor. 
In the same spirit the predictions are de-vectorized. 

We also add support for 2 default priors, frechet mean and geodesic regression. 

- [ ] support for tensor of arbitrary dim
- [ ] Fréchet mean as default prior
- [ ] Geodesic regression as possible prior

@Tom-Szwagier


ninamiolane (approved):
Great, thanks!

Issue #1198, opened 2021-12-03
Add GitHub actions that highlight, directly in the code review, which lines have failed the tests


ninamiolane:
This workflow:
https://github.com/facebookresearch/ParlAI/blob/main/.github/workflows/lint.yml
should allow to do it, see e.g. this PR:
https://github.com/facebookresearch/ParlAI/pull/4134/files

ninamiolane:
Hi Monsij, so good to see you here 🤩 ! 

Nothing has been done on this yet. The workflow you are referring to does mark the lines that have failed the linting tests (code style), but it does not show directly in the code review, i.e. not directly in the PR: one needs to click on the workflow to go to the GitHub actions page. It would be easier for maintainers if it could be shown directly in the PR, e.g. like here:
https://github.com/facebookresearch/ParlAI/pull/4134/files
![Screen Shot 2022-02-20 at 11 02 47 AM](https://user-images.githubusercontent.com/8267869/154859618-c380d349-94ee-4b36-b5f1-4003e63cd87e.png)

monsij:
If I understand properly, the test workflow also needs to run `flake8` and make annotations ? @ninamiolane

ninamiolane:
Sorry if this was not clear. The test workflow needs to run, i.e. run pytest, (not `flake8`) and make annotations from pytest.

PR #1444, opened 2022-04-02
Add action for pytest annotation

monsij:
Addresses #1198 
Here is a preview of how it works: (`assert 3==4` is a dummy line)

Note that it comments for every error found for each job in the matrix.

Screen Shot 2022-04-02 at 2 07 52 AM


ninamiolane (approved):
Great, thank you so much!

PR #1449, opened 2022-04-05
Add KMeans++ initialization

alebrigant:
Improve KMeans by:
- adding KMeans++ initialization (choose one centroid randomly, then choose the others by maximizing the distance to the centroids already chosen)
- adding prints in the verbose mode, in particular the evolution of the convergence criterion
-  returning the variance of the clustering, to be able to compare different solutions obtained by running kmeans several times and picking best one.
- returning the clustering labels at the `.fit` method so that one does not need to run the `.predict` method if the set of points is the same.


ninamiolane (approved):
Great, thanks!

PR #1450, opened 2022-04-06
Parallel transport of the BuresWasserstein metric

nguigs:


ninamiolane (approved):
Great, thanks! Should we put `skip_test_parallel_transport_ivp_is_isometry = True` to tf_backend() or torch_backend(), since this should pass in autograd and numpy now?

If not, I'll let you merge.

PR #1451, opened 2022-04-06
change then by after

nguigs:
The composition is clearer when expressed as f after g than by g then f because the former respects the order of writing with \circ


SaitejaUtpala:
@nguigs  Following needs to  be changed as well.
 
- basis_representation , matrix representation
- intrinsic, extrinic

PR #1452, opened 2022-04-07
Van der Reycken metric on GLn

Iksiri:


ninamiolane (changes requested):
Thank you for your contribution! This looks very good, but we need to following before we can merge:
- a reference to the Van der Reycken metric in the docstring of the class
- the code should follow international guidelines: can you install the pre-commit mentioned on the README, so that it would clean your code automatically?
- the code should be tested. Could you add tests in `tests/tests_geomstats/`? @SaitejaUtpala  or @lpereira95 may be able to assist you if you have trouble in the tests.
Many thanks!

ninamiolane:
@Iksiri @lpereira95 any updaate on this?

PR #1455, opened 2022-04-07
WaldSpace without Geometry

JonasLueg:
The Wald Space as a PointSet and the corresponding point class that is a Wald (the elements of the space).
Plus tests (work in progress).


JonasLueg:
Waiting for the main testing framework for stratified_geometry to be included in the main repository.
After this is done, we can start pushing this one.

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+
+        Returns
+        -------
+        hash_of_topology : int
+            Return the hash of the topology.
+        """
+        return hash((self.n, self.partition, self.split_sets))
+
+
+class Wald(Point):
```
lpereira95, line 670: Isn't `WaldPoint` a better name? `Wald` seems to dry to me...

tests/tests_geomstats/test_wald_space.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class TestWald(PointTestCase, metaclass=Parametrizer):
+
+    _Point = Wald
+    testing_data = WaldTestData()
+
+
+@geomstats.tests.np_only
+class TestSplit(geomstats.tests.TestCase):
```
lpereira95, line 21: @JonasLueg, what are your thoughts on creating `SplitData` and make a better split (;)) between test and data as in the other cases? Do you think it would be too much complexity for such a simple case?

tests/tests_geomstats/test_wald_space.py
```diff
@@ -0,0 +1,246 @@
...
+from tests.stratified_geometry_test_cases import PointSetTestCase, PointTestCase
+
+
+class TestWaldSpace(PointSetTestCase, metaclass=Parametrizer):
+
+    _PointSet = WaldSpace
+    testing_data = WaldSpaceTestData()
+
+
+class TestWald(PointTestCase, metaclass=Parametrizer):
```
lpereira95, line 14: Do not forget to add tests to (at least) all public methods: `e.g. `Wald.generate_wald`. Sometimes the tests are trivial, but at least they force as to keep the API over time.

lpereira95:
Is there any example notebook you can ship with this code? It would be pretty nice to have a full working tutorial (even if very basic).

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+            Geometric Science of Information, Lecture Notes in Computer Science,
+            pages 710–717, Cham, 2021.
+            https://doi.org/10.1007/978-3-030-80209-7_76.
+"""
+
+
+import functools
+import itertools as it
+
+import numpy as np
```
lpereira95, line 54: `import geomstats.backend as gs`

I know it only works with `np`, but maybe soon we will be able to extend to different backends (e.g. `jax`).

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        ``part1``.
+    """
+
+    def __init__(self, part1, part2):
+        part1, part2 = tuple(sorted(list(part1))), tuple(sorted(list(part2)))
+        if set(part1) & set(part2):
+            raise ValueError(
+                f"A split consists of disjoint sets, those are not: {part1}, {part2}."
+            )
+        if part1 and part2:
```
lpereira95, line 92: I think this code does the same, but with half of the logical statements:

```
if part1 and part2:
    self.part1 = part1 if part1[0] < part2[0] else part2
    self.part2 = part2 if part1[0] < part2[0] else part1
else:
    self.part1 = part1 or part2
    self.part2 = part2 if part1 else ()
```

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+            The subset that the split is restricted to.
+
+        Returns
+        -------
+        restr_split : Split
+            The restricted split, if the split is :math:`A\vert B`, then the split
+            restricted to the subset :math:`C` is :math:`A\cap C\vert B\cap C`.
+        """
+        return Split(
+            part1=tuple(set(self.part1) & subset),
```
lpereira95, line 120: Is there any need to "setify" `self.part1` and `self.part2`? We can ensure they are sets from the beginning (or force them to be, if the case). Probably does not have performance impact, but is an unnecessary operation.

This repeats several times.

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        Returns
+        -------
+        is_compatible_with : bool
+            Return ``True`` if the splits are compatible, else ``False``.
+        """
+        p1, p2 = set(self.part1), set(self.part2)
+        o1, o2 = set(other.part1), set(other.part2)
+        return sum([bool(s) for s in [p1 & o1, p1 & o2, p2 & o1, p2 & o2]]) < 4
+
+    def __eq__(self, other):
```
lpereira95, line 225: In terms of class organization, I would probably move all the dunder methods (e.g. `__eq__`) to the beginning of the class, after `__init__`. Seems unnatural to find public methods first...

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        x : array-like
+            Takes a vector of length 'number of total splits' of the structure.
+
+        Returns
+        -------
+        corr : array-like, shape=[n, n]
+            Returns the corresponding correlation matrix.
+        """
+        corr = np.zeros((self.n, self.n))
+        for path_dict in self.paths():
```
lpereira95, line 455: What do you think about making `paths` a property? I'm not sure about this suggestion, but `self.paths` seems more natural than `self.paths()`.

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+            for (u, v), path in path_dict.items():
+                corr[u][v] = np.prod([1 - x[self.where[split]] for split in path])
+                corr[v][u] = corr[u][v]
+        np.fill_diagonal(corr, 1)
+        return gs.array(corr, dtype=gs.float32)
+
+    def support(self):
```
lpereira95, line 462: Same as for `paths`.

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        """Compute the gradient of the correlation matrix, differentiated by x.
+
+        Parameters
+        ----------
+        x : array-like
+            The vector x at which the gradient is computed.
+
+        Returns
+        -------
+        gradient : list of array-like
```
lpereira95, line 498: Why list of array like instead of array-like? What are the advantages?

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        other : Topology
+            The structure to which self is compared to.
+
+        Returns
+        -------
+        is_less_than_or_equal : bool
+            Return ``True`` if (i), (ii) and (iii) are satisfied, else ``False``.
+        """
+
+        class MyCheckException(Exception):
```
lpereira95, line 576: Better naming?

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+        hash_of_topology : int
+            Return the hash of the topology.
+        """
+        return hash((self.n, self.partition, self.split_sets))
+
+
+class Wald(Point):
+    r"""A class for wälder, that are phylogenetic forests, elements of the Wald Space.
+
+    Parameters
```
lpereira95, line 673: Better naming?

e.g. `self.edge_weights` instead of `self.x`?

`st` is hard to grasp also.

geomstats/stratified_geometry/wald_space.py
```diff
@@ -0,0 +1,1024 @@
...
+    ----------
+    n : int
+        Integer determining the number of labels in the forests, and thus the shape of
+        the correlation matrices: n x n.
+    """
+
+    def __init__(self, n):
+        super(WaldSpace).__init__()
+        self.n = n
+        self.a = spd.SPDMatrices(n=self.n)
```
lpereira95, line 949: What is `self.a`? Again, naming. Attributes of the class can also go to the class docstrings (under `Attributes`).

geomstats/stratified_geometry/wald_space.py
JonasLueg: Wald = German for forest and each point is a forest. I would wanted to have something very short.. I can change it to WaldPoint though, maybe more consistent with WaldSpace.

geomstats/stratified_geometry/wald_space.py
JonasLueg: Does this mean I should never import numpy?

lpereira95: Exactly. `gs` will have all the functions and methods you need though.

geomstats/stratified_geometry/wald_space.py
lpereira95: `NotPartialOrderException` maybe?

geomstats/stratified_geometry/wald_space.py
JonasLueg: I'll call it self.ambient now, that's ok

geomstats/stratified_geometry/wald_space.py
JonasLueg: named x now weights and st now top as the attribute and topology as the parameter one has to pass the __init__.

geomstats/stratified_geometry/wald_space.py
JonasLueg: So, for example, it has ``gs.maximum``, but not ``gs.minimum``.
Should I

1. ask you to add it to the backends;
2. add it myself to the backends;
3. add it myself only to the numpy backend;
4. or try to find a workaround (for example ``-gs.maximum(-x) == gs.minimum``?

geomstats/stratified_geometry/wald_space.py
JonasLueg: There are no particular advantages except that I know what a list of arrays does; I was not sure how gs.array([list of arrays]) would behave. That's why I've left it as a list for now. Should I do something like gs.array([list of arrays], axis=-1), or can I safely do gs.array([list of arrays]) and I'll get a three dimensional array?

tests/tests_geomstats/test_wald_space.py
JonasLueg: Yes, I can certainly do that, the main motivation is probably the consistency with the testing framework.

geomstats/stratified_geometry/wald_space.py
lpereira95: `gs.minimum` is now available (see [https://github.com/geomstats/geomstats/blob/master/geomstats/_backend/numpy/__init__.py]). I'll see what I can do for `np.fill_diagonal`.

JonasLueg: thank you :)

JonasLueg:
> @JonasLueg, I'll merge very soon the last updates on spider and stratified spaces. Then, I think it is a good idea that you bring the master to this branch (via merge), so you can access the latest testing tools.

Indeed, I'm following your contributions very closely and can't wait to merge these into the wald space branch! 👍🏽 
I'm really happy to see this rapid progress :) so thanks a bunch! 🥳

ninamiolane:
@lpereira95 @JonasLueg  any update on this? 🙌 Is it outdated?

geomstats/stratified_geometry/wald_space.py
lpereira95: `gs.array([list of arrays])` will return a 3d array. Same behavior as `np.array`.

lpereira95:
Great work @JonasLueg (very high quality code)! I've made small improvements (mostly related with naming) and split tests and data for `Split` and `Topology`. I'll merge as soon as the tests finish.

I'm looking forward for the metric implementation!

PR #1457, opened 2022-04-07
harmonize access to metric property and setting at init

### LINKED ISSUE #1453: Change init of SPDMatrices to allow for a custom metric

https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/spd_matrices.py

### What should be modified?

The following code:
```
from geomstats.geometry.spd_matrices import SPDMatrices, SPDMetricBuresWasserstein
bw_metric = SPDMetricBuresWasserstein(n=28)
spd = SPDMatrices(n=28, metric=bw_metric)
```
gives the error:
```
----> 5 spd = SPDMatrices(n=28, metric=bw_metric)

~/code/geomstats/geomstats/geometry/spd_matrices.py in __init__(self, n, **kwargs)
     28
     29     def __init__(self, n, **kwargs):
---> 30         super(SPDMatrices, self).__init__(
     31             dim=int(n * (n + 1) / 2),
     32             metric=SPDMetricAffine(n),

TypeError: geomstats.geometry.base.OpenSet.__init__() got multiple values for keyword argument 'metric'
```

As suggested by @nguigs , we should have an if that only uses SPDMetricAffine if `metric=None`

### Additional information

_No response_

### LINKED ISSUE #1461: Froebenius metric in special_orthogonal needs to be scaled by 1/2 to corresponds to classical rotation angles.

_No response_

### What should be modified?

Currently, the distance from identity of a rotation of angle theta is \sqrt{2} \theta instead of theta. The Froebenius metric used on the tangent space at identity needs to be scaled by 1/2 to obtain usual angles in radians. For instance, the code bellow gives: found 1.4142135623730951 instead of 1.

```python
import numpy as np
from geomstats.geometry.special_orthogonal import SpecialOrthogonal
so3 = SpecialOrthogonal(n=3, point_type="matrix")
rota = np.array([
            [1., 0., 0.],
            [0., 1., 0.],
            [0., 0., 1.]])
theta = 1.
tg_v = np.array([
            [0., 0., 0.],
            [0., 0., theta],
            [0., -theta, 0.]])
rot_theta = gs.linalg.expm(tg_v)

result_dist = so3.bi_invariant_metric.dist(rota, rot_theta)
expected_dist = theta
print('found', result_dist, 'instead of', expected_dist)
```

### Additional information

_No response_

nguigs:
closes #1453 
closes #1461 

This PR changes the remove the `metric` attribute that was coexisting with the `metric` property and the `_metric` attribute.
It also changes the way `_metric` is set to allow to pass other metric as kwarg when initializing a manifold. Some exceptions are made for the spaces that make sense for a given metric only, e.g. Poincare halfspace.

@lpereira95 this could be of interest to you

### DIFF
--- a/geomstats/geometry/full_rank_matrices.py
+++ b/geomstats/geometry/full_rank_matrices.py
@@ -9,7 +9,7 @@ from geomstats.geometry.matrices import Matrices, MatricesMetric
 
 
 class FullRankMatrices(OpenSet):
-    r"""Class for `math:`R_*^{m\times n}` matrices of dimension m x n and full rank.
+    r"""Class for `math:`R_*^{n\times k}` matrices of dimension n x k and full rank.
 
     Parameters
     ----------
--- a/geomstats/geometry/full_rank_matrices.py
+++ b/geomstats/geometry/full_rank_matrices.py
@@ -22,12 +22,12 @@ class FullRankMatrices(OpenSet):
     def __init__(self, n, k, **kwargs):
         if "dim" not in kwargs.keys():
             kwargs["dim"] = n * k
-        super(FullRankMatrices, self).__init__(
-            ambient_space=Matrices(n, k), metric=MatricesMetric(n, k), **kwargs
-        )
+        super(FullRankMatrices, self).__init__(ambient_space=Matrices(n, k), **kwargs)
         self.rank = min(n, k)
         self.n = n
         self.k = k
+        if self._metric is None:
+            self._metric = MatricesMetric(n, k)
 
     def belongs(self, point, atol=gs.atol):
         r"""Check if the matrix belongs to `math:`R_*^{n \times k}`.
--- a/geomstats/geometry/full_rank_matrices.py
+++ b/geomstats/geometry/full_rank_matrices.py
@@ -20,14 +20,12 @@ class FullRankMatrices(OpenSet):
     """
 
     def __init__(self, n, k, **kwargs):
-        if "dim" not in kwargs.keys():
-            kwargs["dim"] = n * k
+        kwargs.setdefault("dim", n * k)
+        kwargs.setdefault("metric", MatricesMetric(n, k))
         super(FullRankMatrices, self).__init__(ambient_space=Matrices(n, k), **kwargs)
         self.rank = min(n, k)
         self.n = n
         self.k = k
-        if self._metric is None:
-            self._metric = MatricesMetric(n, k)
 
     def belongs(self, point, atol=gs.atol):
         r"""Check if the matrix belongs to `math:`R_*^{n \times k}`.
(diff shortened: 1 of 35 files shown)


geomstats/geometry/full_rank_matrices.py
```diff
@@ -22,12 +22,12 @@ class FullRankMatrices(OpenSet):
...
         if "dim" not in kwargs.keys():
             kwargs["dim"] = n * k
-        super(FullRankMatrices, self).__init__(
-            ambient_space=Matrices(n, k), metric=MatricesMetric(n, k), **kwargs
-        )
+        super(FullRankMatrices, self).__init__(ambient_space=Matrices(n, k), **kwargs)
         self.rank = min(n, k)
         self.n = n
         self.k = k
+        if self._metric is None:
```
lpereira95, line 29: I wonder if instead, we should do something like:

```
kwargs.setdefault('metric', MatricesMetric(n, k)
```

For dim we can do similarly:
```
kwargs.setdefault('dim', n*k)
```

geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -38,6 +38,13 @@ def __init__(self, n, k, **kwargs):
         self.rank = k
         self.sym = SymmetricMatrices(self.n)
 
+    @property
```
lpereira95, line 41: For consistency, shouldn't we override `_metric` if not defined at initialization and do not override the property at this level?

nguigs: It is not possible in that case because of circular/recursive initialisation: the metric here needs a `FiberBundle` which itself needs the space `RankKPSDMatrices` initialized.

geomstats/geometry/full_rank_matrices.py
nguigs: that's a good point, I didn' t know that syntax. There are places where we do 
```python
if dim not in kwargs:
    kwargs['dim'] = ...
```
And also for `shape` and `default_point_type`

ninamiolane (changes requested):
The new syntax is making the notebooks on discrete curves fail:
```
CURVES_SPACE = DiscreteCurves(R2)
SRV_METRIC = CURVES_SPACE.square_root_velocity_metric

geodesic = SRV_METRIC.geodesic(initial_curve=cell_start, end_curve=cell_end)

n_times = 80
times = gs.linspace(0.0, 1.0, n_times)
geod_points = geodesic(times)
------------------

AttributeError                            Traceback (most recent call last)
Input In [18], in <cell line: 6>()
      3 CURVES_SPACE = DiscreteCurves(R2)
      4 SRV_METRIC = CURVES_SPACE.square_root_velocity_metric
----> 6 geodesic = SRV_METRIC.geodesic(initial_curve=cell_start, end_curve=cell_end)
      8 n_times = 80
      9 times = gs.linspace(0.0, 1.0, n_times)

AttributeError: 'NoneType' object has no attribute 'geodesic'
AttributeError: 'NoneType' object has no attribute 'geodesic'
```

Did it run fine for you?

ninamiolane:
Additionally, introducing tests that checks that this runs for many manifolds&metrics would be good.

PR #1463, opened 2022-04-07
Spider Example and Tests for Stratified Spaces

annacalissano:


lpereira95:
@SaitejaUtpala, the test structure for StratifiedSpaces is just a trial. Your feedback will be very valuable!

tests/data/spider_data.py
```diff
@@ -0,0 +1,185 @@
...
+                start_point=SpiderPoint(1, 2.0),
+                end_point=SpiderPoint(2, 3.0),
+            ),
+        ]
+
+        return self.generate_tests(smoke_data)
+
+    def geodesic_test_data(self):
+        pts_start = [SpiderPoint(10, 1.0), SpiderPoint(10, 2.0), SpiderPoint(3, 1.0)]
+
```
SaitejaUtpala, line 120: There is no corresponding test in `test_spider.py` but the data is present?

SaitejaUtpala:
@annacalissano  @lpereira95 Great Work. Looks good to me from testing side.

tests/data/spider_data.py
lpereira95: The method to test this data is in the parent class (`PointSetGeometryTestCase`). I'm trying to generalize these cases somehow, but still a work in progress.

tests/stratified_geometry_test_cases.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertAllClose(result, True)
+
+    def test_belongs(self, space_args, points, expected):
+        space = self._PointSet(*space_args)
+        self.assertAllClose(space.belongs(points), expected)
+
+    def test_set_to_array(self, space_args, points, expected):
+        space = self._PointSet(*space_args)
+
+        self.assertAllClose(space.set_to_array(points), expected)
```
JonasLueg, line 26: We could (and should in my opinion) automate `test_set_to_array` using the `test_to_array` method from `PointTestCase`; as the expected output of `set_to_array` is just the method `to_array` from `Point` applied to each element, no?
Not sure if this is trivial to do, as one would need to draw the smoke_data from `test_to_array`. I think it is worth giving a thought.

lpereira95: In the case of spider the outputs are different: `point.to_array()` only translates ray and R value to an array, whereas `point_set.set_to_array(point)` creates an array of zeros with R from point assigned to the ray position.

@annacalissano may add something here.

geomstats/stratified_geometry/stratified_spaces.py
```diff
@@ -4,43 +4,48 @@
 """
 
 from abc import ABC, abstractmethod
-from typing import TypeVar
 
 
-def belongs_vectorize(fun):
-    r"""Vectorize the belongs acting on Point as lists."""
+def _vectorize_point(*positions):
```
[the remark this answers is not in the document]
annacalissano: great!

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,294 @@
...
+        Parameters
+        ----------
+        initial_point : SpiderPoint or list of SpiderPoint, shape=[...]
+             Point in the Spider.
+        end_point : SpiderPoint or list of SpiderPoint, shape=[...]
+             Point in the Spider.
+
+        Returns
+        -------
+        geo : function
```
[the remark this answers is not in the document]
annacalissano: Check if the outputs are consistent in spider and graphs and wald

tests/stratified_geometry_test_cases.py
```diff
@@ -0,0 +1,63 @@
...
+@geomstats.tests.np_only
+class PointSetGeometryTestCase(TestCase):
+    def test_dist(self, space_args, start_point, end_point, expected):
+        space = self._PointSet(*space_args)
+        geom = self._SetGeometry(space)
+        results = geom.dist(start_point, end_point)
+
+        self.assertAllClose(results, expected)
+
+    def test_geodesic(self, space_args, start_point, end_point, t, expected):
```
annacalissano, line 45: check the output shape structure

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+    _vectorize_point,
+)
+
+
+class SpiderPoint(Point):
+    r"""Class for points of the Spider.
+
+    Parameters
+    ----------
+    s : int
```
ninamiolane, line 22: It is recommended to avoid single-letter for variable names.

Because:
- while "s" means something to the person who wrote the code, it might not mean much to the new user. The user has to go back to the documentation each time they forget what "s" meant.
- during big refactorings of the library, "Find & Replace" variables with a single-letter name can become a nightmare.

I would suggest:
- s --> stratum
- x --> stratum_coord or coord

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+    def __eq__(self, other):
+        """Compare two points."""
+        return self.s == other.s and abs(self.x - other.x) < gs.atol
+
+    def to_array(self):
+        """Return the hash of the instance."""
+        return gs.array([self.s, self.x])
+
+
+class Spider(PointSet):
```
ninamiolane, line 54: Could you add a reference with the mathematical formulas related to the spider?

We are trying to enforce that references are put in the docstrings, to ease the work of the technical writers who might refactor our documentation pages.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+    r"""Spider: a set of rays attached to the origin.
+
+    Parameters
+    ----------
+    rays : int
```
ninamiolane, line 59: For consistency with the codebase, I would suggest renaming this to n_rays. In the current codebase, variables that contain "number of" something, are prefixed with `n_`.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+        for single_point in point:
+            results += [
+                self._value_check(single_point)
+                and self._rays_check(single_point)
+                and self._zero_check(single_point)
+                and type(single_point) is SpiderPoint
+            ]
+        return gs.array(results)
+
+    def _rays_check(self, single_point):
```
ninamiolane, line 113: n_rays

ninamiolane: I was wondering what was `single_point`, but I understand that you have to make the distinction between `single_point` and `point`. The single_points are arranged into lists, and not into an array.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+        -------
+        belongs : boolean
+            Boolean denoting if the point has zero length when it has zero ray.
+        """
+        if single_point.s == 0 and single_point.x != 0:
+            return False
+        return True
+
+    @staticmethod
+    def _value_check(single_point):
```
ninamiolane, line 149: This is effectively checking x, i.e. the coordinate on the ray: should it be called _coord_check? 

Using the same name several times is easier for users who are discovering the file.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+        point_array : array-like, shape=[...,rays]
+            An array with the x parameter in the s position.
+        """
+        point_to_array = gs.zeros((len(point), self.rays))
+        for i, pt in enumerate(point):
+            point_to_array[i, pt.s - 1] = pt.x
+        return point_to_array
+
+
+class SpiderGeometry(PointSetGeometry):
```
ninamiolane, line 186: I don't think we concluded on something for this name, my own personal preference would be: `PointSetMetric` and `SpiderMetric`.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+    """Geometry on the Spider, induced by the rays Geometry."""
+
+    def __init__(self, space, ambient_metric=EuclideanMetric(1)):
+        super(SpiderGeometry, self).__init__(space=space)
+        self.rays_geometry = ambient_metric
```
ninamiolane, line 191: It is confusing to have two different names: `ambient_metric` and `rays_geometry` that refer to the same object.

When a user initializes a new object and gives inputs (e.g. ambient_metric), they expect that these inputs are transformed into the object's attributes, without being renamed.

I would suggest:
`self.ray_metric = ray_metric`

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @property
+    def rays(self):
```
ninamiolane, line 194: n_rays

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+        for point_a_, point_b_ in values:
+            if point_a_.s == point_b_.s or point_a_.s == 0 or point_b_.s == 0:
+                result += [self.rays_geometry.norm(gs.array([point_a_.x - point_b_.x]))]
+            else:
+                result += [point_a_.x + point_b_.x]
+        return gs.array(result) if len(result) != 1 else result[0]
+
+    @_vectorize_point((1, "initial_point"), (2, "end_point"))
+    def geodesic(self, initial_point, end_point):
+        """Return the geodesic between two lists of Spider points.
```
ninamiolane, line 231: Why between two lists of spider points?

I thought:
- the space is the spider
- one point in this space is a spiderpoint
- a curve on that space is a curve between 2 spiderpoints

Are we infact implicitly considering a product space of N spiderspaces? In this case, this is very linked to the ProductManifold and NFoldManifold that we have. 

We can leave this for further discussion, but it is a bit confusing to me to unify the space of SpiderPoints and the space of N SpiderPoints.

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
...
+        if len(initial_point) == 1 and len(end_point) != 1:
+            values = itertools.zip_longest(
+                initial_point, end_point, fillvalue=initial_point[0]
+            )
+        else:
+            values = zip(initial_point, end_point)
+        fncs = [self._point_geodesic(pt_a, pt_b) for (pt_a, pt_b) in values]
+        return lambda t: _vec(t, fncs=fncs)
+
+    def _point_geodesic(self, initial_point, end_point):
```
ninamiolane, line 261: Would `ray_geodesic` work here?

geomstats/stratified_geometry/stratified_spaces.py
```diff
@@ -4,43 +4,47 @@
...
```
ninamiolane, line 9: This could go in the `geomstats/vectorization.py` file instead of being hidden in stratified_space.py?

tests/data_generation.py
```diff
@@ -1051,3 +1052,102 @@ def _parallel_transport_bvp_is_isometry_test_data(
             )
 
         return self.generate_tests([], random_data)
+
+
+class _PointSetTestData(TestData):
+    # TODO: add docstrings
```
ninamiolane, line 1058: Remove leftover comments and address the TODO

tests/data_generation.py
```diff
@@ -1051,3 +1052,102 @@ def _parallel_transport_bvp_is_isometry_test_data(
...
+        smoke_data = [
+            dict(space=space, points=pts[0]),
+            dict(space=space, points=pts),
+        ]
+
+        return self.generate_tests(smoke_data)
+
+
+class _PointTestData(TestData):
+    # TODO: add docstrings
```
ninamiolane, line 1096: Ditto

tests/data_generation.py
```diff
@@ -1051,3 +1052,102 @@ def _parallel_transport_bvp_is_isometry_test_data(
...
+    pass
+
+
+class _PointGeometryTestData(TestData):
+    # TODO: add docstrings
```
ninamiolane, line 1101: Ditto

tests/stratified_geometry_test_cases.py
```diff
@@ -0,0 +1,123 @@
+"""Core parametrizer classes for Stratified Spaces.
```
ninamiolane, line 1: Nice!!!

geomstats/stratified_geometry/spider.py
```diff
@@ -0,0 +1,303 @@
+"""Class for the spider.
+
+Lead authors: Anna Calissano & Jonas Lueg
+"""
+import itertools
+
+import geomstats.backend as gs
+from geomstats.geometry.euclidean import EuclideanMetric
+from geomstats.stratified_geometry.stratified_spaces import (
```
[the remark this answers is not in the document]
ninamiolane: (names of test files would have to be changed too)

ninamiolane (commented):
Very nice!! Many thanks for the huge work. I put some comments on the names and structures to enforce consistency within the code base and ease the work of the people who will refactor it down the line :)

geomstats/stratified_geometry/spider.py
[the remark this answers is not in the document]
annacalissano: Done!

geomstats/stratified_geometry/spider.py
annacalissano: I added the formulas and a reference

geomstats/stratified_geometry/spider.py
annacalissano: Exactly. Multiple points ara arrenged into lists. We built a decorator which allows the functions to work for both a single point and a list of points, to mantain the same "vectorized" fashion of the rest of the package.

JonasLueg:
> @JonasLueg, we've made several updates in the tests. They're more automatic now. Can you please check what you think? Notice that for spider we don't need to define any test, even though all the methods are thoroughly tested.

I think it's great and especially the output shape tests save some time for the contributor. 
Also, the tests on distance and geodesics are nice and definitely needed.

One week point might be that, depending on the space you're testing, the computation time for dist/geodesic might be somewhat long, and thus the test time could blow up. So, as a contributor of some space, one has to specify the dimensions in the way that the computation times are bearable. Also, keep in mind that in general spaces one is not able to analytically compute the geodesic and implement a nice formula to return geodesic(t), but rather, one is able to return a list of, say, `m` points, and then one needs to convert this list as precise as possible to a map ``geodesic``, and a given time ``t`` would yield the point ``int(t*m)`` in the list of ``m`` points.

With this comment I'm not implying that something needs to be changed, but rather that sometimes the contributor might find it hard to be able to pass the tests (as for Wald Space, I think one might pass the tests at the cost of longer computation times, although that should be short enough still).

I think the framework is in a very good shape now! 👍🏽🌱

lpereira95:
Thanks for the great feedback @JonasLueg!

To address the weak points:
1. dist/geodesic computing time: i) the user can always skip a given test (e.g. `skip_<test_name>` in the respective test class) - we should avoid this, but it is ok if skipped tests are well replaced by other tests; ii) these tests are using `space_args_list[0]` to build the space, so the user can control which are the space parameters (ideally with some randomness), and hopefully use one example that is not computationally expensive; iii) the test methods can be overridden.
2. for the "list-output geodesic", we can in fact retrieve a geodesic that behaves like a continuous function (for coherence with what is expected) by creating some kind of interpolation (e.g. internally store a list, and retrieve `int(t*m)` when user passes `t`).

In fact, we are trying to be very generic, but some cases really have to be adapted (you'll seen soon we are not using most of this test machinery in the graph spaces). This is mostly a pitty for the user, as it will force him/her to write more tests.

geomstats/stratified_geometry/stratified_spaces.py
lpereira95: @ninamiolane, maybe for now we should keep in `stratified` only, because it can give the wrong impression in the context of the all library, as `vectorize` here is more "transform the input in a list".

What may be interesting is to review the decorators in `vectorize`, because I believe there's some overlap with the definition of this one and there's probably some code we can simplify there.

geomstats/geometry/stratified/point_set.py
```diff
@@ -3,44 +3,50 @@
...
 """
 
+import functools
 from abc import ABC, abstractmethod
-from typing import TypeVar
 
 
-def belongs_vectorize(fun):
-    r"""Vectorize the belongs acting on Point as lists."""
+def _vectorize_point(*positions):
```
ninamiolane, line 10: Sorry I missed this one before: is `positions` different than `points`? It seems to me that they refer to the same concept. If this is the case, they should have the same name: consistent names really helps users.

geomstats/geometry/stratified/spider.py
```diff
@@ -0,0 +1,337 @@
...
+        for single_point in point:
+            results += [
+                self._coord_check(single_point)
+                and self._rays_check(single_point)
+                and self._zero_check(single_point)
+                and type(single_point) is SpiderPoint
+            ]
+        return gs.array(results)
+
+    def _rays_check(self, single_point):
```
ninamiolane, line 129: For consistency:
- `_n_rays_check`

geomstats/geometry/stratified/spider.py
```diff
@@ -0,0 +1,337 @@
...
+
+        Parameters
+        ----------
+        point : SpiderPoint or list of SpiderPoint, shape=[...]
+             Points to be checked.
+
+        Returns
+        -------
+        point_array : array-like, shape=[...,rays]
+            An array with the x parameter in the s position.
```
ninamiolane, line 194: x parameter -> stratum_coord
s -> stratum

why is the shape of this array [..., rays]? Should it be n_rays?

geomstats/geometry/stratified/spider.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        point_to_array = gs.zeros((len(point), self.n_rays))
+        for i, pt in enumerate(point):
+            point_to_array[i, pt.stratum - 1] = pt.stratum_coord
+        return point_to_array
+
+
+class SpiderMetric(PointSetMetric):
+    """Geometry on the Spider, induced by the rays Geometry."""
```
ninamiolane, line 203: Missing parameter description from the init in the class docstring

geomstats/geometry/stratified/spider.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def __init__(self, space, ray_metric=EuclideanMetric(1)):
+        super(SpiderMetric, self).__init__(space=space)
+        self.ray_metric = ray_metric
+
+    @property
+    def n_rays(self):
+        """Get rays."""
```
ninamiolane, line 211: Get number of rays

tests/data/spider_data.py
```diff
@@ -0,0 +1,107 @@
+import random
+
+import geomstats.backend as gs
+from geomstats.geometry.stratified.spider import Spider, SpiderMetric, SpiderPoint
```
ninamiolane, line 4: 🙌 Many thanks!

tests/data/spider_data.py
```diff
@@ -0,0 +1,107 @@
...
+    def to_array_test_data(self):
+        smoke_data = [
+            dict(point_args=(1, 2.0), expected=gs.array([1.0, 2.0])),
+            dict(point_args=(2, 1.0), expected=gs.array([2.0, 1.0])),
+        ]
+
+        return self.generate_tests(smoke_data)
+
+
+class SpiderGeometryTestData(_PointGeometryTestData):
```
ninamiolane, line 69: This should be SpiderMetricTestData, PointMetricTestData etc

tests/data_generation.py
```diff
@@ -1702,3 +1703,96 @@ def _log_is_horizontal_test_data(
...
+        ]
+
+        return self.generate_tests(smoke_data)
+
+
+class _PointTestData(TestData):
+    pass
+
+
+class _PointGeometryTestData(TestData):
```
ninamiolane, line 1747: PointMetricTestData

ninamiolane (approved):
Nice!! Many thanks for your patience, and for changing the structure of the files 🙌 

I have added very minor comments (residual names and docstrings items), then we can merge! 🎉 congrats!!

geomstats/geometry/stratified/point_set.py
lpereira95: This is a decorator that accepts keyword arguments. In order to use it, the user should specify the arg position and names. Check l105 here (https://github.com/annacalissano/geomstats/blob/spider/geomstats/geometry/stratified/spider.py). 

Maybe the name name should be adapted, I couldn't come up with better ideas (args_positions?).

The idea with this decorator is that we can address very different input orders and names with a very robust decorator.

geomstats/geometry/stratified/spider.py
lpereira95: The idea here is to show that we can pass one or several points.

geomstats/geometry/stratified/spider.py
lpereira95: This one I'll leave to @annacalissano.

lpereira95:
Need to check issues in testing before merging.

@SaitejaUtpala, do you have any idea why the tests are failing. Specially:

1. `testing_data object doesn't have 'triangle_inequality_of_dist_test_data' function for pairing with 'test_triangle_inequality_of_dist' for `test_discrete_curve`
2. `tests/tests_geomstats/test_spider.py` for tensorflow and pytorch should not run, as parent classes are marked with `@geomstats.tests.np_only`: should we also mark the children class? (locally they are skipped)

I had a conflict while merging (`data_generation` file), which I think was solved well (I just had to add the new classes at the end).

PR #1468, opened 2022-04-08
Add Flag manifold class

Tom-Szwagier:
Implement a new class for flag manifolds.

Representation (reduced projection coordinates), notations and formulas inspired from **[Ye2021]**. Work in progress. 

References
----------
**[Ye2021]** Ye, K., Wong, K.S.-W., Lim, L.-H.: Optimization on flag manifolds. Preprint, arXiv:1907.00949 [math.OC] (2019)


Tom-Szwagier:
@lpereira95 I have noticed an error on the shape of the points in the manifold. As I represented them, it is actually (d, n, n) and not (d * n, d * n). Should I correct that in this pull request, or correct that in the next pull request I am currently working on (to implement the remaining abstract methods)? Thank you.

lpereira95:
I think you should try to address it in this pull request. Also, I don't know what will be the content of your next pull request, but if it is closely related with this one (e.g. you are adding methods to any of the created classes), I would say you should do it in this.

Tom-Szwagier:
As discussed with @annacalissano and @lpereira95, as my current work is open and research oriented, I should wait to get results before making my PR. Before that, the current notations, representation and methods might evolve, so it's maybe not a good idea to make the PR now. I will let it open, but if you don't find it useful, you can close it @ninamiolane.

PR #1474, opened 2022-04-11
Make setup declarative

### LINKED ISSUE #1292: Install backend using idiomatic python/pip syntax
# This is a feature request

I find the way of choosing the backend could be done better by using the pythonic syntax:

```bash
pip install geomstats[torch]
```

or

```bash
pip install geomstats[tensorflow]
```

By default, the installed backend would be numpy.

_Reference: https://stackoverflow.com/a/60307740/11348232_

It could also be useful to use this feature to have a nicer syntax for the  development environment setup command (which is currently `pip3 install -r opt-requirements.txt` and could become `pip install geomstats[dev]`).

lpereira95:
When installing `geomstats` from `setup.py` we run `geomstats.__init__` to get the version (with `runpy`). This may be problematic, as some dependencies of `geomstats` are imported when running it (and some users will install them only when installing `geomstats` -> circular).

There's alternative ways to work around this. In my opinion, the `setup.cfg` approach (see [this](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html)) is the best way to install a python package, as it is declarative. (Long discussions around this topic are common, specially regarding the handling of extra dependencies - see e.g. [this](https://github.com/pypa/setuptools/issues/1951)).

With this approach we remove all the dependency files in the repo (to avoid repeated information; notice it would be trivial to build them automatically in case of need), but we gain by having them all clearly stated in a file. The installation of extra dependencies is also straightforward (e.g. `pip install -e .[test]`).

Regarding other packages, I could identify that [pandas](https://github.com/pandas-dev/pandas) follows this structure.

In terms of implementation, I believe the proposed solution is equivalent to the previous one (i.e. produces same output), but your review will confirm it.

What are your thoughts @ninamiolane, @nguigs and @SaitejaUtpala?

Edit:
closes #1292


lpereira95:
Great suggestions @SaitejaUtpala!

In fact #1292 becomes trivial to solve now.

P.S. I'll also have to adapt the actions.

ninamiolane:
Many thanks @lpereira95 ! LGTM.

lpereira95:
It seems to be working. From my side I think its all. Feel free to add something else, otherwise I believe we can merge.

ninamiolane (changes requested):
Thanks! I see that the .txt requirements are deleted which may cause the following problems:
- these files are referenced many times in the documentations (README, doc website, PR templates, etc): this should be adapted too.
- how do users who download geomstats from github know what to install?

lpereira95:
@ninamiolane I've updated README and docs to be consistent.

For the PR template and see no reference to requirements (I've made a full search in the repo for the keyword "requirements").

All the requirements, from mandatory to optional, can be found in the file setup.cfg. Notice how easy (and aesthetical, I dare to say) it is to combine different extra requirements to create a category of extra requirements (e.g. autograd+tf+pytorch=opt).

Issue #1475, opened 2022-04-11
Add properties to manifolds and metrics


SaitejaUtpala:
### Where should the content be modified?

https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/*

### What should be modified?

add properties to manifolds and metrics

for manifolds 

```
- compact 
- connected manifold (stiefel is disconnected if I am not wrong)
```

for metrics
```
- geodesic completeness
```

Ryan-Vaughn:
Is this issue still open? I've been thinking about helping on this one. I also had a couple questions regarding compactness and completeness. For example, since all open manifolds are not compact, is it desirable to specify is_compact() in the OpenManifold class? Or is it better to do it in each individual class? I imagine the first is more simple, but may lead to confusion for others reading the code. The same goes for completeness (i.e. all compact manifolds must have complete metrics.)

ninamiolane:
@Ryan-Vaughn thank you for reaching out! 🙌 

Yes, this issue is still open, you can take over the PR here https://github.com/geomstats/geomstats/pull/1478, as I believe that @SaitejaUtpala does not have bandwidth to continue it.

Yes, I agree that is_compact() should be implemented in the parent class OpenSet and return False.

Completeness should be implemented in a class inheriting from the parent class RiemannianMetric (not in the manifold).

Looking forward to this 🤩

ninamiolane:
@Ryan-Vaughn when including properties at the parent Python class level, (.e.g is_compact(): return False) in OpenSet, it would be good to cite the mathematical theorem that allows us to do so (e.g. Heine-Borel theorem) and give a reference.

PR #1478, opened 2022-04-11
Add properties to Manifold and metrics

SaitejaUtpala:
This PR collects and adds properties to manifolds and metrics. addressing #1475

### DIFF
--- a/geomstats/geometry/manifold.py
+++ b/geomstats/geometry/manifold.py
@@ -198,3 +198,34 @@ class Manifold(abc.ABC):
                 gs.random.normal(size=(n_samples,) + self.shape), base_point
             )
         )
+
+    @abc.abstractmethod
+    def is_connected(self):
+        """Return whether manifold is connected.
+
+        Returns
+        -------
+        _ : boolean
+            True if manifold is connected, False otherwise.
+        """
+
+    @abc.abstractmethod
+    def is_compact(self):
+        """Return whether manifold is compact.
+
+        Returns
+        -------
+        _ : boolean
+            True if manifold is compact, False otherwise.
+        """
+
+    @abc.abstractmethod
+    def is_geodesically_complete(self):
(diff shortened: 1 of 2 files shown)


geomstats/geometry/manifold.py
```diff
@@ -198,3 +198,34 @@ def random_tangent_vec(self, base_point, n_samples=1):
...
+    @abc.abstractmethod
+    def is_geodesically_complete(self):
+        """Return whether manifold with metric is geodesically complete.
+
+        Returns
+        -------
+        _ : boolean
+            True if manifold is geodesically complete, False otherwise.
+        """
+        return self.metric.is_geodesically_complete()
```
nguigs, line 231: This means it should be added to to the metric too right?

SaitejaUtpala:
Just to list out  manifold properties. @ninamiolane  @nguigs  @ythanwerdas  

* Euclidean : Connected : True, Compact : False
* General Linear : Connected : True, Compact: False (general linear as whole is disconnected, but since we are only instantiating with +det or -det it is always connected)
* Grasmannian : Connected : True, Compact : True (ref : https://en.wikipedia.org/wiki/Grassmannian)
* Heisenberg : Connected : True, Compact : True  (ref : https://math.stackexchange.com/questions/434384/the-heisenberg-manifold)
* Hermitian : Connected : True, Compact : False
* Hyperbolic : Connected : True, Compact : False
* Full rank matrices : Not sure 
* Full rank correlation matrices : Not Sure
* Hypersphere : Connected : True, Compact : True
* Lower triangular matrices : connected : True, compact : False
* Positive Lower triangular matrices : connected : True, compact : False
* Preshape space : Not Sure
* rank k psd matrices : Not Sure 
* skew symmetric : Connected : True, Compact : False

geomstats/geometry/manifold.py
ninamiolane: +1

PR #1487, opened 2022-04-12
Separate data from tests

lpereira95:
This pull request separates data from tests.

This is only the first step towards some more simplifications in tests. I think with these changes the tests will be perceived as less overwhelming (even if it is only perception, as nothing has changed). I also think they put in evidence that the test design is pretty powerful (great job @SaitejaUtpala and @ninamiolane).

In my view, the great advantage of this approach is that the tests are (or better, will be) data-agnostic, meaning we can run the same sets of tests with different data. Te path to achieve this is still halfway, as a lot of tests still do not separate data and tests.

The pull request is huge, but it is mostly copy-paste.


ninamiolane (approved):
LGTM, many thanks! @SaitejaUtpala I will let you merge if you are good with this too.

SaitejaUtpala (approved):
LGTM!

Issue #1283, opened 2022-01-13
Errors in FrechetMean (methods: default, adaptive) on discrete curves


ninamiolane:
The following code:

```
from geomstats.geometry.discrete_curves import R2, DiscreteCurves, SRVMetric
from geomstats.learning.frechet_mean import FrechetMean

import geomstats.backend as gs
import matplotlib.pyplot as plt

# Create toy cells

cell1 = gs.array([[
    [2, 2], [3, 3], [2, 4], [1, 3], [1.8, 1.8]
]])


cell2 = gs.array([[
    [3, 2], [2, 1], [-1, 4], [3, 3], [3.2, 2.2]
]])

cell1 = cell1 - gs.mean(cell1, axis=1)
cell2 = cell2 - gs.mean(cell2, axis=1)

plt.figure()
plt.plot(cell1[0, :, 0], cell1[0, :, 1], "blue");
plt.plot(cell2[0, :, 0], cell2[0, :, 1], "red");

cell_shapes = gs.concatenate([cell1, cell2])
print(cell_shapes.shape)

# Compute Frechet Mean

CURVES_SPACE = DiscreteCurves(R2)
SRV_METRIC = CURVES_SPACE.square_root_velocity_metric

mean = FrechetMean(metric=SRV_METRIC, point_type="matrix", method="adaptive")
mean.fit(cell_shapes)

mean_estimate = mean.estimate_
plt.figure()
plt.plot(mean_estimate[:, 0], mean_estimate[:, 1], "black");
```
returns:
```
(2, 5, 2)
/Users/ninamiolane/Google Drive/My Drive/code/geomstats/geomstats/geometry/discrete_curves.py:455: RuntimeWarning: divide by zero encountered in true_divide
  srv = gs.einsum("...i,...->...i", velocity, 1.0 / gs.sqrt(velocity_norm))
/Users/ninamiolane/Google Drive/My Drive/code/geomstats/geomstats/geometry/discrete_curves.py:586: RuntimeWarning: divide by zero encountered in true_divide
  "...ij,...i->...ij", velocity_vec, 1 / velocity_norm
/Users/ninamiolane/Google Drive/My Drive/code/geomstats/geomstats/geometry/discrete_curves.py:540: RuntimeWarning: divide by zero encountered in true_divide
  "...ij,...i->...ij", velocity_vec, 1 / velocity_norm
[...]

PR #1496, opened 2022-04-14
Add check in SRVMetric

alebrigant:
This PR adresses issue #1283.

Before computing the srv transform of a discrete curve, check that consecutive sample points are distinct are raise an error if not.


alebrigant:
@ninamiolane the curves in the notebook on cell shapes analysis need to be "pre processed" to remove duplicate consecutive points. Can you take a look ?

ninamiolane:
@alebrigant yes! we will take over this PR from now on. Many thanks!

ninamiolane:
This has been solved by #1525

PR #1501, opened 2022-04-18
Allow passing starting centroids to RiemannianKMeans

### LINKED ISSUE #1500: Allow passing initial centroids to RiemannianKMeans via init argument
The class `RiemannianKMeans` has an `init` argument but its value appears to be ignored and the only documented choice is `"random"`. It would be very helpful if initial centroids could be set via this argument. That is, the argument should accept an array of shape `(self.n_clusters, X.shape[1])` that is assigned to `self.centroids` at the start of `fit`.

Viech:
## Checklist

- [x] I have made sure the code passes all unit tests. (refer to comment below)
- [x] I have linked to issues and PRs that are relevant to this PR.

## Description 

Sometimes users have domain-specific knowledge on where to expect cluster centers. This MR lets them provide initial cluster centers to nudge k-means towards this guess. If the guess is good, this improves the convergence rate and produces a clustering matching the user's expectation.

**I updated the documentation but failed to compile it due to many unrelated errors.** I also had to run the relevant unit tests manually due to issues with the current version of the `werkzeug` library.

## Issue

This closes #1500.

### DIFF
--- a/geomstats/learning/kmeans.py
+++ b/geomstats/learning/kmeans.py
@@ -26,12 +26,15 @@ class RiemannianKMeans(TransformerMixin, ClusterMixin, BaseEstimator):
         Optional, default: 8.
     metric : object of class RiemannianMetric
         The geomstats Riemannian metric associate to the space used.
-    init : str or array-like, shape=[n_clusters, n_features]
+    init : str or callable or array-like, shape=[n_clusters, n_features]
         How to initialize centroids at the beginning of the algorithm. The
         choice 'random' will select training points as initial centroids
         uniformly at random. The choice 'kmeans++' selects centroids
-        heuristically to improve the convergence rate. By providing an array the
-        starting centroids can be specified externally.
+        heuristically to improve the convergence rate. When providing an array
+        of shape ``(n_clusters, n_features)``, the centroids are chosen as the
+        rows of that array. When providing a callable, it receives as arguments
+        the argument ``X`` to :meth:`fit` and the number of centroids
+        ``n_clusters`` and is expected to return an array as above.
         Optional, default: 'random'.
     tol : float
         Convergence factor. Convergence is achieved when the difference of mean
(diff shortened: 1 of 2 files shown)


Viech:
The test failures seem to be for `tests/tests_geomstats/test_discrete_curves.py`, which should be unrelated to this MR.

geomstats/learning/kmeans.py
```diff
@@ -103,24 +105,43 @@ def fit(self, X):
...
+                    "Dimensions of initial centroids and training data do not match."
+                )
+
             centroids = [
-                gs.expand_dims(X[randint(0, n_samples - 1)], 0)
-                for i in range(self.n_clusters)
+                gs.expand_dims(self.init[i], 0) for i in range(self.n_clusters)
             ]
+
         self.centroids = gs.concatenate(centroids, axis=0)
```
lpereira95, line 145: Here I think we can do:

```
self.centroids = gs.copy(centroids)
self.init_centroids = gs.copy(centroids)
```

And above, within the `if` branch that handles `str`, do: `centroids = gs.concatenate(centroids, axis=0)`.

So we simplify the new if branch by simply doing: `centroids = init` (no need for `gs.expand_dims`).

tests/tests_geomstats/test_riemannian_kmeans.py
```diff
@@ -83,3 +83,23 @@ def test_hypersphere_kmeans_initialization(self):
         result = centroids.shape
         expected = (n_clusters, 3)
         self.assertAllClose(expected, result)
+
+    def test_hypersphere_kmeans_init_array(self):
+        gs.random.seed(55)
+
+        n_features = 3
+        n_clusters = 3
```
lpereira95, line 91: Use a different number of clusters and features, otherwise the shape is `3x3`, which tells very little about the correctness of the result.

lpereira95 (changes requested):
Thanks for the nice contribution @Viech! I have small comments I would like you to address, but they're very small details.

@ninamiolane, this contributions helps to make this method closer to [`sklearn`'s](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html?highlight=kmeans#sklearn.cluster.KMeans). In terms of `init`, we are missing the option `callable`, but it should be trivial to add.

I have some follow-ups for this case, specially regarding testing. (I would also like to create a `_init_centroids` method - following sklearn -, that will simplify testing) If you agree with, I'll implement it after we merge this.

geomstats/learning/kmeans.py
Viech: Makes sense. I also implemented the `callable` option. Note that my approach requires Python >= 3.2 as I'm using the `callable()` built-in function.

tests/tests_geomstats/test_riemannian_kmeans.py
Viech: Sure. I also added a helper function to make it less redundant.

geomstats/learning/kmeans.py
lpereira95: Thanks! We assume Python >= 3.7, so it is perfectly fine.

geomstats/learning/kmeans.py
```diff
@@ -103,26 +108,50 @@ def fit(self, X):
...
+                raise ValueError("Need as many initial centroids as clusters.")
+
+            if centroids.shape[1] != X.shape[1]:
+                raise ValueError(
+                    "Dimensions of initial centroids and training data do not match."
+                )
+
+            centroids = gs.copy(centroids)
+
+        self.centroids = centroids
```
lpereira95, line 153: Here we need to do:

```
self.centroids = gs.copy(centroids)
self.init_centroids = gs.copy(centroids)
```

otherwise both `self.centroids` and `self.init_centroids` will be the same object (at least in numpy the arrays are mutable).

tests/tests_geomstats/test_riemannian_kmeans.py
```diff
@@ -66,20 +66,46 @@ def test_hypersphere_kmeans_predict(self):
         )
         self.assertAllClose(expected, result)
 
-    def test_hypersphere_kmeans_initialization(self):
-        gs.random.seed(55)
+    def _test_hypersphere_kmeans_init(
```
lpereira95, line 69: Nice!

lpereira95 (changes requested):
After changing the little detail about mutability of `self.centroids` and `self.init_centroids`, it is ready to go from my side.

Thanks for the last additions, they really complete this part of the code.

lpereira95 (approved):

PR #1504, opened 2022-04-19
end of binomial and exponential

Jules-Deschamps:


alebrigant:
@ninamiolane @nguigs I would like your opinion on this: with this PR, Jules is implementing 1-dimensional geometries. I have recommended that his implementation accepts floats as well as arrays for points. The idea is to accept `1.` as well as `gs.array(1.)` and `gs.array([1.])`. Currently, if `R1 = Euclidean(dim=1)`, `R1.belongs` returns `True` only for the last one. I think it would be nice to have this flexibility. To obtain it, Jules had to modify `belongs` in geomstats.geometry.base. So I want to check with you first: do you think it is a good idea ?

ninamiolane:
@alebrigant yes, that works on my side!

PR #1505, opened 2022-04-20
Automatize notebook identification for testing

luisfpereira:
This PR automatizes notebook identification for testing (everything within "notebooks" will be tested; note that if desirable, filters can be added or other folders selected).

Most of notebooks run with particular backends. This strategy relies on notebook metadata to identify with which backend to test the notebook. The following should be added: e.g. `"backends": ["numpy", "autograd"]`. This can be done both by opening the ipynb file with a text editor or in the "Edit" menu when the jupyter notebook is open.

Does this solution fit you @ninamiolane?

### DIFF
--- a/notebooks/10_practical_methods__shape_analysis.ipynb
+++ b/notebooks/10_practical_methods__shape_analysis.ipynb
@@ -802,9 +802,14 @@
   }
  ],
  "metadata": {
+  "backends": [
+   "numpy",
+   "autograd",
+   "pytorch"
+  ],
   "celltoolbar": "Tags",
   "kernelspec": {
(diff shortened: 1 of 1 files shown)


ninamiolane (approved):
LGTM if the tests pass, thank you! Much easier to test notebooks indeed.

Should we add indications (in the doc, or in the PR checklist) to the contributors mentioning that, if they are contributing a notebook, they should put their backends in the metadata?

lpereira95:
Yes, I think the contribution guide should contain this information. Probably we can tackle it when starting writing about how the tests are organized.

PR #1507, opened 2022-04-21
gamma updates

Jules-Deschamps:


alebrigant:
Hi @nguigs, we are having trouble with `Connection`. @Jules-Deschamps has implemented the Christoffel symbols for the geometry of Gamma distributions, and is trying to use the `exp`, `log` and `geodesic` functions of the Connection class. But we get an error whenever using `geodesic`, e.g. for
```from geomstats.information_geometry.gamma import GammaDistributions
space = GammaDistributions()
base_point = space.random_point()
end_point = space.random_point()
geod = space.metric.geodesic(initial_point=base_point, end_point=end_point)
geod(0.5)
```
Any easy fix for this error breaks everything else. I think this comes from the fact that `exp` is not vectorized in Connection. We will probably need to reimplement `exp` and `log` anyway, but it would still be important to fix this error. Any ideas ?

nguigs:
Hi Alice and Jules, sorry for seeing this so late. First, Connection.exp **is** vectorized, it is tested on the sphere. I think the problem comes from the fact that the time argument of the geodesic is expected to be an array and not just a scalar. Try `geod([.5])`.

alebrigant:
Hi @nguigs, @lpereira95  what I meant by Connection.exp is not vectorized, is that it does not work for one base point and several tangent vectors. This induces a bug in the Connection.geodesic, when we try to compute it at several times, because `connection.geodesic(initial_point=base_point, initial_tangent_vec=tangent_vec)` tries to compute `connection.exp(t*tangent_vec, base_point)` for different values of t, in a vectorized manner which is not supported by `connection.exp`. In fact, I also get a bug when trying to compute only one time. Here is a minimal failing example:
```
from geomstats.geometry.connection import Connection

dim = 2
sphere = Hypersphere(dim)
connection = Connection(dim)
base_point = gs.array([gs.pi/2, 0])
tangent_vec = gs.array([0.25, 0.5])
connection.christoffels = sphere.metric.christoffels
geod = connection.geodesic(initial_point=base_point, initial_tangent_vec=tangent_vec)
geod([0.5])
geod(gs.linspace(0., 1., 10))
```
From what I've seen, this is not detected in the tests because the `exp` called in `geodesic` is the one reimplemented in the spaces that are tested (e.g. Hypersphere) instead of `connection.exp`.

PR #1517, opened 2022-05-02
TestData inheritance and automatic tests

lpereira95:
@SaitejaUtpala, this is a trial PR where I start developing some ideas regarding the test data simplification.

Can you please take a look and let me know if I'm missing something? You can assume I have a solution to handle (at least temporarily) tolerances.

Update:
closes #1540.


SaitejaUtpala (commented):
Nice! This is good way of cleaning

lpereira95:
The last commit changes the way we handle tolerances in tests to make it easier to inherit test data from parents (as in a lot of cases we were constrained in inheriting by the need to control tolerances).

Now, if a test needs `atol` and `rtol`, the code looks for `atol`/`rtol` definition in the following order:

1. user defined test data
2. `tolerances` dict within `<func name>` key
3. `gs.atol`/`gs.rtol`

1. and 2./3. can be combined: e.g. in a test a user specifies the tolerance for a particular set of parameters (1.), but not for the others (2. or 3.).

The commit also solves another issue (hope it was an issue and not a design choice). The names used in the `dict` when creating data have to be in accordance with test argument names (for the previous implementation only order matters).

What do you think @SaitejaUtpala, @ninamiolane?

P.S. tests are failing because they need to be adapted for the current changes (I've ensured everything is working by correcting everything for symmetric matrices... now if you agree with the changes, I extend to the other tests).

SaitejaUtpala:
@ninamiolane  @lpereira95  currently `data_generation.py` is huge. I think we should split into multiple files

 Current implicit testing properties we added  are some simple properties  while prototyping with design.  @ninamiolane  and me were thinking  we should add even more advanced properties regarding curvature identities. with this files will become humongous and needlessly overwhelming for new contributors

so we should have following structure

* there will be single folder named `structural_tests`  (@ninamiolane  you can suggest just better name)
* into that there will multiple folders based core geometric testing classes. specifically one for each geometry it will contain corresponding test and data generation properties.   ex structural_tests/fiber_bundle/fiber_bundle_properties.py, structural_tests/fiber_bundle/fiber_bundle_data.py 

I will let @ninamiolane and you  choose naming conventions.

SaitejaUtpala:
@lpereira95   True. I agree that is the reason we didn't do it in the first phase. what happens after more geometries, more advanced tests in the file. It will be huge to even scroll. so we need to split based on some structure.

lpereira95:
@ninamiolane, as we discussed privately, I think we will benefit from merging this now, knowing that the PR with the ability to set the precision of the backend is coming soon.

Meanwhile, the tests in pytorch and tensorflow will fail more than usually, but we should expect the ones in `numpy` and `autograd` to have a similar behavior as previously.

ninamiolane:
LGTM as soon as Lint and DeepSource errors are solved!

PR #1521, opened 2022-05-03
Fix torch import in tests_backends

### LINKED ISSUE #1519: Fix the suite of test cases to be run for a specific backend

tests/tests_geomstats/test_backends.py:12

### What should be modified?

Ensure that the tests only for a specific backend mentioned are run and not the tests for other backend are also tested.

### Additional information

Screen Shot 2022-05-02 at 3 12 37 PM

Even after specifying an exact backend the test suite expected the developer to have installed Pytorch on their local machine.

lpereira95:
Closes #1519 (thanks for spotting this @shubhamtalbar96).

@ninamiolane I've just changed slightly the logic of the test by hardcoding the expected results instead of "collecting" results with `gs` and comparing it with `torch.`

### DIFF
--- a/tests/tests_geomstats/test_backends.py
+++ b/tests/tests_geomstats/test_backends.py
@@ -11,12 +11,12 @@ import pytest
 import scipy.linalg
 
 import geomstats.backend as gs
-import geomstats.tests
+import tests.conftest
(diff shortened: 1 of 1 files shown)


ninamiolane (approved):
Thank you so much!! 🙌

PR #1514, opened 2022-04-28
Simple new pullbackdiffeo

ymontmarin:
## Checklist

- [x] My pull request has a clear and explanatory title.
- [x] If neccessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [x] I have added apropriate unit tests.
- [x] I have made sure the code passes all unit tests. (refer to comment below)
- [x] My PR follows [PEP8](https://peps.python.org/pep-0008/) guidelines. (refer to comment below)
- [x] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#coding-style-guidelines) and API.
- [x] My code is properly documented and I made sure the documentation renders properly. ([Link](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#documentation))
- [x] I have linked to issues and PRs that are relevant to this PR.

PR #1529, opened 2022-05-06
vectorization of connection.exp, addition of n_steps as argument in geodesic

Jules-Deschamps:
We modify exp of connection to make it vectorized: it did not work with the use case (one base_point, n tangent_vectors) before. It failed to raise errors before because the exps returned in the test files were not that of connection but those of the test geometries.
Moreover, we add n_steps as a potential argument in geodesic, as it was =N_STEPS=(10?) before.

### DIFF
--- a/geomstats/geometry/connection.py
+++ b/geomstats/geometry/connection.py
@@ -126,7 +126,7 @@ class Connection(ABC):
             Point on the manifold.
         """
         base_point = gs.broadcast_to(base_point, tangent_vec.shape)
-        n_base_point, n_tangent_vec = base_point.shape[0], tangent_vec.shape[0]
+        """n_base_point, n_tangent_vec = base_point.shape[0], tangent_vec.shape[0]
 
         base_point = gs.squeeze(base_point)
         tangent_vec = gs.squeeze(tangent_vec)
@@ -138,7 +138,7 @@ class Connection(ABC):
                     " of tangent vectors."
                 )
             else:
-                base_point = gs.tile(base_point, (n_tangent_vec, 1))
+                base_point = gs.tile(base_point, (n_tangent_vec, 1))"""
 
         initial_state = gs.stack([base_point, tangent_vec])
         flow = integrate(
@@ -560,7 +560,7 @@ class Connection(ABC):
         )
 
     def geodesic(
-        self, initial_point, end_point=None, initial_tangent_vec=None, **kwargs
+        self, initial_point, end_point=None, initial_tangent_vec=None, **exp_kwargs
     ):
         """Generate parameterized function for the geodesic curve.
 
@@ -634,7 +634,7 @@ class Connection(ABC):
                 tangent_vecs = gs.einsum("i,...kl->...ikl", t, initial_tangent_vec)
 
             points_at_time_t = [
-                self.exp(tv, pt, **kwargs)
+                self.exp(tv, pt, **exp_kwargs)
                 for tv, pt in zip(tangent_vecs, initial_point)
             ]
             points_at_time_t = gs.stack(points_at_time_t, axis=0)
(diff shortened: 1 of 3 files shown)


geomstats/geometry/connection.py
```diff
@@ -125,6 +125,22 @@ def exp(
         exp : array-like, shape=[..., dim]
             Point on the manifold.
         """
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
```
lpereira95, line 128: Following a similar [implementation](https://github.com/geomstats/geomstats/pull/1514#discussion_r866068967) by @ymontmarin, I think all this new code can be replaced by:

```base_point = gs.broadcast_to(base_point, tangent_vec.shape)```

I think it accomplishes the goals we're looking for.

File: geomstats/geometry/connection.py [hunk 2]
```diff
@@ -544,7 +560,9 @@ def directional_curvature_derivative(
             tangent_vec_a, tangent_vec_b, tangent_vec_a, tangent_vec_b, base_point
         )
 
-    def geodesic(self, initial_point, end_point=None, initial_tangent_vec=None):
+    def geodesic(
+        self, initial_point, end_point=None, initial_tangent_vec=None, **kwargs
```
lpereira95, line 564: Good idea! Should we call them `exp_kwargs` to know where they'll be used?

tests/tests_geomstats/test_connection.py
```diff
@@ -118,6 +118,37 @@ def test_log_connection_metric(self, dim, point, base_point):
...
+        connection = Connection(dim)
+        connection.christoffels = sphere.metric.christoffels
+        geo = connection.geodesic(
+            initial_point=point, initial_tangent_vec=tangent_vec, n_steps=n_steps
+        )
+        times = gs.linspace(0, 1, n_times)
+        geo = geo(times)
+        result = geo.shape
+
+        self.assertAllClose(result, expected, atol=1e-6)
```
lpereira95, line 134: Can you please pass `atol` as an input argument? (This need will be more clear after merging #1517)

tests/tests_geomstats/test_connection.py
```diff
@@ -118,6 +118,37 @@ def test_log_connection_metric(self, dim, point, base_point):
...
+        connection = Connection(dim)
+        connection.christoffels = sphere.metric.christoffels
+        geo = connection.geodesic(
+            initial_point=point, end_point=end_point, n_steps=n_steps
+        )
+        times = gs.linspace(0, 1, n_times)
+        geo = geo(times)
+        result = geo.shape
+
+        self.assertAllClose(result, expected, atol=1e-6)
```
lpereira95, line 150: Ditto.

lpereira95 (changes requested):
Thanks @Jules-Deschamps!

Please address the suggestion in `exp`. I think we should really avoid using `gs.to_ndarray` there.

tests/tests_geomstats/test_connection.py
Jules-Deschamps: As an input argument of the test function? Only for the two tests you mentioned?

lpereira95: Thanks!

For now the idea is that new tests pass the `tol` when required because it will allow us to really check what tols will be required by using randomized data. (Later I'll have to review all the tests to see where tols are hard-coded)

geomstats/geometry/connection.py
lpereira95: Yes, I remember. I think this solution probably works there also.

geomstats/geometry/connection.py
```diff
@@ -125,6 +125,21 @@ def exp(
         exp : array-like, shape=[..., dim]
             Point on the manifold.
         """
+        base_point = gs.broadcast_to(base_point, tangent_vec.shape)
+        n_base_point, n_tangent_vec = base_point.shape[0], tangent_vec.shape[0]
```
lpereira95, line 129: I think we can now remove the code from this line till l142: this portion of code does not really do anything and the line 129 does not work properly for the case 1 base point, 1 tangent vec

PR #1525, opened 2022-05-04
SRV changes

shubhamtalbar96:
1. Put a check in SRV Transform method to ensure that there are no consecutive duplicate points in a curve. 
2. Made change in SRV Transform method to change the dimensionally of encoding ambient space if the Metric used is of instance type SRV Metric.
3. Introduced a new method predict_labels in Riemannian Mean Shift class to predict labels for the points passed as an argument.

PR #1530, opened 2022-05-06
Vectorize RiemannianMetric closest_neighbor_index

lpereira95:
This vectorizes `RiemannianMetric.closest_neighbor_index`.

Before we needed to do:

```
indices = []
for i in range(n_samples):
  index = metric.closest_neighbor_index(X[i], clusters)
  indices.append(index)
```

Now, it suffices:

```
indices = metric.closest_neighbor_index(X, clusters)
```

The solution uses a combination of `gs.tile`, `gs.concatenate` and some reshaping. It should be faster than the `for`, but more memory consuming.

This will simplify some code in #1525.


alebrigant (approved):
Great, thanks ! Everything looks good to me.

lpereira95:
Need to correct implementation for `tf`before merging.

PR #1536, opened 2022-05-12
Remove _space of Hypersphere and projection of tangent vector in exp

luisfpereira:
Removes unused `self._space` and does not project tangent vectors when computing the `exp` (for consistency with other spaces).


luisfpereira:
not relevant after @1836.

luisfpereira:
not relevant after @1836.

PR #1540, opened 2022-05-19
Fix test method inheritance

lpereira95:
Inheritance in test classes was not working properly: parent method was overriding child's, instead of the opposite.


lpereira95:
This is addressed here #1517.

PR #1552, opened 2022-06-01
Prototype numerical solvers for log, exp, geodesic, ...

luisfpereira:
When solving numerically `log` or `exp`, we are currently relying on `scipy` or "in-house" implementations of symplectic euler/rk, ..., respectively. For the `geodesic`, we simply compute `exp` for "differently-normed" initial tangent vectors.

Alternatively, we can solve a value problem for the geodesic and get `exp` and `log` from there (see [gamma](https://github.com/geomstats/geomstats/blob/master/geomstats/information_geometry/gamma.py)). This is, I believe, @alebrigant and @Jules-Deschamps preferred solution.

The solutions above, being so different, are tricky to implement with the current design of `Connection` (e.g. the parameters of the solvers are distinct: how should them be passed? in an `options` dict like scipy? it is possible, but dirty and very implicit...).

The proposed design circumvents the above problems by defining "solvers". The main advantage of this design is that "solver" arguments are defined at instantiation time, instead of being passed to `exp`, `log`, or `geodesic`. This greatly simplifies the flow of arguments in the code. It also provides a way changing the solver parameters for the learning algorithms (notice that currently e.g. `OnlineKMeans` computes `log` with default parameters and there's no way of changing them).

A small example on how the solvers can be changed with this design:

```
import os
os.environ['GEOMSTATS_BACKEND'] = 'autograd'

from geomstats.information_geometry.gamma import GammaDistributions
from geomstats.solvers import GammaGeodesicVPSolver

space = GammaDistributions()
metric = space.metric

# change exp n_steps
metric.exp_solver.n_steps = 100

base_point = space.random_point()
tangent_vec = space.random_point()

# all solved with current geomstats approach
end_point = metric.exp(tangent_vec, base_point)
metric.log(end_point, base_point)
geodesic = metric.geodesic(base_point, initial_tangent_vec=tangent_vec)

# change for VP solver
geodesic_vp_solver = GammaGeodesicVPSolver(n_steps=100)

metric.geodesic_solver = geodesic_vp_solver
metric.log_solver, metric.exp_solver = geodesic_vp_solver.get_log_exp_solvers()

# all solved with VP approach
end_point = metric.exp(tangent_vec, base_point)
metric.log(end_point, base_point)
geodesic = metric.geodesic(base_point, initial_tangent_vec=tangent_vec)
```

`GammaGeodesicVPSolver` is not working properly yet (and needs to be cleaned at a great extent). The idea would be to extend this solver to handle generic problems, but the implementation right now is particular to distributions.

This is just the prototyping of an idea. I'm eager to hear your opinions @alebrigant, @Jules-Deschamps, @ninamiolane, and @nguigs.

Edit:
Relates to #1224, which itself relates to #1118.


lpereira95:
Thanks for the feedback @Jules-Deschamps!

1. Exactly, I think not having to pass the args for `log`, `exp`, and `geodesic` is super important (specially when learning is considered).
2. No. They can be independent (as long as there's not circularity as you mention). For `GammaGeodesicVPSolver` I put `LogSolver` and `ExpSolver` within because they're completely dependent on the geodesic solver and I didn't want to allow them to be instantiated independently. Nothing forbids the user to set any other `exp` and `log` solvers, even using the current `GammaGeodesicVPSolver`. It is a super flexible solution.
3. Yes, we can change objective function and so on. The solvers should inherit, at a first level, from the abstract classes I've created, because they define the "blueprints" of a solver (what methods should be defined). Nevertheless, we can then work on a hierarchy of solvers that has much more flexibility than the current ones (e.g. to pass a different objective function). This solution will allow us to play with the solvers without changing at all `Connection` code, which in my opinion is very important in order to avoid introducing bugs.

For the unrelated:
1. I've thought about the solution proposed by @alebrigant. In fact, she's proposing a different integrator, because the current one has fixed time step, which means that we may not have exactly the "state" (solution for a given t) the user is asking for. I guess a solution should consist in allowing the integrator "dt" to change in order to pass by all asked `t`s?
2. Where? If in `exp`, it will not work, because the norm is something you're optimizing for.

ninamiolane:
Thank you, I think this is an excellent idea, and a great design! 🎉 

Note: It reminds me of Pymanopt's "optimizers", see their abstract class [here](https://github.com/pymanopt/pymanopt/blob/851e37f7b5ecd22a2447d70b2adbc85f3140b198/pymanopt/optimizers/optimizer.py#L23) and an example code [here](https://pymanopt.org/docs/stable/quickstart.html) --- which might give additional design ideas.

Question: Do the solvers really depend on our geometric objects (e.g. a geodesic equation), i.e. on the fact that the equation that we are trying to solve is a geodesic equation? If not, solvers (the way we solve an equation) are disentangled from which equation is getting solved and in this case, they should have names (class names) that reflect what they do internally, and *not* how we will use them.

Jules-Deschamps:
(just to answer to Luís because I don't know how to comment his answer directly)

Ok thanks for your answers!

As for the unrelated:
- Don't you think it would be more efficient to rewrite `geodesic(initial_point, initial_tangent_vec)`? Because at the moment it is calling `Connection.exp` for every step of the way, and therefore it is using `integrate, for every of those steps` when we could just integrate the equation once and, as in `Connection.exp`, call `flow` at the steps that are of interest to us. I think it might be less complex computationally.
- It was simply in reference to the instantiation of your `tangent_vec`: initializing `tangent_vec` as a `random_point` for a given `base_point` may give a norm of 10000 for the tangent vector and then `exp` can become far less precise or even raise errors: a minimal reproducible example is

Jules-Deschamps:
(just to answer to Luís because I don't know how to comment his answer directly)

Ok thanks for your answers!

As for the unrelated:
- Don't you think it would be more efficient to rewrite `geodesic(initial_point, initial_tangent_vec)`? Because at the moment it is calling `Connection.exp` for every step of the way, and therefore it is using `integrate`, for every of those steps when we could just integrate the equation once and, as in `Connection.exp`, call `flow` at the steps that are of interest to us. I think it might be less complex computationally.
- It was simply in reference to the instantiation of your `tangent_vec`: initializing `tangent_vec` as a `random_point` for a given `base_point` may give a huge norm for the tangent vector and then `exp` can become far less precise or even raise errors: a minimal reproducible example is:
`from geomstats.information_geometry.gamma import *

space = GammaDistributions()

base_point = gs.array([1.0,.01])
vec = gs.array([0.0, 5.0])

print(space.metric.norm(vec, base_point))
print(space.metric.exp(vec, base_point))` (computation of christoffels can explode)

Jules-Deschamps:
(just to answer to Luís because I don't know how to comment his answer directly)

Ok thanks for your answers!

As for the unrelated:
- Don't you think it would be more efficient to rewrite `geodesic(initial_point, initial_tangent_vec)`? Because at the moment it is calling `Connection.exp` for every step of the way, and therefore it is using `integrate`, for every of those steps when we could just integrate the equation once and, as in `Connection.exp`, call `flow` at the steps that are of interest to us. I think it might be less complex computationally.
- It was simply in reference to the instantiation of your `tangent_vec`: initializing `tangent_vec` as a `random_point` for a given `base_point` may give a huge norm for the tangent vector and then `exp` can become far less precise or even raise errors: a minimal reproducible example is:
`from geomstats.information_geometry.gamma import *

space = GammaDistributions()

base_point = gs.array([1.0,.01])
vec = gs.array([0.0, 5.0])

print(space.metric.norm(vec, base_point))
print(space.metric.exp(vec, base_point))` (computation of christoffels can explode)

nguigs:
Hi all,

Thank you Luis for the proposal, here are my thoughts after reading the above discussion.
- I was surprised by the name `GammaGeodesicVPSolver` because solvers should be general and take the geometry as input (and then use its `geodesic_equation` method. It seems you have taken this into considerations and will change the solver in that direction?
- About Jules suggestion on `random_tangent_vecs`, I don't think this is a good idea because tangent vectors need not a metric to be defined. I agree that the norm can be a problem, but there are other ways to tackle it:
   - have a keyword option `normalize=True` that would use the `Manifold.metric` attribute. The `injectivity_radius` method can help.
   - Ensure that the sampled vector is not big for one metric, and as they are all equivalent this will be enough to ensure they are not too big for most other metrics .
- I agree with Nina's comment on naming.

geomstats/solvers.py
```diff
@@ -0,0 +1,723 @@
...
+        Returns
+        -------
+        tangent_vec : array-like, shape=[..., dim]
+            Tangent vector at the base point.
+        """
+        max_shape = point.shape
+        if len(point.shape) <= len(base_point.shape):
+            max_shape = base_point.shape
+
+        def objective(velocity):
```
nguigs, line 61: This could be a method of the solver class. This would answer @Jules-Deschamps's which to enable changing the objective.

geomstats/solvers.py
```diff
@@ -0,0 +1,723 @@
...
+            """Define the objective function."""
+            velocity = gs.array(velocity, dtype=base_point.dtype)
+            velocity = gs.reshape(velocity, max_shape)
+            delta = metric.exp(velocity, base_point) - point
+            return gs.sum(delta**2)
+
+        objective_with_grad = gs.autodiff.value_and_grad(objective, to_numpy=True)
+
+        tangent_vec = gs.flatten(gs.random.rand(*max_shape))
```
nguigs, line 70: There should also be an `initialization` method.

alebrigant:
Hi everyone,

Thanks Luis for this great start ! I think it's really important that we can control in a simple fashion the way exp, log and geodesics are computed in all the functions that call them.

My comments: I believe that the solvers defined in `GammaGeodesicVPSolver` are general (with the exception of the jacobian of the cost function in `_approx_geodesic_bvp`, but I am sure there is a simple solution to this). It should be possible to define them for any geometry. In fact, could we define general solvers for solving IVP and BVP of ODEs (this addresses @ninamiolane's comment about names), and use them in exp, log and geodesic when exp and log do not have closed form expressions ? This would give the following general structure:
```
class ODEIVPSolver
	def __init__(...):
		self.integrator = geomstats.integrator.integrate or scipy.odeint

class ODEBVPSolver
	def __init__(...):
		self.integrator = scipy.solve_bvp 
		self.initialization = ...

class LogODESolver
	def solve(...):
		call ODEBVPSolver
		keep only the difference between first 2 points.

class LogShootingSolver

class ExpODESolver
	def solve(...):
		call ODEIVPSolver
		keep only the last point.

class GeodesicODESolver
	def solve(...):
		if end_point is None:
			determine step size according to times given in argument 
			(smallest common divisor of all consecutive time differences)
			call ODEIVPSolver 
			keep only the points corresponding to the times given in argument.

		if initial_tangent_vec is None:
			determine step size according to times given in argument
			call ODEBVPSolver 
			keep only the points corresponding to the times given in argument.
```
and then the current way of doing computations in `Connection` could be retrieved with
```
class Connection

def _set_default_solvers(self):
	self.exp_solver = ExpODESolver()
	self.log_solver = LogShootingSolver()
	self.geodesic_solver = None

def exp(...)
	return self.exp_solver.solve(...)

def log(...)
	return self.log_solver.solve(...)

def geodesic(...):
	if self.geodesic_solver is None:
		# When exp and log have closed form expressions
		call exp and log

	else:
		# When exp and log don't have closed form expressions
		call GeodesicODESolver
```

lpereira95:
Thanks @ninamiolane, @nguigs and @alebrigant for the feedback.

@alebrigant your suggestion is pretty elegant and in fact addresses @ninamiolane's comments and a later discussion I had with @nguigs. What's curious is that you suggest a layer between ODESolvers and log and exp solvers, which was the direction I was converging with @nguigs. We also thought about following the same strategy for parallel transport (for a user, it does not matter if parallel transport is computed by ladder methods or any other methodology -> it only matters it provides accurate results).

I'll bring your feedback to the code and we iterate again on it later.

PR #1568, opened 2022-06-07
Sasaki metric

ambellan:
## Checklist

- [x] I have made sure the code passes all unit tests. (refer to comment below)
- [x] My code is properly documented ~~and I made sure the documentation renders properly.~~ ([Link](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#documentation)) (Code is documented, but I cannot get _sphinx-build docs/ docs/html_ running, as it fails immediately.)

## Description
A natural choice of metric on the tangent bundle of a Riemannian manifold is the Sasaki metric and this PR implements it.

## Additional context

This implementation was originally part of the submission 'Sasaki Metric and Applications in Geodesic Analysis' to the ICLR Computational Geometry & Topology Challenge 2022.


ninamiolane:
Very nice, thank you! 

May I let you address the lint errors in the Github workflows:
- Linting
- Deep Source
or let us know if you cann't/shouldn't address these.

I will let @lpereira95 comment on the tests, and use of pytorch/tensorflow backends.

ambellan:
I fixed the linting/deepsource issues (Originally I kept the --ignoreD flag for flake8; that introduced errors I did not notice in advance). However, it is a bit tricky that DeepSource somehow spots similar but different problems as flake8 (i.e. ). Can DeepSource somehow be run before initiating a PR? The tests for the tf and torch backends are failing because right now the implementation only supports np and autograd. However, I used flags to enable testing for np/autograd only and I have no idea why the tf and torch test are still carried out. 
Cheerio

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+            for j in range(n_s - 1):
+                p1, _ = pu[j][0], pu[j][1]
+                p2, u2 = pu[j + 1][0], pu[j + 1][1]
+                p3, u3 = pu[j + 2][0], pu[j + 2][1]
+                v, w = metric.log(p3, p2), par_trans(u3, p3, None, p2) - u2
+                gp = metric.log(p3, p2) + metric.log(p1, p2) + metric.curvature(
+                    u2, w, v, p2)
+                gu = par_trans(u3, p3, None, p2) - 2 * u2 + par_trans(u0, p0, None, p2)
+                g.append([gp, gu])
+            return -n_s * gs.array(g)
```
lpereira95, line 196: This can be vectorized:

```
def grad(pu):
  pu = gs.vstack([gs.expand_dims(initial_point, axis=0),
            pu_ini,
            gs.expand_dims(end_point, axis=0)])

  p = gs.take(pu, 0, axis=1)
  u = gs.take(pu, 1, axis-1)

  p1, p2, p3 = p[:-2], p[1:-1], p[2:]
  u2, u3 = u[1:-1], u[2:]

  v = metric.log(p3, p2)
  w = par_trans(u3, p3, end_point=p2) - u2

  gp = metric.log(p3, p2) + metric.log(p1, p2) + metric.curvature(
      u2, w, v, p2)
  
  gu = par_trans(u3, p3, end_point=p2) - 2 * u2 + par_trans(u0, p0, end_point=p2)

  return -n_s*gs.stack([gp, gu], axis=1)
```

The expressions are slightly different from the ones in the paper. Do you have any reference with these expressions?

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+    L = len(x_ini)
+    x = x_ini
+    for _ in range(max_iter):
+        grad_x = grad(x)
+        grad_norm = gs.linalg.norm(grad_x)
+        if grad_norm < tol:
+            break
+        grad_x = -lrate * grad_x
+        for j in range(L):
+            x[j] = exp(grad_x[j], x[j])
```
lpereira95, line 26: Can be vectorized: `self.exp = exp(grad_x, x)`.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+            p = metric.exp(eps * v0, p0)
+            u = par_trans(u0 + eps * w0, p0, None, p)
+            v = par_trans(v0 - eps * (metric.curvature(u0, w0, v0, p0)), p0, None, p)
+            w = par_trans(w0, p0, None, p)
+            p0, u0 = p, u
+            v0, w0 = v, w
+
+        return gs.reshape(gs.array(list(zip(p, u))), base_point.shape)
+
+    def log(self, point, base_point, n_steps=None, **kwargs):
```
lpereira95, line 109: I have some more comments on this, which I'll address after the changes in `geodesic_discrete`.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Returns
+        -------
+        inner_product : array-like, shape=[..., 1]
+            Inner-product.
+        """
+        # unflatten
+        vec_a = gs.reshape(tangent_vec_a, (-1, 2) + self.metric.shape)
+        vec_b = gs.reshape(tangent_vec_b, (-1, 2) + self.metric.shape)
+        pt = gs.reshape(base_point, (-1, 2) + self.metric.shape)
```
lpereira95, line 233: `gs.reshape` can be avoided by coding as above.

lpereira95:
@ambellan, for deep source I don't know if it is possible to run locally (I couldn't find a way by going through their website). Do you know if it is possible @ninamiolane?

About the docs, can you provide the error `sphinx-build` raises? It should work...

For the tests, the best is to use the flag `skip_all` as it is done [here](https://github.com/geomstats/geomstats/blob/6e169c3fdee5b17135063816a48722124ddd0eb9/tests/tests_geomstats/test_spider.py#L20) (sorry for the lack of documentation). Nevertheless, I would prefer to have it working in all the backends.

ninamiolane:
I don't know if DeepSource can be run locally, they do have a CLI https://deepsource.io/docs/cli/usage/#commands but it seems to do something different. :/

Yet, note that DeepSource's GitHub workflow usually completes before all of the others, taking ~2min to complete, thus it is not too too long.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+
+    References
+    ----------
+    .. [1] Muralidharan, P., & Fletcher, P. T. (2012, June).
+    Sasaki metrics for analysis of longitudinal data on manifolds.
+    In 2012 IEEE conference on computer vision and pattern recognition (pp. 1027-1034).
+    https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4270017/
+    """
+
+    def __init__(self, metric: RiemannianMetric, n_s=3):
```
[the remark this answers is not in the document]
ambellan: Ok, I adopt what I found in gamma.py

ambellan:
@lpereira95 regarding sphinx: I got the following errors:
`Extension error:
Could not import extension nbsphinx (exception: No module named 'nbsphinx')`
`Extension error:
Could not import extension nbsphinx_link (exception: No module named 'nbsphinx_link')`
`Extension error:
Could not import extension sphinx_gallery.load_style (exception: No module named 'sphinx_gallery')`
`Theme error:
no theme named 'pydata_sphinx_theme' found (missing theme.conf?)`
After installing the missing packages I was able to proceed, however, running `sphinx-build docs/ docs/html` carried out in the geomstats  project folder did build a documentation for the geomstats package intalled in the respective python environment. I additonally had to make shure that the right geomstats structure is seen first. I did this by `export PYTHONPATH="${PYTHONPATH}:<...>/projects/07geomstats/geomstats"`. After doing so sphinx did build the correct docs,  however, I could not find anything relating to the added Sasaki metric. Is there anything that needs to added to the sphinx configuration?

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+        self.metric = metric  # Riemannian metric of underlying space
+        self.n_s = n_s  # Number of discretization steps
+        shape = (2, gs.prod(metric.shape))
```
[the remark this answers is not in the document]
ambellan: Well, this would introduce a difference if `metric.shape` consists of more than one element. I am not shure whether this should be changed as it my introduce problems.  Ok, I just checked back with @vontycowicz and he confirmed that `(2, *metric.shape)` did not work together with the Frechet-mean in Kendall shape space. For the moment we would like to keep it as is, unrolling `metric.shape`. This may be changed later on. 

Replacing `(2, gs.prod(metric.shape)` by `(2, gs.prod(gs.array(metric.shape))` already allows to successfully pass the unit tests employing the pytorch backend.

geomstats/geometry/sasaki_metric.py
ambellan: Ok

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+import os
+from joblib import Parallel, delayed
+
+import geomstats.backend as gs
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+
+
+def _gradient_descent(x_ini, grad, exp, lrate=0.1, max_iter=100, tol=1e-6):
+    """Apply a gradient descent until max_iter or a given tolerance is reached."""
+    L = len(x_ini)
```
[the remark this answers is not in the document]
ambellan: Ok

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+    x = x_ini
+    for _ in range(max_iter):
+        grad_x = grad(x)
+        grad_norm = gs.linalg.norm(grad_x)
+        if grad_norm < tol:
+            break
+        grad_x = -lrate * grad_x
+        for j in range(L):
+            x[j] = exp(grad_x[j], x[j])
+    return list(x)
```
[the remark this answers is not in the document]
ambellan: Ok

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+            """Calculate the logarithm."""
+            pu = self.geodesic_discrete(bs_pt, pt, n_s)
+            p1, u1 = pu[1][0], pu[1][1]
+            p0, u0 = bs_pt[0], bs_pt[1]
+            w = (par_trans(u1, p1, None, p0) - u0)
+            v = metric.log(point=p1, base_point=p0)
+            return n_s * gs.array([v, w])
+
+        n_jobs = min(os.cpu_count(), len(pts))
+        with Parallel(n_jobs=n_jobs, verbose=0) as parallel:
```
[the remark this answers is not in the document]
ambellan: We think the is the right place for parallelization as the loop here is potentially way larger (hundreds/thousands) as they depend on the size of your dataset, if you e.g. consider statistical operations. Whereas in the `geodesic_discrete` we have a default n_steps of 3.

geomstats/geometry/sasaki_metric.py
ambellan: The formulas used here can all be juggled out of the ones in the paper by using definitions and basic properties.  However, and here we are really thankful for pointing on this, there was introduced some mistake when porting the code form Matlab to Python.  I checked back with Esfandiar and @martinhanik and put in some corrections.

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
ambellan: Ah, ok, now I got your point. You basically suggest to wrapper the whole `geodesic_discrete`  s.t. it becomes vectorized through an interal parallel loop.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Initial guess for gradient_descent
+        v = metric.log(pL, p0)
+        s = gs.linspace(0., 1., n_s + 1)
+        pu_ini = []
+        for i in range(1, n_s):
+            p_ini = metric.exp(s[i] * v, p0)
+            u_ini = (1 - s[i]) * par_trans(u0, p0, None, p_ini) + s[i] * par_trans(
+                uL, pL, None, p_ini)
+            pu_ini.append(gs.array([p_ini, u_ini]))
```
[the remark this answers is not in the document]
ambellan: _scalar_vec_mult does raise an error on return in Parallel when used with the Kendall example from the challenge notebook:
`ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (2,)->(2,newaxis) (8,2)->(8,2)`. I will ignore it for now.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Minimization by gradient descent
+        x = _gradient_descent(pu_ini, grad, self.exp)
+
+        return [initial_point] + x + [end_point]
```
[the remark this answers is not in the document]
ambellan: Ok

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+            Optional, default: self.n_s.
+
+        Returns
+        -------
+        exp : array-like, shape=[..., 2, M.dim]
+            Point on the tangent bundle TM.
+        """
+        # unflatten
+        bs_pts = gs.reshape(base_point, (-1, 2) + self.metric.shape)
+        tngs = gs.reshape(tangent_vec, bs_pts.shape)
```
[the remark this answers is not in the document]
ambellan: Is reshaping such a big problem? As I understand it just changes attributes and does not touch the memory. For the moment I would want to keep the reshaping as it is, simply because I know it works for all relevant cases. Or is there another technical reason I am missing?

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,238 @@
...
+
+        metric = self.metric
+        par_trans = metric.parallel_transport
+        n_s = self.n_s if n_steps is None else n_steps
+        eps = 1 / n_s
+
+        v0, w0 = tngs[:, 0], tngs[:, 1]
+        p0, u0 = bs_pts[:, 0], bs_pts[:, 1]
```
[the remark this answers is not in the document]
ambellan: Hm, here we would loose what we could have gotten omiting reshaping.

ambellan:
Dear @lpereira95, for now I incorporated as many of your suggestions as to me seem to be reasonable/feasible. Please take a look at my changes and comments. Thank you for all your valuable remarks, I think they are really helpful to further develop the code. However, there is one thing I don't understand: Why don't you accept any additional comments? I think they usually simply clarify what happens in the code.

ninamiolane:
I can give my view on this:

Good code is self-documenting. A function should do exactly one thing, and that one thing should be obvious by the functions name, the names of its parameters, its local variable names and, of course, its docstring. Needing comments in the code explaining the logic suggests that (i) e.g. variables are poorly named, or (ii) the function should be broken up into single responsibility sub-functions with docstrings.

1. Allowing inline comments makes contributors become more lazy, as they allow themselves to write less understandable code because they rely on the comments.

2. Comments are not tested by unit-tests, and often get outdated: having a comment explain a line of code that is not there anymore is very detrimental.

3. Comments do not appear in the docstrings, thus they do not show up on the doc website.

That being said, there can be exceptions, and we do have some inline comments in the code base :)

ninamiolane:
Many thanks for this 🙌 !! @lpereira95 what do you think?

As for the lint errors, these can be fixed from your command line via:
```
pip install black
black .
```

Very excited to merge this 🎉

lpereira95:
@ninamiolane I'll review it again very soon (I'm really eager to see the changes done by @ambellan).

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
lpereira95: You're write about reshaping. My alternative would be:

```
base_point = gs.broadcast_to(base_point, tangent_vec.shape)
```

In my opinion this solution is superior because:
* `gs.reshape` requires `self.metric.shape`. broadcast simply requires the arrays that are being manipulated.
* every time reshape is used we need to think "why am I reshaping?", "is this a proper reshaping?". whereas `gs.broadcast_arrays` clearly indicates the goal is to (properly) create arrays with the same shape.
* it is easier to manage the shape of the output (normally nothing needs to be done, whereas with reshape we need to update the shape).
* it is more compact.

A negative point is that the code will need to assume that we will handle one or several points, which sometimes may be tricky if we are not used to think in terms of vectorization.

Anyway, I'm OK with the solution you have.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,236 @@
...
+
+        metric = self.metric
+        par_trans = metric.parallel_transport
+        eps = 1 / n_steps
+
+        v0, w0 = gs.take(tngs, 0, axis=1), gs.take(tngs, 1, axis=1)
+        p0, u0 = gs.take(bs_pts, 0, axis=1), gs.take(bs_pts, 1, axis=1)
+        for _ in range(n_steps):
+            p = metric.exp(eps * v0, p0)
+            u = par_trans(u0 + eps * w0, p0, None, p)
```
lpereira95, line 94: for clarity, can you change `par_trans(u0 + eps * w0, p0, None, p)` to `par_trans(u0 + eps * w0, p0, end_point=p)` in all the places we use it? I find it more readable.

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
lpereira95: My main idea here is exactly what you refer: having a parallelized internal loop. Why?
* the parallelism is placed exactly where it belongs, i.e. in the optimization problem for `geodesic_discrete`.
* `geodesic_discrete` starts accepting multiple `end_point` and `initial_point`. So if for some reason we want to compute several discrete geodesics, then we know the parallelization is taken care of for us.
* we can vectorize operations in `log` after getting the discrete geodesics (e.g. parallel transport and metric.log), which may be much faster if we have tons of points.

Why do you think it may be a problem for later polishing? btw, we are thinking about a change in code architecture (#1152) that will play a role here.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,236 @@
...
+        -------
+        inner_product : array-like, shape=[..., 1]
+            Inner-product.
+        """
+        vec_a = gs.reshape(tangent_vec_a, (-1, 2) + self.metric.shape)
+        vec_b = gs.reshape(tangent_vec_b, (-1, 2) + self.metric.shape)
+        pt = gs.reshape(base_point, (-1, 2) + self.metric.shape)
+
+        inner = self.metric.inner_product
+        return inner(vec_a[:, 0], vec_b[:, 0], pt[:, 0]) + inner(vec_a[:, 1],
```
lpereira95, line 235: We need to reshape back here, for the case of only 1 point.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,236 @@
...
+            return n_steps * gs.array([v, w])
+
+        n_jobs = min(os.cpu_count(), len(pts))
+        with Parallel(n_jobs=n_jobs, verbose=0) as parallel:
+            rslt = parallel(_log(pt, bs_pts[i % len(bs_pts)])
+                            for i, pt in enumerate(pts))
+
+        return gs.reshape(gs.array(rslt), point.shape)
+
+    def geodesic_discrete(self, initial_point, end_point, n_steps=N_STEPS, **kwargs):
```
lpereira95, line 147: To bring the parallelization here, we can simply make this method private and create a public one with the same name that handles  the parallelization and calls this method.

tests/tests_geomstats/test_sasaki_metric.py
```diff
@@ -0,0 +1,33 @@
+"""Unit tests for the Sasaki metric."""
+
+import geomstats.tests
+from tests.conftest import Parametrizer
+from tests.data.sasaki_metric_data import SasakiMetricTestData
+from tests.geometry_test_cases import TestCase
+
+
+class TestSasakiMetric(TestCase, metaclass=Parametrizer):
```
lpereira95, line 9: Could you add a test specifically for the `grad` (since we were discussing the rightness of the formulas). I think the `geodesic_discrete` test is not enough to validate the gradient because if I'm not mistaken it evaluates to zero there.

lpereira95 (changes requested):
Thanks for the code changes @ambellan. We are almost there!

To finish:
* I think we should move the place of the parallelization (if you disagree let me know why).
* there's a little inconsistency in the output shape of `inner_product` that needs to be corrected
* a (robust) test should be added to `grad`.

Thanks for your patience!

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
lpereira95: Do you know why the unrolling hasn't worked with Frechet mean in Kendal shape space? Can you share a small reproducible example of the error?

The goal of unrolling would be that instead of reshaping to `(-1, 2) + self.metric.shape` later in the code, we could do `(-1,) + self.shape`. It would be more robust as we could control reshaping shape from one place only.

geomstats/geometry/sasaki_metric.py
lpereira95: Do you have any notes with such manipulations so we can point our users to them in the documentation?

ninamiolane:
Thanks for this great PR. Any update on addressing the remaining comments? Thanks!

geomstats/geometry/sasaki_metric.py
ambellan: We created a little tech-report with detailed explainations (https://nbn-resolving.org/urn/resolver.pl?urn:nbn:de:0297-zib-87174) and reference it now in the docsting of the sasaki class.

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
ambellan: @lpereira95, When taking `(2, *metric.shape)` instead of  `(2, gs.prod(metric.shape)` problems in the FrechetMean class at different locations, here with two examples:

1) In the constructor:
`        if point_type is None:
            self.point_type = metric.default_point_type
        error.check_parameter_accepted_values(
            self.point_type, "point_type", ["vector", "matrix"]
        )`
We would have something like 3-tensor as point_type and that is not handled.

2) In the _default_gradient_descent
`if point_type == "vector":
        points = gs.to_ndarray(points, to_ndim=2)
        einsum_str = "n,nj->j"
    else:
        points = gs.to_ndarray(points, to_ndim=3)
        einsum_str = "n,nij->ij"`
There is no handling for point_types different from vector and matrix. 

Apart form that, vectorizing the shape shape of the base manifold has the advantage that the iteration of taking the sasaki_metrix (i.e. looking at the tangent bundle of the tangent bundle) does not introduce new dimensions to the shape.

geomstats/geometry/sasaki_metric.py
ambellan: Done

tests/tests_geomstats/test_sasaki_metric.py
ambellan: Good catch! I added some more data, s.t. the geodesic_discrete test now actually performs some steps, i.e. the gradient is not zero right away.

ambellan:
General note: Apart form `black `I also had to install and run `isort `in order to pass the linting test. Maybe this could be added to the guidelines.

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
ambellan: Ok, so we will keep the reshaping for the moment. And you (or anybody else) are/is free to refactor it in the future.

geomstats/geometry/sasaki_metric.py
ambellan: Done

geomstats/geometry/sasaki_metric.py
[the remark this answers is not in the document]
ambellan: > Why do you think it may be a problem for later polishing?

I think my sentence was a little misleading and I don't think this is an issue for later polishing. Instead I meant that it is something that could be done later when performing code polishing. However, I will put in the requested changes.

ambellan:
@lpereira95 and @ninamiolane,  I think I incorporated all requested changes. Please have a look and iterate/accept the PR.

ninamiolane:
@ambellan Many thanks!! There is an issue with tensorflow, see here: https://github.com/geomstats/geomstats/runs/7131533833?check_suite_focus=true

Would you know what is happening? Good to merge on my side otherwise!

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,277 @@
...
+    Technical Report Zuse-Institut Berlin (no. ZIB-22-12) (urn:nbn:de:0297-zib-87174).
+    https://nbn-resolving.org/urn/resolver.pl?urn:nbn:de:0297-zib-87174
+    """
+
+    def __init__(self, metric: RiemannianMetric):
+        self.metric = metric
+        shape = (2, gs.prod(gs.array(metric.shape)))
+
+        super(SasakiMetric, self).__init__(
+            2 * metric.dim, shape=shape, default_point_type="matrix"
```
lpereira95, line 69: `default_point_type` should be `self.metric.default_point_type`. 

I know this will raise problems in Frechet mean. I'lll open a PR to solve that issues there (still today). I think it is the same problem we've been discussing with shape.

lpereira95 (changes requested):
Thanks a lot for addressing the requested changes @ambellan! The code is in great shape now. I'm also very pleased by your effort in creating the gradient document.

Can you please address these last comments?

(If you feel we are iterating too much, let me know and I'll address them myself)

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,277 @@
...
+
+Lead authors: E. Nava-Yazdani, F. Ambellan, M. Hanik and C. von Tycowicz.
+"""
+import os
+
+from joblib import Parallel, delayed
+
+import geomstats.backend as gs
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.tests import tf_backend
```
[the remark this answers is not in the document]
ambellan: Done

geomstats/geometry/sasaki_metric.py
ambellan: @lpereira95, well I think `default_point_type=self.metric.default_point_type` is not a good idea as it potentially would introduce wrong configurations. If we e.g. take an input metric with `self.metric.default_point_type="vector"`the SasakiMetric will have 'shape = (2, gs.prod(gs.array(metric.shape)))' that, as I understand, suggests that `default_point_type="matrix"`. In fact, in the discussion about the shape (above) I wrote (to argue why 'shape = (2, gs.prod(gs.array(metric.shape)))' is a good idea):

> ... vectorizing the shape shape of the base manifold has the advantage that the iteration of taking the sasaki_metrix (i.e. looking at the tangent bundle of the tangent bundle) does not introduce new dimensions to the shape.

This essentially says that we always get the same `default_point_type` and that is "matrix".

Long story short: I would prefer keeping the current code here, as long as SasakiMetric vectorizes its input.

ambellan:
@ninamiolane and @lpereira95, thank your for all your valuable feedback. I think I handled all comments raised so far. Please let me know if there is something more left open, or else please accept the PR.

geomstats/geometry/sasaki_metric.py
```diff
@@ -0,0 +1,279 @@
...
+    On Gradient Formulas in an Algorithm for the Logarithm of the Sasaki Metric
+    Technical Report Zuse-Institut Berlin (no. ZIB-22-12) (urn:nbn:de:0297-zib-87174).
+    https://nbn-resolving.org/urn/resolver.pl?urn:nbn:de:0297-zib-87174
+    """
+
+    def __init__(self, metric: RiemannianMetric):
+        self.metric = metric
+        shape = (2, gs.prod(gs.array(metric.shape)))
+
+        self.n_jobs = (
```
lpereira95, line 67: Pass `n_jobs` as input with default value 1 (for consistency with other uses of `joblib` in the library).

lpereira95:
Thanks @ambellan. Please address the last comment from my side.

Also, can you please try out `FrechetMean` again, after merging the changes from master? I've made some changes to avoid the use of `default_point_type` there. I just want to ensure `shape` works fine for `SasakiMetric` (do you have a minimal test to add to Frechet mean tests?).

ambellan:
@lpereira95, I tested `FrechetMean` for `SasakiMetric` on Kendall shape space and it worked as expected. However, I guess the test case (the one from the ICLR challenge) is, at least in its current form, not suited as minimal working example. As we also planned to contribute our challenge notebook as example on how to use `SasakiMetric` to Geomstats the respective PR might be the right place to also set up a slimmed version as additional test for `FrechetMean`. What do you think?

lpereira95:
Thanks a lot @ambellan!

I agree with the idea of addressing the additional test in another PR. Looking forward to it!

ambellan:
@ninamiolane and @lpereira95, thank you both again for handling the PR! I learned a lot about Life, the Universe, and Everything (well, and of course about Geomstats and good code):grinning:.

PR #1569, opened 2022-06-08
Real world application: optimization of traffic in Sao Paulo

Jules-Deschamps:
This notebook presents a simple use case of *information geometry*, in the context of *traffic optimization* in Sao Paulo.
We rely on a dataset listing all traffic jams in Sao Paulo for the past two decades to propose a solution as for what roads to renovate.


Jules-Deschamps:
Ok thanks! As for the unit vector, I thought it could be a notation to simplify future equations instead of normalizing the tangent vector every time (if that's what you are referring to).

ninamiolane (approved):
Very nice, thanks! Just rename the notebook 18_real_world... instead of n_real_world... :D then, good to merge!

lpereira95:
@Jules-Deschamps, I've added the download of the dataset from a remote source. Let me know if I've broken anything!

Later, please discuss with @alebrigant about the creation of a proper figshare record to ensure we give the credit of the dataset to the right people.

PR #1570, opened 2022-06-09
Clean backend (only a bit)

lpereira95:
This PR cleans slightly the backend.

A better distinction between private and public methods is done (it makes it easier to understand what will be imported to `gs`).

Additionally, the `geomstats._backend_common.py` is created to avoid definition the same function for all backends. The handling of imports is changed to ensure `BackendImporter` looks into this new file.

This is the starting point of more meaningful changes in the backends related to precision.


ninamiolane (approved):
Excellent, thank you! Can't wait for our next-generation backends 🎉

PR #1571, opened 2022-06-10
Add cache flag for dependency caching in CI

lpereira95:
Avoid installing dependencies all the time (since they rarely change) in the CI (it will run faster). (Follows [GitHub Actions: setup-python now supports dependency caching](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/)).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
           cache-name: cache-pip-modules
         with:
           path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-{{ python-version }}-${{ hashFiles('setup.cfg') }}
+          key: ${{ runner.os }}-pip-{{ matrix.python-version }}-${{ hashFiles('setup.cfg') }}
 
       - name: Build using Python ${{matrix.python-version}} and Backend ${{matrix.geomstats-backend}}
         if: steps.cache-pip.outputs.cache-hit != 'true'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,13 +52,11 @@ jobs:
           key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('setup.cfg') }}
 
       - name: install dependencies [pip]
-        if: steps.cache-pip.outputs.cache-hit != 'true'
         run:  |
           pip install --upgrade pip setuptools wheel
           pip install -e .[ci,dev,doc,opt]
 
       - name: Add annotations [pytest]
-        if: steps.cache-pip.outputs.cache-hit != 'true'
         run: pip install pytest-github-actions-annotate-failures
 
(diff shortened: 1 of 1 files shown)


SaitejaUtpala:
@lpereira95  Hey.. do you have stats about how much faster?. Cause, I tried it's not helpful, since our dependencies are pretty small.

lpereira95:
Hi @SaitejaUtpala. `tensorflow` and `pytorch` are huge (~1gb). They're downloaded all the time. Do you feel this will not have an impact in CI time?

SaitejaUtpala:
@lpereira95. see the screenshot 
image

Both pytorch and tf takes ~0sec to downloaded. Two explanations could be possible

- VM's were CI is run have high bandwidth connectivity 
- common things like pytorch, tf are already baked into VM's. 

without caching it takes ~2min  sec. with  caching I remember it was same or worse.

lpereira95:
Thanks @SaitejaUtpala, very helpful comment.

Nevertheless, these 2 minutes are too much in my opinion (but I agree that probably caching does not solve the issue)... because it's 2min x 8 (or am I missing something?).  

I would say that ideally we should have a runner with everything installed (manually once?), so we do not spend 16min every time. (on the other hand, this approach will not test if the dependencies are managed properly -> maybe we can have a different approach for master?). Do you think this is feasible (and worthy to pursue)?

SaitejaUtpala:
@lpereira95. couple of questions

- why 2 * 8 ? these things run in parallel more of min(  CI time for 8 runs ) ~ 2 min

- can run CI with cahcing and let me know stats (perferrably average of mulitple runs) ? I still think we don't get any meaningful speedup. In fact I remember worse in some runs with caching.

also above screenshot I attached is for **without** caching

In fact I tried concurrent testing as well (see for closed PR https://github.com/geomstats/geomstats/pull/1288) I didn't see any meaningful gains.

lpereira95:
From a user perspective it is 2min, since it runs in parallel as you mention. Nevertheless, from a resource use perspective, it still takes 16min for something that should be straightforward.

I'll try it out with caching to see how it goes.

I'm wondering if we should install all the backends when the other backends are being tested... here is more to ensure that it still runs when they're not installed.

lpereira95:
@SaitejaUtpala from the very basic caching I've done in the last commit, it ends up more and less  taking the same amount of time, but differently distributed: build ~20s and install dependencies ~1m40s.

Still, I think I can make it faster by creating `venv` ([Caching dependencies to speed up workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)). We'll soon see...

lpereira95:
You're right @SaitejaUtpala: no real gain from caching here.

In a next PR I'll use the minimal dependencies for each backend, which will alleviate a bit the running time and increase the robustness of the tests.

PR #1572, opened 2022-06-10
Improve dependency handling and CI installations

lpereira95:
This PR will:
-handle dependencies more clearly
-do not install all the backends during the test of a specific backend

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,9 +17,9 @@ jobs:
         python-version: [3.7]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


lpereira95:
@SaitejaUtpala, the reduction of time is achieved not by caching (failed #1571), but by installing only the required dependencies.

lpereira95:
@ninamiolane if you agree with the changes in the handling of the dependencies (only name changes), I think this is ready to go. (Main accomplish here is the reduction of installation time - e.g. by ~1min20s for `numpy` and `autograd`... it is not a lot, but it accumulates over all the times we run the CI).

ninamiolane (approved):
Fantastic, thank you!!

PR #1573, opened 2022-06-13
Add utils to download remote datasets

lpereira95:
Large datasets should not be shipped with `geomstats` to avoid having a unnecessarily large source code. Besides, we can argue they should not even be part of the repo (e.g. 100mb is the limit in a push, if I'm not mistaken).

sklearn uses [figshare](https://figshare.com/account/home) to host datasets (notice it is very "academically" oriented, even requiring connection to ORCID). I think this is a very good solution.

As example, I've created a record for emg data available in `geomstats` (see [this](https://figshare.com/s/9a81e8f83a55c5e55e49)). With this, we can have code such as:

```
from geomstats.datasets._base import _fetch_remote

url = "https://figshare.com/ndownloader/files/35875382?private_link=9a81e8f83a55c5e55e49"
file_path = _fetch_remote(url, 'test.csv')

data = data_utils.load_emg(file_path=file_path)
```

(Notice `geomstats.datasets._base` is private because this should be handle by us, not by a user. Here it is just an example of what could be internal code)

The code in this PR is deeply inspired by [`sklearn.datasets`](https://github.com/scikit-learn/scikit-learn/blob/80598905e517759b4696c74ecc35c6e2eb508cff/sklearn/datasets/_base.py).

This PR appears at this point in time due to the needs of @Jules-Deschamps.

Some design choices:
* if user passes a `data_dir`, then such directory is created (if not existing).
* if the user does not pass `data_dir`, then `~/.geomstats_data` is (created and) used.
* the download is only performed if the file does not exist yet in the data_dir.

### DIFF
--- a/geomstats/datasets/_base.py
+++ b/geomstats/datasets/_base.py
@@ -10,15 +10,17 @@ from urllib.request import urlretrieve
 DEFAULT_DATA_DIR = os.path.expanduser(os.path.join("~", ".geomstats_data"))
 
 
-def _get_data_home(data_home=DEFAULT_DATA_DIR):
+def _get_data_home(data_home=None):
+    if data_home is None:
+        data_home = DEFAULT_DATA_DIR
+
     os.makedirs(data_home, exist_ok=True)
 
     return data_home
 
 
 def _fetch_remote(url, filename, dirname=None):
-    if dirname is None:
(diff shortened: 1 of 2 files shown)


ninamiolane (approved):
Very nice, thanks!

PR #1582, opened 2022-06-15
Refactor test_examples following test_notebooks strategy

lpereira95:
This PR creates a new testing strategy for examples following what was already done for notebooks. In particular:
* any file added to `examples` will be automatically tested (e.g. currently two example were being completely ignored).
* files can be ignored by adding the to `SKIP` set
* backends in which to run can be set in `METADATA`
* any argument to pass to `example.main()` can be set in `METADATA`


ninamiolane (approved):
Great, thanks!

PR #1583, opened 2022-06-15
Change location of test_examples and test together with notebooks

lpereira95:
Moves test_examples to main tests folder and tests `notebooks` and `examples` together in a test workflow called "scripts".

Reasoning: notebooks and examples are script-like code. I think they should be tested together (or, at least, examples should not be tested within tests_geomstats, as it is of a different nature - e.g. they only check if it runs, they do not validate the results).


lpereira95:
@SaitejaUtpala, do you see a better way of doing this? I don't really like that if...

.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         os: [ubuntu-18.04]
         python-version: [3.8]
         geomstats-backend : ['autograd','numpy','pytorch','tensorflow']
-        test-folder : ['tests/tests_geomstats/' , 'tests/test_notebooks.py']
+        test-folder : ['tests/tests_geomstats/' , 'scripts']
```
ninamiolane, line 34: maybe test-folder should be renamed, since "scripts" is not a folder?

ninamiolane (approved):
LGTM! I let you resolve the github conflicts and address my micro comment?

.github/workflows/test.yml
lpereira95: ah your comment made me realize that the better way to go may be to create a folder `test_scripts` and move there both test files, so that the workflow code becomes much simpler. What do you think?

SaitejaUtpala:
> @SaitejaUtpala, do you see a better way of doing this? I don't really like that if...

I think what you said is good idea, have separate folder for the test files of examples and notebooks.

PR #1584, opened 2022-06-17
Improve matching handling in graph space

lpereira95:
This PR creates the object `_Matcher` (similar strategy as solver in #1552). This greatly simplifies the development of learning algorithms (e.g. AAC), as matcher is not anymore passed as an argument. It will also simplify the parallelization of the matching.

Additionally, I've added the `align` method (matching + permutation). Syntax sugar for later.

If you agree with the changes @annacalissano, it is ready to merge.

I'll open another PR with the first draft implementation of AAC.

### DIFF
--- a/geomstats/geometry/stratified/graph_space.py
+++ b/geomstats/geometry/stratified/graph_space.py
@@ -4,6 +4,7 @@ Lead author: Anna Calissano.
 """
 
 import functools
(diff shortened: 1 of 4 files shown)


lpereira95:
After discussion with @annacalissano: change `align` to `align_point_to_point`

PR #1587, opened 2022-06-19
Calculating Geometric Median on Manifolds

shubhamtalbar96:
Calculate Geometric Median on Manifolds using Weiszfeld Algorithm.


ninamiolane:
The Lint, Deep Source and Tests do not pass. Marking this as draft for now. @shubhamtalbar96 I let you have a look? If you do not understand the errors, let us know. Thanks.

shubhamtalbar96:
@lpereira95 could you please advise on the test cases failing above? I don't think they are because of the Geometric Median changes.

PR #1589, opened 2022-06-21
Draft for visualizer module

Jules-Deschamps:
This is merely a draft for a visualizer in the information_geometry module.

I noticed some days ago that @lpereira95 was in the process of refactoring the whole visualisation module but we still thought it could be a good idea to see what I could make of it at the moment, regarding some distributions.

Namely, I think it would be great to be able to embed (or immerse) isometrically 1- and 2-D distributions in 2- or 3-D euclidean spaces because it would make them easier to appreciate: for example, seeing the Beta manifold as a hypersurface in R^3 could be super helpful to visualize in my opinion. We talked about that a little with @alebrigant. 

PS: I modified binomial.py because the metric matrix was not implemented.

### DIFF
--- a/geomstats/information_geometry/binomial.py
+++ b/geomstats/information_geometry/binomial.py
@@ -174,6 +174,24 @@ class BinomialFisherRaoMetric(RiemannianMetric):
         super(BinomialFisherRaoMetric, self).__init__(dim=1)
         self.n_draws = n_draws
 
+    def metric_matrix(self, base_point):
+        """Compute the inner-product matrix.
+
+        Compute the inner-product matrix of the Fisher information metric
+        at the tangent space at base point.
+
+        Parameters
+        ----------
+        base_point : array-like, shape=[..., 1]
+            Base point.
+
+        Returns
+        -------
+        mat : array-like, shape=[..., 1, 1]
+            Inner-product matrix.
+        """
+        return gs.array([self.n_draws / (base_point * (1 - base_point))])
+
     def squared_dist(self, point_a, point_b, **kwargs):
         """Compute squared distance associated with the binomial Fisher Rao metric.
 
--- a/geomstats/information_geometry/visualization_draft.py
+++ b/geomstats/information_geometry/visualization_draft.py
@@ -1,0 +1,99 @@
+"""Draft of visualizer (plotter) for information geometry."""
+
+import geomstats.backend as gs
+
+
+class Visualizer2D:
+    """Visualizer for 2-D parameter distributions."""
+
+    def __init__(self, space):
+        self.space = space
+
+    def scatter(self, ax, point, **kwargs):
+        """Plot points on the manifold."""
+        ax.scatter(point[..., 0], point[..., 1], **kwargs)
+
+    def plot_vector_field(self, ax, point, tangent_vec, **kwargs):
+        """Quiver plot on the manifold."""
+        point, tangent_vec = gs.to_ndarray(point, to_ndim=2), gs.to_ndarray(
+            tangent_vec, to_ndim=2
+        )
+        n_points, n_vecs = point.shape[0], tangent_vec.shape[0]
+        if n_points > 1 and n_vecs != n_points:
+            raise ValueError("Each tangent vector must be associated to a base point.")
+
+        ax.quiver(
+            point[..., 0],
+            point[..., 1],
+            tangent_vec[..., 0],
+            tangent_vec[..., 1],
+            **kwargs
+        )
+
+    def plot_geodesic(
+        self,
+        ax,
+        initial_point,
+        end_point=None,
+        initial_tangent_vec=None,
+        times=gs.linspace(0, 1, 10),
+        **kwargs
+    ):
+        """Plot geodesic on the manifold."""
+        geodesic = self.space.metric.geodesic(
+            initial_point, end_point, initial_tangent_vec
+        )
+        geod_at_t = geodesic(times)
+        ax.plot(*gs.transpose(geod_at_t)[::-1], **kwargs)
+
+    def plot_grid(
+        self, ax, lower_left, upper_right, n_cells=[5, 5], steps=None, **kwargs
+    ):
+        """Plot geodesic grid on the manifold."""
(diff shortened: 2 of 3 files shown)


ninamiolane (commented):
Converting to draft since the PR is titled draft.

Some comments/questions:

1. The PR seems to create parent classes for visualizer of manifolds in 1D or 2D: are these meant to be used to visualize manifolds beyond manifolds from information geometry? Or can they? If so, the docstrings should be adapted.
2. Visualization should go in the folder `visualization`. The name of the python file should refer to the name of the class that is inside, as per @emaignant design document.
3. Very good idea to allow 1D manifolds to be visualized in 3D. I think this could be done with the design in this PR: e.g. the visualization/beta_manifold.py file could have two classes: `class BetaManifold2D(Visualizer2D)` and `BetaManifold3D(Visualizer3D)` that allow both vizus.

Yes, the PR #1512  by @lpereira95 would need to be merged so that we can properly build on it.

Jules-Deschamps:
> Converting to draft since the PR is titled draft.
> 
> Some comments/questions:
> 
> 1. The PR seems to create parent classes for visualizer of manifolds in 1D or 2D: are these meant to be used to visualize manifolds beyond manifolds from information geometry? Or can they? If so, the docstrings should be adapted.
> 2. Visualization should go in the folder `visualization`. The name of the python file should refer to the name of the class that is inside, as per @emaignant design document.
> 3. Very good idea to allow 1D manifolds to be visualized in 3D. I think this could be done with the design in this PR: e.g. the visualization/beta_manifold.py file could have two classes: `class BetaManifold2D(Visualizer2D)` and `BetaManifold3D(Visualizer3D)` that allow both vizus.
> 
> Yes, the PR #1512 by @lpereira95 would need to be merged so that we can properly build on it.

Thanks for the feedback!

1. The child class is sort of an embedding visualizer, I figured it would make more sense to have it inside the original Visualizer class but maybe it is not the way to go. I don't really know if all this can be used to visualize manifolds not from information geometry because the geometries can be too different in my opinion?
2. The idea was to input the geometry in the visualizer so that we wouldn't have to create a visualizer class for each geometry because (in information geometry at least) some geometries can only depend on their metric matrix for example, that's why I gave this name to the file, but yes I should have put in visualization.

Jules-Deschamps:
I have added an `overlay_scatter` method in the 2D visualizer in order to have an easier visualization of the manifold for information geometry: when hovering over a point on the manifols, we can see a plot of the corresponding p.d.f. (I hope the modules needed to do so are not too heavy).

Additionnally, `isometric_beta` is only to be run in order to see what could eventually be the visualizer in information geometry (that's why it is not in `visualization`) (unfortunately this is not in a euclidean space but in a Minkowski space)

Jules-Deschamps:
The `visualizer_draft_infogeo` file implements visualizer classes specifically for information geometry (but so far there is no reason for it not to be used elsewhere).

A reproducible example to see how it works is:

```
import matplotlib.pyplot as plt

from geomstats.information_geometry.binomial import BinomialDistributions
from geomstats.information_geometry.beta import BetaDistributions
from geomstats.visualization.information_geometry.visualization_draft_infogeo import (
    Visualizer,
    Visualizer1D,
    Visualizer2D,
)

### 1D with 7-binomial distributions

fig1d = plt.figure(figsize=(12,8))
space = BinomialDistributions(7)

visu = Visualizer1D

point = space.random_point()

visu(fig1d, 121, space).plot_geodesic_ball(point, radius=.5)
iso = visu(fig1d, 122, space).iso_visualizer1d(left_bound=.1, right_bound=.9)
iso.scatter(point)
iso.plot_geodesic_ball(point)

plt.show()

### 2D with Beta distributions

fig2d = plt.figure(figsize=(12,8))
space = BetaDistributions()

visu = Visualizer2D

point = space.random_point()

tangent_vec = space.metric.random_unit_tangent_vec(point)

visu(fig2d, 221, space).scatter(point, color='r')
visu(fig2d, 221, space).plot_geodesic(initial_point=point, initial_tangent_vec=tangent_vec)
visu(fig2d, 122, space).plot_geodesic_ball(center=point)
plt.show()
```

3D visualizers are incoming!

Jules-Deschamps:
I have been working on this for the past few days to eventually end with unconclusive results :(.

The aim of `plot_geodesic_ball` in `3DVisualizer` is to plot the geodesic sphere (I think I should change the name now actually) for a space of dim 3: to do that, I think it is easier on the eyes to do a triangle surface plot between a set of points at geodesic distance `radius` from the center.

What I originally wanted with this in mind is to generate uniformly distributed (in terms of cartesian coordinates) tangent vectors of norm `radius` at `center` using the Fibonacci sphere algorithm, and then to plot the triangles. So the surface that needs to be plot depends entirely on these tangent vectors (we use `exp` for each of them at `center` and that's it). The triangle surface that needs to be plot has to be parametrized by 2 elements using `mtri.triangles`, so it is natural to go from cartesian coordinates to spherical coordinates to parametrize all the tangent vectors in the sphere of (euclidean) radius `radius` by (theta, phi). The issue with this is we have to be able to generate all the initial tangent vectors from all thetas and phis only (to be able to do a triangle surface plot, the "surface points" must be 2-parametrized), and that is actually not possible (up to a sign problem) surprisingly.

The other way to do it (which I did here) is simply to generate ranges of thetas and phis and then the associated tangent vectors. It works well, although we are losing the "uniformly distributed set of points" of the Fibonacci sphere algorithm. Maybe it is not that important? I haven't been able to find a fix for uniformity.

A working example on the 3-Dirichlet manifold:

```
import matplotlib.pyplot as plt
from matplotlib import cm

from geomstats.information_geometry.dirichlet import DirichletDistributions
from geomstats.visualization.information_geometry.visualization_draft_infogeo import (
    Visualizer3D
)

fig3d = plt.figure(figsize=(12,8))
space = DirichletDistributions(3)

visu = Visualizer3D(fig3d, 111, space)

point = space.random_point()

visu.plot_geodesic_ball(point, n_rays=400, cmap=cm.viridis)
plt.show()

```

PR #1599, opened 2022-06-30
Refactor backend and corresponding tests (part 1)

lpereira95:
This PR updates several functions in the backends and starts the refactoring of the corresponding tests to follow new (data-driven) strategy.

Highlights include:
* vectorization of functions: `outer`, `dot`, `cross`, `trace`
* replace several `einsum` expressions by corresponding functions (for non-vectorized cases, this will improve performance)
* reduce ambiguity of functions: `matmul` does not work with `ndim == 1` ; `dot` assumes vectors (if `ndim==2`, then is several vectors, not a matrix)
* add new functions: `matvec` (replaces several einsums; for particular cases, new implementation is fast)
* simplification of `einsum` for `tf` (new tests added and apparently `tf` native way works with all of them)

Several new implementations call "native" functions if one of the inputs is not vectorized, otherwise `einsum` (I'll soon provide a way of comparing several implementations of the same function to show `geomstats` is the fastest).

This precedes a PR that will allow to globally control precision.

P.S. need to address failures in tests before merge.


ninamiolane (approved):
LGTM! How did you evaluate that such changes will improve performances?

`matvec` is a good addition, although since numpy does not have it I wonder if users will think of it. Generally, the fewer modifications with make to numpy's API the better (we don't want to have to explain all the exceptions to users), but I trust you that matvec is needed here!

lpereira95:
For evaluating the performance of some changes, I've created different implementations of the functions that I've updated:
1) a "backend native" way that looks for a solution other than einsum for vectorization (not always possible)
2) a "backend native" way that performs vectorization via python for loop
3) einsum

Then I run them in all the possible input combinations (e.g. for `matvec`: (1 mat, 1 vec), (n mat, 1 vec), (1 mat, n vec), (n mat, n vec).

My conclusions is that usually the native way is much faster for the case (1, 1) and may be faster for one or both combinations (1, n) or (n, 1). For (n, n) einsum is tipically much faster. I chose gs implementation to use the fastest way fr each case.

(I'll share soon my results in a nice format.)

I would say for the current object sizes used in geomstats we will probably not notice the difference. What I think is more worthy to stress is that we are trying to ensure all the functions are vectorized, so that a contributor can develop its code almost without concern about vectorization. Besides, before some contributors were going for `einsum` for vectorization. Now we are trying to avoid explicitly using it (because i) it can be harder to read, and ii) for the not vectorized case "native" solutions are usually faster).

The tests are also now testing all the vectorization cases in a general way.


I completely agree with you about going away from `numpy` API. I'll ensure the changes will be properly documented. About e.g. `matvec`, notice that also `tf` defines it. A way of justifying our changes relative to `numpy` (which I think will help developers) is that we are removing ambiguities that arise through vectorization when using some `numpy` functions. `geomstats` always assume the first dimension may be a "vectorization" dimension.

PR #1601, opened 2022-07-04
Refactor FrechetMean

lpereira95:
This PR removes `point_type` argument from `FrechetMean` and updates dependent algorithms to not use it either if related with FrechetMean call.

**Why?** Only reason for `point_type` is a scalar multiplication. This can be done with einsum without really knowing if it is a vector or a matrix.  More interestingly, FrechetMean can be now applied even for ndim > 2 (may be relevant for e.g. Sasaki metric (#1568)).

A strong assumption now, that was already being used in another learning algorithms, but not in `FrechetMean`, is that the first dimension of `X` represents `n_samples`. This follows `sklearn` behavior. Therefore, if a single point is passed to `fit`, it will not work.

I've also taken advantage of this PR to clean a bit the code in `FrechetMean`.

**Points that deserve further attention/should be discussed**

1) For the code branch `if "HypersphereMetric" in metric_str and self.metric.dim == 1:` a mean was being computed with `_circle_mean`, but later being recomputed with one of the methods. When correcting the flow, i.e. computing `mean` only once, the tests in circular mean do not pass (see test workflow). What's wrong?

2) `_batch_gradient_descent` used to verify the dimensions of `X` and use `default_gradient_descent` if not batch. I removed this behavior because I think it is misleading (e.g. some code was setting `batch` as default method and in practice using `default_gradient_descent`). I think users should know batch requires a transformation of the input and take care of it themselves.

### DIFF
--- a/examples/empirical_frechet_mean_uncertainty_sn.py
+++ b/examples/empirical_frechet_mean_uncertainty_sn.py
@@ -66,6 +66,9 @@ def empirical_frechet_var_bubble(n_samples, theta, dim, n_expectation=1000):
         estimator = FrechetMean(
             sphere.metric, max_iter=32, method="adaptive", init_point=north_pole
         )
+        if n_samples == 1:
+            data = gs.expand_dims(data, 0)
+
         estimator.fit(data)
         current_mean = estimator.estimate_
         var.append(sphere.metric.squared_dist(north_pole, current_mean))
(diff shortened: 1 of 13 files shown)


ninamiolane (approved):
LGTM! The circle mean was contributed by an independent collaborator, I am not sure why it fails.

lpereira95:
Regarding point 1. (circular mean vs mean computed with gd), the problem is that for some cases gd does not converge to the global minimum, and when it converges, it is the same as circular mean (therefore it shows the correctness of the circular mean implementation).

PR #1606, opened 2022-07-15
Curvature implementation

Jules-Deschamps:
Implementation of methods for metric and connection, so that riemannian metrics can return :
- riemannian tensor
- curvature
- ricci tensor
- scalar curvature
- sectional curvature (was already implemented if curvature was)

Tests on shapes of curvature tensors symmetries of riemannian tensor on test files (geometry_test_cases and data_generation), and on negativity of sectional curvature of dirichlet distributions (on another script file, I did not know if I should add it in the test file).


ninamiolane:
@Jules-Deschamps Looking good! You also need to modify:https://github.com/geomstats/geomstats/blob/master/tests/data/connection_data.py
to solve this error appearing in the tests:
```
testing_data object doesn't have 'ricci_tensor_shape_test_data' function for pairing with 'test_ricci_tensor_shape'
```

You can run `pytest tests/test_geomstats/test_connection.py` on your machine directly (so that you do not wait for the GitHub actions to finish).

geomstats/geometry/riemannian_metric.py
```diff
@@ -453,6 +453,25 @@ def normal_basis(self, basis, base_point=None):
 
         return gs.einsum("i, ikl->ikl", 1.0 / gs.sqrt(norms), basis)
 
+    def covariant_riemannian_tensor(self, base_point):
+        r"""Compute purely covariant version of Riemannian tensor at base_point.
```
ninamiolane, line 457: Thanks, could you also add the formula for it, i.e. R_{ijkl} = g_{ll'}R_{ijk}^l' ?

ninamiolane (approved):
LGTM when test's infrastructure problem is solved.

geomstats/geometry/connection.py
```diff
@@ -458,7 +458,90 @@ def curvature(self, tangent_vec_a, tangent_vec_b, tangent_vec_c, base_point):
...
+            Point on the group. Optional, default is the identity.
+
+        Returns
+        -------
+        riemann_curvature : array-like, shape=[..., {dim, [n, m]}, {dim, [n, m]},
+                                                    {dim, [n, m]}, {dim, [n, m]}]
+            R_ijk^l = riemann_tensor[i,j,k,l]
+            Riemannian tensor curvature.
+        """
+        base_point = gs.to_ndarray(base_point, to_ndim=2)
```
lpereira95, line 486: Does this work when `base_point` is a matrix?

Can we somehow avoid its use? Because we create a new array every time we do it. `christoffels` should work with a base point or multiple base points, no?

geomstats/geometry/connection.py
```diff
@@ -458,7 +458,90 @@ def curvature(self, tangent_vec_a, tangent_vec_b, tangent_vec_c, base_point):
...
+                - gs.transpose(prod_christoffels, [0, 4, 3, 2, 1])
+            )
+        else:
+            riemann_curvature = (
+                gs.transpose(jacobian_christoffels, [2, 3, 1, 0])
+                - gs.transpose(jacobian_christoffels, [2, 1, 3, 0])
+                + gs.transpose(prod_christoffels, [3, 1, 2, 0])
+                - gs.transpose(prod_christoffels, [3, 2, 1, 0])
+            )
+        return riemann_curvature
```
lpereira95, line 510: I think we can avoid the if by using `gs.move_axis` if I'm not mistake and choosing axis from the end.

lpereira95 (changes requested):
Thanks @Jules-Deschamps.

I'm just reviewing `Connection.riemann_tensor` now. Can you please take a look to my comments.

I'll come back to this PR soon.

geomstats/geometry/riemannian_metric.py
Jules-Deschamps: Yes will do! As for the testing thing I thought I had resolved this but turns out I haven't yes, thanks!

geomstats/geometry/connection.py
Jules-Deschamps: Yes I was using it to know the number of points in base_point (to be able to permute the axes using gs.transpose), but if gs.move_axis (which I did not know existed) works I will get rid of these lines.

geomstats/geometry/connection.py
Jules-Deschamps: Neither `gs.moveaxis` and `gs.move_axis` seem to exist as attributes of `geomstats.backend` but yes it would greatly simplify things

geomstats/geometry/connection.py
lpereira95: But if `base_point` is a matrix (representing only a point in the manifold), you will get wrong information, no?

Your branch is too far behind, that's why you cannot find `moveaxis`. Please merge master after pulling upstream.

Jules-Deschamps: It works apparently because there is a squeeze in `christoffels`. So for `base_point` of shape (1,dim), `christoffels` will return shape (dim, dim, dim) and `jacobian_christoffels` (dim, dim, dim, dim) because it is defined with a squeeze. So these lines are really just to get the number of points in `base_point`. I hope this is what you are talking about. This is a minimum reproducible example showing it works for several types of `base_point` (and namely a matrix representing only a point in the manifold).
```
from geomstats.information_geometry.dirichlet import DirichletDistributions

space = DirichletDistributions(3)

point_1 = space.random_point()
point_2 = gs.array([[1.0, 1.0, 1.0]])
point_3 = gs.array([[1.0, 2.0, 2.0], [1.0, 2.0, 1.0]])

print(space.metric.riemann_tensor(point_1))
print(space.metric.riemann_tensor(point_2))
print(space.metric.riemann_tensor(point_3))
```

Ok I will merge thanks!
Do you know if the methods I added in `data_generation` should rather be in `connection_data` and `riemannian_metric_data`? That way I won't have to rewrite the data generation for every manifold?

geomstats/geometry/riemannian_metric.py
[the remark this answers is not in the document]
Jules-Deschamps: Yes. I just checked because the tests were wrong but the convention used in the literature (or at least on wikipedia) is VERY weird (`R_{ijkl} = <R(x_j, x_k)x_i, x_l>`) so I think the previous successful tests were just a fluke. I will make a commit as soon as possible with your feedback!

geomstats/geometry/connection.py
Jules-Deschamps: So turns out I am keeping `base_point = gs.to_ndarray(base_point, to_ndim=2)` in my last commit also because I want to compute the jacobian christoffels for each of the points in `base_point`, which is not possible otherwise (or so far I haven't found anything).

Jules-Deschamps:
I have added modifications following your feedbacks and have tested for dirichlet distributions (the sectional curvature is indeed everywhere negative), but for hyperbolic spaces it is more complicated because sectional and scalar curvatures require `metric_matrix` which has not yet been implemented. As for testing on the hypersphere, it is a bit complicated considering the extrinsic and intrinsic coordinates but I will try.

Also, I was wondering where I should add the tests and data generators so that I don't have to rewrite all the tests for every manifold? For example for dirichlet distributions, in `dirichlet_data`:

```
def ricci_tensor_shape_test_data(self):
        return self._ricci_tensor_shape_test_data(
            self.metric_args_list, self.space_list
        )
```, this looks like it could be replaced up above the architecture? Thanks

**Comment by ninamiolane** [comment]
Fantastic, thanks for addressing the comments. There are still some testing issues, which you can access by clicking on "details" in the github workflows above, or e.g. here:
https://github.com/geomstats/geomstats/runs/7400372430?check_suite_focus=true
for example:
```
    raise Exception(
E   Exception: testing_data object doesn't have 'covariant_riemann_tensor_bianchi_identity_test_data' function for pairing with 'test_covariant_riemann_tensor_bianchi_identity'
```

Yes, I believe we have to have this bit of redundant code for now:
```
def ricci_tensor_shape_test_data(self):
        return self._ricci_tensor_shape_test_data(
            self.metric_args_list, self.space_list
        )
```
but @lpereira95 was working on removing it so that you do not have to copy-paste it like this, in the future. @lpereira95 do you confirm?

File: geomstats/geometry/riemannian_metric.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (alebrigant) [reply geomstats/geometry/riemannian_metric.py:520]: You must beware that the sign convention taken in Wikipedia for the Riemann curvature tensor (https://en.wikipedia.org/wiki/Riemann_curvature_tensor) is not the same as the one taken in geomstats (see [this docstring](https://github.com/geomstats/geomstats/blob/d1aa07d97cd5aa6b18373eb91d64969700c32463/geomstats/geometry/connection.py#L443)) and this has effects on the rest of the definitions. So all definitions should be taken from the *same reference book*, and not Wikipedia.

**Comment by alebrigant** [comment]
To test sectional curvature more precisely than just the sign, you can check that for beta distributions, at any point `gs.array([x, y])`, it is given by

```
K_xy = gs.polygamma(2, x) * gs.polygamma(2, y) * gs.polygamma(2, x+y) * (
    gs.polygamma(1, x) / gs.polygamma(2, x) 
  + gs.polygamma(1, y) / gs.polygamma(2, y) 
  - gs.polygamma(1, x+y) / gs.polygamma(2, x+y)
) / detg ** 2
```
where
```
detg = gs.polygamma(1, x) * gs.polygamma(1, y) - gs.polygamma(1, x+y) * (gs.polygamma(1, x) + gs.polygamma(1, y))
```
See Appendix in https://www.sciencedirect.com/science/article/pii/S2405896321005851?via%3Dihub.

Jules-Deschamps:
Thanks again!

I don't really understand how to address the issues pointed out by pytest because they are the same for every manifold (the data generation methods are not implemented without the '_' underscore whent they inherit from the tests so they don't know with what data to run the tests). This is not only for the bianchi identity but for all the new tests I have added in `geometry_test_cases`. So should I add the test methods without underscore for every manifold/geometry?

As for the testing of the sectional curvature on the Beta manifold, it works perfect! I have added the associated test.

lpereira95:
For the tests, I suggest for the moment that we add a skip to them (which we unskip as soon as I merge the tests PR). Nevertheless, notice that most of the metrics do not have `metric_matrix` implemented, which means most of new implementations will not work.

We still need to go over some parts of the new code (see above) before merging, as it may not be working for several cases.

geomstats/geometry/riemannian_metric.py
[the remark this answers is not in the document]
alebrigant: Looking at all the formulae, I'm not completely sure that they are all compatible, i.e. take the same conventions. The only way to avoid mistakes in my view, is to pick a reference, and to take all formulas from this reference. This reference should be added to all docstrings.

alebrigant (changes requested):
Thanks for making the changes ! I still recommend further checks of the docstrings, to avoid all mistakes. See the detailed comments.

Jules-Deschamps:
Still some issues to fix apparently (with some shape testing specifically, the rest should work now)!
Sorry for all these pushes but I am not getting the same errors when I am running pytest on `tests_geomstats`.

lpereira95:
@Jules-Deschamps, I've merged the tests PR. It should be easier now to make what you want with the tests (no need to use `_` anymore). Let me know if you need help solving the conflicts that were raised.

Jules-Deschamps:
Oh sorry I did not see your message! Ok thank you, I will see how it works and then push for last update. Tests regarding curvature all work fine now (even shape tests) I just tweaked some things.

ninamiolane:
> Tests regarding curvature all work fine now (even shape tests) I just tweaked some things.

Excellent, congrats @Jules-Deschamps on fixing them! Good to merge on my side when the tests pass.

ninamiolane:
TODO: Modify files like: test_gamma and gamma_data and adapt to new tests' architectures by Luis.

PR #1609, opened 2022-07-25
Add ability of using kwargs in test functions

lpereira95:
This change will allow to do something like:

```python
def test_func(self, a, b, c=3):
  pass
```

This is achieved by changing the function signature to e.g. `(self, a, b, c)` and passing the default as data value if not encountered in the datum.

This feature is handy sometimes to reduce verbose in data creation.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -105,7 +105,20 @@ def copy_func(f, name=None):
         f.__code__, f.__globals__, name or f.__name__, f.__defaults__, f.__closure__
     )
     fn.__dict__.update(f.__dict__)
-    return fn
+
+    # transform keyword arguments in positional arguments (due to pytest)
+    sign = inspect.signature(fn)
+    defaults, new_params = {}, []
+    for param in sign.parameters.values():
+        if isinstance(param.default, inspect._empty):
+            new_params.append(param)
+        else:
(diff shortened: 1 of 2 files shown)


SaitejaUtpala:
@lpereira95 Nice addition! I just have one comment. Once all thing are completed in testing end. I would suggest to write up how different pieces are interacting and how test data is built in flow chart so that if someone one wants to look under hood or suggest improvement it becomes easy. Also congrats on merging giant PR #1517

tests/conftest.py
```diff
@@ -105,7 +105,20 @@ def copy_func(f, name=None):
         f.__code__, f.__globals__, name or f.__name__, f.__defaults__, f.__closure__
     )
     fn.__dict__.update(f.__dict__)
-    return fn
+
+    # transform keyword arguments in positional arguments (due to pytest)
```
SaitejaUtpala, line 109: remove leftover comments

SaitejaUtpala (approved):

lpereira95:
Thanks @SaitejaUtpala! I think it is a great suggestion and I'll definitely do it as soon as I finish the changes (there's still some ideas to be developed). Thank you also again for all the work you've put in developing this framework (very unique and powerful).

PR #1610, opened 2022-07-26
Make `array_from_sparse` more general

lpereira95:
This PR makes `array_from_sparse` more general by allowing creation of arrays with dimensions other than 2. Handling of dtype is also improved in pytorch.

Following this change:
* the computation of basis for some cases is simplified by avoiding using of `gs.stack`.
* `get_mask_i_float` is removed from backend as it is a particular case (this pruning of backend functions is valuable to ensure we keep it slim and tested properly)


lpereira95:
@ninamiolane, test failures are unrelated (checked with care).

PR #1612, opened 2022-07-29
mixins for information geometry

Jules-Deschamps:
This is a draft for the mixins for the information geometry module.

`InformationManifold` has methods `sample`, `_hard_pdf`, `point_to_pdf`, `point_to_cdf`. Following a conversation with @ninamiolane , we believe hard-coding in `_hard_pdf` the pdf of the distribution one would want to work on would allow to derive all necessary quantities to do so (namely, `metric_matrix`, because it all stems from here).

To that end, the class `FisherRaoMetric` (the name already exists so maybe to change) is initiated with arguments `pdf` (that's where `_hard_pdf` can come in handy), `support` (where to integrate to compute the expectance to get the Fisher information matrix, but I think we can find a more generic way of addressing this problem), and `dim` (I think we can bypass this argument). The method `metric_matrix` returns the metric at a given base point on the manifold.

It all still needs some changes mainly because I'm not so sure `FisherRaoMetric` and the metrics already implemented for some information manifolds (like Beta, Gamma, ...) are entirely compatible.


Jules-Deschamps:
A reproducible example (to see how it would work) is:

`import geomstats.backend as gs
from geomstats.information_geometry.information_manifolds import InformationManifold, FisherRaoMetric

info = InformationManifold()

def pdf(point, x):
    mu, sigma = point
    return gs.exp(-(x-mu)**2 / (2 * sigma**2)) / (sigma * gs.sqrt(2 * gs.pi))

metric = FisherRaoMetric(pdf, [-20, 20], 2)

print(metric.metric_matrix(gs.array([2.0,2.0])))`

Jules-Deschamps:
A reproducible example (to see how it would work) is:

```
from geomstats.information_geometry.information_manifolds import InformationManifold, FisherRaoMetric

info = InformationManifold()

def pdf(point, x):
mu, sigma = point
return gs.exp(-(x-mu)2 / (2 * sigma2)) / (sigma * gs.sqrt(2 * gs.pi))

metric = FisherRaoMetric(pdf, [-20, 20], 2)

print(metric.metric_matrix(gs.array([2.0,2.0])))
```

geomstats/information_geometry/binomial.py
```diff
@@ -10,9 +10,10 @@
 from geomstats.geometry.base import OpenSet
 from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.information_geometry.information_manifolds import InformationManifold
```
ninamiolane, line 13: NIT: that would be information_manifold

We only use plural when the name used for the manifold is the name of its elements, e.g. THE manifold of binomial distributionS, which is however A Information Manifold.

geomstats/information_geometry/information_manifolds.py
```diff
@@ -0,0 +1,170 @@
...
+            cdf_at_x : array-like, shape=[..., n_points]
+                Values of cdf at x for each value of the parameters provided
+                by point.
+            """
+            raise NotImplementedError("The cdf method has not yet been implemented.")
+
+        return cdf
+
+
+class FisherRaoMetric(RiemannianMetric):
```
ninamiolane, line 125: NIT: This could be its own file fisher_rao_metric.py

geomstats/information_geometry/information_manifolds.py
```diff
@@ -0,0 +1,170 @@
...
+    """Class to derive the information metric from the pdf in InformationManifold."""
+
+    def __init__(self, pdf, support, dim):
```
ninamiolane, line 128: How about passing the information manifold as input to the constructor directly?

This would make the logic more consistent with invariant_metric.py i.e. a class of Riemannian metrics that are defined specifically on Lie groups. For which the lie group is passed in the constructor.
https://github.com/ninamiolane/geomstats/blob/master/geomstats/geometry/invariant_metric.py

Likewise, the Fisher rao metric is defined specifically for information manifolds. Thus the information manifold can be passed in the constructor.

geomstats/information_geometry/information_manifolds.py
```diff
@@ -0,0 +1,170 @@
...
+        self.pdf = pdf
+        self.support = support
+        self.dim = dim
+
+    def metric_matrix(self, base_point):
+        """Compute the inner-product matrix.
+
+        Compute the inner-product matrix of the Fisher information metric
```
ninamiolane, line 136: Can you add:
- formula for the definition
- reference
? 

Thnks!

geomstats/information_geometry/information_manifolds.py
```diff
@@ -0,0 +1,170 @@
...
+            Optional, default: 1.
+
+        Returns
+        -------
+        samples : array-like, shape=[..., n_samples]
+            Sample from the probability distributions.
+        """
+        raise NotImplementedError("The sample method is not yet implemented.")
+
+    def _hard_pdf(self, point, x):
```
ninamiolane, line 34: I don't believe we need this. Because the information manifolds that inherit from InformationManifold will "hard code" the method `pdf`, in _their_ `pdf` method, if they need to.

ninamiolane (commented):
Very nice, thanks! 

How about having FisherRaoMetric taking information_manifold as input:
```
class FisherRaoMetric(RiemannianMetric):
    def __init__(self, information_manifold):
        self.information_manifold = information_manifold # --> gives access to self.information_manifold.pdf
```

We would get this behavior, for a user introducing a new information manifold:

```
from geomstats.information_geometry.information_manifolds import InformationManifold, FisherRaoMetric

class MyInformationManifold(InformationManifold):
    def __init__(self, mu, sigma):
        self.mu = mu
        self.sigma = sigma
    def pdf(self, x):   # will give self.information_manifold.pdf in the fisher rao metric
        return gs.exp(-(x-self.mu)2 / (2 * self.sigma2)) / (self.sigma * gs.sqrt(2 * gs.pi))

my_information_manifold = MyInformationManifold(mu=1, sigma=2)
metric = FisherRaoMetric(my_information_manifold)
```

And for existing classes such as CategoricalMetric:

```
class CategoricalMetric(FisherRaoMetric): 
    def __init__(self, dim):
        super(CategoricalMetric, self).__init__(dim=dim, information_manifold=CategoricalDistributions(dim=dim))
        self.sphere_metric = HypersphereMetric(dim)
```
where `CategoricalDistributions`:
- either does not have `pdf`, but this is fine since we have hard-coded the formula of its fisher rao metric
- can have its own method called `pdf` (not _hard_coded_pdf) for completeness.

Jules-Deschamps:
Sorry for the late response and thanks very much for your feedback!

- About the `pdf` name: for information manifolds (the ones that are already coded) a method already exists called `point_to_pdf`, which takes as input a point on the manifold and returns the pdf function associated with that parameter point. So I thought about using this function to derive the Fisher-Rao metric, the only issue is that it is often coded (most often in the higher scipy architecture) in a way that is non-differentiable by autograd, so we have to hard code the pdf for the information manifold as well (the way I did it for normal distributions). But if we keep the same name `point_to_pdf` it would maybe make more sense?
- I think `InformationManifold` should not take parameters as inputs, for instance we would want to compute the metric at `point_1` and `point_2` without having to instantiate 2 different information manifolds.

Jules-Deschamps:
I made some changes following Nina's comments, so that it can become easier to implement an information manifold on the basis of its parametric pdf only (I am giving an example at the end of the following script with the Laplace distribution).

I added the method `inner_product_derivative_matrix` directly in the `FisherRaoMetric` class for 2 reasons:
- it is less computationally expensive to compute it as the integral of the derivative of pdf * (partial(log_pdf)^2) (integral and derivation commute) than as the derivative of the `metric_matrix` method (which would be the case without coding this method);
- it is required to compute christoffels and essentially all quantities relevant to geodesics.

The following script gives examples of what the new classes `InformationManifold` and `FisherRaoMetric` can be set to do. Geodesic quantities (such as `exp` for information manifolds implemented with a pdf only, for instance the last line of the following script for Laplace distributions) take far too much to compute.

```
from geomstats.information_geometry.fisher_rao_metric import FisherRaoMetric
from geomstats.information_geometry.information_manifold import InformationManifold

# beta distributions

from geomstats.information_geometry.beta import BetaDistributions

beta = BetaDistributions()

def pdf_at_point(point):
    alpha, beta = point
    def pdf(x):
        return gs.gamma(alpha + beta) * x**(alpha - 1) * (1 - x)**(beta - 1) / (gs.gamma(alpha) * gs.gamma(beta))
    return pdf

beta.point_to_pdf = pdf_at_point

metric = FisherRaoMetric(beta, [0,1])

point = gs.array([2.0, 2.0])

print(metric.metric_matrix(point), beta.metric.metric_matrix(point), f"L2 error = {gs.sum((metric.metric_matrix(point) - beta.metric.metric_matrix(point))**2)}")

# gamma distributions

from geomstats.information_geometry.gamma import GammaDistributions

gamma = GammaDistributions()

def pdf_at_point(point):
    kappa, gamma = point
    def pdf(x):
        return kappa**kappa / (gamma**kappa * gs.gamma(kappa)) * x**(kappa-1) * gs.exp(-kappa * x / gamma)
    return pdf

gamma.point_to_pdf = pdf_at_point

metric = FisherRaoMetric(gamma, [0,20])

point = gs.array([2.0, 2.0])

print(metric.metric_matrix(point), gamma.metric.metric_matrix(point), f"L2 error = {gs.sqrt(gs.sum((metric.metric_matrix(point) - gamma.metric.metric_matrix(point))**2))}")

# normal distributions

from geomstats.information_geometry.normal import NormalDistributions

normal = NormalDistributions()

def pdf_at_point(point):
    mu, sigma = point
    def pdf(x):
        return gs.exp(-(x-mu)**2 / (2 * sigma**2)) / (sigma * gs.sqrt(2 * gs.pi))
    return pdf

normal.point_to_pdf = pdf_at_point

metric = FisherRaoMetric(normal, [-20,20])

point = gs.array([2.0, 2.0])

print(metric.metric_matrix(point))

# laplace manifold

laplace = InformationManifold()

def pdf_at_point(point):
    mu, b = point
    def pdf(x):
        return 1 / (2*b) * gs.exp(-gs.abs(x - mu) / b)
    return pdf

laplace.point_to_pdf = pdf_at_point

metric = FisherRaoMetric(laplace, [-50,50])

point = gs.array([2.0, 3.0])

print(metric.metric_matrix(point))

laplace.metric = metric

tangent_vec = laplace.metric.random_unit_tangent_vec(point)

print(laplace.metric.exp(tangent_vec, point))
```

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,106 @@
...
+        mat : array-like, shape=[..., dim, dim]
+            Inner-product matrix.
+
+        References
+        ----------
+        .. [AS1985] Amari, S (1985)
+            Differential Geometric Methods in Statistics, Berlin, Springer – Verlag.
+        """
+
+        def log_pdf_at_x(x):
```
ninamiolane, line 42: The functions here need their docstrings.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,106 @@
+"""Class to implement simply the Fisher-Rao metric on information manifolds."""
+
+from scipy.integrate import quad_vec
+
+import geomstats.backend as gs
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+
+
+class FisherRaoMetric(RiemannianMetric):
+    """Class to derive the information metric from the pdf in InformationManifold."""
```
ninamiolane, line 10: Can you give details on the information_manifold and support input parameters?

geomstats/information_geometry/information_manifolds.py
```diff
@@ -0,0 +1,170 @@
+"""Mixin for manifolds of probability distributions."""
```
ninamiolane, line 1: This file is probably not needed anymore?

ninamiolane (approved):
LGTM! Just a few docstrings to fix and then good to merge. Thank for the great work 🎉

ninamiolane:
@Jules-Deschamps thanks! Note that you will need to fix the github conflicts too (otherwise we can't merge).

Jules-Deschamps:
Should be good now! Had a minor issue with a modification in `init` but I think it's ok now

alebrigant:
Great work !

@Jules-Deschamps did you check the compatibility of the hard coded metric matrices in classes like `DirichletDistributions` with the method `metric_matrix` of `FisherRaoMetric` you implemented ?

To do (from what I've seen):
- remove file `information_geometry.information_manifolds.py`
- address Nina's comment on the docstring of `metric_matrix` method of `FisherRaoMetric` (formula for the definition and reference)
- rename `FisherRaoMetric` in file `normal_distributions.py` to `NormalMetric` (not related to the changes of this PR but avoids that two different classes share the same name)

ninamiolane:
@alebrigant @Jules-Deschamps I am tackling this now.

ninamiolane:
@alebrigant @Jules-Deschamps 

I have done the following modifications:
- [] rename FisherRaoMetric into NormalMetric in normal.py
- [] remove duplicate information_manifoldS.py
- [] add docstring with the formula for fisher rao metric
- [] add `gamma` in all backends
- [] test fisher rao metric by creating its test file and test data file
- [] test the values of fisher rao metric with normal distributions: I had to stop using scipy.stats to get the pdf, because we need automatic differentiation to do fisher rao.
  - here, computing the inner product with the fisher rao metric and with the normal metric give two different results :/

Nevertheless, I would argue that we merge this PR, because it is already giant, and we go solve this last point in a new one.

what do you think?

ninamiolane:
@lpereira95 the failing tests are not related to this PR

```
FAILED tests/tests_geomstats/test_stiefel.py::TestStiefelCanonicalMetric::test_triangle_inequality_of_dist[metric_args1-point_a1-point_b1-point_c1-1e-06]
==== 1 failed, 3853 passed, 2032 skipped, 20 warnings in 194.87s (0:03:14) =====
Error: Process completed with exit code 1.
```
and
```

FAILED tests/tests_geomstats/test_stiefel.py::TestStiefelCanonicalMetric::test_dist_is_positive[metric_args0-point_a0-point_b0-1e-12]
===== 1 failed, 5544 passed, 417 skipped, 11 warnings in 170.73s (0:02:50) =====
Error: Process completed with exit code 1.
```

geomstats/_backend/pytorch/__init__.py
```diff
@@ -31,6 +31,7 @@
 from torch import mean, meshgrid, moveaxis, ones, ones_like, polygamma, quantile
 from torch import repeat_interleave as repeat
 from torch import reshape, stack, trapz, uint8, unique, vstack, zeros, zeros_like
+from torch.special import gammaln  # NOQA
```
lpereira95, line 34: for consistency, let's import as `_gammaln` (for readability, everything  that is public here is assumed to be part of the backed).

geomstats/_backend/tensorflow/__init__.py
```diff
@@ -64,6 +64,7 @@
 imag = _tf.math.imag
 isnan = _tf.math.is_nan
 polygamma = _tf.math.polygamma
+lgamma = _tf.math.lgamma
```
lpereira95, line 67: same as above

geomstats/information_geometry/information_manifold.py
```diff
@@ -0,0 +1,107 @@
+"""Mixin for manifolds of probability distributions."""
```
lpereira95, line 1: For consistency with geometry, shouldn't we call this file `base.py`? It may be more clearer for a user that is something to be inherited instead of instantiated.

geomstats/information_geometry/information_manifold.py
```diff
@@ -0,0 +1,107 @@
...
+            Optional, default: 1.
+
+        Returns
+        -------
+        samples : array-like, shape=[..., n_samples]
+            Sample from the probability distributions.
+        """
+        raise NotImplementedError("The sample method is not yet implemented.")
+
+    def point_to_pdf(self, point):
```
lpereira95, line 31: We can simplify this (and all the inheritances then), by doing:

```python
def point_to_pdf(self, point):
    geomstats.errors.check_belongs(point, self)

   return lambda x: self.pdf(x, point=point)
```

And then each distribution needs to define `pdf(x, point)`.

This way we now `point_to_pdf` is consistent in all distributions by design.

Same applies for `cdf`.

It also simplifies handling of docstrings.

`pdf` and `cdf` don't need necessarily to be public.

geomstats/information_geometry/dirichlet.py
```diff
@@ -16,11 +16,12 @@
 from geomstats.geometry.base import OpenSet
 from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.information_geometry.information_manifold import InformationManifold
 
 N_STEPS = 100
 
 
-class DirichletDistributions(OpenSet):
+class DirichletDistributions(OpenSet, InformationManifold):
```
lpereira95, line 24: May not be very relevant, but is this the order of inheritance we want? I think the opposite may be more robust. Meaning, if it happens (which hopefully not) we have methods with same name in `OpenSet` and `InformationManifold`, then with the current order, `OpenSet` methods have priority.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,157 @@
...
+    ----------
+    information_manifold : InformationManifold object
+        Riemannian Manifold for a parametric family of (real) distributions.
+    support : list, shape = (2,)
+        Left and right bounds for the support of the distribution.
+        But this is just to help integration, bounds should be as large as needed.
+    """
+
+    def __init__(self, information_manifold, support):
+        dim = information_manifold.dim
```
lpereira95, line 27: we may not need to define this variable. just to below `shape=(information_manifold.dim,)`.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,157 @@
...
+            This is the variable whose expectance is the Fisher-Rao information.
+
+            Parameters
+            ----------
+            x : float, shape (,)
+                Point on the support of the distribution
+            """
+            return gs.einsum("...i, ...j -> ...ij", dlog(x), dlog(x))
+
+        metric_mat = quad_vec(
```
lpereira95, line 92: Does this work well with the different backends? e.g. the output is `torch.Tensor` if `torch` backend is used? (looks unlikely.)

I think we may have to consider soon to have a more controlled access to scipy. I have some ideas, we can discuss them later.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,157 @@
...
+            """Compute the square (in matrix terms) of dlog.
+
+            This is the variable whose expectance is the Fisher-Rao information.
+
+            Parameters
+            ----------
+            x : float, shape (,)
+                Point on the support of the distribution
+            """
+            return gs.einsum("...i, ...j -> ...ij", dlog(x), dlog(x))
```
lpereira95, line 90: compute `dlog(x)`  only once, as it is computationally expensive.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,157 @@
...
+                * gs.einsum(
+                    "...ij, ...k -> ...ijk",
+                    gs.autodiff.jacobian(gs.autodiff.jacobian(pdf(x)))(base_point),
+                    gs.autodiff.jacobian(pdf(x))(base_point),
+                )
+                + gs.einsum(
+                    "...i, ...j, ...k -> ...ijk",
+                    gs.autodiff.jacobian(pdf(x))(base_point),
+                    gs.autodiff.jacobian(pdf(x))(base_point),
+                    gs.autodiff.jacobian(pdf(x))(base_point),
```
lpereira95, line 152: do it only once.

geomstats/information_geometry/fisher_rao_metric.py
```diff
@@ -0,0 +1,157 @@
...
+                )
+            )
+        )
+
+        return quad_vec(function, *self.support)[0]
```
lpereira95, line 157: same as above in terms of output type.

geomstats/information_geometry/normal.py
```diff
@@ -11,17 +11,19 @@
...
+class NormalDistributions(PoincareHalfSpace, InformationManifold):
     """Class for the manifold of univariate normal distributions.
 
     This is upper half-plane.
     """
 
     def __init__(self):
         super().__init__(dim=2)
-        self.metric = FisherRaoMetric()
+        self.shape = (2,)
```
lpereira95, line 25: Can be deleted (comes for free by initializing `PoincareHalfSpace`).

geomstats/information_geometry/normal.py
```diff
@@ -11,17 +11,19 @@
...
+        self.metric = NormalMetric()
```
lpereira95, line 26: I think we should change `PoincareHalfSpace`  to accept metric as input as pass this in `super`.

geomstats/information_geometry/normal.py
```diff
@@ -105,7 +109,12 @@ def pdf(x):
                 Points at which to compute the probability density function.
             """
             pdf_at_x = [
-                gs.array(norm.pdf(x, loc=mean, scale=std))
+                gs.array(
```
lpereira95, line 112: can't we code this as a function? looks reusable.

lpereira95 (changes requested):
Thanks @ninamiolane.

Some comments that I believe will improve the code.

(I haven't reviewed the tests yet. I'll look to them after lunch.)

geomstats/_backend/pytorch/__init__.py
ninamiolane: Good point, done.

geomstats/_backend/tensorflow/__init__.py
ninamiolane: Done.

geomstats/information_geometry/normal.py
ninamiolane: Done.

geomstats/information_geometry/dirichlet.py
ninamiolane: Done, I also just read that it is recommended to put the Mixins first:
https://dev.to/bikramjeetsingh/write-composable-reusable-python-classes-using-mixins-6lj

geomstats/information_geometry/fisher_rao_metric.py
ninamiolane: Done.

geomstats/information_geometry/fisher_rao_metric.py
ninamiolane: Yes! Thanks for catching this. Done.

geomstats/information_geometry/fisher_rao_metric.py
ninamiolane: It worked in all backends in the tests from fisher_rao_metric.py
I think it converts the tensors to numpy.
However, it will not to "detach" thus we might get bug there.

Agree that we should think about this more very curious to hear your thoughts on this!

geomstats/information_geometry/fisher_rao_metric.py
ninamiolane: Yes!! Thank you.

ninamiolane: Done.

geomstats/information_geometry/fisher_rao_metric.py
ninamiolane: same answer :))

geomstats/information_geometry/information_manifold.py
ninamiolane: Good point. Done. I also renamed InformationManifold into InformationManifoldMixin to follow scikitlearn conventions

geomstats/information_geometry/normal.py
ninamiolane: Done.

geomstats/information_geometry/normal.py
ninamiolane: PoincareHalfSpace already has its own metric, which the PoincareHalfSpaceMetric from which NormalMetric inherits.
I think we can leave it this way to clarity in the inheritances

geomstats/information_geometry/information_manifold.py
ninamiolane: Good idea, done.

ninamiolane (commented):
Thanks for the great code reviews! I've addressed everything except metric in PoincareHalfSpace (see comment.

ninamiolane:
Failing tests are not related to this PR.

PR #1608, opened 2022-07-22
AAC for graph space

lpereira95:
Implements AAC for graph space following [calissano2020](https://www.mate.polimi.it/biblioteca/add/qmox/14-2020.pdf) and [calissano2022](https://ideas.repec.org/a/eee/jmvana/v190y2022ics0047259x22000021.html) and the corresponding required machinery (e.g. point to geodesic aligner).

PR #1614, opened 2022-08-02
Apply vectorization data strategy to graph space

lpereira95:
Applies concepts introduced in #1613 to simplify graph space tests.

P.S. can only be merged after #1608 and #1613.


lpereira95:
It will be part of #1608.

PR #1615, opened 2022-08-02
Add ability to globally control default dtype

lpereira95:
This PR adds the ability to control default float dtype globally by doing e.g. `gs.set_default_dtype('float64')`.

A battery of tests is added to ensure the functions outputs are according to the default dtype or, when applicable, the inputs dtype.

The main strategy consists in wrapping the backends functions with decorations to:

1. update default `dtype` at "set default dtype time" when function takes `dtype` as input (decorator: `_update_func_default_dtype`)
2. update default `dtype` at running time when, for some reason, function takes `dtype` as input but it is not available in the signature (decorator: `_update_dtype`)
3. cast function output if float type when, for some reason, it is not possible to use default `dtype` as input (decorator: `_cast_fout_from_dtype`)
4. add default dtype to functions that do not have it. In such case, backend output will be casted to desired dtype (decorator: `_add_default_dtype`)
5. cast function output based on input dtype (specially relevant for wrappers around scipy). In this case it is expected that output dtype is same as input dtype. Need for this decorator arises because the expectation is not fulfilled every time (decorator: `_cast_fout_from_input_dtype`). Some care needs to be taken with complex dtype.


ninamiolane (changes requested):
Very nice, thanks!

Two comments:
- could you add detailed docstrings (including parameters' descriptions) even for the private functions that regulate the behavior of the global dtype? It would help understand the method. 
- We still have test_backends2 -> why?

Thank you for this huge contribution !! Can't wait for the green badges :)))))

lpereira95:
@ninamiolane, I think this PR is mostly ready to go.

Overall, there's still some tests that fail once in a while (in some I need to increase tolerances, but in most of them it is really bug related, so I need to review some parts of code to understand the reasons). Most of the times, the tests pass (e.g. running numpy 100x, there's some tests that fail 5x, but not more).

There's still a lot of work to be done in the backends, but I think it should be address in other PRs.

Let me know if the strategy makes sense (I've added some docstrings to help a bit the understanding).

lpereira95:
@ninamiolane, I've update the docstrings. Can you please check and let me know what you feel is still missing?

`test_backends2.py` still exists because I'm slowing migrating tests from `test_backends.py` there.

ninamiolane (approved):
LGTM (as soon as deepsource passes, if the tests can be fixed!).

Thank you so much!!

Issue #1183, opened 2021-10-22
Refactor/clean module of discrete curves


ninamiolane:
There are a few tasks that remain to be done to clean-up or refactor the module of discrete curves:

- [x] Address the comments from @Florent-Michel 's PR #1078 to clean the code:
    - [x] In particular, we need to have consistency between SRVMetric and ElasticMetric(1, 0.5), which is not the case at the moment.
- [ ] Address the problem that "A. Srivastava, E. Klassen, S. H. Joshi and I. H. Jermyn, in  "Shape Analysis of Elastic Curves in Euclidean Spaces," quotient the space of curves by the action of the translation, but our implementation does not (while we are citing their paper).
   - [ ] We could made the quotient by the translation as an option.
- [ ] Use the structure of quotient metric to organize the code.

ninamiolane:
The class QuotientSRVMetric should not inherit from SRVMetric.

Indeed, with this inheritance, whenever a method is not found in QuotientSRVMetric, it is taken from its parent class SRVMetric - which is not correct.

For example, QuotientSRVMetric does not have a `log` implemented, and should not take the `log` from SRVMetric.

ninamiolane:
Same for ClosedSRVMetric. Using QuotientSRVMetric or ClosedSRVMetric within FrechetMean will run, but the estimated mean will not be the frechetmean corresponding to these metrics --- because of this inheritance.

ninamiolane:
`ElasticMetric` does not have log, exp and geodesics, which should be added.

ninamiolane:
'DiscreteCurves' object has no attribute 'random_uniform'

ninamiolane:
DiscreteCurves should also inherit from Landmarks and have a fixed number of sampling points. This would be more consistent with the code base, and avoid math.inf in the dimensions of the manifold.

ninamiolane:
- [ ] SRVMetric should inherit from ElasticCurve
- [ ] srv_transform and srv_transform inverse should overwrite f_transform and f_transform_inverse

ninamiolane:
- [ ] the API of the discrete curves module should be cleaned up: so far, it alternates between calling curves `curve_a, curve_b` and `point_a, point_b` where the latter comes from the consistency with the parent classes which refer to elements on manifolds as "points".

ninamiolane:
PR #1632 addresses the API issue.

ninamiolane:
Copy-pasting here a comment from PR #1632 @alebrigant 

[...]

PR #1632, opened 2022-08-24
Refactor discrete curves using FiberBundle and QuotientMetric 

alebrigant:
This PR addresses issue #1183.

- It uses `FiberBundle` and `QuotientMetric` to structure the code: (1) it defines a new class `SRVShapeBundle` that inherits from `FiberBundle` in which the methods linked to verticality and horizontality (previously in `QuotientSRVMetric`) are moved. (2) It makes `QuotientSRVMetric` inherit from `QuotientMetric`. No changes were made to the classes `FiberBundle` and `QuotientMetric`, except some renaming. 

- It cleans the API for the discrete curves module.


geomstats/geometry/discrete_curves.py
```diff
@@ -1772,66 +1743,73 @@ def space_derivative(curve):
         return space_deriv
 
 
-class QuotientSRVMetric(SRVMetric):
-    """Metric on shape space induced by the SRV metric.
+class SRVShapeBundle(DiscreteCurves, FiberBundle):
```
ninamiolane, line 1746: NIT (not necessarily for this PR): I wonder if we should put the name of the group inducing the quotient in the name of the class.

We could have a SRVShapeBundle only by quotienting the action of rotations.

Actually, in these shape manifolds, we will have a series of quotient, for each additional group applied.

Thus: `SRVRotationBundle`, `SRVParamBundle`?

--> I'll ask the question in our issue on DiscreteCurves.

geomstats/geometry/discrete_curves.py
```diff
@@ -1694,17 +1665,17 @@ def path(t):
             t = gs.cast(t, gs.float32)
             t = gs.to_ndarray(t, to_ndim=1)
             t = gs.to_ndarray(t, to_ndim=2, axis=1)
-            new_initial_curve = gs.to_ndarray(initial_curve, to_ndim=curve_ndim + 1)
+            new_initial_point = gs.to_ndarray(initial_point, to_ndim=point_ndim + 1)
             new_initial_tangent_vec = gs.to_ndarray(
-                initial_tangent_vec, to_ndim=curve_ndim + 1
+                initial_tangent_vec, to_ndim=point_ndim + 1
```
ninamiolane, line 1670: Thank you so much for the renamings curve -> point, it was overdue!

We should also think about a strategy on how to handle these APIs:
- from the codebase perspective, we want to use `point` everywhere,
- from the user's perspective, they probably want to see `curve`
---> We could write an "API remapping" for the manifolds where this happens.

geomstats/geometry/discrete_curves.py
```diff
@@ -1887,18 +1865,40 @@ def split_horizontal_vertical(self, tangent_vec, curve):
...
         tangent_vec_ver = gs.squeeze(tangent_vec_ver)
-        tangent_vec_hor = tangent_vec - tangent_vec_ver
+        if return_norm:
+            return tangent_vec_ver, vertical_norm
+
+        return tangent_vec_ver
 
-        return tangent_vec_hor, tangent_vec_ver, vertical_norm
+    def horizontal_projection(self, tangent_vec, point):
+        """Compute vertical part of tangent vector at base point.
```
ninamiolane, line 1881: NIT: Compute horizontal part

ninamiolane (approved):
LGTM, thank you very much! Just one docstring where vertical --> horizontal, and then one question about naming that we do not have to tackle now, I'll put it on our issue on the refactor of discrete curves :))

ninamiolane:
Merging now, to avoid github conflicts since it's been 7 days, I'll fix the docstring typo.

geomstats/geometry/discrete_curves.py
alebrigant: Yes I was also thinking of a similar "API remapping" solution, I completely agree it would be helpful.

PR #1633, opened 2022-08-26
Create multinomial distributions

alebrigant:
This PR creates the classes `MultinomialDistributions` and `MultinomialMetric`.

The manifold `MultinomialDistributions` is a generalization of `CategoricalDistributions` with an additional parameter `n_draws`.  The parameter manifold stays the same (the simplex). The `MultinomialMetric` is the `CategoricalMetric` multiplied by `n_draw`. The code related to categorical distributions has been moved to the classes related to multinomial distributions, with this additional parameter. The classes related to categorical distributions now inherit from those related to multinomial distributions.


ninamiolane (approved):
LGTM! Only comment: do we want to add a test_multinomial.py that would test something specific to the multinomial case (or we instead assume that the tests of the child classes are enough?).

alebrigant:
> LGTM! Only comment: do we want to add a test_multinomial.py that would test something specific to the multinomial case (or we instead assume that the tests of the child classes are enough?).

I added one test that is specific to the multinomial case, that uses the attribute `n_draws` and checks compatibility with the case `n_draws=1`. Then it remains to test the case `n_draws=1`, which is what is already tested in `TestCategoricalDistributions`.

PR #1644, opened 2022-09-16
Make default_point_type a property

lpereira95:
This PR makes `default_point_dtype` a property (computed from shape).

`shape` has been corrected for several spaces and/or metrics and a general test to validate shape against the shape of a generated random point was added.

### DIFF
--- a/examples/plot_kmeans_manifolds.py
+++ b/examples/plot_kmeans_manifolds.py
@@ -44,7 +44,7 @@ def kmean_poincare_ball():
         space="H2_poincare_disk",
         marker=".",
         color="black",
-        point_type=manifold.coords_type,
+        coords_type=manifold.default_coords_type,
     )
 
     for i in range(n_clusters):
@@ -54,7 +54,7 @@ def kmean_poincare_ball():
             space="H2_poincare_disk",
             marker=".",
             color=colors[i],
-            point_type=manifold.coords_type,
+            coords_type=manifold.default_coords_type,
         )
 
     ax = visualization.plot(
@@ -64,7 +64,7 @@ def kmean_poincare_ball():
         marker="*",
         color="green",
         s=100,
-        point_type=manifold.coords_type,
+        coords_type=manifold.default_coords_type,
     )
 
     ax.set_title("Kmeans on Poincaré Ball Manifold")
--- a/examples/plot_kmeans_manifolds.py
+++ b/examples/plot_kmeans_manifolds.py
@@ -44,7 +44,7 @@ def kmean_poincare_ball():
         space="H2_poincare_disk",
         marker=".",
         color="black",
-        point_type=manifold.point_type,
+        point_type=manifold.coords_type,
     )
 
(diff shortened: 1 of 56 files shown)


lpereira95:
@ninamiolane failures are unrelated with this PR. In fact, the failure in the example is related with last release of matplotlib (to be address in another PR).

examples/plot_kmeans_manifolds.py
```diff
@@ -44,7 +44,7 @@ def kmean_poincare_ball():
         space="H2_poincare_disk",
         marker=".",
         color="black",
-        point_type=manifold.point_type,
+        point_type=manifold.coords_type,
```
ninamiolane, line 47: Why the change of name?

point_type is a purely computational attribute, but coords_type often refers to why coordinate system can be used on a manifold, which is a mathematical attribute.

ninamiolane (changes requested):

geomstats/geometry/_hyperbolic.py
```diff
@@ -23,7 +23,7 @@ class _Hyperbolic:
     ----------
     dim : int
         Dimension of the hyperbolic space.
-    point_type : str, {'extrinsic', 'intrinsic', etc}
+    coords_type : str, {'extrinsic', 'intrinsic', etc}
```
ninamiolane, line 26: Ah, never mind, this is indeed a coords_type 🤔

geomstats/geometry/hyperboloid.py
```diff
@@ -39,10 +39,8 @@ class Hyperboloid(_Hyperbolic, LevelSet):
         Optional, default: 1.
     """
 
-    default_coords_type = "extrinsic"
-    default_point_type = "vector"
-
     def __init__(self, dim, coords_type="extrinsic", scale=1, **kwargs):
+        # TODO: coords_type vs default_coords_type?
```
ninamiolane, line 43: left over comment: address in this PR?

examples/plot_kmeans_manifolds.py
ninamiolane: Nevermind, it seems that point_type was wrongfully used here: it is indeed a coords_type

ninamiolane:
Great, thank you! Why not removing default_point_type and point_type completely? (instead of making it a property?)

It could be simpler to have only one variable ("shape") referring to that one concept.

lpereira95:
I think it may be valuable to keep `default_point_type`, because its current use case is the bool `self.default_point_type == "vector"` or `self.default_point_type == "matrix"`, which in my opinion is more readable than `len(self.shape) == 1` or `len(self.shape)>1`. But if you think it is better to use `len(self.shape)`, it is completely fine to me.( Anyway, `default_point_type` is never assigned, except for the product manifold case.)

Regarding `default_point_type`/`point_type` and `default_coords_type`/`coords_type`, I think it needs homogenization (remove `default`?). I guess this naming exists due to some history and the distinction is not worthy anymore.

PR #1646, opened 2022-09-16
Simplify super()

lpereira95:
Simplify uses of `super()` by making it more python3 friendly (basically remove class name and self).

Can only be merged after #1644 as it builds on top of it (to avoid merge conflicts).


ninamiolane (approved):
I thought we needed the super(ClassName, self) instead of super() for some reason but I cannot remember: multiple inheritance problems, just a legacy of older ways of writing in python?

I can't remember though.

Thus, good for me!

PR #1189, opened 2021-11-08
Improve MDM, 2

qbarthelemy:
I am sorry, but #1188 has been merged too quickly for me: I had not commited my last developments.

PR #1647, opened 2022-09-17
Improve MDM, adding weights and removing n_classes from parameters

qbarthelemy:
## Description

Class parameter `n_classes` is not necessary to fit the model: it is estimated from labels of training set.
It is an attribute, not a parameter.

This PR:
- ~changes n_classes from parameters to attributes;~
- removes n_classes from parameters;
- add weights for samples in `fit`;
- add inheritance from `TransformerMixin`, `ClassifierMixin`, `BaseEstimator`;
- updates tests.

I don’t know the depreciation policy of geomstats, but, at the risk of breaking the API, we might as well simplify the very long name `RiemannianMinimumDistanceToMeanClassifier` into `MinimumDistanceToMeanClassifier` or `MinimumDistanceToMean`.

## Issue

Related to #1188 and #1189.


geomstats/learning/mdm.py
```diff
@@ -11,26 +11,25 @@
...
         Riemannian metric to be used.
-    n_classes : int
-        Number of classes.
 
     Attributes
     ----------
-    mean_estimates_ : list
-        If fit, centroids computed on training set.
+    n_classes_ : int
+        If fit, number of classes.
```
lpereira95, line 28: Do you think this attribute is necessary? I would simply remove it as we can simply compute this variable from `len(estimator.mean_estimates_)` or, simpler, `len(estimator.classes_)`

lpereira95 (approved):
Thanks @qbarthelemy, your cleanings have been very helpful.

I've just updated a notebook that was breaking due to the changes.

I think changing the name to `MinimumDistanceToMean` is a good idea. Do you agree with this @ninamiolane?

geomstats/learning/mdm.py
qbarthelemy: You are right: we can remove this parameter.

geomstats/learning/mdm.py
```diff
@@ -4,33 +4,36 @@
...
 
 import geomstats.backend as gs
+from geomstats.learning._template import TransformerMixin
 from geomstats.learning.frechet_mean import FrechetMean
 
 
-class RiemannianMinimumDistanceToMeanClassifier:
-    r"""Minimum Distance to Mean (MDM) classifier on manifolds.
+class RiemannianMinimumDistanceToMeanClassifier(
+    TransformerMixin,
```
lpereira95, line 16: We haven't defined a `.transform` method, so this mixin will not work properly.

lpereira95 (changes requested):
@qbarthelemy, can you please address my last comments?

PR #1651, opened 2022-09-21
Add gs.matrices and simplify Matrices

lpereira95:
This is is prototype PR which goal is to show that having a module `gs.matrices` may be more interesting than having a bunch of static methods in `geomstats.geometry.matrices.Matrices`.

Basically, I moved most of the methods of `Matrices` to `gs.matrices`.

Why do I think this solution is better:
* `gs.matrices` will be documented within the backends docs (still underdeveloped), so it will be more obvious that we have a bunch of nice methods implemented for matrices (which now are a bit hidden within `Matrices`)
* duplication with backend will be easier to identify
* imports in geometry will be less entangled (which will, probably, allow us to simplify imports)
* tests will be simplified as they will be backend tests, not the tests of a space (this still needs to be performed)
* `Matrices` becomes much leaner and defines methods that are really related with geometry and less with linear algebra
* inheritance from `Matrices` becomes cleaner 
* is is more natural for python users (e.g. `gs.matrices.mul` is more common than `geomstats.geometry.matrices.Matrices.mul`)

Some disadvantages:
* some methods in `Matrices` or `MatricesMetric` will simply call `gs.matrices` ones

What you think @nguigs, @ninamiolane?


ninamiolane:
I like it! The disadvantage you cite:
> some methods in Matrices or MatricesMetric will simply call gs.matrices ones
is actually a strength in my perspective, and very important to keep.

For the following reason:
Even if the core code will be in the backend and gs.matrices, users will still have the option to call matrices.compose if they wish to. 

This is important because it is consistent with how other manifolds are used, and is crucial for the following use case.

If a user wants to test a statistical algorithm over many Lie groups, including matrices, they can do:
```
for lie_group in lie_group_list:
    lie_group.compose
```
without having to put an `if` statement to treat matrices independently.

PR #1658, opened 2022-09-30
Add BHVSpace and WaldSpaceMetric

luisfpereira:
This PR ports `WaldSpaceMetric` from `waldspaces` (developed by @JonasLueg) to `geomstats`.


geomstats/geometry/stratified/wald_space.py
```diff
@@ -47,949 +15,973 @@ class :class:`spd.SPDMatrices`.
...
+)
+from geomstats.geometry.stratified.vectorization import (
+    _manipulate_output,
+    broadcast_lists,
+    vectorize_point,
+)
+from geomstats.numerics.optimizers import ScipyMinimize
+
+
+def _manipulate_input_with_array(arg):
```
ninamiolane, line 44: docstring?

geomstats/geometry/stratified/wald_space.py
```diff
@@ -47,949 +15,973 @@ class :class:`spd.SPDMatrices`.
...
+
+    return arg, False
+
+
+def _manipulate_output_wald(out, to_list):
+    return _manipulate_output(out, to_list, manipulate_output_iterable=WaldCollection)
+
+
+def make_splits(n_labels):
+    """Generate all possible splits of a collection."""
```
ninamiolane, line 61: parameter description missing

geomstats/geometry/stratified/wald_space.py
```diff
@@ -47,949 +15,973 @@ class :class:`spd.SPDMatrices`.
...
+    if n_labels <= 1:
+        raise ValueError("`n_labels` must be greater than 1.")
+    if n_labels == 2:
+        yield Split(part1=[0], part2=[1])
+    else:
```
ninamiolane, line 66: is else needed after yield?

geomstats/geometry/stratified/point_set.py
```diff
@@ -1,91 +1,61 @@
...
+        is_equal : array-like, shape=[...]
+        """
 
-class Point(ABC):
-    r"""Class for points of a set."""
 
-    @abstractmethod
-    def __repr__(self):
-        """Produce a string with a verbal description of the point."""
+class PointCollection(ABC, list):
```
ninamiolane, line 30: Explain what this is, explain that this is the analogous of a batch.
Maybe rename it PointBatch?

ninamiolane (changes requested):
What is the plan to use the learning algorithms (that are compatible with distance-only structures) on these spaces?

NIT: not sure we need `_space.py` in file names

geomstats/geometry/stratified/wald_space.py
luisfpereira: Yes, otherwise also runs code after else.

luisfpereira:
@ninamiolane, I'm merging this one now as, while probably this still needs some refinement, it is in a much better state than we have available in main. Let's revisit it if you disagree with something.

PR #1663, opened 2022-10-03
separate template for spaces and metrics

nanjekyejoannah:
Based on what we discussed this is the new PR, but I am still confused about what the new template should have.


lpereira95:
Will come back here later after the refactoring decided during the hackathon is finished.

PR #1664, opened 2022-10-05
Fix bug in the autodiff of pullback metric matrix

ninamiolane:
This PR:
- overwrites the method `inner_product_derivative_matrix` in PullbackMetric, childclass of RiemannianMetric
to fix a bug in the automatic differentiation required in this method,
- adds corresponding unit tests.

The parallel transport of the pullback metric is extremely slow, thus its test is skipped in autograd and tensorflow to avoid slowing down the CI.

The bug:
In pullback metric, the metric matrix comes from the automatic differentiation of the immersion defining the pullback.
Differentiating again the result of this differentiation was creating an error.

The fix:
In the case of the pullback metric, we have a closed form for the derivative of the metric matrix that we can use.
This closed form requires the computation of the hessian of the immersion.
For this reason, `hessian` has been added in the autodiff backends.

## Checklist

- [ ] If neccessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [ ] I have linked to issues and PRs that are relevant to this PR.

### DIFF
--- a/geomstats/_backend/autograd/__init__.py
+++ b/geomstats/_backend/autograd/__init__.py
@@ -58,7 +58,6 @@ from autograd.numpy import (
     shape,
     sort,
     split,
-    squeeze,
     stack,
     std,
     sum,
@@ -134,6 +133,15 @@ tanh = _box_unary_scalar(target=_np.tanh)
 arctan2 = _box_binary_scalar(target=_np.arctan2)
 mod = _box_binary_scalar(target=_np.mod)
 power = _box_binary_scalar(target=_np.power)
+
+
+def squeeze(x, axis=None):
+    if axis is None:
+        return _np.squeeze(x)
+    x_shape = x.shape
+    if x_shape[axis] != 1:
+        return x
+    return _np.squeeze(x, axis=axis)
 
 
 def angle(z, deg=False):
--- a/geomstats/_backend/autograd/__init__.py
+++ b/geomstats/_backend/autograd/__init__.py
@@ -138,8 +138,7 @@ power = _box_binary_scalar(target=_np.power)
 def squeeze(x, axis=None):
     if axis is None:
         return _np.squeeze(x)
-    x_shape = x.shape
-    if x_shape[axis] != 1:
+    if x.shape[axis] != 1:
         return x
     return _np.squeeze(x, axis=axis)
 
--- a/geomstats/_backend/autograd/autodiff.py
+++ b/geomstats/_backend/autograd/autodiff.py
@@ -1,5 +1,6 @@
 """Wrapper around autograd functions to be consistent with backends."""
 
+from autograd import hessian as _hessian
 from autograd import jacobian as _jacobian
 from autograd import value_and_grad as _value_and_grad
 from autograd.extend import defvjp as _defvjp
@@ -107,6 +108,23 @@ def jacobian(func):
         the jacobian of func at x.
     """
     return _jacobian(func)
+
+
+def hessian(func):
+    """Return a function that returns the hessian of func.
+
+    Parameters
+    ----------
+    func : callable
+        Function whose Hessian is computed.
+
+    Returns
+    -------
+    _ : callable
+        Function taking point as input and returning
+        the hessian of func at point.
+    """
+    return _hessian(func)
 
 
 def value_and_grad(func, to_numpy=False):
--- a/geomstats/_backend/autograd/autodiff.py
+++ b/geomstats/_backend/autograd/autodiff.py
@@ -1,7 +1,6 @@
 """Wrapper around autograd functions to be consistent with backends."""
 
-from autograd import hessian as _hessian
-from autograd import jacobian as _jacobian
+from autograd import hessian, jacobian
 from autograd import value_and_grad as _value_and_grad
 from autograd.extend import defvjp as _defvjp
 from autograd.extend import primitive as _primitive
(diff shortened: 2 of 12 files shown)


geomstats/geometry/pullback_metric.py
```diff
@@ -111,6 +112,47 @@ def pickable_inner_product(i, j):
...
+            Base point.
+            Optional, default: None.
+
+        Returns
+        -------
+        inner_prod_deriv_mat : array-like, shape=[..., dim, dim, dim]
+            Inner-product derivative matrix.
+        """
+        hessian_aij = []
+        jacobian_ai = []
```
lpereira95, line 130: Since we need the jacobian and the hessian, I think we should add a function to the backends called e.g. `jacobian_and_hessian` (same idea as e.g. `value_and_grad`), because the hessian needs to compute the jacobian and when we need both, we are basically repeating a computation that we have already done (we are committing this "repeating" error in `value_and_grad`... I'm preparing a PR to fix it).

For `autograd` it should be simple, because the hessian is simply implemented as: `jacobian(jacobian(fun))(x)`.

For `tensorflow` it is similar, because we are the ones computing the jacobians.

For `torch` it may require a little more thought, as they have additional arguments to (supposedly) run things faster, but it should not be very difficult (they're code is relatively straighforward).

What do you think?

ninamiolane:
@lpereira95 the failing tests in pytorch are not related to this PR.

geomstats/_backend/autograd/__init__.py
```diff
@@ -136,6 +135,15 @@
 power = _box_binary_scalar(target=_np.power)
 
 
+def squeeze(x, axis=None):
```
lpereira95, line 138: We should probably do the same in `numpy.squeeze`.

geomstats/_backend/autograd/__init__.py
```diff
@@ -136,6 +135,15 @@
...
+    if axis is None:
+        return _np.squeeze(x)
+    x_shape = x.shape
+    if x_shape[axis] != 1:
```
lpereira95, line 142: Do we need to create the variable x_shape? Why not `x.shape[axis] != 1`?

lpereira95 (changes requested):
Thanks @ninamiolane!

These `autodiff` related things are very interesting. I'm having some troubles with them while developing the "solvers" PR (there's some inconsistencies, performance problems). I think it may also be a great topic for discussion in the hackthon.

geomstats/_backend/autograd/autodiff.py
```diff
@@ -109,6 +110,23 @@ def jacobian(func):
...
+
+    Returns
+    -------
+    _ : callable
+        Function taking point as input and returning
+        the hessian of func at point.
+    """
+    return _hessian(func)
+
+
```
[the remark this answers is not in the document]
ninamiolane: Indeed! Done, I also did it for the `jacobian`.

geomstats/geometry/pullback_metric.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         tangent_immersion=None,
     ):
         super().__init__(dim=dim)
+        self.embedding_dim = embedding_dim
```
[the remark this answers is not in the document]
ninamiolane: True! Done.

geomstats/_backend/autograd/__init__.py
ninamiolane: True, done!

geomstats/_backend/autograd/__init__.py
ninamiolane: Good eye, that was a leftover. Done.

geomstats/_backend/tensorflow/autodiff.py
```diff
@@ -163,3 +163,49 @@ def jac(x):
...
+        -------
+        _ : array-like
+            Value of the hessian of func at x.
+        """
+        # Note: this is a temporary implementation
+        # that uses the jacobian of the gradient.
+        # inspired from https://github.com/tensorflow/tensorflow/issues/29781
+        # waiting for the hessian function to be implemented in GradientTape.
+        if isinstance(x, _np.ndarray):
+            x = _tf.Variable(x)
```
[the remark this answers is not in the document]
ninamiolane: There is some optimization that is done with scipy, and for which we convert back to numpy... :/
Not the best, I agree. Probably not something to tackle in this PR though?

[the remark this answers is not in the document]
lpereira95: OK, let's leave it for later. (btw, I've developed a solution for this `scipy.minimize` conversion problem. It is in the (new) PR for the solvers (which I haven't opened yet) - since some of them are basically optimization problems.)

geomstats/geometry/pullback_metric.py
ninamiolane: Good point. I've implemented it for autograd and tensorflow (and numpy, returning the error message).

I left a placeholder for torch, which basically calls jacobian and hessian separately. I'll leave that one for future optimization if that is fine with you. I will raise an issue.

geomstats/_backend/tensorflow/autodiff.py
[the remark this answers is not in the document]
ninamiolane: Awesome!!! :D :D

ninamiolane:
Failing tests are not related to this PR:
```
FAILED tests/tests_geomstats/test_stiefel.py::TestStiefelCanonicalMetric::test_triangle_inequality_of_dist[metric_args0-point_a0-point_b0-point_c0-1e-12]
===== 1 failed, 5533 passed, 415 skipped, 11 warnings in 144.84s (0:02:24) =====
```

PR #1667, opened 2022-10-05
Improve jacobian_and_hessian in autograd

lpereira95:
Adds to #1664.

@ninamiolane, if I'm not overlooking anything, the suggested implementation reduces the number of computations of `jacobian(x)` by one. The jacobian of `x` is still computed twice: one within `jacobian(func)(x)` and another in `_value_and_jacobian`, but I think it is just the way automatic differentiation works (at least in `autograd`).

### DIFF
--- a/geomstats/_backend/autograd/autodiff.py
+++ b/geomstats/_backend/autograd/autodiff.py
@@ -1,5 +1,7 @@
 """Wrapper around autograd functions to be consistent with backends."""
 
+import autograd as _autograd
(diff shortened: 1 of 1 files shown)


ninamiolane:
LGTM, but why aren't tests triggered?

lpereira95:
Made a dummy commit to check if tests would trigger. They've triggered now. Hard to understand these workflows sometimes.

Let's see if everything passes smoothly and then we merge.

PR #1676, opened 2022-10-17
Fix wrong scaling in NormalMetric

alebrigant:
This PR:

- redefines the metric on normal distributions using `PullbackDiffeoMetric` 
- changes the scaling of the embedding metric `PoincareHalfSpaceMetric` to 2.


ninamiolane (approved):
LGTM! Thanks for fixing this.

PR #1677, opened 2022-10-17
Consistency fix in visualization of Sphere/Circle

ballerin:
## Description

The visualization library for the Circle and the Sphere is inconsistent in the use of ```gs.linspace``` and ```gs.arange``` in the creation of the mesh.

In addition there is an overlap in the mesh in the circle class (n_angles are actually n_angles-1 unique angles, since two are overlapping) and there is a visible gap in both meridian and latitude circles due to the ```gs.arange``` function excluding the end value.

According to the numpy [1.23] documentation:
> When using a non-integer step, such as 0.1, it is often better to use ```numpy.linspace```.

therefore ```gs.linspace``` is used for consistency.

### DIFF
--- a/geomstats/visualization/hypersphere.py
+++ b/geomstats/visualization/hypersphere.py
@@ -16,7 +16,7 @@ class Circle:
     """Class used to draw a circle."""
 
     def __init__(self, n_angles=100, points=None):
-        angles = gs.linspace(0, 2 * gs.pi, n_angles + 1)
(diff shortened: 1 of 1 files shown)


lpereira95 (changes requested):
Thanks @ballerin (it is a small contribution, but helps you out understanding how all the process of opening a PR and merging code works).

There's an issue arising in tensorflow coming from your change. The issue is probably related with some inconsistency in the backend. We can take a look to it tomorrow.

Issue #1680, opened 2022-10-17
Design: Disentangle concepts of Abstract Classes and Factories


ninamiolane:
### What should be modified?

Following discussions with @ymontmarin , @alebrigant , @sophiaas  : 

We currently have two designs in Geomstats when we try to create a new manifold:
- either we have to create a new class that subclasses an AbstractClass and overwrites its abstract methods (for example in RiemannianMetric)
- or we do not create a class, but we use the instantiator of the AbstractClass and pass it some functions as inputs, effectively using the AbstractClass as a factory. (for example LevelSet)

It would be nice to decide on a clear design of what it means to create a new manifold, (i) either as a child class of an Abstract class, or (ii) using a class as a factory.

From our discussions, Yann, Sophia, Alice and me seem to prefer subclassing an AbstractClass.

ninamiolane:
Interesting use case coming from Juliane: Sometimes, we might have a Riemannian metric for which the exp and log are known, but not the metric matrix nor inner-product. 

Which means that we do not want to force each child of Riemannian metric to overwrite inner-product and metric_matrix nec essarily.

ymontmarin:
Hummm

PR #1681, opened 2022-10-17
Make submersion and tangent_submersion abstract in LevelSet

lpereira95:
Partially answers #1680 for `LevelSet`.

Missing:
* docstrings

Need to discuss:
* should value somehow also be forced to be defined or passing as argument to parent works fine? (I think it should be OK as argument)
* is the API naming OK? e.g. `point` and `vector`, `point`, respectively?

@ninamiolane @ymontmarin


ymontmarin:
Looks promising !!!

- I think `embedding_space` should follow the same logic, as you know the submersion and its tangent before any instantiation you know the space in which it submerse.

maybe with something like 
```
@abs.abstract
def define_embedding_space(self):
    return

@property
def embedding_space(self):
    if not hasattr(self, '_embedding_space'):
         self._embedding_space = self.define_embedding_space()
    return self._embedding_space
```

- Maybe tangent does not necessary needs to be abstract  and we can implement a default autodiff implementation instead, for class with no close form of the tangent immersion

Let us discuss !

geomstats/geometry/base.py
```diff
@@ -324,24 +324,23 @@ class LevelSet(Manifold, abc.ABC):
...
         self.embedding_metric = embedding_space.metric
-        self.submersion = submersion
         if isinstance(value, float):
             value = gs.array(value)
         self.value = value
-        self.tangent_submersion = tangent_submersion
+
+    @abc.abstractmethod
+    def submersion(self, point):
+        pass
```
ninamiolane, line 339: Do you need `pass` ? I thought abstract methods can be empty. However, it might be better to keep pass so that child classes do not necessarily have to provide these? (see my last comment on the related github issue).

ninamiolane (approved):
This is great! I also like the fact that having them as abstract method enforces a consistent API of these functions.

ninamiolane:
Re: "should value somehow also be forced to be defined or passing as argument to parent works fine? (I think it should be OK as argument)"

I think that value could be hardcoded to 0. because the submersion can always be redefined (by subtracting `value`) to have value zero.

ninamiolane:
Another inconsistency in `LevelSet`: it has default_coords_type being intrinsic, but it uses the `shape` of the embedding space.

ninamiolane:
Another comment: tangent_submersion should me a method but not be abstract but it does not have to be overwritten. If it is not provided, then we use gs.autodiff.jacobian(submersion). <--- that piece of code would go in the LevelSet method.

geomstats/geometry/base.py
ymontmarin: I don't know the rule for abc.abstractmethod

ymontmarin:
> Another inconsistency in `LevelSet`: it has default_coords_type being intrinsic, but it uses the `shape` of the embedding space. The docstring also mention that the shapes of points should be [..., dim] which seems to refer to intrinsic coordinates.

Maybe my question is stupid but: we agree we have intrinsic => `shape = [..., dim]`, but what about the other way around ? I guess there is some case in Poincaré disk where extrinsinc coordinate are also `shape = [..., dim]` ? Should we do a check.

ymontmarin:
> Another comment: tangent_submersion should me a method but not be abstract but it does not have to be overwritten. If it is not provided, then we use gs.autodiff.jacobian(submersion). <--- that piece of code would go in the LevelSet method.

I think it is what is written in `PullbackDiffeo`

ymontmarin:
> Another question: we want to put `define_embedding_space` to prevent the users to change the embedding space once the object (say sphere) of a class (say Hypersphere) subclassing LevelSet has been instantiated.
> 
> But what prevents users to still do "sphere.embedding_space = new_space" again?

Because `embedding_space` will be a property without a setter hook, so it can't be set by the user

geomstats/geometry/base.py
lpereira95: After adding docstrings it will be possible to have no code, but not without.

lpereira95:
@ninamiolane @ymontmarin I think this may be ready. A new review from your side may be worthwhile, as several things have changed. I think it implements "our vision".

geomstats/geometry/base.py
```diff
@@ -320,32 +320,33 @@ class LevelSet(Manifold, abc.ABC):
...
-        dim,
-        embedding_space,
-        submersion,
-        value,
-        tangent_submersion,
-        default_coords_type="intrinsic",
-        **kwargs
-    ):
-        kwargs.setdefault("shape", embedding_space.shape)
+    def __init__(self, dim, default_coords_type="intrinsic", **kwargs):
```
ninamiolane, line 328: I think the default should be extrinsic

geomstats/geometry/base.py
```diff
@@ -320,32 +320,33 @@ class LevelSet(Manifold, abc.ABC):
...
-        self.value = value
-        self.tangent_submersion = tangent_submersion
+
+    @abc.abstractmethod
+    def _define_embedding_space(self):
+        pass
+
+    @abc.abstractmethod
+    def submersion(self, point):
+        """Submersion that defines the manifold."""
```
ninamiolane, line 338: Maybe add that the value should be zero, potentialy with a equation: submersion(x) = 0 defines the manifold

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -22,17 +22,41 @@ class FullRankCorrelationMatrices(LevelSet):
...
+        ----------
+        point : array-like, shape=[..., n, n]
+
+        Returns
+        -------
+        submersed_point : array-like, shape=[..., n]
+        """
+        return Matrices.diagonal(point) - gs.ones(self.n)
+
+    def tangent_submersion(self, vector, point):
```
ninamiolane, line 46: Yesss, that feels much cleaner!

geomstats/geometry/grassmannian.py
```diff
@@ -184,16 +147,70 @@ def __init__(self, n, p, **kwargs):
...
-            tangent_submersion=lambda v, x: 2
-            * Matrices.to_symmetric(Matrices.mul(x, v))
-            - v,
-            **kwargs
+        super().__init__(dim=dim, **kwargs)
+
+    def _define_embedding_space(self):
+        return SymmetricMatrices(self.n)
+
+    def submersion(self, point):
```
ninamiolane, line 155: Yess !

geomstats/geometry/special_euclidean.py
```diff
@@ -242,6 +196,58 @@ def __init__(self, n, **kwargs):
         if self._metric is None:
             self._metric = self.left_canonical_metric
 
+    def _define_embedding_space(self):
+        return GeneralLinear(self.n + 1, positive_det=True)
+
+    def submersion(self, point):
+        """Define SE(n) as the pre-image of identity.
```
ninamiolane, line 203: This should be the preimage of 0 by ....

geomstats/geometry/special_euclidean.py
```diff
@@ -242,6 +196,58 @@ def __init__(self, n, **kwargs):
         if self._metric is None:
             self._metric = self.left_canonical_metric
 
+    def _define_embedding_space(self):
+        return GeneralLinear(self.n + 1, positive_det=True)
+
+    def submersion(self, point):
```
ninamiolane, line 202: Yesss !

ninamiolane (approved):
Very nice, thank you! I've just put minor comments. Maybe double check the docstrings of the submersions so that they match the fact that the value is now 0 (I've put one comment to that effect in SE(n).

I'm still not a huge fan of the _define_embedding_space, because:
- this would be the only method across all manifolds that is "talking to the user" (with the keyword "define") and this feels wrong. All the other methods are purely impersonal and geometry-facing.
- if it's main role is to prevent the user from changing the embedding_space attribute, then this goes against our vision of "not preventing anything, but letting the user use the manifolds at their own risk". This would also be prevented by the new gold rule which is "never modify the attributes of a manifold, unless it's its numerics objects (ExpSolver)".

But I forget the details of our debate about this : happy to go with it if I'm forgetting important aspect of this choice.

geomstats/geometry/base.py
lpereira95: Completely agree! (I think we have inconsistencies in some parts of the library due to this) Later I guess we are changing this to `intrisinc = False` (see #1725).

lpereira95:
@ninamiolane, regarding `_define_embedding_space` I would say the main strength of this implementation is that it **forces** the contributor to define an embedding space (since we are asking `_define_embedding_space` as abstract method). In other words, the "contract" for creating a new child of `LevelSet` is very clear for the contributor.

I can only see an alternative to this, but without the strength we have above (i.e. a less knowledgeable contributor may not know he/she has to define `embedding_space`):

* embedding space is define in `__init__` as `self.embedding_space = ...`


The alternative of having a property (or function) simply called `embedding_space` does not work because it would instantiate the embedding space everytime we do `space.embedding_space`.


Do you have any alternatives in mind? Would we prefer the alternative I mentioned to the current implementation?

ninamiolane:
I would indeed prefer the alternative of the attribute. To tell the less knowledgeable contributor he/she has to define embedding_space, we could have a clear error message: the first time self.embedding_space is needed, if it is not defined then the error tells the user to define it.

I know we talked a lot about this already. Maybe we could poll everybody at the next meeting to get a vote?

ymontmarin:
From what i remember from the discussion the starting point was that in python methods implementation are defined when child class are written, meaning it is decided before any instantiation. Whereas argument of `__init__` may not (it allows definition to be outside the object). In addition, having a method defining an attribute only once ensure it canot be modified in the same way method should not be modify.

As we discuss, it is more natural to have immersion (and its derivative) as a method, since it is a function anyway, and it is known when child class are written. Mathematically, the immersion is the tuple $(\phi, E)$ so it makes sense the the embedding space should be at the same level as the immersion, hence a method.

But if we step back, what we really want is that `embedding_space` implementation is written when a child clas is written, at the same moment as the immersion (it is weird to require that immersion should be implemented when child class is written and not requiring the definition of the space it goes in).
To be more precise, in the current implementation it is possible to have code such as:
```
class Child(LevelSet):
    def immersion(self, x):
          XXX

e = Euclidean(n)
a = Child(dim=m, embedding_space=e)
```
which is pretty weird to me, because immersion implementation is written inside the object, and the space it goes to is written outside.
The most natural way of ensuring this cannot happen, is to use method (or some python dark magic) and mark it as abstract.

But I agree with @ninamiolane that a method `_define_embedding_space` feel weird and different. But actually, I do not see the word define as a name talking to the coder, but as a description of what happen in the method, something like `_instantiate_embedding_space` is more accurate. Still, it is different from the other methods name that are mathematic. So lexically i totally agree that it does not feel right !

So for a better use experience, it could be a property ! But then the instantiation would occur every time the attribute is asked.

A python magic that can work is this:
```python
class LevelSet(ABC):
    def __init__(self, dim, default_coords_type="intrinsic", **kwargs):
        super().__init__(dim=dim, default_coords_type=default_coords_type, **kwargs)

    def __getattribute__ (self, name):
        if name == 'embedding_space':
            if not hasattr(self, '_embedding_space'):
                self._embedding_space = super().__getattribute__ ('embedding_space')
            return self._embedding_space
        return super().__getattribute__ (name)

    @property
    @abstractmethod
    def embedding_space(self):
        pass
```
In this way, `embedding_space` is defined using a property so it is closer to mathematic naming, it is marked as abstract and should be written in the child class with `immersion` but we still have the lazy computation which call it only once and it does not impact the user experience.

We will have
```python
class Hypersphere(LevelSet):
    def __init__(self, dim):
        super().__init__(dim=dim, default_coords_type='extrasinc', **kwargs)

    @property
    def embedding_space(self):
        return Euclidean(self.dim + 1)

    def immersion(self, x):
        return gs.norm(x) - 1
```

lpereira95:
I think @ymontmarin's `Child` example shows clearly why it may be inadequate to have the definition at `__init__` level.

I fully disagree with the idea of instantiating it every time: it will negatively impact performance and will not allow to change any numerical parameter of algorithms associated with a given embedding space.

I don't like so much the tricks of checking if it `hasattr` (for me it looks like an unnecessary condition - the impact in performance is negligible, but there's still a dictionary key lookup every time we access embedding space). Besides, we know it will have, for sure, the attribute after instantiation. This is the reason why I started this PR by having `embedding_space` has a property and changed it (today) to simply have it has attribute (instantiated in the init of `LevelSet`) - also, we need to instantiate it in `LevelSet` anyway to have access to `shape`. This is strong due to our "new" way of thinking stating we allow mathematical attributes to change, but at user's risk.

The semantics of `_define_embedding_space` do not look so strange to me. It is not so far from `_create_basis` we already have in several other classes (the reason of being of that method is performance).

I agree with the poll @ninamiolane, but after our discussions, I kind of think the solution current implemented in the PR is the strongest technically speaking.

In order to move on, do you agree I merge this solution for now (after addressing @ninamiolane's comments) and then we update it based on our poll? (The implementation on this PR is by far much cleaner than the currently available in master.)

ymontmarin:
@lpereira95 I also add some minor comments in the code.

About the main discussion:
- under the hood, any attribute access is a key lookup, which is a constant time operation. Having 2 instead of one does not change performance
- I also fully disagree with the idea of instantiating it every time, hence the classical trick of using `xxx` and `_xxx`
- "also, we need to instantiate it in LevelSet anyway to have access to shape." I don't understand the difference you are talking about. If you use `self.embbeding_space.shape` with the property implementation, it will call `_define_embedding_space` so the instantiation also happen in `__init__`. In any case, none of the above solve the problem of circular need: `_define_embedding_space` may need attributes (e.g. `dim`) that are defined in `__init__` but `__init__` need `embedding_space.shape` that is define in `_define_embedding_space`
- So maybe the semantic should simply be `_create_embedding_space` ?

I think the circular issue (and other comments) need to be addressed before merging. I only give suggestion for the love <3 of geomstats; let me know for curiosity, what people want :)

An idea for circular loop issue while following @lpereira95 implementation choice:
```
class LevelSet(ABC):
    def __init__(self, dim, shape=None, default_coords_type="extrinsic", **kwargs):
        self.embedding_space = self._create_embedding_space(dim, shape, default_coords_type, **kwargs)
        super().__init__(dim=dim, default_coords_type=default_coords_type, **kwargs)

    @abstractmethod
    def _create_embedding_space(self, dim, shape, default_coords_type, **kwargs):
        pass

    @abstractmethod
    def immersion(self):
        pass
```
Such that implementation of `Hypersphere` looks like
```
class Hypersphere(LevelSet):

    def _create_embedding_space(self, dim, **kwargs):
        return Euclidean(self.dim + 1)

    def immersion(self, x):
        return gs.norm(x) - 1
```

lpereira95:
Thanks a lot @ymontmarin, we deeply appreciate your comments and suggestions. All this discussion is to try to come up with a solution that pleases everyone and is technically robust.

Thanks for the clarification on the constant time operation for key lookup (in fact I should have thought better before my comment before).

For the third point, I mean that since we definitely have to instantiate `embedding_space` in the init of `LevelSet` to access `shape`, it looks unnecessary to me to instantiate it "implicitly" by having a property verifying if `_embedding_space` already exist and instantiating it if not. The current solution is more explicit in my opinion.

Your point on circularity is in fact interesting, but I would argue there's not really a circularity. What's happening is that we normally need the arguments passed in child init to instantiate the embedding space. I would say it is a good practice to store init args as attributes (e.g. `sklearn` forces you to do that to allow a better `__repr__`), so this solution actually forces us to follow a good practice. Passing arguments to `_define/create_embedding_space` does not feel right to me. The case of the `Hypersphere` is a little particular in the sense that it asks for `dim`, which is also an attribute of `Manifold`. (But I can fully understand why you are raising this issue, and I'm completely open to find out a better solution)

Both `_create` and `_define` work for me.

ymontmarin:
Ok I get the implicit reasoning for the instanciation, I agree.
I agree with the sklearn practice, it is also the practice i follow ! 

But in our case the argument are setted as attribute of the object during the call `super().__init__(...)` but this line is AFTER the line `self.embedding_space = self._create_embedding_space()`, which mean than when `_create_embedding_space` is called the arguments have not been setted as attribute yet ! So `self` has no `dim` attribute and `_create_embedding_space` can't access `dim`.

The natural solution (which is what happen in sklearn etc...) would be to have `super().__init__(...)` BEFORE `self.embedding_space = self._create_embedding_space()`, but in this situation, we also need to have `self.embedding_space` to access its shape as default shape. Hence the circularity.

Maybe I am wrong but for me this code will fail:
```python
class LevelSet(ABC):
    def __init__(self, dim, shape=None, default_coords_type="extrinsic", **kwargs):
        self.embedding_space = self._create_embedding_space()
        if shape is None:
            shape = self.embedding_space.shape
        super().__init__(dim=dim, shape=shape, default_coords_type=default_coords_type, **kwargs)

    @abstractmethod
    def _create_embedding_space(self):
        pass

    @abstractmethod
    def immersion(self, x):
        pass

class Hypersphere(LevelSet):

    def _create_embedding_space(self):
        return Euclidean(self.dim + 1)

    def immersion(self, x):
        return gs.norm(x) - 1
```

ymontmarin:
@lpereira95 Nevermind, I dit not see you were resetting `dim` as attribute manually (it is set twice but I don't think it is a problem). So I think it is ok then !

ymontmarin:
```python
class LevelSet(Manifold):
    def __init__(self, dim, shape=None, default_coords_type="extrinsic", **kwargs):
        self.embedding_space = self._create_embedding_space()
        if shape is None:
            shape = self.embedding_space.shape
        super().__init__(dim=dim, shape=shape, default_coords_type=default_coords_type, **kwargs)

    @abstractmethod
    def _create_embedding_space(self):
        pass

    @abstractmethod
    def immersion(self, x):
        pass

class Hypersphere(LevelSet):
    def __init__(self, dim):
        self.dim = dim
        super().__init__(dim)

    def _create_embedding_space(self):
        return Euclidean(self.dim + 1)

    def immersion(self, x):
        return gs.norm(x) - 1
```
will indeed work and looks pretty clean to me :) Sorry for the misunderstanding !

lpereira95:
@ninamiolane I think I've addressed everything here. I'll just ensure tests are passing and then I'll merge.

ymontmarin:
@lpereira95 
What about the shapes computation ?
image

lpereira95:
Hi @ymontmarin!

It looks like that comments are pending or I really missed them out...

ymontmarin:
@lpereira95 What does the label "pending" mean ? I am not so used to review on github

lpereira95:
Your review has not been submitted @ymontmarin. Only you have access to it. There should be a button on top with "Submit review".

ymontmarin (commented):
First time using the review tool !

ymontmarin:
@lpereira95  Ok !!! Thank you :) sorry I miss that, I left out the only two points (qhape stuff) that was not adress with the discussion and so on

geomstats/geometry/base.py
```diff
@@ -366,12 +367,18 @@ def belongs(self, point, atol=gs.atol):
...
-        elif value.ndim == 1:
-            constraint = gs.all(constraint, axis=-1)
+
+        submersed_point = self.submersion(point)
+        n = len(submersed_point.shape)
+        axis = (
+            tuple(-n + 1 + i for i in range(n - 1))
+            if gs.ndim(point) > len(self.shape)
+            else tuple(-n + i for i in range(n))
+        )
```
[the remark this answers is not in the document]
lpereira95: Thanks @ymontmarin!

This solution fails for:

```python
space = Hyperboloid(3)

space.belongs(space.random_point())
```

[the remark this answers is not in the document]
lpereira95: I think this may work in all cases:

```python
axis = tuple(range(n - len(self.shape), n)) if (n - len(self.shape)) >= 0 else ()
```

What do you think?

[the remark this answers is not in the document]
lpereira95: Actually I'm not so sure this version increases readability (it may even make it harder to understand). Besides, we can easily achieve general batching by replacing 1 by `gs.ndim(point) - len(self.shape)` in the first range...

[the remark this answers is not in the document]
lpereira95: a simplified and more readable version may be:

```python
n_batch = gs.ndim(point) - len(self.shape)
tuple(range(-n + n_batch, 0))
```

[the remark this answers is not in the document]
ymontmarin: After thinking about it, none of the two solutions are very universal.

`point` is of shape [*batch_shape, *self.shape]
`immersed_point` is of shape [*batch_shape, *sub_value_shape]

If the subermesion only give a real sub_value_shape = () otherwise we want the `all` on the sub_value_shape axis.

So maybe a more short and universal way that work for any batching, shape and output of submersion space would be:
```python
    def belongs(self, point, atol=gs.atol):
        """Evaluate if a point belongs to the manifold.

        Parameters
        ----------
            Point to evaluate.
        atol : float
            Absolute tolerance.
            Optional, default: backend atol.

        Returns
        -------
        belongs : array-like, shape=[...,]
            Boolean evaluating if point belongs to the manifold.
        """
        belongs = self.embedding_space.belongs(point, atol)
        if not gs.any(belongs):
            return belongs

        submersed_point = self.submersion(point)
        constraint = gs.isclose(submersed_point, 0.0, atol=atol)

        n_sub_dim = len(submersed_point.shape) - len(point.shape) + len(self.shape)
        if n_sub_dim:
            constraint = gs.all(constraint, axis=tuple(range(-n_sub_dim, 0)))

        return gs.logical_and(belongs, constraint)
```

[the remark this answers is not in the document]
ymontmarin: My pessage was written before i saw your two messages ! I think it is almost the same :)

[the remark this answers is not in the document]
lpereira95: I'm glad we came up with similar solutions independently. These ones I think are pretty universal. I'll merge now.

lpereira95:
Test failures are unrelated... merging.

Issue #1655, opened 2022-09-23
Bug: Product manifold is not able to project if one of the components is `Euclidean`


miguelgondu:
### Describe the bug

If one of the components of a product of manifolds is $\mathbb{R}^n$, the implementation of `projection` throws an error at `is_intrinsic`.

More specifically, `is_intrinsic` in `ProductRiemannianMetric` is implemented like this:

```python
    def is_intrinsic(self, point):
        """..."""
        if self.default_point_type != "vector":
            raise ValueError("Invalid default_point_type: 'vector' expected.")

        if point.shape[-1] == self.dim:
            return True
        if point.shape[-1] == sum(dim + 1 for dim in self.dims):
            return False
        raise ValueError("Input shape does not match the dimension of the manifold")
```

That is, it is expecting *all* of the manifolds to be either intrinsic or extrinsic. This means that I cannot combine $\mathbb{R}^n$ with other manifolds that I want to define as extrinsic.

Also, if I read it correctly, we are assuming that if a point is not intrinsic, then it must be embedded in $\mathbb{R}^{\text{dim}(M) + 1}$. Isn't this problematic if the ambient space is actually higher-dimensional?

I include a small example using `torch` as a backend, and the manifolds $\mathbb{S}^2\times \mathbb{S}^2$ and $\mathbb{R}^2\times \mathbb{S}^2$.

### Steps/Code to Reproduce

```python
import torch

from geomstats.geometry.hypersphere import Hypersphere
from geomstats.geometry.euclidean import Euclidean
from geomstats.geometry.product_manifold import ProductManifold

r2_s2 = ProductManifold([Euclidean(2), Hypersphere(2)])
s2_s2 = ProductManifold([Hypersphere(2), Hypersphere(2)])

# Some point almost on the product R2 x S2
point_to_be_projected_to_r2_s2 = torch.Tensor([0.0, 0.0] + [1.01, 0.0, 0.0])

# Some point almost on the product S2 x S2
point_to_be_projected_to_s2_s2 = torch.Tensor([1.01, 0.0, 0.0] + [1.01, 0.0, 0.0])

# Trying to project to S2 x S2 works wonderfully...
s2_s2_projection = s2_s2.projection(point_to_be_projected_to_s2_s2)
print(f"Projection to S2xS2: {s2_s2_projection}")

# ...but projecting to R2 x S2 fails.
r2_s2_projection = r2_s2.projection(point_to_be_projected_to_r2_s2)
[...]

PR #1689, opened 2022-10-18
ProductManifold and ProductRiemannianMetric

johnharveymath:
## Description

Significant rewriting of both ProductManifold and ProductRiemannianMetric. ProductManifold now allows mixtures of manifolds with vector or matrix representation or with explicit or implicit coordinates. Individual factors can be rescaled.

## Issue

This resolves the bug in #1655.


geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
-    def __init__(
-        self, manifolds, metrics=None, default_point_type="vector", n_jobs=1, **kwargs
-    ):
+    def __init__(self, factors, metrics=None, default_point_type="vector", **kwargs):
         geomstats.errors.check_parameter_accepted_values(
             default_point_type, "default_point_type", ["vector", "matrix"]
         )
 
-        self.dims = [manifold.dim for manifold in manifolds]
+        self.factors = factors
```
lpereira95, line 83: maybe do `tuple(factors)` so we flag that they're not supposed to be changed.

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+                    "A default_point_type of \'matrix\' can only be used if all "
+                    "manifolds have the same shape."
+                )
+
+        if "extrinsic" in self.factor_default_coords_types:
+            default_coords_type = "extrinsic"
+        else:
+            default_coords_type = "intrinsic"
+
         if metrics is None:
```
lpereira95, line 110: by removing `metrics` from input, this part of the code simplifies.

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
-            metrics = [manifold.metric for manifold in manifolds]
+            metric_scales = None
+            metrics = [manifold.metric for manifold in factors]
+        elif gs.all([is_positive(metric) for metric in metrics]):
+            metric_scales = metrics
```
lpereira95, line 114: call it `metrics_scales` in argument. Very confusing otherwise. The fact we will not have `metrics` there anymore makes things more clear.

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
-        self.manifolds = manifolds
-        self.n_jobs = n_jobs
+
+        if self.default_coords_type == "extrinsic":
+            factor_embedding_spaces = [
+                manifold.embedding_space if manifold.default_coords_type == "extrinsic"
+                else manifold
+                for manifold in factors
+            ]
+            self.embedding_space = ProductManifold(factor_embedding_spaces)
```
lpereira95, line 138: the `scale` of factors has no importance for the scales of the embedding spaces?

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+    def factor_default_coords_types(self):
+        """List containing the default_coords_type of each factor."""
+        return [factor.default_coords_type for factor in self.factors]
 
     @staticmethod
     def _get_method(manifold, method_name, metric_args):
         return getattr(manifold, method_name)(**metric_args)
 
-    def _iterate_over_manifolds(self, func, args, intrinsic=False):
+    def _validate_args_for_iteration(self, args):
```
lpereira95, line 159: the name of this function is a bit misleading, as it does more than validation: `_validate_and_prepare_args_for_iteration` instead?

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+                gs.cumsum(self.factor_dims)[:-1]
+                if self.default_coords_type == "intrinsic"
+                else gs.cumsum(self.embedding_space.factor_dims)[:-1]
             )
-            for i in range(len(self.manifolds))
+            for key, value in args.items():
+                if not gs.is_array(value):
+                    numerical_args[key] = value
+                else:
+                    msg = (f"Argument {key}: {value} could not be broadcast to "
```
lpereira95, line 193: Move message within `if`, so it is not defined when not needed.

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+        elif self.default_point_type == "matrix":
+            for key, value in args.items():
+                if not gs.is_array(value):
+                    numerical_args[key] = value
+                else:
+                    arguments[key] = value
+            for key, value in arguments.items():
+                if value.shape[-2:] != self.shape[-2:]:
+                    raise ValueError(
+                        "Arguments did not have correct trailing dimensions"
```
lpereira95, line 209: Put in present.

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+        """List containing the shape of each factor."""
+        return [factor.shape for factor in self.factors]
+
+    @property
+    def factor_default_coords_types(self):
+        """List containing the default_coords_type of each factor."""
+        return [factor.default_coords_type for factor in self.factors]
 
     @staticmethod
     def _get_method(manifold, method_name, metric_args):
```
lpereira95, line 156: we can change slightly the signature, so below we don't have to create a dictionary again:

`def _get_method(manifold, methods_name, **metric_args)`

geomstats/geometry/product_manifold.py
```diff
@@ -14,101 +14,288 @@
...
+        if self.default_point_type == 'vector':
+            # The individual factors might have matrix type, so their responses must be
+            # flattened before proceeding. Since it is not straightforward to see which
+            # type has been returned, we are going to have to compare the shape of the
+            # return to the leading shape of all arguments broadcast against each otehr
+            for response in out:
+                if (gs.is_array(response) and
+                        len(response.shape) > len(leading_dimensions) + 1):
+                    response.reshape(response.shape[:-2] + (-1,))
+            out = gs.concatenate(out, axis=-1)
```
lpereira95, line 287: This part of the code fails for this simple case (the problem is in the concatenate), but changing it may not be trivial, as it may break other cases:

```python

from geomstats.geometry.product_manifold import ProductManifold

factors = [Euclidean(3), Euclidean(2)]
space = ProductManifold(factors)

points = space.random_point(2)  # works

space.belongs(points)  # outputs True. Should output gs.array([True, True])
```

Another example is `space.is_tangent` with n `tangent_vec` and n `base_point`.

geomstats/geometry/product_manifold.py
```diff
@@ -3,7 +3,7 @@
 Lead author: Nicolas Guigui.
```
lpereira95, line 3: Add your name here.

lpereira95 (changes requested):
Thanks again @johnharveymath. This part of the code is finally getting pretty neat.

I haven't done a full review yet, but we have already several things to iterate on in `geomstats.geometry.product_manifold`. Let's do it and then think about the other changed stuff.

lpereira95:
Additionally, the tests now are covering several cases: `extrinsic-extrinsic`, `extrinsic-intrisic`, `intrinsic-intrinsic`, `vector-vector`, `matrix-matrix`, `vector-matrix`. Everything seems to, at least, run.

PR #1693, opened 2022-10-19
Vectorization of metric_matrix

johnharveymath:
## Description

In some subclasses of `RiemannianMetric`, the `metric_matrix` method only returned a single copy of the matrix, regardless of the array passed as `base_point`. This PR corrects that in `HermitianMetric`, `HypersphereMetric` and `MinkowskiMetric`. It amends the method in `EuclideanMetric` to use `gs.broadcast_to` instead of `gs.tile` as this method permits more flexibility in the array shape.


lpereira95 (approved):
Looks great. Thanks @johnharveymath!

PR #1710, opened 2022-10-21
Add statistical manifold of multivariate diagonal normal distributions

antoinecollas:
## Description

Add the statistical manifold of multivariate diagonal normal distributions.

It is a product manifold of the statistical manifold of univariate normal distributions.

## Additional context

`MultivariateDiagonalNormalDistributions` and `MultivariateDiagonalNormalMetric` store data points and tangent vectors by stacking the location (mean vector) and the diagonal of the covariance matrix in a vector.

### DIFF
--- a/geomstats/information_geometry/multivariate_normal.py
+++ b/geomstats/information_geometry/multivariate_normal.py
@@ -1,4 +1,5 @@
-"""Statistical Manifold of multivaraite normal distributions with the Fisher metric.
+"""Information Manifold of multivariate normal distributions with
+the Fisher information metric.
 
 Lead author: Antoine Collas.
 """
@@ -17,7 +18,19 @@ from geomstats.information_geometry.normal import NormalMetric
 
 
 class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin):
-    """Class for the manifold of diagonal multivariate normal distributions."""
+    """Class for the manifold of diagonal multivariate normal distributions.
+
+    This is the class for multivariate normal distributions with diagonal
+    covariance matrices and samples on the $n$-dimensional Euclidean space.
+    Each distribution is represented by a vector of size $2n$ where the first
+    $n$ elements contain the mean vector and the $n$ last elements contain
+    the diagonal of the covariance matrix.
+
+    Parameters
+    ----------
+    n : int
+          Dimension of the sample space of the multivariate normal distribution.
+    """
 
     def __init__(self, n):
         self.n = n
@@ -30,6 +43,9 @@ class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin)
 
     def belongs(self, point, atol=gs.atol):
         """Evaluate if the point belongs to the manifold.
+
+        First $n$ elements contain the mean vector and the $n$ last
+        elements contain the diagonal of the covariance matrix.
 
         Parameters
         ----------
@@ -48,7 +64,7 @@ class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin)
         return belongs
 
     def random_point(self, n_samples=1):
-        """Sample parameters of normal distributions.
+        """Generate random parameters of multivariate diagonal normal distributions.
 
         Parameters
         ----------
@@ -73,7 +89,10 @@ class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin)
         return point
 
     def projection(self, point):
-        """Floor the eigenvalues to gs.atol.
+        """Project a point on the manifold of
+        diagonal multivariate normal distributions.
+
+        Floor the eigenvalues of the diagonal covariance matrix to gs.atol.
 
         Parameters
         ----------
@@ -92,7 +111,7 @@ class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin)
         return projected
 
     def sample(self, point, n_samples=1):
-        """Sample from the diagonal mutltivariate normal distribution.
+        """Sample from the diagonal multivariate normal distribution.
 
         Parameters
         ----------
@@ -164,7 +183,13 @@ class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin)
 
 
 class MultivariateDiagonalNormalMetric(RiemannianMetric):
-    """Class for the Fisher information metric of diagonal normal distributions."""
+    """Class for the Fisher information metric of diagonal normal distributions.
+
+    Parameters
+    ----------
+    n : int
+          Dimension of the sample space of the multivariate normal distribution.
+    """
 
     def __init__(self, n):
         self.n = n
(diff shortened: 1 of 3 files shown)


geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+import geomstats.backend as gs
+from geomstats.geometry.base import OpenSet
+from geomstats.geometry.euclidean import Euclidean
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.information_geometry.base import InformationManifoldMixin
+from geomstats.information_geometry.normal import NormalMetric
+
+
+class MultivariateDiagonalNormalDistributions(OpenSet, InformationManifoldMixin):
+    """Class for the manifold of diagonal multivariate normal distributions."""
```
alebrigant, line 20: Add some information in the docstring, in particular: what is the shape of a point in this class ? If it is a vector of size $2n$, what does it contain, and in what order ?

e.g.
```
This is the class for multivariate normal distributions on the $n$-dimensional Euclidean
space with diagonal covariance matrices. Each distribution is represented by a vector
of size $2n$ where the first $n$ elements contain the mean vector and the $n$ last
elements contain the diagonal of the covariance matrix.

Parameters
----------
n : int
      Integer representing the number of variables of the multivariate normal distributions.
```

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+        self._unstack_location_diagonal = self.metric._unstack_location_diagonal
+        self._stack_location_diagonal = self.metric._stack_location_diagonal
+
+    def belongs(self, point, atol=gs.atol):
+        """Evaluate if the point belongs to the manifold.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., 2*n]
+            Point to test.
```
alebrigant, line 37: Specify what is contained in point, in what order.

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+        location = self.euclidean_n.random_point(n_samples=n_samples, bound=bound)
+        if n_samples == 1:
+            diagonal = gs.array(norm.rvs(size=(n,)) ** 2)
+        else:
+            diagonal = gs.array(norm.rvs(size=(n_samples, n)) ** 2)
+        point = self._stack_location_diagonal(location, diagonal)
+        return point
+
+    def projection(self, point):
+        """Floor the eigenvalues to gs.atol.
```
alebrigant, line 76: Docstring title should be "Project a point on the manifold of diagonal multivariate normal distributions", and then in the description, specify how: the diagonal elements/eigenvalues of the covariance matrix are floored.

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+            Boolean evaluating if point belongs to the space.
+        """
+        point_dim = point.shape[-1]
+        belongs = point_dim == self.dim
+        _, diagonal = self._unstack_location_diagonal(point)
+        belongs = gs.logical_and(belongs, gs.all(diagonal >= atol, axis=-1))
+        return belongs
+
+    def random_point(self, n_samples=1):
+        """Sample parameters of normal distributions.
```
alebrigant, line 51: Maybe "Generate random parameters of multivariate diagonal normal distributions" would be less confusing ?

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+            pdf_at_x = tmp_0 * tmp_1
+            pdf_at_x = gs.squeeze(pdf_at_x)
+            return pdf_at_x
+
+        return pdf
+
+
+class MultivariateDiagonalNormalMetric(RiemannianMetric):
+    """Class for the Fisher information metric of diagonal normal distributions."""
+
```
alebrigant, line 168: Add
```
Parameters
----------
n : int
      Integer representing the number of variables of the multivariate normal distributions.
```

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+        ----------
+        point: array-like, shape=[..., 2*n]
+            Point.
+        apply_sqrt: bool
+            Determine if a square root is applied to the diagonals.
+
+        Returns
+        -------
+        pairs : array-like, shape=[..., n, 2]
+            Pairs of locations and standard deviations.
```
alebrigant, line 225: By default, pairs of locations and variances ? When apply_sqrt is True, then standard deviations instead ?

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
...
+        ----------
+        point: array-like, shape=[..., 2*n]
+            Point.
+        apply_square: bool
+            Determine if a square is applied to the diagonals.
+
+        Returns
+        -------
+        pairs : array-like, shape=[..., n, 2]
+            Pairs of locations and standard deviations.
```
alebrigant, line 246: Same as previous comment.

alebrigant (changes requested):
Thanks for the great addition ! The code looks good to me, just some minor comments on the docstrings.

tests/tests_geomstats/test_multivariate_normal.py
```diff
@@ -0,0 +1,72 @@
...
+import geomstats.backend as gs
+from tests.conftest import Parametrizer
+from tests.data.multivariate_normal import (
+    MultivariateDiagonalNormalDistributionsTestData,
+    MultivariateDiagonalNormalMetricTestData,
+)
+from tests.geometry_test_cases import OpenSetTestCase, RiemannianMetricTestCase
+
+
+class TestMultivariateDiagonalNormalDistributions(
```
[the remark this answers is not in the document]
ninamiolane: (Could be, since you inherit from the OpenSetTestCase... I just forget)

ninamiolane (commented):
Very nice! Thank you for this very important addition. And super clean code. I mostly have minor comments, see directly in the code, and the following questions:

1. Do we want to use the word "mean" instead of "location" here? Or diagonal_cov? It could be more consistent with the terms used to refer to the parameters of the other distributions.

2. Can you add references (e.g. research papers) from where you have taken the formulas used, e.g. for exp, log, etc.
Thanks!

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -0,0 +1,348 @@
+"""Statistical Manifold of multivaraite normal distributions with the Fisher metric.
```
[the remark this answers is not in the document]
alebrigant: Yes good point ! To be consistent with the paper, we should replace "Statistical manifold" with "Information manifold". We made this choice of terminology because in differential geometry, statistical manifold usually refers to a manifold with dually flat affine structure, which is not implemented in geomstats (yet).

geomstats/information_geometry/multivariate_normal.py
antoinecollas: Not necessarily since this function can also be applied to tangent vectors.

tests/data/multivariate_normal.py
```diff
@@ -0,0 +1,186 @@
+import geomstats.backend as gs
```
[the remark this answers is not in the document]
antoinecollas: it seems to me that no file in "test/data" contains a docstring at the top of the file (?)

antoinecollas:
Thanks for the review ! I think I addressed all your comments.

PR #1712, opened 2022-10-21
Make tests check for correct shapes and add support for multidimensional batches

jubrau:
## Checklist

- [] I have made sure the code passes all unit tests. (refer to comment below)
- [] I have linked to issues and PRs that are relevant to this PR.

## Description
Tests for the methods `belongs` and `is_tangent` did not check that the correct shape was returned. This PR adds tests checking for the shape, where the batch shape of tested points can have several dimensions. 
Support for multidimensional shapes was added.

### DIFF
--- a/geomstats/geometry/matrices.py
+++ b/geomstats/geometry/matrices.py
@@ -54,7 +54,7 @@ class Matrices(VectorSpace):
         if ndim == 2:
             return belongs
         if belongs:
-            return gs.cast(gs.ones(point.shape[:-2]), bool)
+            return gs.ones(point.shape[:-2], dtype=bool)
         return False
 
     @staticmethod
--- a/geomstats/geometry/matrices.py
+++ b/geomstats/geometry/matrices.py
@@ -51,7 +51,11 @@ class Matrices(VectorSpace):
             return False
         mat_dim_1, mat_dim_2 = point.shape[-2:]
         belongs = (mat_dim_1 == self.m) and (mat_dim_2 == self.n)
-        return belongs if ndim == 2 else gs.tile(gs.array([belongs]), [point.shape[0]])
+        if ndim == 2:
+            return belongs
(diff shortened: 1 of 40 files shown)


lpereira95 (commented):
Thanks @jubrau, this is a great PR!

Nevertheless, I will not merge it, because I think we are not prepared to have any batch shape yet (and I'm not sure we want this behavior: users can always explicitly reshape batch dimensions by themselves). This was a discussion I had with @ymontmarin.

Anyway, I think we should provide this behavior when it is easy to provide (e.g. for methods that receive only one point), but do not test it explicitly. We ensure everything works when we have one batch dimension, but we do not promise anything for more.

Why this? Because things become harder when:

* we have a method that receives two arrays. e.g. your implementation of `FullRankCorrelationMatrices.diag_action` breaks the case 2 diagonal vectors, 1 point. This behavior is more important for us then having several batch dimensions.

* we implement "vectorization" through a loop


I'll cherry pick several of your changes and open a new PR because I think they improve the code substantially.

geomstats/geometry/matrices.py
```diff
@@ -320,7 +326,10 @@ def is_pd(cls, mat):
         """
         if mat.ndim == 2:
             return gs.array(gs.linalg.is_single_matrix_pd(mat))
-        return gs.array([gs.linalg.is_single_matrix_pd(m) for m in mat])
+        shape = mat.shape
+        mat = gs.reshape(mat, (-1,) + shape[-2:])
+        is_pd = gs.array([gs.linalg.is_single_matrix_pd(m) for m in mat])
+        return gs.reshape(is_pd, shape[:-2])
```
lpereira95, line 332: Later we should extend update `gs.linalg.is_single_matrix_pd` to be vectorized.

lpereira95:
See #1719.

PR #1714, opened 2022-10-22
Make SRVMetric on DiscreteCurves inherit from PullbackDiffeoMetric

alebrigant:
This PR:
- makes `SRVMetric` inherit from `PullbackDiffeoMetric`, so that the code is shortened.
- rewrites some of the tests, and deletes some unnecessary tests.

In order to make this possible, the `SRVMetric` is now only defined on curves starting at the origin. 

TODO:

- [ ] `DiscreteCurves` should be renamed to something like `DiscreteCurvesStartingAtOrigin`
- [ ] `ElasticMetric` should also inherit from `PullbackDiffeoMetric`
- [ ] General `DiscreteCurves` can be implemented as a `ProductManifold` of `Euclidean(dim)` times `DiscreteCurvesStartingAtOrigin` where the first factor gives the starting point. More general metrics can then be implemented as `ProductMetric`s to take into account translations. This can also be useful to visualize geodesics.


geomstats/geometry/discrete_curves.py
```diff
@@ -1229,17 +1256,45 @@ class SRVMetric(ElasticMetric):
...
+            if hasattr(ambient_manifold, "metric"):
+                self.ambient_metric = ambient_manifold.metric
+            else:
+                raise ValueError(
+                    "Instantiating an object of class "
+                    "ElasticMetric requires either a metric"
+                    " or an ambient manifold"
+                    " equipped with a metric."
+                )
+        self.l2_curves_metric = L2CurvesMetric(ambient_manifold=ambient_manifold)
```
luisfpereira, line 1281: can be deleted since it is defined in `define_embedding_metric`. then in the code use `self.embedding_metric`.

luisfpereira (changes requested):
Looks great @alebrigant. Nice to see so much code being replaced by the inheritance structure.

Just small comments to address.

geomstats/geometry/discrete_curves.py
alebrigant: Yes you're right, much better, thanks !

alebrigant:
Thanks @luisfpereira for your review and your previous help with this ! If you approve, we can merge so that Souhail can start a similar refactoring of `ElasticMetric`.

luisfpereira (approved):
Thanks for addressing the comments @alebrigant. From my side, it looks good!

PR #1716, opened 2022-10-25
"Outputs" of ihp hackathon

lpereira95:
An initial draft of a document explaining the decisions made during the last week hackathon.

@ninamiolane, @alebrigant, @nguigs, @xpennec, @sophiaas.

### DIFF
--- a/docs/hackathons/ihp_10_2022.rst
+++ b/docs/hackathons/ihp_10_2022.rst
@@ -1,0 +1,276 @@
+
+
+Hackathon at Institut Henri Poincaré, 17th-21st Oct 2022
+========================================================
+
+This hackathon happened at IHP within the thematic trimester on `Geometry and Statistics in Data Sciences <https://indico.math.cnrs.fr/category/498/>`_.
+
+
+Several participants made their first contributions to ``geomstats``. In no particular order: Francesco Ballerin (University of Bergen), Juliane Braunsmann (WWU Munster), Antoine Collas (Paris-Saclay University), John Harvey (Cardiff University), Emmanuel Hartman (Florida State University), and Joan Glaunès (Université Paris Descartes).
+
+
+Besides the usual informal schedule, we had a very fruitful presentation by Jean Feydy (INRIA), entitled "Fast geometric libraries for vision and data sciences". 
+
+
+It was a prolific week in terms of (great) discussions about math and programming. Several actions to improve our code base originated from them. The next section describes them and the rationale behind the changes that will be implemented.
+
+
+
+Future changes
+--------------
+
+
+Equip with metric
+*****************
+
+The following way of instantiating a space and equipping it with a metric will become standard:
+
+
+.. code-block:: python
+
+    from geomstats.geometry.hypersphere import (
+        Hypersphere,
+        HypersphereMetric,
+    )
+
+    space = Hypersphere(dim=2, equip=False)
+    space.equip_with_metric(HypersphereMetric)
+
+
+This means metrics are no more objects that should be instantiated by themselves, i.e. metrics are associated with spaces. To use methods available within the metric, it is now preferred to do ``space.metric.<method>`` instead of simply instantiating a metric and do ``metric.<method>``.
+
+
+Mathematically, this means that ``space.equip_with_metric`` transforms our smooth manifold :math:`M` into the tuple :math:`(M, g)`. The boolean ``equip`` allows to have the possibility to instantiate a non-equipped manifold. In order to keep the current behavior and for practical reasons (e.g. it can be cumbersome to equip a space that only has one "valid" metric; it is nice to have embedding spaces already equipped), ``equip=True`` by default, which means the following code is equivalent to the above:
+
+
+.. code-block:: python
+
+    space = Hypersphere(dim=2)
+
+
+We will now ask contributors to add ``_default_metric`` method to their classes, which returns the default metric for a given class (not instantiated). This will be used to equip the manifold when ``equip=True``.
+
+
+With this (seemingly small) change, we can now have (benign) circularity in our design, i.e. the space "knows" about the metric, and the metric "knows" about the space. This is achieved by requiring the first input argument of a metric to be the space:
+
+
+.. code-block:: python
+
+    class HypersphereMetric(RiemannianMetric):
+        def __init__(self, dim):
+            # init operations
+
+will become:
+
+
+.. code-block:: python
+
+    class HypersphereMetric(RiemannianMetric):
+        def __init__(self, space):
+            self._space = space
+            # other init operations
+
+
+Notice that now ``HypersphereMetric`` has access to ``dim`` via ``self._space``, which means there's only one ``dim`` (the one from ``space``) and no inconsistencies are allowed by design (e.g. equip a space with a metric that has an inappropriate ``dim``). This also removes the toll of checking input arguments for consistency.
+
+
+The ``space`` should be stored privately in the metric to avoid the recursion ``space.metric.space.metric.<...>``. The recursion is still allowed, but a user that accesses a private method from "outside" knows he/she's making a mistake.  
+
+
+Additionally, a difficult design problem is solved: how to access methods of the space within the metric? This is particularly relevant for quotient spaces (e.g. ``QuotientMetric`` requires ``FiberBundle``, and ``GraphSpaceMetric`` needs ``GraphSpace``) and was creating a design divergence where some metrics required the space, whereas others no: now all the metrics require the space.
+
+
+Moreover, we force each space to contain only one metric, since an equipped instance of a space represents :math:`(M, g)`.
+
+
+
+Similar actions will be done regarding equipping a space with a group action.
+
+
+
+Intrinsic vs extrinsic coordinates (manifold)
+*********************************************
+
+At a lower level, a manifold is expected to implement points via ``extrinsic`` OR ``intrinsic`` coordinates, never both.   
+
+
+This means that code as the following will not be allowed:
+
+
+.. code-block:: python
+
+    space = MySpace(default_coords_type="extrinsic")
+
+    space.default_coords_type = "intrinsic"
+
+
+That is, instances of a manifold are expected to be static, i.e. if you want a different point representation of your space, you should instantiate a new object.
+
+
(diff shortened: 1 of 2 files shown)


docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,276 @@
...
+
+.. code-block:: python
+
+    from geomstats.geometry.hypersphere import (
+        Hypersphere,
+        HypersphereMetric,
+    )
+
+    space = Hypersphere(dim=2, equip=False)
+    space.equip_with_metric(HypersphereMetric)
```
ninamiolane, line 37: NIT: 

- In a future version,  maybe this could become space.equip(HypersphereMetric) : the space would automatically recognize that the class is a child of RiemannianMetric and perform an equip_with_metric.

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,276 @@
...
+
+It was a prolific week in terms of (great) discussions about math and programming. Several actions to improve our code base originated from them. The next section describes them and the rationale behind the changes that will be implemented.
+
+
+
+Future changes
+--------------
+
+
+Equip with metric
```
ninamiolane, line 23: - This paragraph could mention the syntax coming from the use case where metrics have parameters.

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,276 @@
...
+
+
+Similar actions will be done regarding equipping a space with a group action.
+
+
+
+Intrinsic vs extrinsic coordinates (manifold)
+*********************************************
+
+At a lower level, a manifold is expected to implement points via ``extrinsic`` OR ``intrinsic`` coordinates, never both.
```
ninamiolane, line 94: I think we can distinguish two aspects:
- the coords type: extrinsic or intrinsic
- the coords: a manifold can accept several coordinate systems that are intrinsic, or several coordinate systems that extrinsic.

With this change:
- a manifold = a manifold implemented in a static coordinate system, which is additionally "tagged" with a flag "intrinsic" or "extrinsic"

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,276 @@
...
+
+.. code-block:: python
+
+    space.metric.exp_solver.n_steps = 20
+
+
+(The above code changes the number of steps performed during integration of the geodesic equation during the solution of an initial value problem.)
+
+
+The important distinction is that this code does not change anything related to mathematical properties (the equipped space still represents the same tuple :math:`(M, g)`): it simply changes the parameters of numerical algorithms used to find solutions to manifold/metric operations that do not have closed-form solutions.
```
ninamiolane, line 180: Very clear explanation!! 💯

ninamiolane (commented):
Very nice, thanks for the write up!

docs/hackathons/ihp_10_2022.rst
lpereira95: `space.equip_with` instead? I think is neat and pretty straightforward to implement.

docs/hackathons/ihp_10_2022.rst
lpereira95: Good point! I'll add it.

docs/hackathons/ihp_10_2022.rst
lpereira95: Sounds good! Do you mind to change the text to reflect this idea?

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+
+This means metrics are no more objects that should be instantiated by themselves, i.e. metrics are associated with spaces. To use methods available within the metric, it is now preferred to do ``space.metric.<method>`` instead of simply instantiating a metric and do ``metric.<method>``.
+
+
+Mathematically, this means that ``space.equip_with_metric`` transforms our smooth manifold :math:`M` into the tuple :math:`(M, g)`. The boolean ``equip`` allows to have the possibility to instantiate a non-equipped manifold. In order to keep the current behavior and for practical reasons (e.g. it can be cumbersome to equip a space that only has one "valid" metric; it is nice to have embedding spaces already equipped), ``equip=True`` by default, which means the following code is equivalent to the above:
+
+
+.. code-block:: python
+
+    space = Hypersphere(dim=2)
```
nguigs, line 48: If I understand correctly, the metric is instantiated in `space.equip_with_metric(HypersphereMetric)`. Is this called by `__innit__` when `equip=True`, but this does not create circularity issues because `space=self` in the metric?

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+When equipping a space with a metric that requires additional parameters, a keyword-argument syntax should be user:
+
+
+.. code-block:: python
+
+    space.equip_with_metric(ElasticMetric, a=1., b=1.)
+
+
+
+With this (seemingly small) change, we can now have (benign) circularity in our design, i.e. the space "knows" about the metric, and the metric "knows" about the space. This is achieved by requiring the first input argument of a metric to be the space:
```
nguigs, line 63: This is a great solution, it allows to be more consistent without changing too much the current api!

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+Intrinsic vs extrinsic coordinates (connection)
+***********************************************
+
+Similarly to the manifold, the connection inheritance structure will also be affected by the clearer distinction between ``extrinsic`` and ``intrinsic`` (in fact, changes in the manifold structure tend to be reflected in the connection structure and vice-versa).
+
+
+Here, this distinction will reduce namespace pollution ("ill-defined" methods will not be available to many children), make testing easier (e.g. currently there are many skips in tests due to "inappropriate" method inheritance), and reduce information overload to a class user (e.g. ``dir(<MyMetric>)`` will only show methods that are properly defined).
+
+
+To make it more concrete, we will keep in ``Connection`` only methods that return coordinate invariant objects (e.g. scalars), and move coordinate dependent methods to ``IntrinsicConnection`` (e.g. ``christoffels``, ``riemannian_tensor``, etc).
```
nguigs, line 149: What actually remains in the extrinsic case?

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+
+
+
+* **history will be easier to turn off/on**. For example, in some use cases of parallel transport we are only interested in parallel transporting a tangent vector and we do not want to access the trajectory. In other cases, the trajectory is important. Having a ``ParallelTransporter`` object allows to handle these use cases easily by having a boolean parameter e.g. ``save_trajectory=True`` that turns off/on the saving of the trajectory. If ``False``, all the additional operations required to save the trajectory will not be done (which may lead to better performance and less memory use). If ``True``, the trajectory should be stored within the object (e.g. by creating the attribute ``self.trajectory_``; naming follows ``sklearn`` naming convention), but never be returned, as we want to keep the same API. A similar behavior can be achieved without the creation of an additional object, but then the question of where to save the trajectory arises. "In ``Connection``" may be a OK answer, but it will lead fast to namespace pollution. The same applies to optimization problems, solution of ODEs, etc. This is especially interesting for research.
```
nguigs, line 233: The `self.trajectory_` attribute is a bit surprising. I do get the point of not returning the trajectory, but I didn't think of the `ParallelTransporter` as a `sklearn` estimator: the trajectory is specific to the input points and tangent_vecs and cannot be reused with different inputs, while a `coef_` attribute in `sklearn` is reused on new data in `predict`.

nguigs: Maybe the compute method of the `ParallelTransporter` should return an object that has those attributes: `transported_vecs_`, `trajectory_`, etc... and ODE solvers could return an object with the final point and the trajectory if it was required. Kind of what scipy optimizers return and what we are currently doing with a dictionnary actually...

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+
+.. code-block:: python
+
+    class MyMetric(RiemannianMetric):
+
+        def _define_exp_solver(self):
+         return MyNiceExpSolver(n_steps=10)
+
+
+If ``_define_exp_solver`` does not exist, we will assume a closed-form solution is implemented. A nice error message will be implemented in ``Connection.exp`` to help contributors know what to change if something goes wrong.
```
nguigs, line 251: This will be useful for the long awaited refactorisation of geodesics that would use the results for shorter time points to compute new ones.

docs/hackathons/ihp_10_2022.rst
```diff
@@ -0,0 +1,285 @@
...
+Methods in ``geomstats.geometry`` will still be able to receive one point or multiple points, but some behaviors will be slightly modified. In particular, the following behaviors are expected for a method that receives a 1d point and returns a 1d point (after some operations):
+
+1. input: 1d -> output: 1d
+
+2. input: 2d (multiple points) -> output: 2d (multiple points)
+
+3. input: 2d (one point) -> output: 2d (one point)
+
+
+The last behavior is new, as currently if we have one point as input, we will always return a 1d array (for the 1d-1d case). This means ``gs.squeeze`` will have to be used at a much lower extent.
```
nguigs, line 278: Good point, although I would argue that the current behavior is not consistent across the code base.

nguigs (commented):
Great work and thanks for the summary!

docs/hackathons/ihp_10_2022.rst
lpereira95: Exactly!

docs/hackathons/ihp_10_2022.rst
lpereira95: Unfortunately it is not. The idea is to set a rule and now ensure it is followed everywhere.

lpereira95:
Thanks for the feedback @nguigs. I'm glad you agree with most everything. Let's iterate for the case of the parallel transport, to ensure we both get what we want.

docs/hackathons/ihp_10_2022.rst
nguigs: actually the current implementations of exp/log in connection are for the intrinsic case, and I don't think we have an extrinsic case in the code base.

PR #1727, opened 2022-10-31
Add product hpd matrices and siegel disks

YannCabanes:
The ProductHPDMatricesAndSiegelDisks manifold is defined as a
product manifold of the HPD manifold and (n-1) Siegel disks.
The HPD Siegel disks product has a product metric.
The product metric on the HPD Siegel disks product space is the usual
HPD matrices affine-invariant metric (with power affine parameter equal 1)
and Siegel metrics multiplied by constants.
This product manifold can be used to represent Block-Toeplitz HPD matrices.

References
----------
.. [Cabanes2022] Yann Cabanes. Multidimensional complex stationary
    centered Gaussian autoregressive time series machine learning
    in Poincaré and Siegel disks: application for audio and radar
    clutter classification, PhD thesis, 2022
.. [Cabanes2021] Yann Cabanes and Frank Nielsen.
    New theoreticla tools in the Siegel space for vectorial
    autoregressive data classification,
    Geometric Science of Information, 2021.
    https://franknielsen.github.io/IG/GSI2021-SiegelLogExpClassification.pdf
.. [JV2016] B. Jeuris and R. Vandebril. The Kahler mean of Block-Toeplitz
    matrices with Toeplitz structured blocks, 2016.
    https://epubs.siam.org/doi/pdf/10.1137/15M102112X


YannCabanes:
Hello @lpereira95, 

One of the difficulties is that the `ProductManifold` and `ProductRiemannianMetric` classes have not been designed for a product manifold of square matrices.
The shape of a point of the `ProductHPDMatricesAndSiegelDisks` manifold is `(n_manifolds, n, n)`.

The TensorFlow tests do not run due to the following error which is not related to this PR:
`ERROR tests/tests_geomstats/test_multivariate_normal.py - TypeError: 'tensorflow.python.framework.ops.EagerTensor' object does not support item assignment`

geomstats/geometry/hpd_matrices.py
```diff
@@ -510,7 +510,7 @@ class HPDAffineMetric(ComplexRiemannianMetric):
     """
 
     def __init__(self, n, power_affine=1, scale=1.0):
-        dim = int(n * (n + 1) / 2)
+        dim = n**2
```
lpereira95, line 513: it was correct. the behavior you want is achieved by making `default_coords_type="intrisinc"` (it was wrongly defined as `"intrinsic"`.

Please revert this change here and in the other places.

For future reference, it is also advised to make corrections for code that is not the focus of the PR in different PRs.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
```diff
@@ -0,0 +1,111 @@
+"""The ProductHPDMatricesAndSiegelDisks manifold.
```
lpereira95, line 1: the filename is too long: `product_hpd_and_siegel` instead? the class can then keep its name.

geomstats/geometry/product_manifold.py
```diff
@@ -179,7 +179,9 @@ def embed_to_product(self, points):
                 points_.append(response)
 
             return gs.concatenate(points_, axis=-1)
-        return gs.stack(points, axis=-2)
+        if len(self.shape) <= 2:
+            return gs.stack(points, axis=-2)
```
lpereira95, line 183: the `if` is unnecessary, the return below is enough.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, n_manifolds, n, **kwargs):
+        hpd_matrices = HPDMatrices(n=n)
+        hpd_matrices.metric = HPDAffineMetric(n=n, scale=n_manifolds**0.5)
+        siegel_disk = Siegel(n=n)
+        list_manifolds = [hpd_matrices] + (n_manifolds - 1) * [siegel_disk]
+        super(ProductHPDMatricesAndSiegelDisks, self).__init__(
+            factors=list_manifolds, **kwargs
+        )
+        self.shape = (n_manifolds, n, n)
```
lpereira95, line 59: this is hacking `ProductManifold`.

We need to refactor `ProductManifold` to handle this situation "natively". Can you please open a PR only with the changes to product manifold, so we can focus on it? (I'll help you moving fast with it, and I also would like to bring @johnharveymath to the discussion, since he worked a lot on its refactoring.)

We can test the behavior you want by e.g. doing a product manifold with 2 `SPDMatrices` or any matrices space combination with same point shape.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
```diff
@@ -0,0 +1,111 @@
...
+        self.metric = ProductHPDMatricesAndSiegelDisksMetric(
```
lpereira95, line 60: this can be passed directly to `super`

geomstats/geometry/product_manifold.py
```diff
@@ -208,8 +210,11 @@ def project_from_product(self, point):
                 for j in range(len(self.factors))
             ]
 
-        else:
+        elif len(self.shape) <= 2:
             projected_points = [point[..., j, :] for j in range(len(self.factors))]
+        else:
+            point = gs.reshape(point, (-1,) + self.shape)
+            projected_points = [point[:, j, ...] for j in range(len(self.factors))]
```
lpereira95, line 217: no need for reshape: `point[..., j, :, :]` should work.

lpereira95 (changes requested):
Thanks @YannCabanes.

As I told, let's move with the `ProductManifold` changes separately (my comments also apply to `ProductManifoldMetric`).

geomstats/geometry/hpd_matrices.py
YannCabanes: Hello @lpereira95, 
I am not sure about the HPD matrices manifold dimension.
I would say that its dimension as a real scalar coordinates is `n**2`: 
- the dimension of the real part is equal to `n * (n+1) / 2` as it entirely determine by the upper triangular part, diagonal included.
- the dimension of the imaginary part is equal to `n * (n-1) / 2` as it entirely determine by the upper triangular part, diagonal excluded.

We can also say that the dimension is equal the `n * (n+1) / 2` as it entirely determine by the upper triangular part (diagonal included) which is composed of `n * (n+1) / 2` complex coefficients.
However, we can not create a complex basis of this dimension.
For example the diagonal coefficients of an HPD matrix are positive reals. If a diagonal coefficient is multiplied by a complex constant, then the matrix is not HPD anymore. 
Anyway, the HPD matrices is not a vector space as multiplying an HPD matrix by minus 1 do not give an HPD matrix.

Even if the HPD matrices manifold is not a vector space, if we had to define a basis of this space, I think that it would be better to use `n**2` basis vectors (real or complex) and using real scalar coordinates than to use `n * (n+1) / 2` complex basis vectors with complex coordinates.
Then I would suggest to consider that the dimension of the HPD matrices is rather `n**2` than `n * (n+1) / 2`.

So from a mathematical point of view, I would say that its dimension is `n**2` : an HPD matrix can be represented by `n**2` real coordinates.

However, if you consider the dimension as the minimum number of coefficients of the matrix that you need to save to be able to recover the entire HPD matrix, then it is `n * (n+1) / 2` complex coefficients.

What do you think about it?

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
YannCabanes: Hello @lpereira95,
Is `product_hpd_and_siegel_disks` fine?
Or is it still too long?
If possible, I would like to show that there are several Siegel spaces.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
johnharveymath: I am happy to get in on this. I can contribute on Friday. Just for clarification - the idea here is that the factors coming in are of shape (n, n) and we then want the product manifold to have a shape like (k, n, n)? I approve of this idea as it reflects best how we would mathematically think about the manifold.

lpereira95: Exactly @johnharveymath. We can move this on in a different PR.

Do you have in mind the required changes to implement this? (I have some ideas in case.) I was thinking that maybe we should replace `default_point_type` by something more generic, like `point_ndim`, so it seems more naturally to make this change (this variable basically describes `len(shape)`. But probably this is not even required.

Let me know you're thoughts.

geomstats/geometry/hpd_matrices.py
YannCabanes: Hello @lpereira95  and @johnharveymath,
Whether you decide that the dimension is `n**2` or `n * (n+1) / 2`, you will have odd dimensions for some values of `n` in both cases. (For example `n=1`.)
If do not think that the HPD matrices manifold should be considered as a complex manifold.
For example, in dimension `n=1`, the HPD manifold is just the set of positive reals.
Then, taking the basis vector `[[1]]`, each one-dimensional HPD matrix `[[a]] = a * [[1]] `can be represented in this basis by the real coordinate `a`.
It is then an open subset of the real space of dimension 1 as the set of HPD matrices corresponds to the coordinate `a > 0`.
However, if we consider complex coordinates `a`, then `a * [[1]]` is not an HPD matrix in general.
The HPD matrices manifold is a manifold which has some complex values, but it is much closer to an open subset of a real vector space (with some real or complex basis vectors) than to a complex vector space.

In dimension `n=2`, the basis that I would choose would be the 4 basis vectors:
- `e1 = [[1, 0], [0, 0]]`
- `e2 = [[0, 0], [0, 1]]`
- `e3 = [[0, 1], [1, 0]]`
- `e4 = [[0, i], [-i, 0]]`

Then each two-dimensional HPD matrix can be represented as a linear combination of these 4 basis vectors using real coordinates: `hpd_matrix = a1 e1 + a2 e2 + a3 e3 + a4 e4`. 
The set of coefficients `(a1, a2, a3, a4) in R^4` corresponding to HPD matrices is an open convex subset of `R^4` which is stable by multiplication by a strictly positive constant.

Then would you agree to consider that its dimension is `n**2`?

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
johnharveymath: Yes, this may be a good idea, but I don't think it's necessary for this particular issue. So far what I've done is add a `default_point_type` of `'other'` for `len(shape)>2`. The `ProductManifold` will try to choose the best shape for the user, which will be `(n_factors, factors.shape)` if they all have the same shape. That's working fine. Now I have to make sure nothing is broken in any other classes. by having this new type. Then I need to to just run through all the methods of `ProductManifold` to make them work with this shape, but I don't think that will be a problem at all.

geomstats/geometry/hpd_matrices.py
johnharveymath: I see - it is embedded in a complex space but is not itself complex. Then `n**2` looks right to me, yes.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
YannCabanes: Hello @lpereira95,
I have changed `product_hpd_matrices_and_siegel_disks` into `product_hpd_and_siegel_disks` in the file names.
Tell me if it is still too long.

YannCabanes: Done.

geomstats/geometry/product_manifold.py
YannCabanes: Hello @lpereira95,
Yes, using `point[..., j, :, :]` is working when `len(self.shape) == 2`.
My solution (reshaping the data) should be working when `len(self.shape) >= 2`.
Which solution would you like me to keep?

geomstats/geometry/product_manifold.py
YannCabanes: Done.

geomstats/geometry/product_hpd_matrices_and_siegel_disks.py
YannCabanes: Done.

lpereira95:
@YannCabanes, do you have some smoke data to add here? I'll merge it for now, but it would be important to add some known data to ensure everything really works fine.

YannCabanes:
Hello @lpereira95,
With my notations, the scale parameter means multiplying the distance by the scale parameter, which means multiplying:

- `dist` --> (multiply by) `scale` 
- `norm` --> `scale`

- `metric_matrix` --> `scale ** 2`
- `inner_product` --> `scale ** 2`
- `squared_norm` --> `scale ** 2`
- `squared_dist` --> `scale ** 2`

However, I see that in the current master branch, the file `product_riemannian_metric.py` is using a different convention for `metric_matrix` and `inner_product` as they are only multiplied by `scale` and not `scale ** 2`.
--> I think that we should change that and specify the changes for each of the 6 elements mentioned above.

I am currently looking at the how the scale influences the curvature, sectional_curvature, riemann_tensor...

When you write:
```
metrics = [scales[0] * HPDAffineMetric(n=n)] + [
            scale * SiegelMetric(n=n) for scale in scales[1:]
        ]
```
do you obtain the changes that I described above?

If yes, then you can merge.

lpereira95:
@YannCabanes, maybe we also need to adapt here the scale parameter as you are discussing in #1761. Do you have any smoke data where we can try this out? Or alternatively, what happens e.g. to the distance when we increase the number of manifolds and e.g. repeat a siegel point many times?

YannCabanes:
Hello @lpereira95,
As discussed in #1761, I just did a complete review of the influence of the scale parameter. 
The scale parameter (related to the metric) has no influence on any function defined in the file `connection.py` (`christoffels`, `exp`, `log`, `riemann_tensor`, `curvature`, `ricci_tensor`, `geodesic`...).
The scale parameter influences the methods defined in the file `riemannian_metric.py`.
Using the convention in which the distance is multiplied by the scale parameter and the metric_matrix by scale ** 2, I now list the influence of the scale parameter on each method of the class `RiemannianMetric`:
- `metric_matrix` --> `* scale ** 2`
- `cometric_matrix` --> `/ scale ** 2`
- `inner_product_derivative_matrix` --> `* scale ** 2`
- `christoffels` --> no influence
- `inner_product` --> `* scale ** 2`
- `inner_coproduct` --> `/ scale ** 2`
- `hamiltonian` --> `/ scale ** 2`
- `squared_norm` --> `* scale ** 2`
- `norm` --> `* scale`
- `normalize` --> `/ scale`
- `random_unit_tangent_vec` --> `/ scale`
- `squared_dist` --> `* scale ** 2`
- `dist` --> `* scale`
- `dist_broadcast` --> `* scale`
- `dist_pairwise` --> `* scale`
- `diameter` --> `* scale`
- `closest_neighbor_index` --> no influence
- `normal_basis` --> `/ scale`
- `covariant_riemann_tensor` --> `* scale ** 2`
- `sectional_curvature` --> `/ scale ** 2`
- `scalar_curvature` --> `/ scale ** 2`

YannCabanes:
Hello @lpereira95,
The `ProductHPDMatricesAndSiegelDisks` with is the product of one HPDMatrices manifold and `(n_manifolds - 1)` Siegel manifolds.
About the metric, when `point_a` and `point_b` have the same values in each of the `(n_manifolds - 1)` Siegel disks, we have:
```
ProductHPDMatricesAndSiegelDisks(n_manifolds, n).dist
= n_manifolds ** 0.5 * HPDAffineMetric(n=n).dist
+ ((n_manifolds - 1) * n_manifolds / 2) ** 0.5 * SiegelMetric(n).dist
```
Can you write a test with this information or do you want me to do it?

lpereira95:
Thanks for all these verifications @YannCabanes. I'll add the test.

YannCabanes:
Hello @lpereira95,
Great, I let you add the test and merge this PR. Thank you!
Once it is merged, I will add another product manifold: `ProductPositiveRealsAndComplexPoincareDisks` which corresponds to the one-dimensional case of `ProductHPDMatricesAndSiegelDisks`.
Indeed, `PositiveReals` is the one-dimensional case of `HPDMatrices` and `ComplexPoincareDisk` is the one-dimensional case of `Siegel`. In these one-dimensional manifolds, many simplifications occur compared with the multidimensional manifolds since matrices commute in dimension 1.

PR #1729, opened 2022-11-02
Create multivariate normal distributions

alebrigant:
This PR adds classes for:
- centered multivariate normal distributions
- general multivariate normal distributions
- the Fisher information metric on centered normal distributions

TODO:
- The Fisher information metric on centered normal distributions is the affine invariant metric **corrected** by a factor $1/\sqrt{2}$
- Address the general case of the Fisher information metric

### DIFF
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -98,6 +98,12 @@ class SPDMatrices(OpenSet):
         size = (n_samples, n, n) if n_samples != 1 else (n, n)
 
         mat = bound * (2 * gs.random.rand(*size) - 1)
+        # if n_samples != 1:
+        #    spd_mat = []
+        #    for matrix in mat:
+        #        spd_mat.append(GeneralLinear.exp(Matrices.to_symmetric(matrix)))
+        #    spd_mat = gs.stack(spd_mat)
+        # else:
         spd_mat = GeneralLinear.exp(Matrices.to_symmetric(mat))
 
         return spd_mat
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -34,7 +34,7 @@ class SPDMatrices(OpenSet):
         )
         self.n = n
 
-    def belongs(self, mat, atol=gs.atol):
+    def belongs(self, point, atol=gs.atol):
         """Check if a matrix is symmetric with positive eigenvalues.
 
         Parameters
@@ -50,8 +50,8 @@ class SPDMatrices(OpenSet):
         belongs : array-like, shape=[...,]
             Boolean denoting if mat is an SPD matrix.
         """
-        is_sym = self.embedding_space.belongs(mat, atol)
-        is_pd = Matrices.is_pd(mat)
+        is_sym = self.embedding_space.belongs(point, atol)
+        is_pd = Matrices.is_pd(point)
         belongs = gs.logical_and(is_sym, is_pd)
         return belongs
 
@@ -98,12 +98,6 @@ class SPDMatrices(OpenSet):
         size = (n_samples, n, n) if n_samples != 1 else (n, n)
 
         mat = bound * (2 * gs.random.rand(*size) - 1)
-        # if n_samples != 1:
-        #    spd_mat = []
-        #    for matrix in mat:
-        #        spd_mat.append(GeneralLinear.exp(Matrices.to_symmetric(matrix)))
-        #    spd_mat = gs.stack(spd_mat)
-        # else:
         spd_mat = GeneralLinear.exp(Matrices.to_symmetric(mat))
 
         return spd_mat
--- a/geomstats/information_geometry/multivariate_normal.py
+++ b/geomstats/information_geometry/multivariate_normal.py
@@ -12,8 +12,99 @@ import geomstats.backend as gs
 from geomstats.geometry.base import OpenSet
 from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.geometry.spd_matrices import SPDMatrices
 from geomstats.information_geometry.base import InformationManifoldMixin
 from geomstats.information_geometry.normal import NormalMetric
+
+
+class MultivariateCenteredNormalDistributions(InformationManifoldMixin, SPDMatrices):
+    """Class for the manifold of centered multivariate normal distributions.
+
+    This is the class for multivariate normal distributions with zero mean
+    on the $n$-dimensional Euclidean space. Each distribution is represented by
+    its covariance matrix, i.e. an $n$-by-$n$ symmetric positive-definite matrix.
+
+    Parameters
+    ----------
+    n : int
+        Dimension of the sample space of the multivariate normal distribution.
+    """
+
+    def __init__(self, n):
+        super().__init__(n=n)
+        # self.metric = MultivariateFixedMeanNormalMetric()
+
+    def sample(self, point, n_samples=1):
+        """Sample from a centered multivariate normal distribution.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., n, n]
+            Symmetric positive definite matrix representing the covariance matrix
+            of a multivariate normal distribution with zero mean.
+        n_samples : int
+            Number of points to sample with each covariance matrix in point.
+            Optional, default: 1.
+
+        Returns
+        -------
+        samples : array-like, shape=[..., n_samples, n]
+            Sample from centered multivariate normal distributions.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
+        point = gs.to_ndarray(point, to_ndim=3)
+        samples = []
+        loc = gs.zeros(self.n)
+        for cov in point:
+            samples.append(gs.array(multivariate_normal.rvs(loc, cov, size=n_samples)))
+        return samples[0] if point.shape[0] == 1 else gs.stack(samples)
+
+    def point_to_pdf(self, point):
+        """Compute pdf associated to point.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., n, n]
+            Symmetric positive definite matrix representing the covariance matrix
+            of a multivariate normal distribution with zero mean.
+
+        Returns
+        -------
+        pdf : function
+            Probability density function of the centered multivariate normal
+            distributions with covariance matrices provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
+        point = gs.to_ndarray(point, to_ndim=3, axis=0)
+        point = point[:, None, :]
+        n = self.n
+        location, diagonal = self._unstack_location_diagonal(n, point)
+
+        def pdf(x):
+            """Generate parameterized function for normal pdf.
+
+            Parameters
+            ----------
+            x : array-like, shape=[n_samples, n]
+                Points at which to compute the probability
+                density function.
+            """
+            x = gs.to_ndarray(x, to_ndim=2, axis=0)
+            x = x[None, :, :]
+            det_cov = gs.squeeze(gs.prod(diagonal, axis=-1))
+            pdf_normalization = 1 / gs.sqrt(gs.power((2 * gs.pi), n) * det_cov)
+            pdf = gs.exp(-0.5 * gs.sum(((x - location) ** 2) / diagonal, axis=-1))
+            while pdf_normalization.ndim < pdf.ndim:
+                pdf_normalization = pdf_normalization[..., None]
+            pdf = pdf_normalization * pdf
+            pdf = gs.squeeze(pdf)
+            return pdf
+
+        return pdf
 
 
 class MultivariateDiagonalNormalDistributions(InformationManifoldMixin, OpenSet):
--- a/geomstats/information_geometry/multivariate_normal.py
+++ b/geomstats/information_geometry/multivariate_normal.py
@@ -9,12 +9,46 @@ from scipy.stats import multivariate_normal, norm
 
 import geomstats
 import geomstats.backend as gs
+import geomstats.errors as errors
 from geomstats.geometry.base import OpenSet
 from geomstats.geometry.euclidean import Euclidean
+from geomstats.geometry.product_manifold import ProductManifold
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 from geomstats.geometry.spd_matrices import SPDMatrices
 from geomstats.information_geometry.base import InformationManifoldMixin
 from geomstats.information_geometry.normal import NormalMetric
+
+
+class MultivariateNormalDistributions:
+    """Class for the multivariate normal distributions.
+
+    This class is a common interface to the following different situations:
+
+    - centered distributions
+    - distributions with diagonal covariance matrix
+    - general distributions
+
+    Parameters
+    ----------
+    n : int
+        Dimension of the sample space of the multivariate normal distribution.
+    distribution_type : str, {'centered', 'diagonal', 'general'}
+        Type of distributions.
+        Optional, default: 'general'.
+    """
+
+    def __new__(cls, n, distribution_type="general"):
+        """Instantiate class that corresponds to the distribution_type."""
+        errors.check_parameter_accepted_values(
+            distribution_type,
+            "distribution_type",
+            ["centered", "diagonal", "general"],
+        )
+        if distribution_type == "centered":
+            return MultivariateCenteredNormalDistributions(n)
+        if distribution_type == "diagonal":
+            return MultivariateDiagonalNormalDistributions(n)
+        return MultivariateGeneralNormalDistributions
 
 
 class MultivariateCenteredNormalDistributions(InformationManifoldMixin, SPDMatrices):
@@ -79,10 +113,8 @@ class MultivariateCenteredNormalDistributions(InformationManifoldMixin, SPDMatri
         geomstats.errors.check_belongs(point, self)
         if point.ndim > 3:
             raise NotImplementedError
-        point = gs.to_ndarray(point, to_ndim=3, axis=0)
-        point = point[:, None, :]
+        point = point[None, :, :] if point.ndim == 2 else point
         n = self.n
-        location, diagonal = self._unstack_location_diagonal(n, point)
 
         def pdf(x):
             """Generate parameterized function for normal pdf.
@@ -94,15 +126,15 @@ class MultivariateCenteredNormalDistributions(InformationManifoldMixin, SPDMatri
                 density function.
             """
             x = gs.to_ndarray(x, to_ndim=2, axis=0)
-            x = x[None, :, :]
-            det_cov = gs.squeeze(gs.prod(diagonal, axis=-1))
-            pdf_normalization = 1 / gs.sqrt(gs.power((2 * gs.pi), n) * det_cov)
-            pdf = gs.exp(-0.5 * gs.sum(((x - location) ** 2) / diagonal, axis=-1))
-            while pdf_normalization.ndim < pdf.ndim:
-                pdf_normalization = pdf_normalization[..., None]
+            det_cov = gs.linalg.det(point)
+            inv_cov = gs.linalg.inv(point)
+            pdf_normalization = 1 / gs.sqrt(gs.power(2 * gs.pi, n) * det_cov)
+            pdf = []
+            for xi in x:
+                pdf.append(gs.exp(-0.5 * gs.transpose(xi) @ inv_cov @ xi))
+            pdf = gs.stack(pdf)
             pdf = pdf_normalization * pdf
-            pdf = gs.squeeze(pdf)
-            return pdf
+            return gs.squeeze(pdf)
 
         return pdf
 
@@ -181,7 +213,8 @@ class MultivariateDiagonalNormalDistributions(InformationManifoldMixin, OpenSet)
         Parameters
         ----------
         point : array-like, shape=[..., 2*n]
-            Point to test.
+            Point to test. First $n$ elements contain the mean vector
+            and the $n$ last elements contain the diagonal of the covariance matrix.
 
         Returns
         -------
@@ -228,7 +261,8 @@ class MultivariateDiagonalNormalDistributions(InformationManifoldMixin, OpenSet)
         Parameters
         ----------
         point : array-like, shape=[..., 2*n]
-            Point to project.
+            Point to project. First $n$ elements contain the mean vector
+            and the $n$ last elements contain the diagonal of the covariance matrix.
 
         Returns
         -------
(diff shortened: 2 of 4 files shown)


geomstats/geometry/product_manifold.py
```diff
@@ -255,7 +255,7 @@ def _iterate_over_factors(self, func, args):
             self._get_method(self.factors[i], func, args_list[i], numerical_args)
             for i in range(len(self.factors))
         ]
-
+        print(out)
```
ninamiolane, line 258: rm

geomstats/geometry/spd_matrices.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, n, **kwargs):
         )
         self.n = n
 
-    def belongs(self, mat, atol=gs.atol):
+    def belongs(self, point, atol=gs.atol):
```
ninamiolane, line 37: Thanks!

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -9,13 +9,137 @@
...
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.geometry.spd_matrices import SPDAffineMetric, SPDMatrices
 from geomstats.information_geometry.base import InformationManifoldMixin
 from geomstats.information_geometry.normal import NormalMetric
 
 
+class MultivariateNormalDistributions:
+    """Class for the multivariate normal distributions.
+
+    This class is a common interface to the following different situations:
```
ninamiolane, line 26: Nice!

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -9,13 +9,137 @@
...
+class MultivariateCenteredNormalDistributions(InformationManifoldMixin, SPDMatrices):
+    """Class for the manifold of centered multivariate normal distributions.
+
+    This is the class for multivariate normal distributions with zero mean
+    on the $n$-dimensional Euclidean space. Each distribution is represented by
+    its covariance matrix, i.e. an $n$-by-$n$ symmetric positive-definite matrix.
+
+    Parameters
+    ----------
+    n : int
```
ninamiolane, line 64: That might be a remark for a few of the information manifolds: do we want to rename this `n` as `sample_dim` or another more explicit variable name? 

Pros:
- since `n` is used in statistics to denote the sample size, n as the dimension of the sample space can be confusing.

Cons:
- `n` here also corresponds to the `n` in SPDMatrices, thus that part is nice.

I do not have a strong opinion on this, we can keep it like this for now :)

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -9,13 +9,137 @@
...
+        -------
+        samples : array-like, shape=[..., n_samples, n]
+            Sample from centered multivariate normal distributions.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
+        point = gs.to_ndarray(point, to_ndim=3)
+        samples = []
+        loc = gs.zeros(self.n)
```
ninamiolane, line 94: NIT: maybe call this `mean` ? It seems we do not use the "location" word at all here.

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -90,7 +214,8 @@ def belongs(self, point, atol=gs.atol):
         Parameters
         ----------
         point : array-like, shape=[..., 2*n]
-            Point to test.
+            Point to test. First $n$ elements contain the mean vector
```
ninamiolane, line 217: Use :math:`n` in docstrings

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -137,7 +262,8 @@ def projection(self, point):
         Parameters
         ----------
         point : array-like, shape=[..., 2*n]
-            Point to project.
+            Point to project. First $n$ elements contain the mean vector
```
ninamiolane, line 265: Ditto

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
         return pdf
 
 
+class MultivariateGeneralNormalDistributions(InformationManifoldMixin, ProductManifold):
+    """Class for the manifold of multivariate normal distributions.
+
+    This is the class for multivariate normal distributions on the $n$-dimensional
```
ninamiolane, line 357: Ditto

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+    n : int
+        Dimension of the sample space of the multivariate normal distribution.
+    """
+
+    def __init__(self, n):
+        super().__init__(factors=[Euclidean(n), SPDMatrices(n)])
+        self.n = n
+
+    def reformat(self, point):
+        """Reformat point."""
```
ninamiolane, line 372: + Missing docstring for parameter point.
+ Missing explanation on what "reformat" means.

how about `from_array_to_tuple` or `unconcatenate` or `unpack` or something like this?

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+        point = gs.to_ndarray(point, to_ndim=2)
+        n_points = gs.shape(point)[0]
+        loc = point[:, : self.n]
```
ninamiolane, line 375: Ditto: avoid "loc"

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+        return gs.squeeze(loc), gs.squeeze(cov)
+
+    def sample(self, point, n_samples=1):
+        """Sample from a multivariate normal distribution.
+
+        Parameters
+        ----------
+        point : array-like, shape=[..., n + n**2]
+            Point representing a multivariate normal distribution. First $n$
+            elements contain the mean vector and the $n**2$ last elements contain
```
ninamiolane, line 386: Ditto

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+        n_samples : int
+            Number of points to sample with each parameter in point.
+            Optional, default: 1.
+
+        Returns
+        -------
+        samples : array-like, shape=[..., n_samples, n]
+            Sample from multivariate normal distributions.
+        """
+        # geomstats.errors.check_belongs(point, self)
```
ninamiolane, line 397: Rm

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+            elements contain the mean vector and the $n**2$ last elements contain
+            the covariance matrix row by row.
+
+        Returns
+        -------
+        pdf : function
+            Probability density function of the multivariate normal
+            distributions with parameters provided by point.
+        """
+        # geomstats.errors.check_belongs(point, self)
```
ninamiolane, line 423: Rm

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+        -------
+        samples : array-like, shape=[..., n_samples, n]
+            Sample from multivariate normal distributions.
+        """
+        # geomstats.errors.check_belongs(point, self)
+        locs, covs = self.reformat(point)
+        locs = gs.to_ndarray(locs, to_ndim=2)
+        covs = gs.to_ndarray(covs, to_ndim=3)
+        samples = []
+        for loc, cov in zip(locs, covs):
```
ninamiolane, line 402: Ditto: no locs

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+                pdf_at_xi = gs.exp(-0.5 * xi0[:, None, :] @ inv_cov @ xi0[:, :, None])
+                pdf.append(gs.squeeze(pdf_at_xi))
+            pdf = gs.stack(pdf)
+            pdf = pdf_normalization * pdf
+            return gs.squeeze(pdf)
+
+        return pdf
+
+
+class MultivariateCenteredNormalMetric(NFoldMetric):
```
ninamiolane, line 456: Maybe not for this PR, but for readability we should create:

`class ScaleMetric(NFoldMetric)` with `n_copies=1` and `scale`. Otherwise it's too much of a "hack" for the user to have to think about creating a nfold manifold with one copy. Maybe raise an issue on this?

@lpereira95

ninamiolane: Why doesn't this inherit from FisherRaoMetric?

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
...
+    """
+
+    def __init__(self, n):
+        super().__init__(
+            base_metric=SPDAffineMetric(n),
+            n_copies=1,
+            scales=[1 / 2],
+        )
+
+
```
ninamiolane, line 472: Why doesn't this inherit from FisherRaoMetric?

geomstats/information_geometry/multivariate_normal.py
```diff
@@ -225,6 +351,125 @@ def pdf(x):
         return pdf
 
 
+class MultivariateGeneralNormalDistributions(InformationManifoldMixin, ProductManifold):
```
ninamiolane, line 354: Should this inherit from openset too?

tests/data/multivariate_normal_data.py
```diff
@@ -0,0 +1,427 @@
...
+                    dict(
+                        n=n,
+                        point=self.Space(n).random_point(n_points),
+                        n_samples=n_samples,
+                    )
+                )
+        return self.generate_tests([], random_data)
+
+
+class MultivariateCenteredNormalMetricTestData(_RiemannianMetricTestData):
```
ninamiolane, line 268: Not necessarily for this PR (would be too much probably), but we could create a parent test _FisherRaoMetricTestData to factor out the tests that are the same across information metrics

ninamiolane: Same for _InformationManifoldTestData

tests/data/product_manifold_data.py
```diff
@@ -267,3 +268,42 @@ def inner_product_shape_test_data(self):
             dict(space=space, n_samples=4, point=point, tangent_vec=tangent_vec)
         ]
         return self.generate_tests(smoke_data)
+
+    def inner_product_scales_test_data(self):
```
ninamiolane, line 272: nice!

tests/tests_geomstats/test_multivariate_normal.py
```diff
@@ -26,21 +68,95 @@ def test_sample(self, n, point, n_samples, expected):
...
+        space = self.Space(n)
+        samples = space.sample(space.random_point(), n_samples)
+        result = space.point_to_pdf(point)(samples)
 
+        samples = gs.to_ndarray(samples, to_ndim=2, axis=0)
+        point = gs.to_ndarray(point, to_ndim=2, axis=0)
         expected = []
         for i in range(point.shape[0]):
-            loc, cov = Space._unstack_location_diagonal(n, point[i, ...])
+            loc, cov = space._unstack_location_diagonal(n, point[i, ...])
```
ninamiolane, line 79: @alebrigant is the function called `reformat` the same as this `_unstack_location_diagonal` ?

ninamiolane: the function `reformat` is also not tested.

ninamiolane (approved):
Looking good, many thanks for this great addition!

geomstats/information_geometry/multivariate_normal.py
ninamiolane: - Should the interface just be called "NormalDistributions"
-  and the univariate normals become "UnivariateNormalDistributions"
- and when n = 1 then UnivariateNormalDistributions is called?
It'd be nicer from a user perspective.

tests/tests_geomstats/test_multivariate_normal.py
alebrigant: > @alebrigant is the function called `reformat` the same as this `_unstack_location_diagonal` ?

No not exactly.

geomstats/information_geometry/multivariate_normal.py
alebrigant: > Should this inherit from openset too?

@ninamiolane what would be the benefit ?

tests/data/multivariate_normal_data.py
alebrigant: Yes I agree, something to do.

geomstats/information_geometry/multivariate_normal.py
alebrigant: > Why doesn't this inherit from FisherRaoMetric?

@ninamiolane It seems that as it is, FisherRaoMetric is only defined for one-dimensional sample spaces ? It would be nice to extend it to multi-dimensional sample spaces, and use it in the tests.

ninamiolane:
@alebrigant LGTM as soon as the errors in the tests are skipped (the one on TF) and the other one is:
```
Error: TestDiagonalNormalDistributions.test_sample[10-point8-5-expected8]

RuntimeError: Some points do not belong to manifold 'DiagonalNormalDistributions' of dimension 20.
```

geomstats/geometry/product_manifold.py
```diff
@@ -321,7 +320,7 @@ def _pool_outputs_from_function(self, outputs):
             and gs.all([gs.is_bool(factor_output) for factor_output in outputs])
             or (not all_arrays)
         ):
-            outputs = gs.stack(outputs)
+            outputs = gs.stack([gs.array(factor_output) for factor_output in outputs])
```
lpereira95, line 323: do you remember what's the case that messed up with this line before you changes?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            Probability density function of the centered multivariate normal
+            distributions with covariance matrices provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
+        point = point[None, :, :] if point.ndim == 2 else point
+        n = self.sample_dim
+
+        def pdf(x):
```
lpereira95, line 229: following `InformationManifoldMixin`, this should be a method with signature `(x, point)`.

after this change, it is very likely that `point_to_pdf` does not need to be defined, as `InformationManifoldMixin` has a general implementation of it.

The same applies for the other manifolds.

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            while pdf_normalization.ndim < pdf.ndim:
+                pdf_normalization = pdf_normalization[..., None]
+            pdf = pdf_normalization * pdf
+            pdf = gs.squeeze(pdf)
+            return pdf
+
+        return pdf
+
+
+class GeneralNormalDistributions(InformationManifoldMixin, ProductManifold):
```
lpereira95, line 461: Isn't `MultivariateNormalDistributions` a better name?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+    ----------
+    sample_dim : int
+        Dimension of the sample space of the multivariate normal distribution.
+    """
+
+    def __init__(self, sample_dim):
+        super().__init__(factors=[Euclidean(sample_dim), SPDMatrices(sample_dim)])
+        self.sample_dim = sample_dim
+
+    def unstack_mean_covariance(self, point):
```
lpereira95, line 478: private?

lpereira95 (changes requested):
Thanks @alebrigant!

Some initial comments that will require some changes.

I'll try to make it work with tf also, so we don't have to skip the tests.

tests/data/normal_data.py
```diff
@@ -0,0 +1,475 @@
...
+        ]
+        sample_dim = self.sample_dim_list[1]
+        n_samples = self.n_samples_list[1]
+        random_data.append(
+            dict(
+                sample_dim=sample_dim,
+                point=self.Space(sample_dim).random_point(n_samples=n_samples),
+                expected=gs.array([True] * n_samples),
+            )
+        )
```
lpereira95, line 40: This data is "repeated" because `random_point_belongs_test_data` does exactly this.

(same applies for other test data)

geomstats/information_geometry/normal.py
lpereira95: looking to this now, I think we should move `pdf` and `cdf` again within `point_to_pdf` and `point_to_cdf` in `InformationManifoldMixin`. (this was my suggestion @ninamiolane, but in this new code I can see what we've done is not so general as I would like.)

That's because e.g. `CenteredNormalDistributions` requires the computation of `det` and `inv` of point (i.e. it is independent of `x`), so we should do it in the scope of `point_to_pdf` instead of doing it in the `pdf` itself.

geomstats/information_geometry/normal.py
```diff
@@ -1,29 +1,73 @@
...
+    """
+
+    def __new__(cls, sample_dim, distribution_type="general"):
+        """Instantiate class that corresponds to the distribution_type."""
+        errors.check_parameter_accepted_values(
+            distribution_type,
+            "distribution_type",
+            ["centered", "diagonal", "general"],
+        )
+        if sample_dim == 1:
```
lpereira95, line 53: what if "centered" and `sample_dim == 1`?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            Parameters
+            ----------
+            x : array-like, shape=[n_samples, sample_dim]
+                Points at which to compute the probability
+                density function.
+            """
+            x = gs.to_ndarray(x, to_ndim=2, axis=0)
+            det_cov = gs.linalg.det(point)
+            inv_cov = gs.linalg.inv(point)
+            pdf_normalization = 1 / gs.sqrt(gs.power(2 * gs.pi, n) * det_cov)
```
lpereira95, line 241: depend only on point: can be moved outside the scope of `pdf` (performance enhancement).

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+
+        def pdf(x):
+            """Generate parameterized function for normal pdf.
+
+            Parameters
+            ----------
+            x : array-like, shape=[n_samples, sample_dim]
+                Points at which to compute the probability
+                density function.
+            """
```
lpereira95, line 237: Can you please add the return shape? It is non obvious for the case of several points.

Additionally, if `n_samples == 1`, can I use an array of dim 1?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+
+        Returns
+        -------
+        pdf : function
+            Probability density function of the centered multivariate normal
+            distributions with covariance matrices provided by point.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
```
lpereira95, line 225: why?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            Optional, default: 1.
+
+        Returns
+        -------
+        samples : array-like, shape=[..., n_samples, sample_dim]
+            Sample from centered multivariate normal distributions.
+        """
+        geomstats.errors.check_belongs(point, self)
+        if point.ndim > 3:
+            raise NotImplementedError
```
lpereira95, line 200: in other parts of the code we decided to not make any check. Should we be consistent?

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+                Points at which to compute the probability
+                density function.
+            """
+            x = gs.to_ndarray(x, to_ndim=2, axis=0)
+            det_cov = gs.linalg.det(point)
+            inv_cov = gs.linalg.inv(point)
+            pdf_normalization = 1 / gs.sqrt(gs.power(2 * gs.pi, n) * det_cov)
+            pdf = []
+            for xi in x:
+                pdf.append(gs.exp(-0.5 * gs.transpose(xi) @ inv_cov @ xi))
```
lpereira95, line 244: we can probably avoid this for.

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            Parameters
+            ----------
+            x : array-like, shape=[n_samples, sample_dim]
+                Points at which to compute the probability
+                density function.
+            """
+            x = gs.to_ndarray(x, to_ndim=2, axis=0)
+            x = x[None, :, :]
+            det_cov = gs.squeeze(gs.prod(diagonal, axis=-1))
+            pdf_normalization = 1 / gs.sqrt(gs.power((2 * gs.pi), n) * det_cov)
```
lpereira95, line 450: independent of `x`. can be moved outside.

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+            pdf = gs.exp(-0.5 * gs.sum(((x - mean) ** 2) / diagonal, axis=-1))
+            while pdf_normalization.ndim < pdf.ndim:
+                pdf_normalization = pdf_normalization[..., None]
```
lpereira95, line 453: this should probably be done with reshape or something similar.

geomstats/information_geometry/normal.py
```diff
@@ -116,8 +160,420 @@ def pdf(x):
...
+                Points at which to compute the probability
+                density function.
+            """
+            x = gs.to_ndarray(x, to_ndim=2, axis=0)
+            mean, cov = self.unstack_mean_covariance(point)
+            mean = gs.to_ndarray(mean, to_ndim=2)
+            cov = gs.to_ndarray(cov, to_ndim=3)
+            det_cov = gs.linalg.det(cov)
+            inv_cov = gs.linalg.inv(cov)
+            pdf_normalization = 1 / gs.sqrt(gs.power(2 * gs.pi, n) * det_cov)
```
lpereira95, line 562: independent of `x`.

lpereira95 (commented):
Some additional comments @alebrigant. Specially related with `point_to_pdf` performance improvement.

geomstats/geometry/product_manifold.py
alebrigant: Unfortunately no, it was quite a long time ago...

geomstats/information_geometry/normal.py
alebrigant: All the classes except `UnivariateNormalDistributions` refer to multivariate distributions, so I'm afraid it might be confusing.

tests/data/normal_data.py
alebrigant: What change do you suggest here ?

geomstats/information_geometry/normal.py
alebrigant: If sample_dim is one, we don't need to distinguish between special cases since everything is explicit for univariate distributions.

geomstats/information_geometry/normal.py
alebrigant: I added the return shape. You can equivalently use `x` of size [1, sample_dim] or [sample_dim,].

geomstats/information_geometry/normal.py
alebrigant: You are right this should not be here, thanks !!

geomstats/information_geometry/normal.py
alebrigant: Maybe this can be replaced by two einsums but I'm afraid this will result in a loss of readability.

geomstats/information_geometry/normal.py
alebrigant: Why is reshape better ? This is Antoine's code, I'm not exactly sure what he wanted to do with this.

geomstats/information_geometry/normal.py
alebrigant: This generates error 'pdf_normalization referenced before assignment' when moved outside.

geomstats/information_geometry/normal.py
alebrigant: I think this should be fixed in a different PR, so that this issue can be debated with @ninamiolane.

geomstats/information_geometry/normal.py
lpereira95: It is more readable and we avoid creating several copies of `pdf_normalization` with decreasing number of dimensions (one for each iteration in the while).

tests/data/normal_data.py
lpereira95: I was suggesting to delete all the cases that generate random points, because `random_point_belongs` does exactly that (check line218 of `data_generation.py`). If I'm not missing anything, the "automatic" code generates exactly the same test cases as your code here (the data points will be different only due to randomness).

PR #1731, opened 2022-11-03
Fix Siegel log and exp errors for zero matrices

YannCabanes:
The Siegel Riemannian logarithm is was failing when `point` equals the null matrix.
The Siegel Riemannian exponential map is failing when `tangent_vec` equals the null matrix.
These errors are due the the matrix inversion of a null matrix.
This PR fix these errors and add the corresponding tests.


YannCabanes:
Hello @lpereira95,
This PR is fixing errors of the Siegel logarithm et exponential map for null matrices.
The failing tests do not seem to be related to this PR.

lpereira95:
Thanks for spotting this @YannCabanes. Merging.

PR #1733, opened 2022-11-09
Add conda install guidelines

nguigs:
Add conda install command

@ymontmarin : does this also work on windows with Anaconda Prompt?


ymontmarin:
@nguigs Yes it works for all architectures !

nguigs:
Good point @ymontmarin, I did the update too.

ymontmarin:
@nguigs looks good !

PR #1736, opened 2022-11-16
Amend errors.check_point_shape 

johnharveymath:
## Description

`geomstats.errors.check_point_shape` is amended so that it does not always raise a `ShapeError` if the shapes do not match. It is possible to simply check the shape by setting the optional parameter `suppress_error=True` and obtain a boolean return value. This is as requested by other developers during the IHP hackathon

## Issue

We can use this to provide checks on inputs in methods for all `Manifold` and `RiemannianMetric` subclasses, if that seems worthwhile. In that case we should probably implement a method `check_point_shape` in those classes which wraps this, so that it can simply be called as `self.check_point_shape(point)`.

### DIFF
--- a/geomstats/errors.py
+++ b/geomstats/errors.py
@@ -85,13 +85,13 @@ def check_tf_error(exception, name):
     return exception
 
 
-def check_point_shape(point, manifold):
-    """Raise an error if the shape of point does not match the shape of a manifold.
+def check_point_shape(point, manifold, suppress_error=False):
+    """Check if the shape of point does not match the shape of a manifold or metric.
 
     If the final elements of the shape of point do not match the shape of manifold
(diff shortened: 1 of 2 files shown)


johnharveymath:
Failing tests are not relevant to this pull request. Ready to merge, unless you want me to expand functionality as mentioned in the first comment.

lpereira95:
Thanks @johnharveymath!

For now I think it is better to not add it in `Manifold` and `RiemannianMetric` because we explicitly decided to not make this kind of checks. Maybe in the future will reconsider this, but it should be discuss with everyone and measure the implications (e.g. if we provide for shape checking, should we also provide other kind of checks?).

One that that may be worthwhile is to extend this idea of `suppress_error` to other `check_` defined in `errors`. We can leave it for a future PR.

I've just added a quick test in this PR.

PR #1737, opened 2022-11-17
Support product manifolds of matrices

YannCabanes:
This PR is related to the PR #1727 adding a product manifold `ProductHPDMatricesAndSiegelDisks` which is a product manifold of matrices.
To add this manifold, we have to modify the way that `ProductManifold` handles the points shape.
This PR is inspired by a discussion with @lpereira95 and @johnharveymath in PR #1727.


YannCabanes:
Hello @lpereira95,

johnharveymath:
I have created a PR on Yann's branch with my own changes so @YannCabanes can consider whether to merge them. These implement the type `'other'` in the `Manifold` class, allow a `default_point_type` of `None` in the `ProductManifold` so that you can leave it to the code to choose the best representation, and include a bit of flexibility for arbitrarily deep shapes.

YannCabanes:
Thank you @johnharveymath, I have merged your changes.

YannCabanes:
Hello @lpereira95, 
Would you like to replace `default_point_type` by `point_ndim` everywhere in geomstats in this PR?

lpereira95:
Maybe not a great idea to do it everywhere in this PR... to avoid doing too much simultaneously, let's move on with the current solution of using `"other"` and I'll make the change everywhere after we merge.

YannCabanes:
Ok, I prefer this solution too.

johnharveymath:
I agree. Best to separate this.

By the way, we forgot to make this change to add the `'other'` option in
`RiemannianMetric` mirroring the change in `Manifold`.

YannCabanes:
Hello @johnharveymath,
For now the word `default_point_type` does not appear in the file `riemannian_metric.py`.
Do you think that we should add something in the file `riemannian_metric.py`?
If you think so, you can do some modifications from my branch and I will merge them as we did before.

johnharveymath:
In that case I reckon we leave it as is. Sorry for the confusion!

YannCabanes:
There is a conflict between the latest changes in the file `manifold.py` due to PR #1739 and the changes proposed by @johnharveymath in this PR.
@lpereira95, I let you decide which version to keep. 
Once you have decided, could you review this PR please?

lpereira95:
@YannCabanes, @johnharveymath, I'm merging this as it accomplishes its proposed goal. There's some follow ups coming from our discussion. We will address them in other PRs.

Thanks for your contributions.

PR #1709, opened 2022-10-21
Add complex poincare disk

YannCabanes:
Add the complex Poincaré disk manifold.

The Poincaré complex disk is a representation of the Hyperbolic space of dimension 2.

References
----------
.. [Cabanes2022] Yann Cabanes. Multidimensional complex stationary
    centered Gaussian autoregressive time series machine learning
    in Poincaré and Siegel disks: application for audio and radar
    clutter classification, PhD thesis, 2022
.. [JV2016] B. Jeuris and R. Vandebril. The Kahler mean of Block-Toeplitz
    matrices with Toeplitz structured blocks, 2016.
    https://epubs.siam.org/doi/pdf/10.1137/15M102112X
.. [Yang2013] Marc Arnaudon, Frédéric Barbaresco and Le Yang.
    Riemannian medians and means with applications to radar signal processing,
    IEEE Journal of Selected Topics in Signal Processing, vol. 7, no. 4, pp. 595-604,
    Aug. 2013, doi: 10.1109/JSTSP.2013.2261798.
    https://ieeexplore.ieee.org/document/6514112

PR #1701, opened 2022-10-20
Add the positive reals manifold

YannCabanes:
Add the manifold of positive reals.

This manifold is a particular case in dimension 1
of the SPD matrices endowed with the SPD affine-invariant metric
with a power affine coefficient equal to one.
This manifold is used as part of a product manifold involving
complex Poincare disks to classify complex stationary centered
Gaussian autoregressive time series in [Cabanes2022], [Cabanes_CESAR_2019],
[Cabanes_GSI_2019] and [Cabanes_RADAR_2019].

References
----------
.. [Cabanes2022] Yann Cabanes. Multidimensional complex stationary
    centered Gaussian autoregressive time series machine learning
    in Poincaré and Siegel disks: application for audio and radar
    clutter classification, PhD thesis, 2022
.. [Cabanes_CESAR_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon,
    Jérémie Bigot. Unsupervised Machine Learning for Pathological Radar
    Clutter Clustering: the P-Mean-Shift Algorithm, C&ESAR 2019, Nov 2019,
    Rennes, France. hal-02875430
    https://hal.archives-ouvertes.fr/hal-02875430/document
.. [Cabanes_GSI_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon,
    Jérémie Bigot. Toeplitz Hermitian Positive Definite Matrix Machine Learning
    based on Fisher Metric. Geometric Science of Information, 2019,
    10.1007/978-3-030-26980-7_27. hal-02875403
    https://hal.archives-ouvertes.fr/hal-02875403/document
.. [Cabanes_RADAR_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon,
    Jérémie Bigot. Non-Supervised High Resolution Doppler Machine Learning
    for Pathological Radar Clutter. RADAR 2019, Sep 2020, Toulon, France.
    10.1109/RADAR41533.2019.171295. hal-02875415
    https://hal.archives-ouvertes.fr/hal-02875415/document
.. [JV2016] B. Jeuris and R. Vandebril. The Kahler mean of Block-Toeplitz
    matrices with Toeplitz structured blocks, 2016.
    https://epubs.siam.org/doi/pdf/10.1137/15M102112X

PR #1739, opened 2022-11-18
Fix point shapes

lpereira95:
This PR:

**1)** sets the expectations for point shapes of manifolds of dimension 1 (relevant for you @YannCabanes):

* point: `(1,)`; multiple points `(n_points, 1)`
* several reasons for this:

i. spaces that also work with higher dimensions naturally represent points for case `dim=1` as `(1,)` (e.g. `Euclidean`)
ii. `()` treatment would require several exceptions in the code: e.g. `ProductManifold`; `gs.random.rand` use to create a generic function to create random points 

* points are represented by `gs.array`, never `float`

**2)** fixes several general functions in abstract classes (building on top of previous point)

**3)** fixes several vectorizations in abstract classes when `point` is ignored

**4)** fixes docstrings of abstract classes regarding points shapes

**5)** modifies `BinomialDistributions` to apply **1)** (@alebrigant can you please review to ensure I do not introduce any error?)

@YannCabanes can you please review the changes in complex manifolds abstract classes to ensure I haven't missed anything out?

### DIFF
--- a/geomstats/_backend/autograd/__init__.py
+++ b/geomstats/_backend/autograd/__init__.py
@@ -398,7 +398,7 @@ def ndim(x):
 
 
 def copy(x):
-    return x.copy()
+    return _np.array(x, copy=True)
 
 
 def array_from_sparse(indices, data, target_shape):
--- a/geomstats/_backend/tensorflow/__init__.py
+++ b/geomstats/_backend/tensorflow/__init__.py
@@ -343,7 +343,7 @@ def _assignment_single_value(x, value, indices, mode="replace", axis=0):
     if use_vectorization:
         full_shape = shape(x)
         n_samples = full_shape[axis]
-        tile_shape = list(full_shape[:axis]) + list(full_shape[axis + 1 :])
+        tile_shape = list(full_shape[:axis]) + list(full_shape[axis + 1:])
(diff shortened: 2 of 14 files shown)


YannCabanes:
Hello @lpereira95, 
At first sight, everything that you have done in the file `complex_manifold.py` looks fine to me.

YannCabanes:
I will take your latest decision about the point shape of manifolds of dimension 1 in PR #1701 adding `PositiveReals` and PR #1709 adding `ComplexPoincareDisk`. I will change back the point shape from `[...]` to `[..., 1]`.

johnharveymath:
I just left this comment on the related issue - I see now it's a little too late @lpereira95 ! But in case you want to revisit, here are my thoughts:

I would suggest that this particular variable name might cause confusion with `manifold.dimension`. 

Perhaps `manifold.point_repr_depth` or `manifold.point_coord_depth` or `manifold.point_array_depth` or `manifold.shape_depth`? I think 'depth' is a good word for indicating how many dimensions an array has without using the word 'dimension'. Something that makes it clear that it is not a mathematical property of the manifold but a property of how we are representing or co-ordinatizing points of the manifold. `manifold.shape_depth` might be my favourite because it suggests, accurately, that the property is derived from `manifold.shape`.

PR #1740, opened 2022-11-20
Two different versions of the scalar product metric.

### LINKED ISSUE #1705: Scaling metrics
The ProductManifold class can be used to rescale a metric by passing only one factor. However, this isn't very efficient so it might be worthwhile improving performance for a single factor or developing an alternative method.

### LINKED ISSUE #1738: Implement scalar product of metric
We need a method to multiply a metric by a scalar. After discussion with @alebrigant we suggest the following. Thoughts from everyone very welcome!

In `RiemannianMetric` implement `__mul__` and `__rmul__` to allow operations of the type `new_metric = 2 * old_metric` which will return a `ScalarProductMetric`. This will give a simple and intuitive way to make scalar products.

Implement `ScalarProductMetric` which will simply wrap `NFoldMetric` with one product.

Then the key work should happen in `NFoldMetric`. This makes sense because `NFoldMetric` should already be little more than a wrapper for the underlying metric. It is already configured to take an argument `scales` which rescales the metric on the different factors. However, it only implements a couple of basic methods from `RiemannianMetric` and it is not set up to take advantage of any of the implementations in the various subclasses. I see a couple of options to remedy that.

1. Dynamic implementation: edit `NFoldMetric.__getattribute__` to obtain all attributes from the underlying metric and then pass it to a wrapper which rescales the output. The scaling factor is hard-coded into `NFoldMetric` (ie, is it linear, square-root, reciprocal) and if `NFoldMetric` doesn't know what to do it throws an `AttributeError` and you have to add your missing method to the list of known methods. It is possible to change the underlying metric on the fly.
2. Generate attributes at `__init__`: the same as above expect that instead of editing `__getattribute__`, at initialization we use the `__dir__` method of the underlying metric to gather all its attributes, we wrap them, and we put them into our `ScalarProductMetric` object.  Now if you change the underlying metric, it has no impact on the `ScalarProductMetric` object, because it has already collected all the attributes. 
3. Implement methods one by one: each method gets its own implementation in `ScalarProductMetric` which calls the underlying method and rescales it. `__init__` will have to collect all variables from the underlying metric.

Does using `NFoldMetric` seem reasonable? As for the implementation, my preference is to go with option 1 or 2. If you want to teach `ScalarProductMetric` a new method you just have to add it to one of the hard-coded lists so that it knows how to scale the result, rather than putting in a whole new method. That keeps the code a bit cleaner, I think. I'm not sure which I like better. I don't see any clear advantages to either. They have different functionality, obviously, but I do not see clearly which functionality is more likely to be useful. In all three cases, one risk to be aware of is the possibility that different child classes of `RiemannianMetric` could have different methods with the same name which have to be scaled differently. That seems unlikely, but it's something to consider.

johnharveymath:
Draft form for considering best possible implementation.

Should dynamically inherit from underlying_metric but that's not implemented yet. One concrete benefit from that is you get tooltips!

Resolves #1738 
Resolves #1705


lpereira95:
Thanks @johnharveymath, great PR!

I prefer version 1, because wrapping is done only once at instantiation.

Regarding dynamic inheritance, I thought it would be the best way to go, but after looking to your solution I think it should not be done: some methods depend on others, if we apply the scale to the underlying computations (which will happen if we inherit), then we are scaling several times and we will get a wrong result (e.g. `dist` calls `squared_dist`).

I have some suggestions to improve version 1, and I think after addressing them, it is almost ready to merge.


**Suggestions**

Changes in init:
```python
    def __init__(self, underlying_metric, scaling_factor):
        """Load all attributes from the underlying metric."""
        self.underlying_metric = underlying_metric
        self.scaling_factor = scaling_factor

        reserved_names = ("underlying_metric", "scaling_factor")
        for attr_name in dir(self.underlying_metric):
            if attr_name.startswith("__"):
                # TODO: should we only jump __init__?
                continue

            if attr_name in reserved_names:
                raise AttributeError(
                    f"The underlying metric has an attribute '{attr_name}' but this name is"
                    f"reserved for the class 'ScalarProductMetric'"
                )

            attr = getattr(underlying_metric, attr_name)
            if not callable(attr):
                try:
                    setattr(self, attr_name, attr)
                except AttributeError as ex:
                    if not isinstance(
                        getattr(type(underlying_metric), attr_name, None), property
                    ):
                        raise ex
            else:
                scaling_factor = _get_scaling_factor(attr.__name__, self.scaling_factor)

                if scaling_factor is None:
                    method = attr
                else:
                    method = _wrap_attr(scaling_factor, attr)

                setattr(
                    self,
                    attr_name,
                    method,
                )

```

where,

```python
def _get_scaling_factor(func_name, scale):
    if func_name in SQRT_LIST:
        return gs.sqrt(scale)

    elif func_name in LINEAR_LIST:
        return scale

    elif func_name in QUADRATIC_LIST:
        return gs.power(scale, 2)

    elif func_name in INVERSE_LIST:
        return 1.0 / scale

    elif func_name in INVERSE_QUADRATIC_LIST:
        return 1.0 / gs.power(scale, 2)

    return None
```

and 

```python
def _wrap_attr(scaling_factor, func):
    @wraps(func)
    def response(*args, **kwargs):
        return scaling_factor * func(*args, **kwargs)

    return response

```

This means:

1. we remove `INVARIANT_LIST`: if a method is not listed in any of the other lists, it is assumed the output is not scaled and we use the method directly (this is much easier to maintain; I think listing everything that is invariant is unfeasible and not generic enough)
2. we uncouple the wrapping and the computation of the scaling factor

I also think that `underlying_metric` should be private and `self.scaling_factor` should be renamed `scale` later.


For testing, we should select some manifolds (let's start with e.g. `Euclidean`) and compare scaled vs non-scaled metrics. We don't need to over test: e.g. if it works for `squared_dist`, it will work for `squared_norm`.


After merging I think we should:
1. remove `scale` in the metrics it is currently used
2. remove `scales` from `ProductManifold`: the user should be responsible to create `ScalarProductMetric` if needed


Let me know what you think!

geomstats/geometry/riemannian_metric.py
```diff
@@ -46,6 +46,42 @@ def __init__(
...
+
+        Returns
+        -------
+        metric : ScalarProductMetric
+            The metric multiplied by the scalar.
+        """
+        from geomstats.geometry.scalar_product_metric import ScalarProductMetric
+        if not isinstance(scalar, (float, int)):
+            raise NotImplementedError
+        return ScalarProductMetric(self, scalar)
```
lpereira95, line 82: should be enough to do `return self*scalar`, right?

johnharveymath: Absolutely

johnharveymath:
OK - I'm on board with all of that and will implement. For the power users, I think we want to bring the wrapping function into the class so they can use some public method to edit those lists as class (or instance?) variables. They were only out at module level because I had so much getting and setting going on in some versions that it was useful to keep the class itself as empty as possible.

lpereira95:
I'll address testing @johnharveymath.

Regarding power users, I'm not sure if bringing the wrapping function into the class or if provide nicer ways to change the lists at module level (they can already do it, but I find it very hacky). Probably I would go for the second.

johnharveymath:
Thanks @lpereira95  - I need to understand testing a lot better! Though I did do enough testing to find some issues with my code.

I have pushed some final improvements to the documentation and then I suppose we can cover the other matters later. Just to summarise what they are:

**Follow-up tasks to do after merging:**
- a method for users to change the lists in `scalar_product_metric.py`
- remove variable `scales` from `ProductRiemannianMetric` and `NFoldMetric`
- rename variable `scaling_factor` to `scale`, making changes to classes that already use `scale`
- check all existing classes that use `scale` to make sure their results match the results from `ScalarProductMetric`, as I think there are a couple of issues.

johnharveymath:
Resolves #1705 - I had a duplicate issue open already.

johnharveymath:
Thanks for all the tests - looks great!  By the way, the reason I was raising a `NotImplementedError` in `__mul__` was that I understood that this was the standard pattern. See [this stackoverflow question](https://stackoverflow.com/questions/5181320/under-what-circumstances-are-rmul-called). 

When `a * b` is used, Python basically does 
```
try:
    return a.__mul__(b)
except NotImplementedError: 
    return b.__rmul__(a)
``` 
If someone developed their own class and wanted to allow multiplication by objects of type `RiemannianMetric`, raising `TypeError` would pose a problem, but this is pretty unlikely!

lpereira95:
Thanks @johnharveymath!

I was not aware of this behavior from Python. I think the `NotImplementedError` is a good idea (I'll revert that change).

I've removed the possibility of `int` because it may create problems with some of the backends. Besides, I'm trying to make people more aware that `int` and `float` are very different and should be used properly.

Regarding the tests, this is not the usual approach in geomstats. But looking to the code, I don't think we gain a lot by overtesting this (e.g. comparing outputs of `underlying_metric` and `metric`). So, I've decided to simply check if the wrapper is working properly. If so, then the results will also be proper.

Again, great work with this! I'll merge after fixing these last things.

johnharveymath:
I can do those follow-up tasks that I referred to.

lpereira95:
That would be great @johnharveymath! I'll open some issues so we can keep track on the tasks and discuss implementations.

I think we should start by "check all existing classes that use scale to make sure their results match the results from ScalarProductMetric, as I think there are a couple of issues." I've started this myself, but the results are super inconsistent. I think the problem may be in the classes that implement `scale`. Ideally, this task is a check, to ensure `ScalarProductManifold` is correctly implemented, followed by the removal of `scale` from all the manifolds.

`ProductManifold` should not be considered for now. Let's focus on it after the first task and the merge of #1737 (which I'll try to do still today).

Issue #1704, opened 2022-10-20
Clean namespaces, especially for RiemannianMetric


ninamiolane:
### What should be modified?

The attributes and methods of a class should be cleaned, e.g.:

Move RiemannianMetric's
- closest neighbor
- distance pairwise
to another place

Similar question for how to handle linear algebra computations in SymmetricMatrices: for example, use a flag

PR #1742, opened 2022-11-22
[DRAFT] Refactoring complex objects with a flag for field in ["real", "complex"]

sophiaas:
## Checklist

- [X] If neccessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [ ] I have added apropriate unit tests.
- [X] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#coding-style-guidelines) and API.
- [X] I have linked to issues and PRs that are relevant to this PR.

## Description

This is a draft of a refactor for complex objects such as ComplexMatrices. Currently, adding a complex object requires duplicating the real object and replacing methods. I propose an alternative in which the basic objects (e.g. Matrices, GeneralLinearGroup) are passed a flag to specify the `field`, which can take a value in `["real", "complex"]`. This flag is checked during certain methods that involve things like the inner product or instantiation of new vectors, which are handled differently for real and complex fields. This will simplify the process of adding new complex-valued objects, and is more consistent with the current strategy for coordinatization. 

Before finishing the refactor, I wanted to raise the proposal and get approval. Please let me know what you think, or if you prefer an alternative approach.

## Issue

This PR addresses the current difficulty of adding new complex-valued objects.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -86,7 +86,6 @@ cos = _box_unary_scalar(target=_torch.cos)
 cosh = _box_unary_scalar(target=_torch.cosh)
 exp = _box_unary_scalar(target=_torch.exp)
 floor = _box_unary_scalar(target=_torch.floor)
-imag = _box_unary_scalar(target=_torch.imag)
 log = _box_unary_scalar(target=_torch.log)
 real = _box_unary_scalar(target=_torch.real)
 sign = _box_unary_scalar(target=_torch.sign)
@@ -762,3 +761,11 @@ def cross(a, b):
 
 def gamma(a):
     return _torch.exp(_gammaln(a))
+
+
+def imag(a):
+    if not _torch.is_tensor(a):
+        a = _torch.tensor(a)
+    if is_complex(a):
+        return _torch.imag(a)
+    return _torch.zeros(a.shape, dtype=a.dtype)
--- a/geomstats/geometry/positive_reals.py
+++ b/geomstats/geometry/positive_reals.py
@@ -1,0 +1,282 @@
+"""The manifold of positive reals.
+
+This manifold is a particular case in dimension 1
+of the SPD matrices endowed with the SPD affine-invariant metric
+with a power affine coefficient equal to one.
+This manifold is used as part of a product manifold involving
+complex Poincare disks to classify complex stationary centered
+Gaussian autoregressive time series in [Cabanes2022], [Cabanes_CESAR_2019],
+[Cabanes_GSI_2019] and [Cabanes_RADAR_2019].
+This product manifold is called ProductPositiveRealsAndComplexPoincareDisks
+in geomstats. It is a product manifold of complex type, therefore the manifold
+PositiveReals is compatible with complex input points and vectors even if the
+input values should be reals.
+
+
+Lead author: Yann Cabanes.
+
+References
+----------
+.. [Cabanes2022] Yann Cabanes. Multidimensional complex stationary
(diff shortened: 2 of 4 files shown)


lpereira95:
I have mixed feelings here.

I agree there's a lot of code duplication that should be avoided. I'm not sure having a string field variable is the best way to go.

My main points:

1. having to add a lot of ifs everywhere (`if self.field == "real"`) does not feel right. Whenever this happens, it means we should split the class (which is what we have already done).

2. almost everyone (@YannCabanes is the only exception I'm aware of) develops code in geomstats thinking about real data. Having to think about complex numbers (or other fields) when creating a manifold (or changing an existing one as matrices) makes it much harder (too much information to keep in the brain simultaneously). The same applies for debugging. Test will also be harder.

3. method naming will become harder: e.g. should we use `transpose` or  `transconjugate`? why should someone not interested in complex fields care about (or even remember) using `transconjugate`.

4. attributes of a manifold may change for complex or real fields: e.g. `dim` of `Matrices` is $n\times m$, but `dim` of `ComplexMatrices` should probably be $2n\times m$ if we want to be rigourous.

5. backends have much less support for complex numbers that for reals.


After addressing #1704, a lot of the code duplication will disappear in `Matrices` and `ComplexMatrices`.

An alternative way I see to reduce code duplication is to transform some methods in functions and then wrap the functions in method calls in the different objects that need the code.

Note also that for the end user perspective we can have a `field` variable that works as an interface in case we want to somehow hide that `Matrices` and `ComplexMatrices` are different objects.

ninamiolane:
1. Agree. That's the main drawback in my opinion.
2. Partially agree: 
- There are users besides Yann and Sophia waiting for the complex version of geomstats: e.g. for quantum physics and quantum computing.
- The possibility of having many fields is done in Manifolds.jl (see screenshot): there is an interest.
- Since the default is field = "real", the change could be invisible for users. However, it might become apparent in tests, i.e. not invisible for contributors.
4. Partially agree: there are not that many name conflicts. transpose/transconjugate are called "adjoint" for both, yet, SPDMatrices versus HermitianMatrices does pose a name conflict. 
5. Agree.
6. Agree, but that might not be a problem.

> An alternative way I see to reduce code duplication is to transform some methods in functions and then wrap the functions in method calls in the different objects that need the code.

That might become too obscure for contributors.
Screen Shot 2022-11-23 at 9 00 18 AM

How about these classes:
- `_Matrices`: factor the code and has computations that work for all fields
- `_RealMatrices`
- `_ComplexMatrices`
- `Matrices`: interface choosing the field.

lpereira95:
2. I agree with adding support to complex fields. I just think we need to think thoroughly HOW to do it

3. in terms of manifold names I think there will be some `SPD`-`HPD`, `Symmetric`-`Hermitian`, etc. But this does not concern be so much though (we can always create `SPD` by instantiating `HPD` with a real field). I more concerned on method names because used to real fields does not have the nomenclature for complex fields (being it `transconjugate` or `adjoint`: I simply expect `transpose` to be available)

5. Probably not a huge problem, but some casting are sometimes needed for complex fields, which means additional ifs in the code.


Which kind of (inheritance) relationships are you thinking about for  `_Matrices`, `_RealMatrices` and `_ComplexMatrices`?

ninamiolane:
Not entirely sure...

class _Matrices(_VectorSpace)
class _RealMatrices(_Matrices)
class _ComplexMatrices(_Matrices)

where _VectorSpace is also field-agnostic?

lpereira95:
That looks good to me. We will probably require all the abstract classes to be field agnostic.

Issue #1753, opened 2022-12-08
Make gs.reshape accept float in pytorch 


tramy1258:
I would like to do something like this:

```python
gs.reshape(2, (-1,))
```

It works with all the backends except `pytorch`. For `pytorch` the error message is the following:

`TypeError: reshape(): argument 'input' (position 1) must be Tensor, not int`

Same problem and error in `pytorch` for the function below so it might come down to a same problem.
```python
gs.expand_dims(2, axis=-1)
```

tramy1258:
I have tested your solution and it works well for all backends. Besides that, it also resolves automatically the problem I had with Tensorflow which is the clashing between `float64` in an array and `float32` with single floats, like when users pass a single float for `path(t)` in `geodesic` case.

I also agree that it would be cumbersome to modify such primary behavior, especially when we have this type of workaround.

Thank you for your reply, I will close the issue.

PR #1755, opened 2022-12-12
Add functions and associated tests for Binomial and Exponential Distributions

tramy1258:
## Checklist

- [x] I have linked to issues and PRs that are relevant to this PR.

## Description

Add metric_matrix and geodesic for Binomial and Exponential Distributions as well as associated tests.
Code requires #1753 resolved to work properly with pytorch.


tramy1258:
1. I have added 
* `metric_matrix`, `geodesic`, `exp` and `log` for `BinomialMetric` and `ExponentialMetric`.
* `test_metric_matrix` and `test_geodesic_symmetry`  and associated data for `TestBinomialMetric` and `TestExponentialMetric`.

2. I have also modified the output shape of `squared_dist` so that `test_dist_is_norm_of_log` is successful, this should not affect other functions and tests.

3. I have skipped
* `test_log_after_exp` for `TestBinomialMetric` because this leads to testing something of format `a==arcsin(sin(a))`, which is not always true because the latter is often restricted to `[-pi/2, pi/2]`.
* `test_exp_geodesic_ivp` for both classes because this uses `tangent_vec` as a vector `(...,)` instead of an array `(...,1)`, which is not coherent with other functions using `tangent_vec`.
* a case of `test_point_to_pdf` where `k` is a 2D array for both classes. Specifically, when something like this `dict(point=self.Space().random_point(2), n_samples=3)` is given as data, then `samples` in `test_point_to_pdf` will be of shape `(2,3)` and we then call `pdf(samples)` while `pdf` should only work with vectors.

geomstats/information_geometry/binomial.py
```diff
@@ -181,18 +175,240 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+            )
+        if n_initial_tangent_vecs > n_initial_points > 1:
+            raise ValueError(
+                "For several initial tangent vectors, "
+                "specify either one or the same number of "
+                "initial points."
+            )
+
+        c = gs.arcsin(gs.sqrt(initial_point))
+        K = initial_tangent_vec / (
```
alebrigant, line 248: We try to avoid single letters for variable names. Maybe change to `constant_a` and `constant_b` or `intercept` and `slope` or anything else that seems relevant.

geomstats/information_geometry/binomial.py
```diff
@@ -181,18 +175,240 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+                "one or the same number of end points."
+            )
+        if n_end_points > n_initial_points > 1:
+            raise ValueError(
+                "For several end points, specify either "
+                "one or the same number of initial points."
+            )
+
+        c = gs.arcsin(gs.sqrt(initial_point))
+        K = gs.arcsin(gs.sqrt(end_point)) - c
```
alebrigant, line 303: Ditto.

tests/data/binomial_data.py
```diff
@@ -86,11 +89,6 @@ def sample_belongs_test_data(self):
 
     def point_to_pdf_test_data(self):
         random_data = [
-            dict(
-                n_draws=5,
-                point=self.Space(5).random_point(2),
-                n_samples=3,
-            ),
```
alebrigant, line 93: Why has this been deleted ?

tests/tests_geomstats/test_binomial.py
```diff
@@ -40,3 +39,41 @@ def test_point_to_pdf(self, n_draws, point, n_samples):
             pmf.append(gs.array([binom.pmf(x, n_draws, point[i]) for x in samples]))
         expected = gs.squeeze(gs.stack(pmf, axis=0))
         self.assertAllClose(result, expected)
+
+
+class TestBinomialMetric(RiemannianMetricTestCase, metaclass=Parametrizer):
+    skip_test_log_after_exp = True
+    # skip_test_exp_after_log = True
```
alebrigant, line 46: Remove commented line. Same for lines below.

tests/tests_geomstats/test_binomial.py
```diff
@@ -40,3 +39,41 @@ def test_point_to_pdf(self, n_draws, point, n_samples):
...
+    def test_metric_matrix(self, n_draws, point, expected):
+        self.assertAllClose(self.Metric(n_draws).metric_matrix(point), expected)
+
+    def test_geodesic_symmetry(self, space_args):
+        space = self.Space(*space_args)
+        point_a, point_b = space.random_point(2)
+        path_ab = space.metric.geodesic(initial_point=point_a, end_point=point_b)
+        path_ba = space.metric.geodesic(initial_point=point_b, end_point=point_a)
+        t = gs.linspace(0.0, 1.0, 10)
+        self.assertAllClose(path_ab(t), path_ba(1 - t))
```
alebrigant, line 79: Could you add a test to verfiy that `dist` gives the same result as `norm` of `log`? I don't think this test is done automatically since the distance is usually implemented using norm of log, but here it is implemented using a direct formula.

tests/tests_geomstats/test_binomial.py
```diff
@@ -40,3 +39,41 @@ def test_point_to_pdf(self, n_draws, point, n_samples):
...
+
+
+class TestBinomialMetric(RiemannianMetricTestCase, metaclass=Parametrizer):
+    skip_test_log_after_exp = True
+    # skip_test_exp_after_log = True
+    skip_test_parallel_transport_ivp_is_isometry = True
+    skip_test_parallel_transport_bvp_is_isometry = True
+    # skip_test_geodesic_ivp_belongs = True
+    # skip_test_geodesic_bvp_belongs = True
+    skip_test_exp_geodesic_ivp = True
```
alebrigant, line 51: Do you know why this one isn't working ? Sorry if we have we discussed this already and I forgot !

tests/data/exponential_data.py
```diff
@@ -69,10 +72,6 @@ def sample_belongs_test_data(self):
 
     def point_to_pdf_test_data(self):
         random_data = [
-            dict(
-                point=self.Space().random_point(2),
-                n_samples=3,
-            ),
```
alebrigant, line 75: Why has this been deleted ?

tests/tests_geomstats/test_exponential.py
```diff
@@ -36,6 +35,42 @@ def test_point_to_pdf(self, point, n_samples):
...
+    def test_metric_matrix(self, point, expected):
+        self.assertAllClose(self.Metric().metric_matrix(point), expected)
+
+    def test_geodesic_symmetry(self):
+        space = self.Space()
+        point_a, point_b = space.random_point(2)
+        path_ab = space.metric.geodesic(initial_point=point_a, end_point=point_b)
+        path_ba = space.metric.geodesic(initial_point=point_b, end_point=point_a)
+        t = gs.linspace(0.0, 1.0, 10)
+        self.assertAllClose(path_ab(t), path_ba(1 - t))
```
alebrigant, line 76: Same as previous comment for BinomialMetric.

alebrigant (changes requested):
Thanks for the nice additions and improvements ! Just a few comments on the tests and some minor comments on the rest.

geomstats/information_geometry/exponential.py
```diff
@@ -168,14 +171,224 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+        if n_initial_points > n_initial_tangent_vecs:
+            raise ValueError(
+                "There cannot be more initial points than " "initial tangent vectors."
+            )
+        if n_initial_tangent_vecs > n_initial_points > 1:
+            raise ValueError(
+                "For several initial tangent vectors, "
+                "specify either one or the same number of "
+                "initial points."
+            )
```
lpereira95, line 236: same as for binomial.

geomstats/information_geometry/exponential.py
```diff
@@ -168,14 +171,224 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+            t : array-like, shape=[n_times,]
+                Times at which to compute points of the geodesics.
+
+            Returns
+            -------
+            geodesic : array-like, shape=[..., n_times, 1]
+                Values of the geodesic at times t.
+            """
+            t = gs.reshape(gs.array(t), (-1,))
+            _base, _t = gs.broadcast_arrays(base, gs.transpose(t))
```
lpereira95, line 254: variable names. no need for transposing `t`.

geomstats/information_geometry/exponential.py
```diff
@@ -168,14 +171,224 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+        if n_initial_points > n_end_points > 1:
+            raise ValueError(
+                "For several initial points, specify either"
+                "one or the same number of end points."
+            )
+        if n_end_points > n_initial_points > 1:
+            raise ValueError(
+                "For several end points, specify either "
+                "one or the same number of initial points."
+            )
```
lpereira95, line 290: same as for binomial.

geomstats/information_geometry/exponential.py
```diff
@@ -168,14 +171,224 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+            t : array-like, shape=[n_times,]
+                Times at which to compute points of the geodesics.
+
+            Returns
+            -------
+            geodesic : array-like, shape=[..., n_times, 1]
+                Values of the geodesic at times t.
+            """
+            t = gs.reshape(gs.array(t), (-1,))
+            _base, _t = gs.broadcast_arrays(base, gs.transpose(t))
```
lpereira95, line 308: variable names. no need to transpose `t`.

geomstats/information_geometry/normal.py
```diff
@@ -682,7 +682,7 @@ def inverse_tangent_diffeomorphism(self, image_tangent_vec, image_point):
         return self.inverse_diffeomorphism(image_tangent_vec)
 
     @staticmethod
-    def metric_matrix(base_point):
+    def metric_matrix(base_point=None):
```
lpereira95, line 685: remove `=None`.

geomstats/information_geometry/normal.py
```diff
@@ -695,6 +695,10 @@ def metric_matrix(base_point):
         mat : array-like, shape=[..., 2, 2]
             Metric matrix.
         """
+        if base_point is None:
+            raise ValueError(
+                "A base point must be given to compute the " "metric matrix"
+            )
```
lpereira95, line 701: remove.

lpereira95 (changes requested):
Thanks @tramy1258! Very well written code.

About the skipped tests:

1. I don't agree with skipping `test_log_after_exp`. What you probably need to guarantee is that whenever `arcsin(a)` is used, `a` belongs to the correct interval (@alebrigant may know better how to handle this).

2. I'll update the test to work properly (for some reason there's a squeeze in there), it shouldn't be skipped.

3. The test needs to be improved (not skipped). This is a test particularly written for this class. You should check it works as you desire.

geomstats/information_geometry/binomial.py
```diff
@@ -101,42 +101,42 @@ def sample(self, point, n_samples=1):
...
-            Number of points to sample with each pair of parameters in point.
+            Number of points to sample with each point.
             Optional, default: 1.
 
         Returns
         -------
         samples : array-like, shape=[..., n_samples]
             Sample from binomial distributions.
         """
+        geomstats.errors.check_belongs(point, self)
```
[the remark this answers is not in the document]
lpereira95: I think, for consistency, we shouldn't verify. That's being done in some other classes in information geometry though (we should remove it for now, I think. If later we decide to do checks, we do it everywhere, as you suggest).

geomstats/information_geometry/binomial.py
```diff
@@ -101,42 +101,42 @@ def sample(self, point, n_samples=1):
         point : array-like, shape=[..., dim]
             Point representing a binomial distribution.
         n_samples : int
-            Number of points to sample with each pair of parameters in point.
+            Number of points to sample with each point.
```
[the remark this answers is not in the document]
tramy1258: Yes, I agree!

alebrigant:
> Thanks @tramy1258! Very well written code.
> 
> About the skipped tests:
> 
>     1. I don't agree with skipping `test_log_after_exp`. What you probably need to guarantee is that whenever `arcsin(a)` is used, `a` belongs to the correct interval (@alebrigant may know better how to handle this).
> 
>     2. I'll update the test to work properly (for some reason there's a squeeze in there), it shouldn't be skipped.
> 
>     3. The test needs to be improved (not skipped). This is a test particularly written for this class. You should check it works as you desire.
> 
> 
> P.S. notice also that the changes are breaking `TestFisherRaoMetric` for some backends.

We decided to skip `test_log_after_exp` with @tramy1258 because exp is not injective: just like on a sphere, different tangent vectors lead to the same results (because the geodesic makes several loops around the sphere). There is no way to retrieve any general vector by doing log after exp in this case. The only possible check (in all generality) would be to redo exp after that, i.e. "exp after log after exp", but that boils down to "exp after log" which is already tested. In other words, exp and log are inverse inside a ball of a certain radius around zero in the tangent space, but not beyond.

Another possibility is to overwrite `test_log_after_exp` and test with vectors that have a norm smaller than this radius.

lpereira95:
@tramy1258, I've fixed the problem with `test_exp_geodesic_ivp`, which was a bug in the test side.

tests/data/binomial_data.py
tramy1258: This generates a test for `point_to_pdf(k)` where `k` is a 2D array, where as `point_to_pdf` should only work when `k` is a vector (as said in the doc string). 
I could modify the code to work for this case if it is neccessary, then for example, when given `point` of shape `(2,dim)` and `k` of shape `(3,4)`, we will get something of shape `(2,3,4)`, but I am not sure if that is what we want?

tests/tests_geomstats/test_binomial.py
tramy1258: Yes I will look into it.

lpereira95:
I see @alebrigant.

Just to see if I fully get the implications:  if I do `log_after_exp` the test is failing because i) we do not get back the same tangent vector - which is OK or ii) `arcsin(a)` raises an error due to arcsin domain issues?

If i), will the vectors have the same direction? (in the case of the sphere I think that's the case) If not, we should simply skip the test. If yes, we can adapt the test for this particular case.

If ii), there's probably something to fix, as this should not happen.

geomstats/information_geometry/binomial.py
```diff
@@ -181,18 +175,240 @@ def squared_dist(self, point_a, point_b, **kwargs):
...
+        if n_initial_points > n_end_points > 1:
+            raise ValueError(
+                "For several initial points, specify either"
+                "one or the same number of end points."
+            )
+        if n_end_points > n_initial_points > 1:
+            raise ValueError(
+                "For several end points, specify either "
+                "one or the same number of initial points."
+            )
```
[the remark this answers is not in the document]
tramy1258: Thanks Luis, that is very neat!

tramy1258:
> Thanks @tramy1258! Very well written code.
> 
> About the skipped tests:
> 
> 1. I don't agree with skipping `test_log_after_exp`. What you probably need to guarantee is that whenever `arcsin(a)` is used, `a` belongs to the correct interval (@alebrigant may know better how to handle this).
> 2. I'll update the test to work properly (for some reason there's a squeeze in there), it shouldn't be skipped.
> 3. The test needs to be improved (not skipped). This is a test particularly written for this class. You should check it works as you desire.
> 
> P.S. notice also that the changes are breaking `TestFisherRaoMetric` for some backends.

1. The test fails because we get another tangent vector in the same direction and normally, the tangent vector after log is the smallest vector in that direction. I agree with @alebrigant that we could add a test where the initial tangent vector is the smallest one (in our case, verify whenever we take `sin(a)`, `a` is in `[-pi/2, pi/2]`.
2. Thank you for fixing the test.
3. I only skipped a case in the test and I think passing all other cases already proves that the function works properly. The skipped case actually does not conform to what we expected from the function, hence why I decided to delete it.

alebrigant:
> I see @alebrigant.
> 
> Just to see if I fully get the implications: if I do `log_after_exp` the test is failing because i) we do not get back the same tangent vector - which is OK or ii) `arcsin(a)` raises an error due to arcsin domain issues?
> 
> If i), will the vectors have the same direction? (in the case of the sphere I think that's the case) If not, we should simply skip the test. If yes, we can adapt the test for this particular case.
> 
> If ii), there's probably something to fix, as this should not happen.

It's i). The vectors all have the same direction in dimension 1 so this is not something interesting to check here, unlike the for the sphere. But we can do as @tramy1258 suggests, tests for tangent vectors chosen so that the angles are in [-pi/2, pi/2].

lpereira95:
Thanks for the clarifications @tramy1258 and @alebrigant. I agree with your proposed solution (I momentarily forgot we were in 1d, hence the question about the direction).

@tramy1258, you probably only need to override the data creation for `test_log_after_exp`, no need to create a new test.

geomstats/information_geometry/exponential.py
tramy1258: By "variables names" you mean I should change `_base` and `_t` right?

tramy1258:
> @tramy1258, I've fixed the problem with `test_exp_geodesic_ivp`, which was a bug in the test side.

Was the bug the `gs.squeeze` in `data_generation::_ConnectionTestData::exp_geodesic_ivp_test_data` or was it something else? Because I still have a shape mismatch error for the test

lpereira95:
@tramy1258, `test_exp_geodesic_ivp` looks to be passing for all the backends, no? There's a bunch of errors (especially for `TestFisherRaoMetric`), but not this one... (have you pulled my changes?)

tests/tests_geomstats/test_binomial.py
tramy1258: @alebrigant There is an automatic `test_dist_is_norm_of_log` where we basically compute and compare `sqrt(squared_dist(point_a, point_b))` and `norm(log(point_b, point_a))`. 

In `RiemannianMetric`, `squared_dist` is computed exactly like you mentionned using `squared_norm` of `log`, however, we override this in `BinomialMetric` using a direct solution so the test is actually done as we wish (using 2 different methods for `dist` and `norm(log)`).

tramy1258:
@lpereira95 Okay I did not pull the right way, it works now!

tramy1258:
I have modified the codes as suggested, there are however still some errors after the tests.
1. Errors in `test_geodesic_ivp_belongs` and `test_geodesic_bvp_belongs` for `tensorflow` at single point cases, `belongs(gs.array([1]))` returns `True` while we expect `[True]`. All other backends consider them equal but not `tensorflow`. I could modify the output shape of `belongs` for Binomial and Exponential for now but this is actually the case for all other distributions.
2. Errors with `TestFisherRaoMetric` for Exponential in `autograd`, which gave this error `VJP of broadcast_arrays wrt argnums (0,) not defined`, which I would like some help to resolve.
3. Errors concerning curves and parallel transport for `autograd`. Should I skip these tests?
4. Other errors in other distributions that are not due to my changes (?).

tests/data/binomial_data.py
alebrigant: Ok I see. I agree this is not what we want. But we still need to check whether `point_to_pmf` is vectorized, i.e. works for several points. I think the test should be rewritten: the shape of the generated samples should not depend on the shape of the input point, it should always be a vector, so that when point is of the shape `(n,dim)` and the samples of shape `(k,)` then the output is of shape `(n,k)`.

tests/data/exponential_data.py
alebrigant: Same as for Binomial distributions.

tests/tests_geomstats/test_binomial.py
alebrigant: Perfect !

alebrigant (changes requested):
Looks good to me ! Just a final thing to change in the test of `point_to_pmf` and I'll approve merging !

tests/data/binomial_data.py
tramy1258: So would it be okay if I generated the samples with something like `gs.random.rand(k)`? Would it make sense mathematically, because I thought the point of using `sample` in `test_point_to_pdf` was to have samples related to the input points/distributions (and so naturally the samples shape depends on the shape of input)?

lpereira95 (changes requested):
Almost there @tramy1258! You're doing a great job, the code is very neat!

Before merging I think we need:

1. update `point_to_pmf` and `point_to_pdf` to be automatically differentiable (I think it is important we are able to use this distributions within `FisherRaoMetric`). This means `FisherRaoMetric` tests should pass. The problem seems to be `gs.broadcast_arrays`: could you find an alternative solution (probably relying in `gs.einsum` that does not use `gs.broadcast_arrays` in that function? Notice that the automatic differentiation works for `pytorch`, but outputs `nan`

2. when we repeat the tests for `exponential` several times, `exp_belongs` fails about 2 out of 10 times. Can you please try to find out why? We are trying to only add new tests that never (or very very rarely) fail (maybe you will need to input the tolerance for belongs)

3. could you add tests where you compare the scipy functions you've replaced with the scipy functions themselves?

geomstats/information_geometry/binomial.py
```diff
@@ -151,19 +149,21 @@ def pmf(k):
...
             """
-
-            def _pmf(param):
-                return binom.pmf(k, self.n_draws, param)
-
-            if n_batch:
-                return gs.array([_pmf(param) for param in point])
-
-            return gs.array(_pmf(point))
+            k = gs.reshape(gs.array(k), (-1,))
```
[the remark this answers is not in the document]
tramy1258: I see that sometimes something like `pmf(0.2)` is called (I think it is the case for `test_fisher_rao_metric`) which will not work with the broadcasting.

alebrigant (approved):
LGTM ! @luisfpereira do you approve the changes ?

geomstats/information_geometry/binomial.py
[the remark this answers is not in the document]
luisfpereira: I see. We should probably think if we want to allow it, since the docstrings say "array-like". Let's keep it like this for now.

geomstats/information_geometry/binomial.py
```diff
@@ -151,19 +149,24 @@ def pmf(k):
...
+            k_shape = gs.ones_like(k)
+            point_shape = gs.ones_like(point)
+            point_aux = gs.einsum("...i,j->...j", point, k_shape)
+            k_aux = gs.einsum("...i,j->...j", point_shape, k)
```
luisfpereira, line 156: why don't we use `gs.power` here? (I think we spoke about it at a very early stage, but I can't remember what was the drawback... maybe it was `tf`, which is not a problem anymore)

`gs.power(point, k)` seem to output the same as this for lines.

geomstats/information_geometry/exponential.py
```diff
@@ -145,8 +145,11 @@ def pdf(x):
             -------
             pdf_at_x : array-like, shape=[..., n_points]
             """
-            x = gs.reshape(gs.array(x), (-1,))
-            point_aux, x_aux = gs.broadcast_arrays(point, x)
+            x = gs.reshape(gs.array(x),(-1,))
+            x_shape = gs.ones_like(x)
+            point_shape = gs.ones_like(point)
+            point_aux = gs.einsum('...i,j->...j',point,x_shape)
+            x_aux = gs.einsum('...i,j->...j',point_shape,x)
```
luisfpereira: The same comment above apply.

luisfpereira (changes requested):
@tramy1258, can you please address my last two comments? If `gs.power` works, then we just update it and merge, otherwise it is ready to go.

Please merge the last version of master, so we can ensure nothing has been broken (e.g. there was a large PR removing tensorflow meanwhile).

geomstats/information_geometry/binomial.py
tramy1258: I guess by using `gs.power` you mean not to broadcast but to use inputs as they are? Indeed, I just tested the version without broadcasting and it works well, I cannot remember either why I chose to broadcast.

luisfpereira (approved):
Looks ready to me @tramy1258!

The test failures are unrelated. I'm fixing them in another PR.

tramy1258:
> Looks ready to me @tramy1258!
> 
> The test failures are unrelated. I'm fixing them in another PR.

Glad to hear that! Thank you for your suggestions.

PR #1761, opened 2022-12-13
Reserve `scale` parameter to ScalarProductMetric

### LINKED ISSUE #1757: Check `ScalarProductMetric` and remove `scale` from manifolds
Following #1740: 

> check all existing classes that use scale to make sure their results match the results from ScalarProductMetric, as I think there are a couple of issues.

After ensuring `ScalarProductMetric` is well implemented, we can remove `scale` from all the manifolds, as we now have a better way to handle scales.

`ProductManifold` should be handled in a different PR. 

(With @johnharveymath.)

### LINKED ISSUE #1758: Rename variable `scaling_factor` to `scale` in `ScalarProductMetric`
Following #1740. #1757 needs to be done first.

(With @johnharveymath.)

johnharveymath:
Closes #1757 and closes #1758 by removing the `scale` parameter from various manifolds.

These are the `Hyperbolic`, `HPDMatrices` and `PoincarePolydisk` manifolds, along with all their related classes.

@YannCabanes The tests are failing on `PoincarePolydisk` and I don't understand the objects well enough to remedy the issue. Maybe you could have a look?

### DIFF
--- a/geomstats/geometry/_hyperbolic.py
+++ b/geomstats/geometry/_hyperbolic.py
@@ -21,7 +21,7 @@ class _Hyperbolic:
     """
 
     def __init__(self, **kwargs):
-        if "scale" in kwargs.keys():
+        if "scale" in kwargs:
             raise TypeError(
                 "Argument scale is no longer in use: instantiate the "
                 "manifold without this parameter and then use "
@@ -459,7 +459,7 @@ class HyperbolicMetric(RiemannianMetric):
     """
 
     def __init__(self, dim, default_coords_type="extrinsic", **kwargs):
-        if "scale" in kwargs.keys():
+        if "scale" in kwargs:
             raise TypeError(
                 "Argument scale is no longer in use: instantiate scaled "
                 "metrics as `scale * RiemannianMetric`. Note that the "
--- a/geomstats/geometry/_hyperbolic.py
+++ b/geomstats/geometry/_hyperbolic.py
@@ -21,6 +21,10 @@ class _Hyperbolic:
     """
 
     def __init__(self, **kwargs):
+        if 'scale' in kwargs.keys():
+            raise TypeError("Argument scale is no longer in use: instantiate the "
+                            "manifold without this parameter and then use "
+                            "`scale * metric` to rescale the standard metric.")
         super().__init__(**kwargs)
 
     @staticmethod
@@ -452,7 +456,11 @@ class HyperbolicMetric(RiemannianMetric):
         Default coordinates to represent points in hyperbolic space.
     """
 
-    def __init__(self, dim, default_coords_type="extrinsic"):
+    def __init__(self, dim, default_coords_type="extrinsic", **kwargs):
+        if 'scale' in kwargs.keys():
+            raise TypeError("Argument scale is no longer in use: instantiate scaled "
+                            "metrics as `scale * RiemannianMetric`. Note that the "
+                            "metric is scaled, not the distance.")
         super().__init__(
             dim=dim,
             signature=(dim, 0),
--- a/geomstats/geometry/_hyperbolic.py
+++ b/geomstats/geometry/_hyperbolic.py
@@ -441,6 +441,9 @@ class _Hyperbolic:
 class HyperbolicMetric(RiemannianMetric):
     """Class that defines operations using a hyperbolic metric.
 
+    This class does not contain any methods and is only defined to act as a parent
+    class for `HyperboloidMetric`, `PoincareBallMetric` and `PoincareHalfSpaceMetric`.
+
     Parameters
     ----------
     dim : int
@@ -455,90 +458,4 @@ class HyperbolicMetric(RiemannianMetric):
             signature=(dim, 0),
             shape=(dim + 1,),
             default_coords_type=default_coords_type,
-        )
-
-    def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
-        """Compute the inner-product of two tangent vectors at a base point.
-
-        Parameters
-        ----------
-        tangent_vec_a : array-like, shape=[..., dim + 1]
-            First tangent vector at base point.
-        tangent_vec_b : array-like, shape=[..., dim + 1]
-            Second tangent vector at base point.
-        base_point : array-like, shape=[..., dim + 1]
-            Point in hyperbolic space.
-            Optional, default: None.
-
-        Returns
-        -------
-        inner_prod : array-like, shape=[..., 1]
-            Inner-product of the two tangent vectors.
-        """
-        inner_prod = self._inner_product(tangent_vec_a, tangent_vec_b, base_point)
-        return inner_prod
-
-    def squared_norm(self, vector, base_point=None):
-        """Compute the squared norm of a vector.
-
-        Squared norm of a vector associated with the inner-product
-        at the tangent space at a base point.
-
-        Parameters
-        ----------
-        vector : array-like, shape=[..., dim + 1]
-            Vector on the tangent space of the hyperbolic space at base point.
-        base_point : array-like, shape=[..., dim + 1]
-            Point in hyperbolic space in extrinsic coordinates.
-            Optional, default: None.
-
-        Returns
-        -------
-        sq_norm : array-like, shape=[..., 1]
-            Squared norm of the vector.
-        """
-        sq_norm = self._squared_norm(vector)
-        return sq_norm
-
-    def _squared_norm(self, vector, base_point=None):
-        """Squared norm.
-
-        Squared norm of a vector associated with the inner-product
-        at the tangent space at a base point.
-
-        Parameters
-        ----------
-        vector : array-like, shape=[..., dim + 1]
-            Vector on the tangent space of the hyperbolic space at base point.
-        base_point : array-like, shape=[..., dim + 1]
-            Point in hyperbolic space in extrinsic coordinates.
-            Optional, default: None.
-
-        Returns
-        -------
-        sq_norm : array-like, shape=[..., 1]
-            Squared norm of the vector.
-        """
-        return super().squared_norm(vector, base_point=base_point)
-
-    def _inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
-        """Compute the inner-product of two tangent vectors.
-
-        Parameters
-        ----------
-        tangent_vec_a : array-like, shape=[..., dim + 1]
-            First tangent vector at base point.
-        tangent_vec_b : array-like, shape=[..., dim + 1]
-            Second tangent vector at base point.
-        base_point : array-like, shape=[..., dim + 1]
-            Point in hyperbolic space.
-            Optional, default: None.
-
-        Returns
-        -------
-        inner_prod : array-like, shape=[..., 1]
-            Inner-product of the two tangent vectors.
-        """
-        return super().inner_product(
-            tangent_vec_a, tangent_vec_b, base_point=base_point
         )
--- a/geomstats/geometry/_hyperbolic.py
+++ b/geomstats/geometry/_hyperbolic.py
@@ -20,7 +20,7 @@ class _Hyperbolic:
     chosen representation: Hyperboloid, Poincare Ball or Poincare Half-space.
     """
 
-    def __init__(self, scaling_factor=1.0, **kwargs):
+    def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     @staticmethod
--- a/geomstats/geometry/_hyperbolic.py
+++ b/geomstats/geometry/_hyperbolic.py
@@ -18,18 +18,10 @@ class _Hyperbolic:
     This class cannot be instantiated in itself, but a public class
     `Hyperbolic` is available, that returns the class that corresponds to the
     chosen representation: Hyperboloid, Poincare Ball or Poincare Half-space.
-
-    Parameters
-    ----------
-    scale : int
-        Scale of the hyperbolic space, defined as the set of points
-        in Minkowski space whose squared norm is equal to -scale.
-        Optional, default: 1.
     """
 
-    def __init__(self, scale=1, **kwargs):
+    def __init__(self, scaling_factor=1.0, **kwargs):
         super().__init__(**kwargs)
-        self.scale = scale
 
     @staticmethod
     def _extrinsic_to_extrinsic_coordinates(point):
@@ -455,20 +447,15 @@ class HyperbolicMetric(RiemannianMetric):
         Dimension of the hyperbolic space.
     default_coords_type : str, {'extrinsic', 'intrinsic', etc}, optional
         Default coordinates to represent points in hyperbolic space.
-    scale : int, optional
-        Scale of the hyperbolic space, defined as the set of points
-        in Minkowski space whose squared norm is equal to -scale.
     """
 
-    def __init__(self, dim, scale=1, default_coords_type="extrinsic"):
+    def __init__(self, dim, default_coords_type="extrinsic"):
         super().__init__(
             dim=dim,
             signature=(dim, 0),
             shape=(dim + 1,),
             default_coords_type=default_coords_type,
         )
-
-        self.scale = scale
 
     def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
         """Compute the inner-product of two tangent vectors at a base point.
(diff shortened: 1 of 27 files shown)


geomstats/geometry/hyperboloid.py
```diff
@@ -392,7 +383,7 @@ def log(self, point, base_point):
             Tangent vector at the base point equal to the Riemannian logarithm
             of point at the base point.
         """
-        angle = self.dist(base_point, point) / self.scale
+        angle = self.dist(base_point, point)
```
lpereira95, line 386: Removing here does not affect the result? That is, are you getting the expected result with `ScalarProductMetric`?

geomstats/geometry/poincare_ball.py
```diff
@@ -203,7 +193,7 @@ def mobius_add(self, point_a, point_b, project_first=True):
         mobius_add : array-like, shape=[..., dim]
             Result of the Mobius addition.
         """
-        ball_manifold = PoincareBall(self.dim, scale=self.scale)
+        ball_manifold = PoincareBall(self.dim)
```
lpereira95, line 196: unrelated, but maybe we can move this to init and call it `self._space`, so later it will simplify the coming up refactoring of `Connection`.

geomstats/geometry/poincare_ball.py
```diff
@@ -273,7 +262,7 @@ def retraction(self, tangent_vec, base_point):
         .. [1] Nickel et.al, "Poincaré Embedding for
             Learning Hierarchical Representation", 2017.
         """
-        ball_manifold = PoincareBall(self.dim, scale=self.scale)
+        ball_manifold = PoincareBall(self.dim)
```
lpereira95, line 265: as above.

geomstats/geometry/_hyperbolic.py
```diff
@@ -455,21 +447,16 @@ class HyperbolicMetric(RiemannianMetric):
         Dimension of the hyperbolic space.
     default_coords_type : str, {'extrinsic', 'intrinsic', etc}, optional
         Default coordinates to represent points in hyperbolic space.
-    scale : int, optional
-        Scale of the hyperbolic space, defined as the set of points
-        in Minkowski space whose squared norm is equal to -scale.
     """
 
-    def __init__(self, dim, scale=1, default_coords_type="extrinsic"):
+    def __init__(self, dim, default_coords_type="extrinsic"):
```
lpereira95, line 452: I think now we can simplify this class my simply removing everything (maybe we just keep it due to mathematical reasoning, i.e. the children are hyperbolic metrics).

In children (at least in `Hyperboloid`) we need then to update `_inner_product` and `_squared_norm` to be simply `inner_product` and `squared_norm`.

geomstats/geometry/poincare_polydisk.py
```diff
@@ -107,6 +107,6 @@ def __init__(self, n_disks, default_coords_type="extrinsic"):
         list_metrics = []
         for i_disk in range(n_disks):
             scale_i = (n_disks - i_disk) ** 0.5
-            metric_i = HyperboloidMetric(2, default_coords_type, scale_i)
+            metric_i = scale_i * HyperboloidMetric(2, default_coords_type)
```
lpereira95, line 110: Great!

tests/data/siegel_data.py
```diff
@@ -169,83 +160,38 @@ def sectional_curvature_test_data(self):
...
                 expected=-1,
             ),
-            dict(
-                n=2,
-                scale=2.0,
-                tangent_vec_a=[[0.25 + 0j, 0j], [0j, 0j]],
-                tangent_vec_b=[[0j, 0.25j], [0j, 0j]],
-                base_point=[[0j, 0j], [0j, 0j]],
-                expected=-0.25,
-            ),
```
lpereira95, line 209: could we keep this by removing scale and changing expected accordingly? (just because I don't see this case without scale and it is nice to keep smoke data.)

tests/data/complex_poincare_disk_data.py
```diff
@@ -91,45 +91,22 @@ class ComplexPoincareDiskMetricTestData(_ComplexRiemannianMetricTestData):
...
                 base_point=[[0.0]],
                 expected=[25],
             ),
-            dict(
-                scale=2,
-                tangent_vec_a=[[1j], [1j], [1j]],
-                tangent_vec_b=[[1j], [2j], [3j]],
-                base_point=[[0j], [0j], [0j]],
-                expected=[4, 8, 12],
-            ),
```
lpereira95, line 113: keep data without scale and updating expected?

lpereira95 (changes requested):
Thanks @johnharveymath! Please take a look to my comments (most of them are little details).

I'll try to understand the tests that are failing. @YannCabanes, your feedback and help would be of great value here, as most of `scale` related things were developed by you.

johnharveymath:
Seems like for most classes it was the distance that was scaled by `self.scale`, except for `PoincareHalfSpace` where it was the inner product that got scaled. So I think it's great that we're taking this out as there was clearly some inconsistency going on.

geomstats/geometry/poincare_polydisk.py
```diff
@@ -102,11 +97,10 @@ class PoincarePolydiskMetric(ProductRiemannianMetric):
...
+    def __init__(self, n_disks):
         self.n_disks = n_disks
-        list_metrics = []
-        for i_disk in range(n_disks):
-            scale_i = (n_disks - i_disk) ** 0.5
-            metric_i = HyperboloidMetric(2, default_coords_type, scale_i)
-            list_metrics.append(metric_i)
+        base_metric = HyperboloidMetric(2)
+        list_metrics = [
+            (n_disks - i_disk) ** 0.5 * base_metric for i_disk in range(n_disks)
```
johnharveymath, line 104: Can we double check this @YannCabanes - we need to use the factor which scales the metric itself, rather than the distance. I suspect it should not have the `** 0.5`.

YannCabanes:
Hello @johnharveymath and @lpereira95,
With my notations, the scale parameter means multiplying the distance by the scale parameter, which means multiplying:

- `dist` --> (multiply by) `scale` 
- `norm` --> `scale`

- `metric_matrix` --> `scale ** 2`
- `inner_product` --> `scale ** 2`
- `squared_norm` --> `scale ** 2`
- `squared_dist` --> `scale ** 2`

However, I see that in the current master branch, the file `product_riemannian_metric.py` is using a different convention for `metric_matrix` and `inner_product` as they are only multiplied by `scale` and not `scale ** 2`.

I am currently looking at the how the scale influences the curvature, sectional_curvature, riemann_tensor...

geomstats/geometry/poincare_polydisk.py
YannCabanes: Hello @johnharveymath and @lpereira95,
Yes, we have to decide which convention we are willing to use.
I was using the convention multiplying the distance by the scale parameter, which means multiplying the `metric_matrix` by `scale ** 2`.
It seems that you would prefer to multiply the `metric_matrix` by `scale` and therefore the distance by `scale ** 0.5`.
Both conventions are fine for me, so I will let you both decide.

YannCabanes: If you choose to multiply the `metric_matrix` by `scale`, then you can remove `** 0.5`.

lpereira95:
Thanks for the feedback @YannCabanes.

I think the implementation of `ScalarProductMatrix` is consistent. Which probably means we just need to remove **.5 as you say.

Nevertheless, are you sure about the correctness of `TestPoincarePolydiskMetric.test_product`? I can't understand why the distances should be the same for these cases. Could you please explain me?

YannCabanes:
Hello @johnharveymath and @lpereira95,
I just did a complete review of the influence of the scale parameter. 
The scale parameter (related to the metric) has no influence on any function defined in the file `connection.py` (`christoffels`, `exp`, `log`, `riemann_tensor`, `curvature`, `ricci_tensor`, `geodesic`...).
The scale parameter influences the methods defined in the file `riemannian_metric.py`.
Using the convention in which the distance is multiplied by the scale parameter and the metric_matrix by scale ** 2, I now list the influence of the scale parameter on each method of the class `RiemannianMetric`:
- `metric_matrix` --> `* scale ** 2`
- `cometric_matrix` --> `/ scale ** 2`
- `inner_product_derivative_matrix` --> `* scale ** 2`
- `christoffels` --> no influence
- `inner_product` --> `* scale ** 2`
- `inner_coproduct` --> `/ scale ** 2`
- `hamiltonian` --> `/ scale ** 2`
- `squared_norm` --> `* scale ** 2`
- `norm` --> `* scale`
- `normalize` --> `/ scale`
- `random_unit_tangent_vec` --> `/ scale`
- `squared_dist` --> `* scale ** 2`
- `dist` --> `* scale`
- `dist_broadcast` --> `* scale`
- `dist_pairwise` --> `* scale`
- `diameter` --> `* scale`
- `closest_neighbor_index` --> no influence
- `normal_basis` --> `/ scale`
- `covariant_riemann_tensor` --> `* scale ** 2`
- `sectional_curvature` --> `/ scale ** 2`
- `scalar_curvature` --> `/ scale ** 2`

YannCabanes:
Hello @lpereira95,
About the `TestPoincarePolydiskMetric.test_product_distance`, you are right, `distance_two_disks` is different from `distance_single_disk`. This is a change introduced in the current PR.
Normally, when `point_a` and `point_b` are duplicated in each disks, we should have: 
- `distance_single_disk` = distance HyperboloidMetric(2)
- `distance_two_disks = 3 ** 0.5 * distance_single_disk`
- For n disks, we should have: `distance_n_disks = ((n * (n+1) / 2) ** 0.5 * distance_single_disk`

johnharveymath:
> Hello @lpereira95, About the `TestPoincarePolydiskMetric.test_product_distance`, you are right, `distance_two_disks` is different from `distance_single_disk`. This is a change introduced in the current PR. Normally, when `point_a` and `point_b` are duplicated in each disks, we should have:
> 
>     * `distance_single_disk` = distance HyperboloidMetric(2)
> 
>     * `distance_two_disks = 3 ** 0.5 * distance_single_disk`
> 
>     * For n disks, we should have: `distance_n_disks = ((n * (n+1) / 2) ** 0.5 * distance_single_disk`

Thanks for the explanation - this was caused by my ignorance of the geometry. I though the 3 had something to do with the scale parameter. This test passes now.

johnharveymath:
> 



> I just did a complete review of the influence of the scale parameter.

I'm glad you did, @YannCabanes! Your review reveals that some parameters were being divided by `scale ** 2` instead of `gs.sqrt(scale)`. I have opened PR #1763 to correct these.

As I mentioned earlier, it's definitely best to just have one implementation that is in `ScalarProductMetric`. This is because we did actually have some inconsistencies about how `scale` was treated in different classes. @alebrigant and I chose this particular implementation because we are expressing it as multiplying the metric (ie, the matrix, or the (0, 2)-tensor if you prefer) by a scalar. Your implementation is far more natural (this is how we think about scaling a space) but I think that, given that we can now write `h = 2 * g`, it is better to go with the more mathematically formalistic approach.

geomstats/geometry/hyperboloid.py
johnharveymath: Yes - it works. I was able to put the removed tests back in but use the `ScalarProductMetric` functionality. So I'll keep doing that with as many removed tests as possible.

geomstats/geometry/poincare_ball.py
johnharveymath: done

geomstats/geometry/poincare_ball.py
johnharveymath: done

geomstats/geometry/poincare_ball.py
johnharveymath: Oops - not done! It's not so straightforward as there is a bit of circularity since `PoincareBall.__init__` calls `PoincareBallMetric.__init__`. Shall I just leave it alone?

tests/data/siegel_data.py
johnharveymath: Retained scale by using the parameter in the test file as `scale * self.Metric(n)`.

tests/data/complex_poincare_disk_data.py
johnharveymath: As above.

geomstats/geometry/_hyperbolic.py
johnharveymath: Yes, I agree. Done.

geomstats/geometry/poincare_ball.py
lpereira95: ah! yes, for now we keep. This kind of circularity will finally disappear with the design changes discussed in the last hackathon.

tests/data/siegel_data.py
lpereira95: @johnharveymath, I would prefer here to fully remove `scale` from `Siegel` tests. Other tests care about the influence of the scale (e.g. tests on any child of `ProductManifold`). The other manifold tests should "ignore" the existence of `ScalarProductMetric`.

tests/tests_geomstats/test_poincare_polydisk.py
```diff
@@ -33,21 +33,17 @@ def test_signature(self, n_disks, expected):
         self.assertAllClose(metric.signature, expected)
 
     @tests.conftest.np_autograd_and_torch_only
-    def test_product_distance_extrinsic_representation(
-        self, n_disks, point_a_extrinsic, point_b_extrinsic
-    ):
+    def test_product_distance(self, point_a_extrinsic, point_b_extrinsic):
```
lpereira95, line 36: in a previous commit I removed `n_disks` from here. but following @YannCabanes comment, we can now edit this test to receive `n_disks` and then we verify `distance_n_disks = ((n * (n+1) / 2) ** 0.5 * distance_single_disk`.

tests/data/siegel_data.py
johnharveymath: Is this particular to `Siegel`? There is `ComplexPoincareDisk` as well, where there are some `scale` parameters in the smoke data.

johnharveymath: In the latest commits I've done it for both classes.

tests/tests_geomstats/test_poincare_polydisk.py
johnharveymath: OK - this is done. In theory it is now possible to use `n_disks` polydisks, each factor having `m_disks`, but the smoke data as it is only has points for `m_disks=1` so I've left it there. This now runs with `n_disks=2` and `n_disks=3`.

johnharveymath:
OK - I think we're ready to go.

In summary, this PR:

- removes the `scale` parameter from `_hyperbolic` and its children, `Hyperboloid`, `PoincareBall`, `PoincareHalfSpace` as well as from `ComplexPoincareDisk`, `HPDMatrices`, `PositiveReals` and `Siegel`.
- adds an informative error message for users who attempt to initialize these manifolds or their metrics while using a `scale` parameter
- makes `ScalarProductMetric` a child of `RiemannianMetric` so that it is possible to write `manifold.metric = 2.0 * manifold.metric`

lpereira95:
Great work @johnharveymath and @YannCabanes! I'm merging now.


@johnharveymath, just two last comments for future reference:

1. I think we should remove the error messages for `scale` soon. I thought about removing it before merging, but I see we should not do it as our inits accept kwargs (which mean code with `scale` will not fail, but will provide unexpected results). Very soon I want to remove `**kwargs` from all the inits, so we can remove this kind of errors (python error messages will be good enough then).

2. I think `ScalarProductMetric` shouldn't inherit from `RiemannianMetric`. The reason is that by doing that we lose control over the methods we are providing: e.g. in __init__ we do not wrap private methods of `underlying_metric`, but we inherit them from `RiemannianMetric`, without providing proper scaling (e.g. imagine in the future we implement `_inner_product` for some reason, it will become available for `ScalarProductMetric`). What do you think? We can simply implement the multiplication in `ScalarProductMetric` also.

johnharveymath:
That's makes sense to me. The only issue is that you cannot actually assign
the `ScalarProductMetric` object to be an attribute of a `Manifold`, ie,
you cannot do `manifold.metric = 2.0 * manifold.metric`. I have changed
`manifold.py` in one additional commit to permit this. I know there has
been discussion about making `manifold.metric` a read-only object in
future, but it seems to me that this might be one time you want to be able
edit it.

lpereira95:
Thanks @johnharveymath!

I added multiplication to `ScalarProductMetric` (because it is not being inherited). It is code duplication, but I think for a good reason here.

Merging now...

johnharveymath:
Well spotted. Thanks for all your advice and help on this one.

PR #1762, opened 2022-12-13
Remove code duplication in ProductManifold and ProductManifoldMetric

lpereira95:
There's a lot of (unnecessary) code duplication in `ProductManifold` and `ProductManifoldMetric`.

This PR uses a mixins approach to remove code duplication. A more functional-based approach could have been used, but it would probably require functions with very long signatures.

`_IterateOverFactorsMixins` is implemented in the file `product_riemannian_metric.py` only to avoid circularity imports. Probably these two files  should be merged soon.

@johnharveymath, do you think this abstraction makes the code easier to manipulate or is it better to maintain the current code duplication (or find an alternative solution)?

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -3,24 +3,19 @@
 Lead author: Nicolas Guigui, John Harvey.
 """
 
-import math
 
 import geomstats.backend as gs
 import geomstats.errors
-from geomstats.errors import ShapeError, check_point_shape
 from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.product_riemannian_metric import (
     NFoldMetric,
     ProductRiemannianMetric,
+    _all_equal,
+    _IterateOverFactorsMixins,
 )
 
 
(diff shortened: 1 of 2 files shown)


johnharveymath:
I'm not sure that I'll get a good look at this until after Christmas!

lpereira95:
@johnharveymath I haven't done much here.

In summary, I've:

1. collected all the methods shared by `ProductManifold` and `ProductMetric` which are responsible for iteration over factors in a class called `_IterateOverFactorsMixins`; I haven't changed the methods besides the addition of a flag `_pool_outputs` because that's required only for `ProductManifold`.

2. make `ProductManifold` and `ProductMetric` inherit from `_IterateOverFactorsMixins`, so all the required methods are inherited.


The goal is simply to remove code duplication. If you find this solution appealing and easy to modify in the future, we can merge. Otherwise I close this and we think about a better strategy to avoid this code duplication later.

johnharveymath:
This makes sense to me - I certainly wasn't happy with the way it was! I
think you can go ahead.

PR #1771, opened 2022-12-16
Fixes and improvements coming from test refactoring

### LINKED ISSUE #1694: Bug: `_SpecialEuclideanMatrices` overrides `MatrixLieGroup` n
`_SpecialEuclideanMatrices`, which is a `MatrixLieGroup`, overrides `n`.

Therefore, all the methods defined in `MatrixLieGroup` that rely on `n` will return wrong results.

The solution is to rename one of the variables.

lpereira95:
This PR brings some fixes and improvements developed while refactoring tests in #1767. Since #1767 is a long PR and will take a while to get finished, some fixes and improvements done there are merged separately here.

Changes include:

1. closes #1694 by using `n_la` and `n_lg` in `MatrixLieAlgebra` and `MatrixLieGroup`, respectively.
2. for consistency, `get_identity` is removed from several files and only `identity` is kept (as a property)
3. vectorization is fixed in `FullRankCorrelationMatrices.tangent_submersion`
4. `left_or_right="left"` is replaced by `left=True` in several places (bools are more robust than strings)
5. `from_vector` is added to `LowerTriangularMatrices` for consistency with other matrix vector spaces
6. special_euclidean `homogeneous_representation` is simplified and vectorization behavior fixed

@ninamiolane, can you please quickly check if you agree with the changes?

### DIFF
--- a/geomstats/geometry/base.py
+++ b/geomstats/geometry/base.py
@@ -145,14 +145,6 @@ class VectorSpace(Manifold, abc.ABC):
             self._basis = self._create_basis()
         return self._basis
 
-    @basis.setter
-    def basis(self, basis):
-        if len(basis) < self.dim:
-            raise ValueError(
-                "The basis should have length equal to the " "dimension of the space."
-            )
-        self._basis = basis
-
     @abc.abstractmethod
     def _create_basis(self):
         """Create a canonical basis."""
--- a/geomstats/geometry/euclidean.py
+++ b/geomstats/geometry/euclidean.py
@@ -23,17 +23,15 @@ class Euclidean(VectorSpace):
             metric=EuclideanMetric(dim, shape=(dim,)),
         )
 
-    def get_identity(self):
-        """Get the identity of the group.
+    @property
+    def identity(self):
+        """Identity of the group.
 
         Returns
         -------
         identity : array-like, shape=[n]
         """
-        identity = gs.zeros(self.dim)
-        return identity
-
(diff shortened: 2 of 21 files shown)


ninamiolane:
@lpereira95 everything looks good to me, if the tests + deep source tests pass (or are ignored).

One question : could you remind me of the conflict between _SpecialEuclideanGroup(MatrixLieGroup) on the variable `n`?

I wonder if the bug in #1694 is symptomatic of a deeper math issue. I do not see any difference between:
- a manifold `MyManifold(Manifold)` inheriting from manifold and overwriting the `dim`
- a matrix lie group `_SpecialEuclideanGroup(MatrixLieGroup)` inheriting from manifold and overwriting the `n`

I am wondering whether the methods that pose a conflict on `n` should be put outside of MatrixLieGroup? (linear algebra backend?)

lpereira95:
After discussion with @ninamiolane:

* in SO(n) and SE(n), `n` refers to the dimension of the defining/fundamental representation, i.e., loosely, the dimension of the vector space on which these matrices act: $\mathbb{R}^n$
* we use homogeneous representations to represent elements in SO(n) and SE(n)
* while the defining and homogeneous representations agree for SO(n), they do not for SE(n), hence the conflict.

We decided to rename `n_lg` and `n_la` (`n` in master) to `representation_dim`.

lpereira95:
@ninamiolane test failures are unrelated and deep source comments are meaningless.

PR #1774, opened 2022-12-22
Refactor Product and NFold

johnharveymath:
## Description

This PR is a straightforward refactoring. It places all code for products together in `product_manifold.py` and all code for n-fold manifolds together in `nfold_manifold.py`.

## Issue

`ProductManifold` and `NFoldManifold` were in one module, while `ProductRiemannianMetric` and `NFoldMetric` were in another. This separation of the manifold from the metric is not in line with practice across the code base.

On the other hand, mixing product classes and n-fold classes together in modules does not seem to have any benefits.


lpereira95:
@nguigs, is there any special reason for you to have decided to organize the code as it was? do you agree with the restructure proposed by @johnharveymath?

luisfpereira:
I think this makes the module more consistent. I'll merge @johnharveymath.

luisfpereira (approved):

PR #1793, opened 2023-01-11
Update setup.cfg

johnharveymath:
## Description

Require more recent version of joblib

## Issue

`test_sasaki_metric.py` failed for me on a fresh installation due to an error in `joblib`. This works fine with any version above 0.17.0. This PR updates `setup.cfg` with the required version.

Ironically, the linting tools picked up trailing whitespace in the line `lint = `.


ninamiolane (approved):
Also just ran into this error. Thank you for the fix!

PR #1764, opened 2022-12-15
Add product positive reals and complex poincare disks

YannCabanes:
Add the `ProductPositiveRealsAndComplexPoincareDisks` manifold.

The `ProductPositiveRealsAndComplexPoincareDisks` manifold is defined as a
product manifold of the positive reals manifold and (n-1) complex Poincaré disks.
The positive reals and complex Poincaré disks product has a product metric.
The product metric on the positive reals and complex Poincaré disks product space is
the positive reals metric and (n - 1) complex Poincaré metrics multiplied by constants.
This product manifold can be used to represent Toeplitz HPD matrices.
The `ProductPositiveRealsAndComplexPoincareDisks` corresponds to
the one-dimensional case of `ProductHPDMatricesAndSiegelDisks`.
Indeed, `PositiveReals` is the one-dimensional case of `HPDMatrices` and
`ComplexPoincareDisk` is the one-dimensional case of `Siegel`.
In these one-dimensional manifolds, many simplifications occur compared with
the multidimensional manifolds since matrices commute in dimension 1.

References
----------
.. [Cabanes_2022] Yann Cabanes. Multidimensional complex stationary
    centered Gaussian autoregressive time series machine learning
    in Poincaré and Siegel disks: application for audio and radar
    clutter classification, PhD thesis, tel-03708515, 2022.
    https://theses.hal.science/tel-03708515
.. [Cabanes_CESAR_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon et
    Jérémie Bigot. Unsupervised Machine Learning for Pathological Radar Clutter
    Clustering: the P-Mean-Shift Algorithm, IEEE, C&ESAR 2019, Rennes, France, 2019.
    https://hal.archives-ouvertes.fr/hal-02875430
.. [Cabanes_RADAR_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon et
    Jérémie Bigot. Non-Supervised High Resolution Doppler Machine Learning for
    Pathological Radar Clutter, IEEE, RADAR 2019, Toulon, France, 2019.
    https://hal.archives-ouvertes.fr/hal-02875415
.. [Cabanes_GSI_2019] Yann Cabanes, Frédéric Barbaresco, Marc Arnaudon et
    Jérémie Bigot. Toeplitz Hermitian Positive Definite Matrix Machine Learning
    based on Fisher Metric, IEEE, GSI 2019, Toulouse, France, 2019.
[...]

PR #1798, opened 2023-01-20
Fix complex output types

luisfpereira:
This PR ensures consistency of output types in complex metrics. In particular, new tests are added for inner_product, dist, log, and exp.

It results from the discussion in #1764.


ninamiolane (approved):
Thanks!

PR #1804, opened 2023-01-29
test linting

Jacqueskim:
Currently have Grassmannian.py and ipynb and test files


ninamiolane (changes requested):
Address my comments from Christine's PR (applied to this PR) and resubmit.

PR #1828, opened 2023-02-20
Fixes and improvements coming from test refactoring (part 2)

luisfpereira:
Following #1771, this PR brings some fixes and improvements developed while refactoring tests  in #1767.

Main fixes/improvements are related with vectorization. In some cases, slight improvements to the API are done.

### DIFF
--- a/geomstats/geometry/fiber_bundle.py
+++ b/geomstats/geometry/fiber_bundle.py
@@ -12,6 +12,7 @@ import geomstats.backend as gs
 from geomstats.geometry.lie_group import LieGroup
 from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.vectorization import get_batch_shape
 
 
 class FiberBundle(Manifold, ABC):
@@ -72,7 +73,7 @@ class FiberBundle(Manifold, ABC):
 
         This is the projection of the fiber bundle, defined on the total
         space, with values in the base manifold. This map is surjective.
-        By default, the base manifold  is not explicit but is identified with a
+        By default, the base manifold is not explicit but is identified with a
         local section of the fiber bundle, so the submersion is the identity
         map.
 
@@ -86,7 +87,7 @@ class FiberBundle(Manifold, ABC):
         projection : array-like, shape=[..., {base_dim, [n, m]}]
             Point of the base manifold.
         """
-        return point
+        return gs.copy(point)
 
     @staticmethod
     def lift(point):
@@ -108,7 +109,7 @@ class FiberBundle(Manifold, ABC):
         lift : array-like, shape=[..., {total_space.dim, [n, m]}]
             Point of the total space.
         """
-        return point
+        return gs.copy(point)
 
     def tangent_riemannian_submersion(self, tangent_vec, base_point):
         """Project a tangent vector to base manifold.
@@ -164,14 +165,12 @@ class FiberBundle(Manifold, ABC):
         aligned : array-like, shape=[..., {total_space.dim, [n, m]}]
             Action of the optimal g on point.
         """
+        # TODO: need to fix vectorization (with use of optimizers)
+
         group = self.group
         group_action = self.group_action
-        initial_distance = self.total_space_metric.squared_dist(point, base_point)
-        if isinstance(initial_distance, float) or initial_distance.shape == ():
-            n_samples = 1
-        else:
-            n_samples = len(initial_distance)
-        max_shape = (n_samples, self.group_dim) if n_samples > 1 else (self.group_dim,)
+
+        max_shape = get_batch_shape(self, point, base_point) + (self.group_dim,)
 
         if group is not None:
 
--- a/geomstats/geometry/full_rank_matrices.py
+++ b/geomstats/geometry/full_rank_matrices.py
@@ -46,8 +46,7 @@ class FullRankMatrices(OpenSet):
         has_right_rank = gs.where(
             gs.linalg.matrix_rank(point) == self.rank, True, False
         )
-        belongs = gs.logical_and(gs.array(has_right_size), has_right_rank)
-        return belongs
+        return gs.logical_and(gs.array(has_right_size), has_right_rank)
 
     def projection(self, point):
         r"""Project a matrix to the set of full rank matrices.
@@ -95,12 +94,12 @@ class FullRankMatrices(OpenSet):
         samples : array-like, shape=[..., n, k]
             Point sampled on :math:`R_*^{n\times k}`.
         """
-        n = self.n
-        k = self.k
         sample = []
         n_accepted, iteration = 0, 0
         while n_accepted < n_samples and iteration < n_iter:
-            raw_samples = gs.random.normal(size=(n_samples - n_accepted, n, k))
+            raw_samples = gs.random.normal(
+                size=(n_samples - n_accepted, self.n, self.k)
+            )
             ranks = gs.linalg.matrix_rank(raw_samples)
             selected = ranks == self.rank
             sample.append(raw_samples[selected])
--- a/geomstats/geometry/hyperboloid.py
+++ b/geomstats/geometry/hyperboloid.py
@@ -10,7 +10,6 @@ import math
 
 import geomstats.algebra_utils as utils
 import geomstats.backend as gs
-import geomstats.vectorization
 from geomstats.geometry._hyperbolic import HyperbolicMetric, _Hyperbolic
 from geomstats.geometry.base import LevelSet
 from geomstats.geometry.minkowski import Minkowski, MinkowskiMetric
@@ -23,19 +22,13 @@ class Hyperboloid(_Hyperbolic, LevelSet):
     Minkowski space as the set of points with squared norm equal to -1. For other
     representations of hyperbolic spaces see the `Hyperbolic` class.
 
-    The default_coords_type parameter allows to choose the representation of the
-    points as input.
-
     Parameters
     ----------
     dim : int
         Dimension of the hyperbolic space.
-    default_coords_type : str, {'extrinsic', 'intrinsic'}
-        Default coordinates to represent points in hyperbolic space.
-        Optional, default: 'extrinsic'.
     """
 
-    def __init__(self, dim, default_coords_type="extrinsic", **kwargs):
+    def __init__(self, dim, **kwargs):
         if "scale" in kwargs:
             raise TypeError(
                 "Argument scale is no longer in use: instantiate the "
@@ -43,8 +36,8 @@ class Hyperboloid(_Hyperbolic, LevelSet):
                 "`scale * metric` to rescale the standard metric."
             )
         self.dim = dim
-        kwargs.setdefault("metric", HyperboloidMetric(dim, default_coords_type))
-        super().__init__(dim=dim, default_coords_type=default_coords_type, **kwargs)
+        kwargs.setdefault("metric", HyperboloidMetric(dim))
+        super().__init__(dim=dim, default_coords_type="extrinsic", **kwargs)
 
     def _define_embedding_space(self):
         return Minkowski(self.dim + 1)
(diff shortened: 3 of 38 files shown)


luisfpereira:
@ninamiolane, I've address the meaningful deep source comments.

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
 ERROR_MSG = "Invalid type: %s."
 
 
-def get_n_points(points, point_type):
+def _get_max_ndim_point(*point):
```
ninamiolane, line 17: What is the philosophy behind the use of `get`? Why not `max_ndim_point` which is shorter + more consistent with the current coding style + more consistent with pythonic style generally speaking?

https://stackoverflow.com/questions/374763/should-i-use-get-set-prefixes-in-python-method-names

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+def get_n_points(space, *point):
     """Compute the number of points.
 
     Parameters
     ----------
-    points : array-like
-        Input points.
-    point_type : str, {'scalar', ''vector', 'matrix'}
-        Type of input points.
+    space : Manifold
```
ninamiolane, line 31: Missing parameter descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+
+    return gs.prod(point_max_ndim.shape[: -space.point_ndim])
+
+
+def check_is_batch(space, *point):
+    """Check if inputs are batch.
+
+    Parameters
+    ----------
+    space : Manifold
```
ninamiolane, line 52: Missing parameter descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    point : gs.array
+
+    Returns
+    -------
+    is_batch : bool
+    """
+    return any(point_.ndim > space.point_ndim for point_ in point)
+
+
+def get_batch_shape(space, *point):
```
ninamiolane, line 62: Ditto re: get

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    """Get batch shape.
+
+    Parameters
+    ----------
+    space : Manifold
```
ninamiolane, line 67: Missing parameter descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
 
-    is_vect_scalar = point_type == "scalar" and points_ndim == 1
-    is_vect_vector = point_type == "vector" and points_ndim == 2
-    is_vect_matrix = point_type == "matrix" and points_ndim == 3
+def repeat_point(point, n_reps=2, expand=False):
+    """Repeat point.
+
+    Parameters
+    ----------
+    point : gs.array
```
ninamiolane, line 83: Missing parameter descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    n_reps : int
+    expand : bool
+        Repeat even if n_reps == 1.
+
+    Returns
+    -------
+    rep_point : gs.array
```
ninamiolane, line 90: Missing return descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    """Check if inputs are batch.
+
+    Parameters
+    ----------
+    space : Manifold
+    point : gs.array
+
+    Returns
+    -------
+    is_batch : bool
```
ninamiolane, line 57: Missing return descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    """Get batch shape.
+
+    Parameters
+    ----------
+    space : Manifold
+    point : gs.array
+
+    Returns
+    -------
+    batch_shape : tuple
```
ninamiolane, line 72: Missing return descriptions

geomstats/vectorization.py
```diff
@@ -14,32 +14,85 @@
...
+    if not expand and n_reps == 1:
+        return gs.copy(point)
 
-    if is_vect_scalar or is_vect_vector or is_vect_matrix:
-        n_points = gs.shape(points)[0]
-    return n_points
+    return gs.repeat(gs.expand_dims(point, 0), n_reps, axis=0)
 
 
 def decorator(input_types):
```
ninamiolane, line 98: are we ready to remove this decorator? 🤞

tests/data/hyperboloid_data.py
```diff
@@ -23,26 +23,24 @@ class HyperboloidTestData(_LevelSetTestData):
 
     Space = Hyperboloid
 
+    def intrinsic_after_extrinsic_test_data(self):
+        return self.generate_tests([])
```
ninamiolane, line 27: Is that generating empty tests? Will it be confusing, as in: we think the tests pass but in fact nothing is being tested? Should we use the skip flags instead. Feel free to ignore this comment if you plan in addressing this in the upcoming test PR.

tests/tests_geomstats/test_complex_matrices.py
```diff
@@ -63,170 +43,30 @@ def test_frobenius_product(self, mat_a, mat_b, expected):
...
-            gs.array(expected),
-        )
-
     def test_is_hpd(self, m, n, mat, expected):
         self.assertAllClose(
             ComplexMatrices(m, n).is_hpd(gs.array(mat, dtype=CDTYPE)),
             gs.array(expected),
         )
 
-    def test_is_upper_triangular(self, m, n, mat, expected):
```
ninamiolane, line 141: We use the methods from `Matrices` to compute on `ComplexMatrices`. Which is great, because it avoids duplicated code. Do we still want to check that these methods work for _complex_ matrices, i.e. for arrays with complex values? 🤔

tests/tests_geomstats/test_hyperboloid.py
```diff
@@ -14,11 +14,13 @@
 class TestHyperboloid(LevelSetTestCase, metaclass=Parametrizer):
     skip_test_extrinsic_after_intrinsic = True
     skip_test_projection_belongs = True
+    skip_test_extrinsic_ball_extrinsic_composition = True
```
ninamiolane, line 17: Why did these skips appear? was something working previously, but not anymore with the new version? If so, what is the plan to recover the previous behavior?

tests/tests_geomstats/test_hyperboloid.py
```diff
@@ -68,6 +70,9 @@ class TestHyperboloidMetric(RiemannianMetricTestCase, metaclass=Parametrizer):
     skip_test_ricci_tensor_shape = True
     skip_test_sectional_curvature_shape = True
     skip_test_exp_belongs = True
+    skip_test_exp_after_log_intrinsic_ball_extrinsic = True
```
ninamiolane, line 73: Ditto re:skips

ninamiolane (approved):
Thank you for this incredible PR!! Such a pleasure to see the codebase cleaned up like this, super grateful to you for doing this 🙌

geomstats/geometry/fiber_bundle.py
```diff
@@ -12,6 +12,7 @@
 from geomstats.geometry.lie_group import LieGroup
 from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.vectorization import get_batch_shape
```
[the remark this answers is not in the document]
luisfpereira: usually when I use this method I'm interested in writing something like `batch_shape = get_batch_shape(space, point)`. Calling the method `batch_shape` will force me to change my variable name to something more incomprehensible. That's the reasoning on why to add a verb at the start.

geomstats/geometry/full_rank_matrices.py
```diff
@@ -95,12 +94,12 @@ def random_point(self, n_samples=1, bound=1.0, n_iter=100):
...
             Point sampled on :math:`R_*^{n\times k}`.
         """
-        n = self.n
-        k = self.k
         sample = []
         n_accepted, iteration = 0, 0
         while n_accepted < n_samples and iteration < n_iter:
-            raw_samples = gs.random.normal(size=(n_samples - n_accepted, n, k))
+            raw_samples = gs.random.normal(
+                size=(n_samples - n_accepted, self.n, self.k)
```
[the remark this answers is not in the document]
luisfpereira: To be address later (#1829).

geomstats/geometry/hyperboloid.py
```diff
@@ -181,37 +138,12 @@ def to_tangent(self, vector, base_point):
...
-
         sq_norm = self.embedding_space.metric.squared_norm(base_point)
         inner_prod = self.embedding_space.metric.inner_product(base_point, vector)
 
         coef = inner_prod / sq_norm
 
-        tangent_vec = vector - gs.einsum("...,...j->...j", coef, base_point)
-        return tangent_vec
-
-    def is_tangent(self, vector, base_point, atol=gs.atol):
```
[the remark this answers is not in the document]
luisfpereira: come's from free from `LevelSet`. it also becomes more readable, in the sense that `self.embedding_space.metric.inner_product` is the `tangent_submersion` of this space.

geomstats/geometry/spd_matrices.py
```diff
@@ -132,12 +129,10 @@ def random_tangent_vec(self, base_point, n_samples=1):
...
         )
 
-        tangent_vec = Matrices.mul(sqrt_base_point, tangent_vec_at_id, sqrt_base_point)
-
-        return tangent_vec
+        return Matrices.mul(sqrt_base_point, tangent_vec_at_id, sqrt_base_point)
 
     @staticmethod
-    def aux_differential_power(power, tangent_vec, base_point):
+    def _aux_differential_power(power, tangent_vec, base_point):
```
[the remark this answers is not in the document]
luisfpereira: I think `aux` is actually helpful because at the end it is not really computing differential power, but intermediate results.

geomstats/geometry/special_orthogonal.py
```diff
@@ -386,22 +380,18 @@ def projection(self, point):
         rot_mat : array-like, shape=[..., n, n]
             Rotation matrix.
         """
+        # TODO: why receiving (n, n)?
```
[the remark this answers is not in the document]
luisfpereira: To be address later (see #1830).

geomstats/vectorization.py
luisfpereira: see comment above. related with this idea that when I call the function I'll likely want to use the same name. having a verb helps avoiding the conflict.

geomstats/vectorization.py
luisfpereira: as above

geomstats/geometry/special_orthogonal.py
```diff
@@ -560,6 +555,7 @@ def regularize_tangent_vec(self, tangent_vec, base_point, metric=None):
         regularized_tangent_vec : array-like, shape=[..., 1]
             Regularized tangent vector.
         """
+        # TODO: remove metric
```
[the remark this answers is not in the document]
luisfpereira: Thanks for the suggestion (see #1831). In fact it is helpful to have it in single issue (at the time of development it was not so clear there were so many related TODOs).

tests/data/hyperboloid_data.py
luisfpereira: This one shouldn't have been removed. I've added back the data. On the other hand, `extrinsic_after_intrinsic` needs to be skipped in this way because it assumes space can receive `default_point_type="intrinsic"`, which is not the case. Notice that data is still created when we skip with "skip_" flag, so I really need to skip it this way.

I'm addressing this in the new tests, but my first thought is that we need to rethink these methods. The reason is that I find counter-intuitive to have methods in a class that can receive point shapes the class is not supposed to handle, and this will always happen because:

* if we have an intrinsic representation, `extrinsic_to_intrinsic_coords` will receive an "unexpected" point type.
* if we have an extrinsic representation, `intrinsic_to_extrinsic_coords` will receive an "unexpected" point type.

This relates with the discussion in the last meeting regarding have this "PointRepresentationTransform" / "CoordsTransform" object.

tests/tests_geomstats/test_complex_matrices.py
luisfpereira: The idea is to move these methods from `Matrices` to the backend and then add complete tests there (i.e. checking also complex if they can handle it). This step towards removing all this methods from `ComplexMatrices` is to make it easier to identify which methods are specific (the ones kept there for now).

tests/tests_geomstats/test_hyperboloid.py
luisfpereira: These tests assume `Hyperboloid` can receive `default_coords_type="intrinsic"`, which is not the case anymore. I'm keeping the tests to check later what is missing now and how could this behavior be replicated in a different way.

tests/tests_geomstats/test_hyperboloid.py
luisfpereira: same as above.

luisfpereira:
Thanks for the valuable comments @ninamiolane, they really helped to put the code in better shape.

I've solved and marked as resolved most of your comments. I dropped some comments in the ones that may need further discussion (not necessarily now).

geomstats/vectorization.py
luisfpereira: almost! there's still some uses of it in some metrics. I haven't addressed that cleaning yet. It is one of the next steps.

Issue #1829, opened 2023-02-21
Change variable names in FullRankMatrices


luisfpereira:
Following @ninamiolane's comment in #1828, for consistency with variable names used in e.g. `Matrices` it is preferable to have `(m, n)` instead of `(n, k)` to represent the matrix shape in `FullRankMatrices`.

PR #1695, opened 2022-10-19
New metrics correlation matrices

ythanwerdas:
## Checklist

- [ ] My pull request has a clear and explanatory title.
- [ ] If neccessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [ ] I have added apropriate unit tests.
- [ ] I have made sure the code passes all unit tests. (refer to comment below)
- [ ] My PR follows [PEP8](https://peps.python.org/pep-0008/) guidelines. (refer to comment below)
- [ ] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#coding-style-guidelines) and API.
- [ ] My code is properly documented and I made sure the documentation renders properly. ([Link](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#documentation))
- [ ] I have linked to issues and PRs that are relevant to this PR.

PR #1836, opened 2023-03-23
Refactor metric to receive space (equip_with_metric behavior)

### LINKED ISSUE #1831: Remove metric from regularize signatures
Remove `metric` from the signature of:

* `geomstats.geometry.special_euclidean`
  * `_SpecialEuclideanVectors.regularize_tangent_vec_at_identity`
  * `_SpecialEuclidean2Vectors.regularize_tangent_vec`
  * `_SpecialEuclidean3Vectors.regularize_tangent_vec`

* `geomstats.geometry.special_orthogonal`
  * `_SpecialOrthogonalVectors.regularize_tangent_vec_at_identity`
  * `_SpecialOrthognalVectors.regularize_tangent_vec`

luisfpereira:
This PR reflects the changes proposed [here](https://geomstats.github.io/hackathons/ihp_10_2022.html#equip-with-metric), i.e. now `RiemannianMetric` receives the space and stores it as `_space`. The suggested way to equip a given  space is using `space.equip_with_metric`.

An example:

```python
space = Hypersphere(dim=2, equip=False)
space.equip_with_metric(HypersphereMetric)
```

Nonetheless, the following code is still possible and useful sometimes:

```python
space = Hypersphere(dim=2, equip=False)

metric = HypersphereMetric(space)
```

An assumption we are making is that `space` (after  equipping) is an equipped manifold, i.e. $(M, g)$. Therefore, I've tried to remove instantiations of metrics within manifolds (e.g. in `MatrixLieGroup` there's no more a `left_canonical_metric` and a `right_canonical_metric`; in `DiscreteCurves` I've removed `srv_metric`, `l2_metric` and so on). This simplifies testing a lot and gives a better mental model to a user of a space: if a metric is not instantiated, then we clearly know it is not used anywhere to make any computation.

Given the size of this PR, it would be nice if several different people could take a look to different parts, to ensure I haven't overlooked anything. I suggest some mapping below, but feel free to look to other parts or even to not provide any feedback.

@ninamiolane, as we've discussed, it would be nice if you take an overall view. I'm keen to see your feedback on the pullback metric and on the changes to autodiff. You probably also have something relevant to say in discrete curves. Also, I've deleted a small part in the notebook `03_practical_methods__data_on_manifolds` because `DiscreteCurves` now asks for the setting of `k_sampling_points`.

@nguigs, as I told you, I'm keen to see your feedback on Lie group stuff and FiberBundle. Your overall feedback would be appreciated.

@alebrigant, it would be great if you take a look to the information geometry module and discrete curves.

@tramy1258, it would be great if you can take a look to information geometry, in particular to `FisherRaoMetric`

@johnharveymath, the changes in `ProductManifold` and `NFoldManifold` are probably of your interest.

@ymontmarin, your feedback on the PullbackMetric and the PullbackDiffeoMetric would be deeply appreciated.

@astamm after this we are kind of close to the state we've discussed in the hackathon, which hopefully will make your R code easier to develop/maintain.

closes #1831

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -74,6 +74,7 @@ class DiscreteCurves(Manifold):
         }
 
     def new(self, equip=True):
+        """Create manifold with same parameters."""
         return DiscreteCurves(
             self.ambient_manifold,
             k_sampling_points=self.k_sampling_points,
--- a/geomstats/geometry/full_rank_correlation_matrices.py
+++ b/geomstats/geometry/full_rank_correlation_matrices.py
@@ -278,10 +278,10 @@ class FullRankCorrelationAffineQuotientMetric(QuotientMetric):
     manifold of full-rank correlation matrices.
     """
 
-    def __init__(self, space):
-        # TODO: change API to receive total space?
-        total_space = SPDMatrices(space.n, equip=False)
-        total_space.equip_with_metric(SPDAffineMetric)
+    def __init__(self, space, total_space=None):
+        if total_space is None:
+            total_space = SPDMatrices(space.n, equip=False)
+            total_space.equip_with_metric(SPDAffineMetric)
 
         super().__init__(
             space=space,
--- a/geomstats/geometry/manifold.py
+++ b/geomstats/geometry/manifold.py
@@ -83,6 +83,13 @@ class Manifold(abc.ABC):
         self.metric = Metric(self, **metric_kwargs)
 
     def equip_with_group_action(self, group_action):
+        """Equip manifold with group action.
+
+        Parameters
+        ----------
+        group_action : str
+
+        """
         self.group_action = group_action
 
     def _check_equip_with_quotient_structure(self):
@@ -94,6 +101,10 @@ class Manifold(abc.ABC):
                 raise ValueError(f"Need to equip with `{structure_name}` first")
 
     def equip_with_quotient_structure(self):
+        """Equip manifold with quotient structure.
+
+        Creates attributes `quotient` and `fiber_bundle`.
+        """
         self._check_equip_with_quotient_structure()
 
         key = type(self.metric), self.group_action
--- a/geomstats/geometry/pre_shape.py
+++ b/geomstats/geometry/pre_shape.py
@@ -62,6 +62,7 @@ class PreShapeSpace(LevelSet):
         )
 
     def new(self, equip=True):
+        """Create manifold with same parameters."""
         return PreShapeSpace(
             k_landmarks=self.k_landmarks, m_ambient=self.m_ambient, equip=equip
         )
@@ -240,6 +241,8 @@ class PreShapeSpace(LevelSet):
 
 
 class PreShapeSpaceBundle(FiberBundle):
+    r"""Class for the Kendall pre-shape space bundle."""
+
     def align(self, point, base_point, **kwargs):
         """Align point to base_point.
 
--- a/geomstats/geometry/rank_k_psd_matrices.py
+++ b/geomstats/geometry/rank_k_psd_matrices.py
@@ -326,10 +326,10 @@ class BuresWassersteinBundle(FiberBundle):
 class PSDBuresWassersteinMetric(QuotientMetric):
     """Bures-Wasserstein metric for fixed rank PSD matrices."""
 
-    def __init__(self, space):
-        # TODO: change API to receive total space?
-        k = space.rank if hasattr(space, "rank") else space.n
-        total_space = FullRankMatrices(space.n, k, equip=False)
-        total_space.equip_with_metric(MatricesMetric)
+    def __init__(self, space, total_space=None):
+        if total_space is None:
+            k = space.rank if hasattr(space, "rank") else space.n
+            total_space = FullRankMatrices(space.n, k, equip=False)
+            total_space.equip_with_metric(MatricesMetric)
 
         super().__init__(space=space, fiber_bundle=BuresWassersteinBundle(total_space))
--- a/tests/tests_geomstats/test_discrete_surfaces.py
+++ b/tests/tests_geomstats/test_discrete_surfaces.py
@@ -32,9 +32,6 @@ class TestDiscreteSurfaces(ManifoldTestCase, metaclass=Parametrizer):
 
     testing_data = DiscreteSurfacesTestData()
 
-    def test_manifold_shape(self, space_args):
-        return _test_manifold_shape(self, space_args)
-
     def test_vertex_areas(self, faces, point):
         """Test vertex_areas.
 


geomstats/geometry/pullback_metric.py
```diff
@@ -86,81 +51,48 @@ def metric_matrix(self, base_point=None, n_jobs=1, **joblib_kwargs):
...
-    def hessian_immersion(self, base_point):
-        """Compute the Hessian of the immersion.
+        else:
+            immersed_basis_elements = gs.matvec(jacobian_immersion, basis_elements)
 
-        Parameters
-        ----------
-        base_point : array-like, shape=[..., dim]
-            Base point.
+        elems = {}
```
ninamiolane, line 76: why a dictionary of tuples? why not some array?

ninamiolane: how about renaming elems and mat to something more specific / understandable for mathematicians?

geomstats/geometry/pullback_metric.py
```diff
@@ -221,38 +139,12 @@ def second_fundamental_form(self, base_point):
...
-            second_fundamental_forms[0]
-            if initial_ndim == 1
-            else second_fundamental_forms
-        )
+        christoffels = self.christoffels(base_point)
+
+        jacobian = self._space.jacobian_immersion(base_point)
+        hessian = self._space.hessian_immersion(base_point)
+
+        return hessian - gs.einsum("...kij,...dk->...dij", christoffels, jacobian)
```
ninamiolane, line 147: good if k denotes the index of the derivative \partial_k

geomstats/geometry/pullback_metric.py
```diff
@@ -86,81 +51,48 @@ def metric_matrix(self, base_point=None, n_jobs=1, **joblib_kwargs):
...
+        is_vec = check_is_batch(self._space, base_point)
 
-        pool = joblib.Parallel(n_jobs=n_jobs, **joblib_kwargs)
-        out = pool(
-            pickable_inner_product(i, j)
-            for i, j in itertools.product(range(self.dim), range(self.dim))
-        )
-        metric_mat = gs.reshape(gs.array(out), (-1, self.dim, self.dim))
-        return metric_mat[0] if base_point.ndim == 1 else metric_mat
+        immersed_base_point = self._space.immersion(base_point)
```
ninamiolane, line 57: it is thus assumed that the space has an immersion function. i.e. is an immersed set. LGTM, I just wonder if we could make it explicit in the documentation somewhere that this is how it works?

geomstats/geometry/pullback_metric.py
```diff
@@ -273,64 +165,30 @@ def mean_curvature_vector(self, base_point):
...
-        Dimension.
-    shape : tuple of int
-        Shape of one element of the underlying manifold.
-        Optional, default : None.
-    """
+    """Pullback metric via a diffeomorphism."""
 
-    def __init__(self, dim, shape=None):
-        super().__init__(dim=dim, shape=shape)
+    def __init__(self, space, signature=None):
```
ninamiolane, line 176: Why does this metric have an init? It seems that the logic is different here?

To match the logic of the other pullback metric, it should be a metric for which the space has to be defined by an immersion that is a diffeomorphism.

And we wouldn't need embedding space bcs that would come from the immersed set.

geomstats/geometry/pullback_metric.py
```diff
@@ -351,26 +209,26 @@ def diffeomorphism(self, base_point):
...
+        :math:`f: M \rightarrow N` of the manifold
+        :math:`M` into the manifold `N`.
 
-        Raw jacobian autodiff of diffeomorphism regardless of vectorization.
+        :math:`f^-1: N \rightarrow M` of the manifold
 
         Parameters
         ----------
-        base_point : array-like, shape=[..., *shape]
+        image_point : array-like, shape=[..., *i_shape]
```
ninamiolane, line 224: I wonder if "image_point" is the more intuitive wording here..

output space // output point?
embedding space // embedded point?
immersed point?

geomstats/geometry/pullback_metric.py
```diff
@@ -425,65 +274,23 @@ def tangent_diffeomorphism(self, tangent_vec, base_point):
         image_tangent_vec : array-like, shape=[..., *i_shape]
             Image tangent vector at image of the base point.
         """
-        base_point = gs.broadcast_to(base_point, tangent_vec.shape)
-        rad = base_point.shape[: -len(self.shape)]
+        # TODO: avoid reshapings?
```
ninamiolane, line 277: Leftover?

geomstats/geometry/pullback_metric.py
```diff
@@ -425,65 +274,23 @@ def tangent_diffeomorphism(self, tangent_vec, base_point):
...
-        Returns
-        -------
-        mat : array-like, shape=[..., *shape, ..., *i_shape]
-            Inner-product matrix.
-        """
-        if self._raw_inverse_jacobian_diffeomorphism is None:
-            self._raw_inverse_jacobian_diffeomorphism = gs.autodiff.jacobian(
-                self.inverse_diffeomorphism
-            )
-        return self._raw_inverse_jacobian_diffeomorphism(image_point)
```
ninamiolane, line 485: OOooo yes! No more `_raw_`?! Love it!!

geomstats/geometry/quotient_metric.py
```diff
@@ -198,13 +175,13 @@ def curvature(self, tangent_vec_a, tangent_vec_b, tangent_vec_c, base_point):
...
         """
-        bundle = self.fiber_bundle
+        bundle = self._space
         fiber_point = bundle.lift(base_point)
         horizontal_a = bundle.horizontal_lift(tangent_vec_a, base_point)
         horizontal_b = bundle.horizontal_lift(tangent_vec_b, base_point)
         horizontal_c = bundle.horizontal_lift(tangent_vec_c, base_point)
 
-        top_curvature = self.total_space_metric.curvature(
+        top_curvature = bundle.total_space_metric.curvature(
```
ninamiolane, line 184: bundle.metric?

geomstats/geometry/quotient_metric.py
```diff
@@ -295,7 +272,7 @@ def curvature_derivative(
         nabla_h_y = bundle.integrability_tensor(hor_h, hor_y, point_fiber)
         nabla_h_z = bundle.integrability_tensor(hor_h, hor_z, point_fiber)
 
-        nabla_curvature_top = self.total_space_metric.curvature_derivative(
+        nabla_curvature_top = bundle.total_space_metric.curvature_derivative(
```
ninamiolane, line 275: Ditto

geomstats/geometry/quotient_metric.py
```diff
@@ -395,15 +372,15 @@ def directional_curvature_derivative(
...
+        bundle = self._space
         point_fiber = bundle.lift(base_point)
         hor_x = bundle.horizontal_lift(tangent_vec_a, point_fiber)
         hor_y = bundle.horizontal_lift(tangent_vec_b, point_fiber)
 
         nabla_x_x = gs.zeros_like(hor_x)
         nabla_x_y = bundle.integrability_tensor(hor_x, hor_y, point_fiber)
 
-        nabla_curvature_top = self.total_space_metric.curvature_derivative(
+        nabla_curvature_top = bundle.total_space_metric.curvature_derivative(
```
ninamiolane, line 383: Ditto.

tests/data/pullback_metric_data.py
```diff
@@ -1,220 +1,574 @@
+import abc
+
 import geomstats.backend as gs
+from geomstats.geometry.euclidean import Euclidean
+from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.pullback_metric import PullbackMetric
 from tests.data_generation import TestData
 
 
+class ImmersedSet(Manifold, abc.ABC):
```
ninamiolane, line 10: Why is this here?

ninamiolane (approved):
I reviewed some files!! It looks purely extraordinary, it is such a relief to see how much the code gets simplified! @luisfpereira you made it happen!!

The code is also very clean, I didn't see anything bad in what I reviewed, thus feel free to merge it if you need to (.e.g before I fully review since I'm gone for 2 weeks).

geomstats/geometry/complex_manifold.py
```diff
@@ -30,31 +29,46 @@ class ComplexManifold(abc.ABC):
...
+            self.default_point_type = "vector"
+        elif self.point_ndim == 2:
+            self.default_point_type = "matrix"
+        else:
+            self.default_point_type = "other"
+
+        if equip:
+            self.equip_with_metric()
+
+    def equip_with_metric(self, Metric=None, **metric_kwargs):
```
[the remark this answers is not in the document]
luisfpereira: the idea with uppercase is to show it receives a non-instantiated object. for me the logical solutions are `Metric` or `MetricClass`.

geomstats/geometry/discrete_curves.py
```diff
@@ -297,8 +280,11 @@ def __init__(self, ambient_manifold, k_sampling_points=10):
         super().__init__(
             dim=dim,
             shape=(k_sampling_points,) + ambient_manifold.shape,
+            equip=False,
```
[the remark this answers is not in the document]
luisfpereira: In master it is not clear what's the `default_metric` here, so I made it `equip=False`.

geomstats/geometry/discrete_curves.py
```diff
@@ -479,7 +465,7 @@ def projection(self, point, atol=gs.atol, max_iter=1000):
 
         is_vec = point.ndim > 2
 
-        srv_metric = self.embedding_space.srv_metric
+        srv_metric = self.embedding_space.metric
```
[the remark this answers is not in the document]
luisfpereira: That's correct. Nevertheless, `embedding_space` is instantiated within `ClosedDiscreteCurves`, so we should not expect anyone to change its metric I guess. We can always add some check though, or make it explicit in the docstrings the embedding space should be equipped with srv. I tend towards the latter.

geomstats/geometry/discrete_curves.py
```diff
@@ -1250,41 +1155,40 @@ class SRVMetric(PullbackDiffeoMetric):
...
-        r"""Define the metric which is pulled back.
+    def _check_ambient_manifold(self, ambient_manifold):
+        if not isinstance(ambient_manifold, Euclidean):
+            raise AssertionError(
+                "This metric is only "
+                "implemented for discrete curves embedded "
+                "in a Euclidean space."
+            )
+
+    def define_embedding_space(self):
```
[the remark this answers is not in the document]
luisfpereira: I'll make it private, because in fact your example is something that should not be possible. With it private, it becomes closer to `LevelSet` and in fact fulfills the same goal.

geomstats/geometry/discrete_curves.py
```diff
@@ -1607,13 +1493,23 @@ class SRVShapeBundle(DiscreteCurves, FiberBundle):
...
             ambient_manifold=ambient_manifold,
             k_sampling_points=k_sampling_points,
+            equip=equip,
         )
-        self.total_space_metric = SRVMetric(ambient_manifold=ambient_manifold)
-        self.l2_curves_metric = L2CurvesMetric(ambient_manifold=ambient_manifold)
+        self.l2_curves_metric = L2CurvesMetric(self)
+
+    @staticmethod
+    def default_total_space_metric():
```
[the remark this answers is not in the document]
luisfpereira: You have a point. Nonetheless, we actually want to put a `QuotientMetric` on the bundle, i.e. if we want `SVRMetric` to be the default metric, we simply use directly `DiscreteCurves`.

I think we are missing a better abstraction here. (for now I simply kept the behavior implemented in master)

geomstats/geometry/fiber_bundle.py
```diff
@@ -67,6 +53,28 @@ def __init__(
         self.group_dim = group_dim
         self.group_action = group_action
 
+        if equip:
+            self.equip_with_total_space_metric()
```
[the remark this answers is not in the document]
luisfpereira: as above.

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -21,11 +21,9 @@ class FullRankCorrelationMatrices(LevelSet):
         Integer representing the shape of the matrices: n x n.
     """
 
-    def __init__(self, n, **kwargs):
+    def __init__(self, n):
         self.n = n
-
-        kwargs.setdefault("metric", FullRankCorrelationAffineQuotientMetric(n))
-        super().__init__(dim=int(n * (n - 1) / 2), **kwargs)
+        super().__init__(dim=int(n * (n - 1) / 2), equip=False)
```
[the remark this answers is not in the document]
luisfpereira: That line was a mistake. A `QuotientMetric` expects a `FiberBundle` space, therefore cannot be used here. For now we don't have metrics for this space. #1695 implements some. I'll try to merge it soon afterwards.

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -266,24 +278,3 @@ def horizontal_lift(self, tangent_vec, base_point=None, fiber_point=None):
         diagonal_point = Matrices.diagonal(fiber_point) ** 0.5
         lift = FullRankCorrelationMatrices.diag_action(diagonal_point, tangent_vec)
         return self.horizontal_projection(lift, base_point=fiber_point)
-
-
-class FullRankCorrelationAffineQuotientMetric(QuotientMetric):
```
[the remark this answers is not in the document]
luisfpereira: With the refactoring, the implementation of this metric would simply be `pass`. I've deleted it and used `QuotientMetric` as default (the docstrings of `CorrelationMatricesBundle` were updated with the information previously in the metric). I can bring it back if you find useful to have such an object.

geomstats/geometry/full_rank_matrices.py
```diff
@@ -19,14 +19,17 @@ class FullRankMatrices(OpenSet):
         Integer representing the shape of the matrices: n x k
     """
 
-    def __init__(self, n, k, **kwargs):
-        kwargs.setdefault("dim", n * k)
-        kwargs.setdefault("metric", MatricesMetric(n, k))
-        super().__init__(embedding_space=Matrices(n, k), **kwargs)
+    def __init__(self, n, k, equip=True):
```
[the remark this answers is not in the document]
luisfpereira: Identified in #1829, but it is probably better to address independently.

geomstats/geometry/pullback_metric.py
luisfpereira: I haven't changed the expression, only added `...` for multiple point.

geomstats/geometry/pullback_metric.py
luisfpereira: There's a notebook that shows how to define this. It is actually expected this to be an `ImmersedSet`. (see below)

geomstats/geometry/pullback_metric.py
luisfpereira: There's indeed a logic difference here. I think this is another part where we are still experimenting and haven't converged to an architectural solution.

This is my interpretation.

An `immersion` defines a manifold (that's why we came up with the notion of `ImmersedSet`). e.g. the point representation we use for computations is dependent on the choice of the immersion. A `PullbackMetric` is defined on a manifold defined by an immersion.

On the other hand, a `PullbackDiffeoMetric` defines a **metric** through a diffeomorphism. So we start with two manifolds $M$ and $N$ and then we "assign" a diffeormorphism ($\phi: M \to N$) between them. $M$ can exist without the diffeormorphism. Therefore, I think the diffeo should be defined in the metric and not in the manifold, which will lead to different logics between `PullbackMetric` and `PullbackDiffeoMetric`.

luisfpereira: Regarding the `init`, all the metrics have an init. Most of the times it is the same as `RiemannianMetric`, so we don't explictly write it. Other times we have to perform some more operations besides the ones done by `RiemannianMetric` and `Connection`. That's the case here.

geomstats/geometry/pullback_metric.py
luisfpereira: I have no preference here. I've simply used the terminology already present in the docstrings.

tests/data/pullback_metric_data.py
luisfpereira: The main reason is that I feel it is still very experimental to move it to the main code. Nonetheless, it may be actually relevant to have access to it already (e.g. it will simplify one of the notebooks; someone may actually want to use it), so I'll move it it. We need to design it a bit better later though.

geomstats/geometry/base.py
```diff
@@ -699,3 +697,195 @@ def projection(self, point):
...
+        Returns
+        -------
+        hessian_immersion : array-like, shape=[..., embedding_dim, dim, dim]
+            Hessian at the base point
+        """
+        return gs.autodiff.hessian_vec(
+            self.immersion, func_out_ndim=self.embedding_space.dim
+        )(base_point)
+
+    def is_tangent(self, vector, base_point, atol=gs.atol):
```
nguigs, line 798: These are inherited from `Manifold`, why override them here and allow not to implement them?

geomstats/geometry/complex_matrices.py
```diff
@@ -19,21 +19,23 @@ class ComplexMatrices(ComplexVectorSpace):
...
-        kwargs.setdefault("default_point_type", "matrix")
-        super(ComplexMatrices, self).__init__(shape=(m, n), **kwargs)
+        super().__init__(shape=(m, n), equip=equip)
         self.m = m
         self.n = n
 
+    @staticmethod
+    def default_metric():
+        """Metric to equip the space with if equip is True."""
+        return ComplexMatricesMetric
```
nguigs, line 32: why not pass `m, n` here as well?

geomstats/geometry/fiber_bundle.py
```diff
@@ -23,42 +21,30 @@ class FiberBundle(Manifold, ABC):
...
     def __init__(
         self,
-        dim: int,
-        group: LieGroup = None,
-        total_space_metric: RiemannianMetric = None,
+        group=None,
         group_action=None,
         group_dim=None,
-        **kwargs
+        equip=True,
```
nguigs, line 43: The option to equip or not seems a bit odd to me as some methods in this class require a metric (mathematically) to be defined. This is the case of `align` and `horizontal_projection`. The idea behind this class was that the total space was always equipped. Could this be enforced? Maybe there should not be a kwarg `equip` here or it should always be True.

geomstats/geometry/full_rank_correlation_matrices.py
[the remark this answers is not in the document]
nguigs: well there is a metric for this space, the `FullRankCorrelationAffineQuotientMetric`. Are you saying it won't be possible to equip this space with a `QuotientMetric`?

geomstats/geometry/quotient_metric.py
```diff
@@ -70,20 +49,20 @@ def inner_product(
         """
         if fiber_point is None:
             if base_point is not None:
-                fiber_point = self.fiber_bundle.lift(base_point)
+                fiber_point = self._space.lift(base_point)
```
nguigs, line 52: This is a bit puzzling as the `QuotientMetric` equips the base manifold, not the total space which is also the fiber bundle. So I think `_space` should be the base here, and we still need to access the fiber_bundle.

nguigs (commented):
Great work! Many things will simplify now. I have a few questions about `FiberBundle` and `QuotientMetric` though. 

More generally, in the case
```python
metric = HypersphereMetric(space)
```
How do you ensure that space is compatible with the metric (i.e. prohibiting passing `Matrices` instead of `Hypersphere`?

geomstats/geometry/quotient_metric.py
luisfpereira: You fully right here @nguigs! I've misunderstood this part: I thought the metric was to put in the `FiberBundle`, as it is done with the `PreShapeSpace`. (I also understand better your comments @ninamiolane.)

I can solve it easily for the case where the `QuotientMetric` is defined by the submersion (scenario 1 in your paper). I still need to think a bit more for scenario 2.

I'll change the code accordingly and the we iterate.

geomstats/geometry/fiber_bundle.py
luisfpereira: Yes, I will enforce this.

geomstats/geometry/complex_matrices.py
luisfpereira: The main idea is that `m` and `n` are available in the `space`. Therefore, when I pass the `space` to the `metric`, I don't need to pass anymore `m` and `n`. This way, we ensure the arguments of `space` and `metric` are always compatible without having to perform any check. We also ensure that `shape`, `default_point_type` and other (computational-oriented) attributes exist only in one place.

But, as you state above, this is the only "compatibility" check that is verified automatically. There's currently no way of prohibiting passing e.g. `Matrices` instead of `Hypersphere` to `HypersphereMetric`. It is also true for the converse, it is possible to wrongly do:

```python
space = Hypersphere(dim, equip=False)

space.equip_with_metric(MatricesMetric)
```

I think we probably don't want to restrict this. The main reason is that it may be hard to make a check that is restrictive enough, but not too restrictive: e.g. it would be easier to check `isinstance(metric, HypersphereMetric)`, but then we would not be able to equip the space with e.g. `ScalarProductMetric`. We would also restrict (power) user of developing their own independent metrics or spaces.

geomstats/geometry/base.py
luisfpereira: This is still an underdeveloped class. The main idea of allowing not implementing it is that for using the `PullbackMetric`, these methods are kind of unnecessary (see for example notebook 9). Later we should probably remove this, but I think for now this approach is convenient.

luisfpereira:
@nguigs, I've refactored `FiberBundle` and `QuotientMetric` to take into account your comments.

For scenario 1 cases, i.e. we have an explicit base, I think it has now the desired behavior and it is clear what is what. @ninamiolane, with this we do not use `space.total_space_metric` anymore as it is the total space as you referred above.

For scenario 2 cases, the situation is a bit trickier. I'm using `pre_shape` as an example of what could be an implementation of this case having only one metric per class.

Basically, I've split `PreShapeSpace` in 2: `PreShapeSpace` and `PreShapeSpaceBundle` (names are not definitive). `PreShapeSpace` tries to emulate somehow the concept of `Base`, and is the class that receives the `QuotientMetric` (e.g. `KendallShapeMetric`).  I've tried to keep there everything that is not `FiberBundle`-related. `PreShapeSpaceBundle` inherites from `PreShapeSpace` and `FiberBundle` and is responsible to implement everything that is `FiberBundle`-related. It is our total space. Some methods in `PreShapeSpace` need access to the bundle, so I've instantiated it there.

Does this makes sense @nguigs? Let's try to iterate on it. I can always revert the last commit if you find it too nonsensical.

geomstats/geometry/discrete_curves.py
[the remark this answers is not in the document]
luisfpereira: I'm addressing this after @nguigs' comments.

geomstats/geometry/discrete_curves.py
```diff
@@ -1607,13 +1493,23 @@ class SRVShapeBundle(DiscreteCurves, FiberBundle):
...
+        """Metric to equip the total space with if equip is True."""
+        return SRVMetric
+
+    @staticmethod
+    def default_metric():
```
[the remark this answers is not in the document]
luisfpereira: Being addressed.

geomstats/geometry/fiber_bundle.py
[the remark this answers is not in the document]
luisfpereira: removed in last changes.

geomstats/geometry/full_rank_correlation_matrices.py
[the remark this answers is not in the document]
luisfpereira: Fixed. Now it is working as expected.

geomstats/geometry/full_rank_correlation_matrices.py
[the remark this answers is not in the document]
luisfpereira: I've brought it back.

geomstats/information_geometry/geometric.py
```diff
@@ -19,13 +19,17 @@ class GeometricDistributions(InformationManifoldMixin, OpenSet):
     i.e. [0,1] segment.
     """
 
-    def __init__(self):
+    def __init__(self, equip=True):
         super().__init__(
             dim=1,
-            embedding_space=Euclidean(dim=1),
-            metric=GeometricMetric(),
+            embedding_space=Euclidean(dim=1, equip=equip),
```
tramy1258, line 25: this should be more like this I think?
```
embedding_space=Euclidean(dim=1, equip=False), 
equip=equip)
```

luisfpereira: Thanks @tramy1258! updated accordingly

geomstats/geometry/discrete_curves.py
```diff
@@ -1577,12 +1458,30 @@ def space_derivative(curve):
         mat_3 = from_vector_to_diagonal_matrix(vec_3, 1)
         mat_space_deriv = mat_1 - mat_2 + mat_3
 
-        space_deriv = n_points * gs.matmul(mat_space_deriv, curve)
+        return n_points * gs.matmul(mat_space_deriv, curve)
+
+
+class SRVPreShapeSpace(DiscreteCurves):
```
alebrigant, line 1464: If I understand correctly, this is supposed to represent the base space ? If so, the name is a bit misleading because the base space is usually called the shape space, equipped with the quotient metric, while the total space is called the pre-shape space, and equipped with e.g. the SRV metric.

alebrigant: More generally, I think it would be more natural if the shape space had the same structure as the space of curves: only one class `DiscreteShapes` or `UnparametrizedDiscreteCurves` (the equivalent of `DiscreteCurves`) and several quotient metrics, among which one (e.g. `QuotientSRVMetric`) is chosen as the default metric, just like `SRVMetric` is the default metric of discrete curves. What do you think @luisfpereira ?

alebrigant (commented):
Great work, this simplifies the code so much !!

General question, I think the same as Nicolas': should we test that the space that the user enters as argument in a metric is compatible, i.e. of the right class ?

geomstats/geometry/discrete_curves.py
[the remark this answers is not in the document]
ninamiolane: Using the regular SRV metric on closed curves (cancer cell boundaries) gave geodesics that were going through open curves, etc. We had to re-project on the space of closed curves each time.

I'd say that we indeed need a ClosedSRVMetric that performs this projection. In the meantime, I agree to put equip=False, as was done originally !

ninamiolane (commented):
I answered re:ClosedSRVMetric

PR #1839, opened 2023-04-15
feat(ci): add `pip` caching to CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument to enable caching to speed up CI times.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,14 +2,12 @@ name: "Docs check and Deployment"
 
 on:
   push:
-    branches: [master,github-actions-test]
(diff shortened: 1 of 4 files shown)


luisfpereira (approved):
Thanks @SauravMaheshkar.

Just a clarification about caching: when is the caching updated when a package releases a new version?

SauravMaheshkar:
> Thanks @SauravMaheshkar.
> 
> Just a clarification about caching: when is the caching updated when a package releases a new version?

AFAIK it changes when the file mentioned in the `cache-dependency-path` is changed.

PR #1842, opened 2023-04-17
feat(build): switch to `pyproject.toml` for packaging

SauravMaheshkar:
`pyproject.toml` is the new standard for packaging python packages, `setup.py` is now deprecated (first introduced in [PEP 518](https://peps.python.org/pep-0518/) and later expanded in [PEP 517](https://peps.python.org/pep-0517/), [PEP 621](https://peps.python.org/pep-0621/) and [PEP 660](https://peps.python.org/pep-0660/)).

This would lead to a overall minimal file structure (one `pyproject.toml` instead of (`setup.cfg` + `setup.py`))

Changes proposed by this PR can be summarised as follows :-

* drop `setup.cfg` and `setup.py` in favour of `pyproject.toml` for packaging.
* update `lint` and `pre-commit` to add [`Flake8-pyproject`](https://pypi.org/project/Flake8-pyproject/) as a dependency to configure `flake8` from within the `pyproject.toml` file itself.


luisfpereira:
I'm not fully sure about this. The main scientific computing libs (`numpy`, `jax`, `sklearn`, ...) are still using `setup.cfg` and `setup.py` (or in the limit a combination of these with a `project.toml`).

I don't mind adopt a standard if the community agrees, but I would like to keep things similar to these projects to ensure people used to them do not have additional friction to use `geomstats`.

@SauravMaheshkar, do you have any example of a reference scientific computing package using uniquely `project.toml`?

luisfpereira:
Thanks @SauravMaheshkar.

What do you think @ninamiolane? Do you want to move on with it?

luisfpereira:
Thanks again @SauravMaheshkar, I'm bring it it!

luisfpereira (approved):

SauravMaheshkar:
> Thanks again @SauravMaheshkar, I'm bring it it!

Thank you @luisfpereira Looking forward to contributing more

PR #1845, opened 2023-05-02
Numerical solvers for log, exp, geodesic, ...

luisfpereira:
Follows the discussion in #1552. Tests have been added in #1767. Joint work with @tramy1258. Nothing is changed in `geomstats` files, with exception of `geomstats.integrator`.

This PR brings the notion of `solver` to `geomstats`. In general terms, a solver is an object capable of solving a numerical problem (ODE - BVP and IVP-, optimization, ...).  The main goal is to have an object that stores parameters (usually numerical parameters) and then a main function with a simple signature that solves the problem. This way:

* we can easily interface with other libraries (e.g. for optimization) while keeping our function signatures clean.
* we have full control when calling a method (e.g. `exp`) within a learning method
* we can centralize the use of `autodiff` for functions optimization
* we can perform all the array conversions we need in only one place (e.g. numpy -> pytorch -> numpy) (and moreover test only once)

For optimization we've implemented `ScipyMinimize` (wraps `scipy.optimize.minimize`).

For boundary value problems we've implemented `ScipySolveBVP` (wraps `scipy.integrate.solve_bvp`).

For initial value problems (`ODEIVPSolver`) we've implemented `GSIntegrator` (uses the machinery available in `gs.integrator`) and `ScipySolveIVP` (wraps `scipy.integrate.solve_ivp`).

For the solution of `exp` and `geodesic_ivp`, we've implemented `ExpIVPSolver` (relies on `ODEIVPSolver` - therefore we can use our in-house IVP solvers or `scipy`).

For the solution of `log` and `geodesic_bvp`, we've implemented `LogShootingSolver` (relies on `ScipyMinimize`), `LogBVPSolver` (relies on `ScipySolveBVP`), and `LogPolynomialApproxSolver` (relies on `ScipyMinimize`).

Different logs solvers can be combined (e.g. use `LogPolynomialApproxSolver` to provide an initial solution to a `LogBVPSolver`).

After merging this, the idea is to replace any use of numerical methods in the library by these objects.

The PR is mostly ready to merge. There's only a bug we need to (find and) fix in `LogPolynomialApproxSolver`, but I think we can start discussing it.


alebrigant (approved):
Thank you for this amazing work, this is super clean and will clarify so much !! I just have some comments about the names, but I don't have a strong opinion on this, so they are just suggestions open for debate.

- The names for the scipy wrappers are chosen to fit the scipy names (e.g. `ScipySolveBVP`), which in a sense is practical. But then they don't fit so well with the rest of the nomenclature. Would `ScipyIVPIntegrator`, `ScipyBVPIntegrator`, `ScipyMinimizer` be more coherent ? And `GSIntegrator` could become `GSIVPIntegrator`. Then for the users, choosing `integrator=ScipyIVPIntegrator` might make more sense ?
- I see that you have an abstract class `ExpSolver` from which `ExpIVPSolver` inherits, even though there is only one of those (contrary to `LogSolver` which has 2 child classes `LogShootingSolver` and `LogBVPSolver`). Similarly, would it be pertinent to create an abstract class `ODEBVPSolver`, just like there is an abstract `ODEIVPSolver`, even though for now there will be only one child class `ScipyBVPIntegrator` (currently `ScipySolveBVP`) ?
- I would be a little more enclined to name `ExpODESolver` and `LogODESolver` instead of `ExpIVPSolver` and `LogBVPSolver`, because by definition exp always solves a geodesic initial value problem  and log a boundary value problem, and replacing that by ODE might give more information on the method. What do you think ?

luisfpereira:
Thanks for the feedback @alebrigant. I agree with everything (I'll change it accordingly) except with the first point.

For that, I understand the names within `geomstats` would become more consistent, but at the expense of clarity for an experienced `scipy` user. Since these objects only wrap specific `scipy` functions, having their name on the object name seems very important to me.

PR #1848, opened 2023-05-05
Fixes and improvements coming from test refactoring (part 3)

luisfpereira:
Follows #1828.

Main additions are:

* `repeat_out`: quick way of repeating an output for vectorization consistency when some of the inputs are not used in the computation

* `AutodiffNotImplementedError` exception: will be used later in the tests in a `try...except` statement to avoid having to skip tests that depend in autodiff when testing for numpy.

Several vectorization inconsistencies are also fixed.


geomstats/_backend/numpy/autodiff.py
```diff
@@ -4,6 +4,8 @@
 The following functions return error messages.
 """
 
+from geomstats.exceptions import AutodiffNotImplementedError
```
ninamiolane, line 7: Why do we need a special exception, since the MSG is not the exception file?

Either:
1. Put the message in exception file
OR:
2. Only use RunTimeError with this MSG?
I'd prefer the second one, as it is easier for the reader.

If we keep that tailored exception, maybe change the name to:
UseAutodiffBackend?
Or
AutodiffNotInNumpy?

geomstats/exceptions.py
```diff
@@ -0,0 +1,5 @@
+"""Geomstats custom exceptions."""
+
+
+class AutodiffNotImplementedError(RuntimeError):
+    """Raised when autodiff is not implemented."""
```
ninamiolane, line 5: I prefered the RuntimeError option, which seems less "over-engineering", but there might be something I am missing there?

geomstats/geometry/euclidean.py
```diff
@@ -85,10 +86,9 @@ def metric_matrix(self, base_point=None):
         inner_prod_mat : array-like, shape=[..., dim, dim]
             Inner-product matrix.
         """
-        mat = gs.eye(self._space.dim)
-        if base_point is not None and base_point.ndim > 1:
-            mat = gs.broadcast_to(mat, base_point.shape + (self._space.dim,))
-        return mat
+        dim = self._space.dim
+        mat = gs.eye(dim)
+        return repeat_out(self._space, mat, base_point, out_shape=(dim, dim))
```
ninamiolane, line 91: NIT: mat could be called `out` for consistency with how the function repeat_out is used elsewhere.

Alternatively (probably preferred), avoid using "out" as a variable and call each variable by its mathematical interpretation?

geomstats/geometry/grassmannian.py
```diff
@@ -117,7 +118,7 @@ def _squared_dist(point_a, point_b, metric):
     _ : array-like, shape=[...,]
         Geodesic distance between point_a and point_b.
     """
-    return metric.private_squared_dist(point_a, point_b)
+    return metric._squared_dist(point_a, point_b)
```
ninamiolane, line 121: Ah nice! How did you solve this? The introduction of the very ugly "private_squared_dist" was to get the automatic differentiation with custom gradient work with all backends.

Maybe it was not needed anymore since we dropped TF?

ninamiolane (commented):
A first batch of comments!

geomstats/geometry/invariant_metric.py
```diff
@@ -49,8 +50,9 @@ def __init__(self, space, metric_mat_at_identity=None, left=True):
         self.metric_mat_at_identity = metric_mat_at_identity
         self.left = left
 
-    def reshape_metric_matrix(self):
-        """Reshape diagonal metric matrix to a symmetric matrix of size n.
+    @property
+    def reshaped_metric_matrix(self):
+        """Diagonal metric matrix reshaped to a symmetric matrix of size n.
```
ninamiolane, line 55: Does the rule "Docstring starts with verb at infinitive" not apply for @property ?

geomstats/exceptions.py
luisfpereira: see above.

geomstats/geometry/pre_shape.py
```diff
@@ -753,16 +756,14 @@ def log(self, point, base_point, **kwargs):
             Tangent vector at the base point equal to the Riemannian logarithm
             of point at the base point.
         """
-        sphere = self._space._sphere
-        sphere_embedding_dim = sphere.embedding_space.dim
-        flat_bp = gs.reshape(base_point, (-1, sphere_embedding_dim))
-        flat_pt = gs.reshape(point, (-1, sphere_embedding_dim))
-        flat_log = sphere.metric.log(flat_pt, flat_bp)
+        batch_shape = get_batch_shape(self._space, point, base_point)
```
ninamiolane, line 759: Why "get_batch_shape"? and not "batch_shape". Is this because it is a defined as a property somewhere?

Even in that case, I don't think that we use "get" for the other properties defined in the library: I would remove to keep the coding style consistent, and conciseness.

https://stackoverflow.com/questions/374763/should-i-use-get-set-prefixes-in-python-method-names

ninamiolane: And we generally don't use "get" in the whole codebase, thus if we can avoid it altogether, it'll be better.

geomstats/geometry/grassmannian.py
luisfpereira: here is really just naming. I'm representing private by `_`. But nothing changes from a logical perspective.

geomstats/geometry/riemannian_metric.py
```diff
@@ -304,7 +304,8 @@ def normalize(self, vector, base_point):
         """
         norm = self.norm(vector, base_point)
         norm = gs.where(norm == 0, gs.ones(norm.shape), norm)
-        return gs.einsum("...i,...->...i", vector, 1 / norm)
+        indices = "ijk"[: self._space.point_ndim]
```
ninamiolane, line 307: Smart!!

geomstats/geometry/spd_matrices.py
```diff
@@ -848,13 +849,15 @@ def squared_dist(self, point_a, point_b, **kwargs):
         trace_b = gs.trace(point_b)
         trace_prod = gs.trace(sqrt_product)
 
-        return trace_a + trace_b - 2 * trace_prod
+        squared_dist = trace_a + trace_b - 2.0 * trace_prod
+
+        return gs.where(squared_dist < 0.0, 0.0, squared_dist)
```
ninamiolane, line 854: Why would the squared_dist be < 0? 

Are we sure that it is a numerical issue, and not a problem in the math in the code?

If it is the later, then the gs.where would hide that bug, that would otherwise be caught by unit tests.

geomstats/vectorization.py
```diff
@@ -106,6 +111,36 @@ def repeat_point(point, n_reps=2, expand=False):
     return gs.repeat(gs.expand_dims(point, 0), n_reps, axis=0)
 
 
+def _is_not_none(value):
```
ninamiolane, line 114: Why do we need this?

geomstats/vectorization.py
```diff
@@ -106,6 +111,36 @@ def repeat_point(point, n_reps=2, expand=False):
...
+def repeat_out(space, out, *points, out_shape=()):
+    """Repeat out shape after finding batch shape.
+
+    Parameters
+    ----------
+    space : Manifold
+        Space to which point belongs.
+    out : array-like
+        Output to be repeated
+    point : array-like
```
ninamiolane, line 127: points?

geomstats/vectorization.py
```diff
@@ -106,6 +111,36 @@ def repeat_point(point, n_reps=2, expand=False):
...
+        Point belonging to the space.
+    out_shape : tuple
+        Indicates out shape for no batch computations.
+
+    Returns
+    -------
+    out : array-like
+        If no batch, then input is returned. Otherwise it is broadcasted.
+    """
+    points = filter(_is_not_none, points)
```
ninamiolane, line 137: I have trouble reading through this because some private functions do not have docstrings (as is conventional) and the public functions have short docstrings.

Any chance you could add details + 1-2 examples in the docstrings, even if private ones, to explain the logic?

geomstats/geometry/pre_shape.py
luisfpereira: this comes from another PR. see [our discussion][https://github.com/geomstats/geomstats/pull/1828#discussion_r1112743175]


I still think having a verb is better for a function (for me a name represents an object, not a callable).  also, as you see in this example, I do ` batch_shape = get_batch_shape(self._space, point, base_point)`. if the function didn't have the verb, I would have to do  `batch_shape_ = batch_shape(self._space, point, base_point)` which is more cumbersome.

maybe we can change this in another PR if you really don't want it, since it is unrelated.

geomstats/vectorization.py
ninamiolane: I was also confused in the _get_max_ndim and the line point_max_ndim = point[0]

--> why is the ndim an element of point? 

docstrings would help for this.

geomstats/vectorization.py
```diff
@@ -106,6 +111,36 @@ def repeat_point(point, n_reps=2, expand=False):
...
+    batch_shape = get_batch_shape(space, *points)
```
ninamiolane, line 138: Why the "get_" wording?

tests/data/sasaki_metric_data.py
```diff
@@ -26,7 +26,9 @@ class SasakiMetricTestData(TestData):
     def inner_product_test_data(self):
         _sqrt2 = 1.0 / gs.sqrt(2.0)
         base_point = gs.array([[_sqrt2, -_sqrt2, 0], [_sqrt2, _sqrt2, 1]])
-        _log = self.sas_sphere_metric.log(gs.array([self.pu0, self.pu1]), base_point)
+        end_point = gs.stack([self.pu0, self.pu1])
```
ninamiolane, line 29: NIT: what is pu0, pu1? what is sqrt2? the names could be more self explanatory.

ninamiolane (approved):
Awesome!

There seem to be vectorization errors in the examples: fix these?

Another high level comments:
- Add small explanations for vectorization logics in docstrings
- Why "get" and "_get"? If we can avoid using "get" that could be great, otherwise what is the clear logic jsutifying its use? Anything that is "engineering related" and not "math related"?

geomstats/geometry/spd_matrices.py
luisfpereira: On my tests the negative values were always super close to zero (e.g. -1e-12). the problem seems to be the tolerance in `gs.linalg.sqrtm`.

Your point is very good though!

geomstats/vectorization.py
luisfpereira: good catch. in fact the parameter should be `point` for consistency with other functions, so I'll change it instead.

geomstats/vectorization.py
luisfpereira: see above.

geomstats/vectorization.py
luisfpereira: I'll improve the docstrings.

The main idea is that I want to have these methods working for the cases where `point` is `None`, because in a lot of our methods we have something like `base_point=None` and then I want to use something like `repeat_out(self.space, out, base_point)` without having to check if there's `None` (it simplifies a lot the vectorization logic there).

therefore, before checking `batch_shape`, I remove the `None` from the list of points (that's the role of `filter`).

tests/data/sasaki_metric_data.py
luisfpereira: This comes from a very old PR. I'll take that into account in the new tests.

PR #1853, opened 2023-05-24
Remove code duplication between numpy and autograd backends

luisfpereira:
This PR removes code duplication for `numpy`-based backends (for now only `numpy` and `autograd`, but hopefully soon we will have more - thinking about `jax`). This will simplify backend development and will ensure consistency by design among `numpy`-based backends.

The technical solution is based on dispatch at import time. In other words, the `numpy` package used in the folder `_shared_numpy` changes accordingly to the backend in use.

All the functions need to be imported in the respective backend code (as later the importer only looks to that folder). The imports are a bit tricky (to avoid circular imports), but are principled and easy to maintain after understood. Functions in `_shared_numpy` correspond to the `numpy` implementation (i.e. they're  not the intersection of common implementations). If a implementation does not work for a given backend, then we just need to implement it in the backend and  do not import the `_shared_numpy` one.

Let's take `autograd` as example to understand the imports structure:

* `_shared_numpy/_common` cannot import from anywhere in the backend

* `autograd/_common` imports from `_shared_numpy/_common`

* `_shared_numpy/__init__` imports from  `autograd/_common` : it is the way to control we use the proper functions when there's some different implementations

*  `autograd/__init__` imports from `autograd/_common` and `_shared_numpy/__init__`

The main motivation for this PR comes from the need to improve some methods in `gs.linalg`. The current code duplication is very cumbersome and hard to maintain.


luisfpereira:
Thanks @ninamiolane! I'm keeping `_shared_numpy` for now because the purpose is really to use it for any numpy-based backend, as a lot of things will be the same. We can adapt in the future if this is not true.

PR #1854, opened 2023-05-25
Bump python 3.11

luisfpereira:
Made the little changes required to run the package in 3.11 (there's an issue with `joblib` we may need to come back later). CI workflow runs completely in 3.11 now.


ninamiolane:
Thank you so much for this! Why are some tests failing?

ninamiolane (approved):
Thanks!

PR #1857, opened 2023-05-25
Fix ProductRiemannianMetric

johnharveymath:
Add to `ProductRiemannianMetric` a method `_pool_outputs_from_functio…n` (which wraps the method from the `ProductManifold`) to correctly structure the output of `exp` and `log`.

## Issue

As previously implemented, this did not correctly concatenate the output of `exp` and `log` if they had different shapes. They need to be flattened first.

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -7,7 +7,9 @@ import math
 import geomstats.backend as gs
 import geomstats.errors
 from geomstats.geometry.complex_manifold import ComplexManifold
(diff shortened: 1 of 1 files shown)


johnharveymath:
This still needs an amendment to tests to check more combinations of factors. Factors of different point_type, factors of different shapes, etc.

johnharveymath:
I have been having some issues with the pre-commit hook. This second commit is just intended to remove all the changes that are not related to the code.

luisfpereira:
Closing this as @Abdellaoui-Souhail continued it in #1862.

PR #1862, opened 2023-05-31
Fix product metric

Abdellaoui-Souhail:
## Checklist

- [ ] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/master/docs/contributing.rst#coding-style-guidelines) and API.

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -679,14 +679,12 @@ class ProductRiemannianMetric(_IterateOverFactorsMixins, RiemannianMetric):
     
     def geodesic(self, initial_point, end_point=None, initial_tangent_vec=None, **kwargs):
         args = {"initial_point": initial_point, "end_point": end_point, "initial_tangent_vec":initial_tangent_vec} | kwargs
-        geodesics = gs.array(self._iterate_over_factors("geodesic", args))
-        dict = {factor:geodesic for factor,geodesic in zip(self._space.factors, geodesics)}
+        geodesics = self._iterate_over_factors("geodesic", args)
         
         def geod_fun(t):
             t = gs.to_ndarray(t, to_ndim=1)
             result = []
-            for factor in self._space.factors:
-                geodesic = dict[factor]
+            for geodesic in geodesics:
                 value = geodesic(t)
                 value = gs.reshape(value, (len(t),math.prod(value.shape[1:])))
                 result.append(value)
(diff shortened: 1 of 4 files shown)


luisfpereira:
@johnharveymath, @Abdellaoui-Souhail I've done small updates on this, by taking advantage of some methods of the space when performing reshapings in the metric.

@Abdellaoui-Souhail, by last suggestion to finish this is to change the `default_point_list` on the tests you have added to the same values as in `ProductManifold`. This way, we can cover more use cases as @johnharveymath suggested in #1857.

geomstats/geometry/product_manifold.py
```diff
@@ -668,27 +655,67 @@ def log(self, point, base_point=None, **kwargs):
...
+            Time parameterized geodesic curve. If a batch of initial
+            conditions is passed, the output array's first dimension
+            represents the different initial conditions, and the second
+            corresponds to time.
+        """
+        args = {
+            "initial_point": initial_point,
+            "end_point": end_point,
+            "initial_tangent_vec": initial_tangent_vec,
+        }
```
Abdellaoui-Souhail, line 713: Sometimes we may want to add additional arguments to geodesic. For example in SRVQuotientMetric the geodesic and distance can be computed through 2 different methods ( "iterative horizontal projection" or "dynamic programming" ) and each method needs it's own parameters to be specified. So, for example if we want to compute a geodesic (or dist) of a ProductMetric where one of the metrics is SRVQuotientMetric we may want to specifie some parameters that will need to be passed on to each geodesic. That's why kwargs was added to args

luisfpereira: We are trying to homogenize all signatures of methods such as `exp`, `log`, `geodesic`, and so on, to receive only the arguments that are shared among all manifolds (which means only points and tangent vectors).

Other behaviors should be controlled at init. The strategy we are using now is to create different objects implementing the different ways of solving a problem, which we are calling solvers (see e.g. #1860, where before we were passing arguments to `log` and `exp` in e.g. `DirichletMetric` and now we don't do it anymore).

Another strategy can be to use a variable at instantiation choosing the strategy (I would suggest this to be temporary though), and then use an if.

This is super important because learning methods only pass tangent vectors and points to these methods, so it is the only way to control how to compute a given quantity when using a space within them.

Abdellaoui-Souhail: Okey I see. Thanks for the explanation !

luisfpereira (approved):

PR #1863, opened 2023-06-01
Add Diffeo abstraction

luisfpereira:
This PR brings the notion of `Diffeo`. This intends to simplify `PullbackDiffeoMetric`, as well as make it more logically sound (both mathematically and computationally). 

In the current implementation everything that  is related with the diffeomorphism (meaning moving points and tangent vectors from one space to the other, in both directions) is implemented within `PullbackDiffeoMetric`. I argue these operations can live by themselves and are independent of `PullbackDiffeoMetric` (this PR moves them to a new class called `Diffeo`). On the other hand, `PullbackDiffeoMetric` needs access to this object.

I've changed the code in `UnivariateNormalMetric` (may interest you @alebrigant and @tramy1258), and `ElasticMetric` and `SRVMetric` (may interest you @alebrigant) to be according to my proposal.

This also simplifies testing, as `PullbackDiffeoMetric` only defines methods that we clearly associate to a metric (`exp`, `log`, `dist`, etc), and nothing related with point transformations.

I think this also interests you @ymontmarin.

This is just a prototype (I haven't changed the tests yet, so their failing). Looking forward for your feedback.

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -15,7 +15,7 @@ from geomstats.geometry.euclidean import Euclidean
 from geomstats.geometry.fiber_bundle import FiberBundle
 from geomstats.geometry.landmarks import Landmarks
 from geomstats.geometry.manifold import Manifold
-from geomstats.geometry.pullback_metric import PullbackDiffeoMetric
+from geomstats.geometry.pullback_metric import Diffeo, PullbackDiffeoMetric
 from geomstats.geometry.quotient_metric import QuotientMetric
 from geomstats.geometry.riemannian_metric import RiemannianMetric
 from geomstats.geometry.symmetric_matrices import SymmetricMatrices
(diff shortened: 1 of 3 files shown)


luisfpereira:
@ninamiolane, I would like to move on with this idea as we are bringing in several `PullbackDiffeoMetric` that would benefit from this design. Do you think the design is proper? Can I move on and fully implement it everywhere it is required?

ymontmarin:
@luisfpereira I just looked, and the design seems coherent with our last discussions on this subject. I am on holiday, and I will be back on the 1st of August and will be available for review if you need me. I did not see your previous invitation; I think it has expired.

geomstats/geometry/discrete_curves.py
```diff
@@ -1096,6 +1051,59 @@ def inverse_diffeomorphism(self, image_point):
...
+
+    def __init__(
+        self,
+        space,
+        a,
+        b,
+    ):
+        self._check_ambient_manifold(space.ambient_manifold)
+
+        diffeo = FTransform(
```
ymontmarin, line 1085: The reading is simple, it is great

luisfpereira:
Closing as new version as been developed more carefully and thoroughly in #1904.

PR #1864, opened 2023-06-12
Update connection to use solvers

luisfpereira:
This PR changes `Connection` to make use of numerical solvers.

Additionally:

* it adds matrix point-representation support for `numerics` `ExpSolver` and `LogShootingSolver`

* makes use of the numerical solvers in `_InvariantMetricMatrix`

* cleans `Gamma` and `Dirichlet`: now `Connection` implements `log`, `exp`, and `geodesic` as they expect

Notice behavior of `geodesic` in `Connection` was kept: if no solvers are defined, then `geodesic` is found by doing `log` + `exp` for different `t`. This is specially useful when there's closed form solutions for `exp`.


ninamiolane:
Is it normal that the tests are failing?

luisfpereira:
Missed some changes @ninamiolane . Fixed it now.

Everything should be fine now. This is the PR I wanted to merge before releasing. After this, we can release from my side.

ninamiolane:
Some tests are still failing: should numerical tolerances be changed maybe?

PR #1866, opened 2023-06-20
Prepare release 2.6.0

luisfpereira:
This PR prepares new release.

@ymontmarin, what should we do for the conda release?

PR #1867, opened 2023-06-21
Fix test_connection due to use of solvers

luisfpereira:
This is a quick fix to `test_connection` due to use of solvers (related with #1864). Basically, we need to use the `equip_with_metric` strategy to test the connection, otherwise we don't use e.g. the proper exp when solving log with shooting method.


luisfpereira:
@ninamiolane, I've added a comment on what you've mentioned. Regarding the other errors, they are now "solved" after #1866.

ninamiolane (approved):
LGTM! Thanks!

PR #1872, opened 2023-06-28
Add invariant metric for positive lower triangular matrices 

olivierbisson:
In this PR we implement an invariant metric for positive lower triangular matrices. Additionally, we endow the PositiveLowerTriangularMatrices with a LieGroup structure and a Lie algebra.

### DIFF
--- a/geomstats/geometry/lower_triangular_matrices.py
+++ b/geomstats/geometry/lower_triangular_matrices.py
@@ -93,6 +93,21 @@ class LowerTriangularMatrices(VectorSpace):
         """
         return gs.einsum("...i,...ijk->...jk", vec, self.basis)
 
+    def matrix_representation(self, vec):
+        """Compute the matrix representation for the given basis coefficients.
+
+        Parameters
+        ----------
+        matrix_representation : array-like, shape=[..., n, n]
+            Matrix.
+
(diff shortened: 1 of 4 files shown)


luisfpereira:
@ninamiolane, we've added the tests for the metric in the new framework. I think it is not relevant to add them in the previous one because this metrics only slightly changes the definition of the `inner_product_at_identity`.

ninamiolane (changes requested):
Looking great! However, each new function needs to come with its tests. May I let you coordinate with @luisfpereira to add corresponding tests? Thanks!

luisfpereira:
@ninamiolane we are already coordinating. I've added tests for this in the new framework (everything works fine). I would say it is not worthy to add them in the current framework because we actually have a lot of tests for the `InvariantMetric` already, and this metric basic relies on those methods with a small variant on the `inner_product_at_identity` to cover the cases where the `matrix_mat_at_identity` is not diagonal.

Moreover, the method added to `LowerTriangularMatrices` is just due to the `inner_product_at_identity` of `InvariantMetric` and simply calls an already existing method with another name.

luisfpereira (approved):

luisfpereira (approved):

PR #1873, opened 2023-06-29
Refactor learning module

luisfpereira:
TODO: update

### DIFF
--- a/geomstats/learning/frechet_mean.py
+++ b/geomstats/learning/frechet_mean.py
@@ -231,6 +231,8 @@ class BatchGradientDescent(BaseGradientDescent):
         iteration = 0
         convergence_old = convergence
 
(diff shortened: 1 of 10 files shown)


ninamiolane (approved):
So much cleaner and interpretable, thank you so much!!

Two remarks:
- the optimizers (e.g. BaseGradientDescent) are currently in the files where they're used (e.g. in frechet_mean): since the logic of optimization is now abstracted, would it make sense to have a separate folder, in the spirit of what you did for the numerics folders?
- pymanopt provides optimizers: could we make our abstraction of optimizers compatible with theirs, so that we can use their optimizers on demand?

These two remarks are not necessarily for this PR, we can talk about them in another one, if they make sense.

Thanks again!!

PR #1875, opened 2023-07-12
Fix complex data type in squared_dist, norm and squared_norm in the manifold ProductPositiveRealsAndComplexPoincareDisks

YannCabanes:
The methods `squared_dist`, `norm` and `squared_norm` of the class `ProductPositiveRealsAndComplexPoincareDisksMetric` return complex data type.
This PR fixes this issue.

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -617,6 +617,32 @@ class ProductRiemannianMetric(_IterateOverFactorsMixins, RiemannianMetric):
         inner_products = self._iterate_over_factors("inner_product", args)
         return sum(inner_products)
 
+    def squared_norm(self, vector, base_point=None):
+        """Compute the square of the norm of a vector.
+
+        Squared norm of a vector associated to the inner product
+        at the tangent space at a base point.
+
+        Parameters
+        ----------
+        vector : array-like, shape=[..., self.shape]
+            Vector.
+        base_point : array-like, shape=[..., self.shape]
+            Base point.
+            Optional, default: None.
+
+        Returns
+        -------
+        sq_norm : array-like, shape=[...,]
+            Squared norm.
+        """
+        args = {
+            "vector": vector,
+            "base_point": base_point,
+        }
+        sq_norms = self._iterate_over_factors("squared_norm", args)
+        return sum(sq_norms)
+
     def exp(self, tangent_vec, base_point):
         """Compute the Riemannian exponential of a tangent vector.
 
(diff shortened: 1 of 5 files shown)


luisfpereira (changes requested):
Thanks @YannCabanes.

I think the problem is not in `ProductPositiveRealsAndComplexPoincareDisksMetric`, but in `ProductManifold` or `ComplexPoincareDisk`. Can you please identify the root cause?

YannCabanes:
Hello @luisfpereira,
I have checked the methods `dist`, `squared_dist`, `norm`, `squared_norm` of the class` ComplexPoincareDiskMetric`, they all return float values. The problem does not come from here.

The class `ProductPositiveRealsAndComplexPoincareDisksMetric` inherits from `ProductRiemannianMetric` which inherits from `RiemannianMetric` in which the `squared_norm` is defined as:
`self.inner_product(vector, vector, base_point)`
instead of:
`gs.real(self.inner_product(vector, vector, base_point))`.
Indeed, the `inner_product` of a complex manifold is complex but the `squared_norm` should be real.

In my opinion, it is cleaner to modify the method `squared_norm` of the class `RiemannianMetric` than to redefine the method `squared_norm` in the class `ProductPositiveRealsAndComplexPoincareDisksMetric`.

Do you allow be to replace 
`self.inner_product(vector, vector, base_point)` 
by `
gs.real(self.inner_product(vector, vector, base_point))`
in the method `squared_norm` of the class `RiemannianMetric`?

I would then remove the method `squared_norm` from the class `ProductPositiveRealsAndComplexPoincareDisksMetric`.

luisfpereira:
Thanks for presenting different solutions @YannCabanes. I think we should go for 2., as it takes advantage of the work we've done in `ProductManifold` and keeps the solution for the problems in the place they arise.

YannCabanes:
Hello @luisfpereira, I think that this PR is ready to be merged. Could you check please?

YannCabanes:
All tests have passed except:
`Testing / build (ubuntu-latest, 3.11, pytorch, tests/tests_geomstats/) (pull_request) `
which is still in progress after a few hours.

luisfpereira:
Thanks @YannCabanes. Probably related with triggering of tests due to lack of resources. In the it says it took a few minutes.

Looks good to me!

luisfpereira (approved):

luisfpereira:
I'll just adapt some tests tolerances and merge after.

YannCabanes:
Perfect, thanks a lot @luisfpereira

PR #1876, opened 2023-07-12
Fix minor error in variable name in method squared_dist of the class ComplexPoincareDiskMetric

YannCabanes:
Fix minor error in variable name in method `squared_dist` of the class `ComplexPoincareDiskMetric`: the variable name `sq_dist` is used instead of `dist`.


luisfpereira (approved):
Looks good. Merging.

PR #1878, opened 2023-07-26
Fix problem in selection of the principal directions in Tangent PCA.

YannCabanes:
This PR is related to PR https://github.com/geomstats/geomstats/pull/1513/ in which a solution is proposed.

The current Tangent PCA algorithm does not necessarily select the axis which corresponds to the largest variance.
The problem is that the current Tangent PCA does not take into account the metric at the mean to sort the principal directions.

To show this problem, I have created an almost Euclidean manifold (e488adb8b6bd8e6c956337f657817ddb1b525fad).
This manifold is the Euclidean space of dimension 2 whose metric has been modified to give more weight to the second axis. It's infinitesimal metric element is defined by: `ds^2 = dx1^2 + 10000 dx2^2`
Then if we run the following codes (e488adb8b6bd8e6c956337f657817ddb1b525fad):

```
almost_euclidean_dim_two = AlmostEuclideanDimTwo()
tpca = TangentPCA(space=almost_euclidean_dim_two, n_components=1)
point_a = gs.array([10, 0], dtype=float)
point_b = gs.array([0, 1], dtype=float)
X = gs.stack([point_a, -point_a, point_b, -point_b], axis=0)
mean = gs.array([0, 0], dtype=float)
tpca.fit(X=X, base_point=mean)
variance_axis_1 = almost_euclidean_dim_two.metric.squared_dist(mean, point_a)  # variance_axis_1 = 100
variance_axis_2 = almost_euclidean_dim_two.metric.squared_dist(mean, point_b)  # variance_axis_1 = 10000
assert variance_axis_2 >= variance_axis_1 # True
axis_1 = gs.array([1, 0])
axis_2 = gs.array([0, 1])
assert gs.all(tpca.components_ == axis_1) # True
```

The tangent PCA returns `axis_2` as the principal axis since it is the axis which 
corresponds to the larger variance. It is returning `axis_1` instead.

To solve this problem, we can multiply the `log` vectors at the `mean` by square root of the `metric_matrix` to select the correct principal directions (see PR https://github.com/geomstats/geomstats/pull/1513/).

### DIFF
--- a/geomstats/learning/pca.py
+++ b/geomstats/learning/pca.py
@@ -3,9 +3,11 @@
 Lead author: Nina Miolane.
 """
 
+import logging
 import numbers
 from math import log
 
+from scipy.linalg import pinvh
 from scipy.special import gammaln
 from sklearn.decomposition._base import _BasePCA
 from sklearn.utils.extmath import stable_cumsum, svd_flip
@@ -188,10 +190,8 @@ class TangentPCA(_BasePCA):
         """
         U, S, _ = self._fit(X, base_point=base_point)
 
-        U = U[:, : self.n_components_]
-
-        U *= S[: self.n_components_]
-        return U
+        # return self.transformed_log_in_riemannian_basis
+        return self.coordinates_on_components
 
     def transform(self, X, y=None):
         """Project X on the principal components.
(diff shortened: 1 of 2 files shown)


YannCabanes:
Hello @luisfpereira and @ninamiolane,
Do you agree that the current behavior of the tangent PCA described above is not the expected one?
(On the example in dimension 2, it does not rank first the direction with the largest variance.)

YannCabanes:
The tests pass except `Linting`, `Testing / build (ubuntu-latest, 3.11, autograd` (errors not related to this PR) and `DeepSource`.

PR #1881, opened 2023-08-03
Fix bug in both online kmeans examples

andreakiro:
A very short pull request to correct a minor bug in example files. This will make it easier for the next ones testing it :)

### DIFF
--- a/examples/plot_online_kmeans_s1.py
+++ b/examples/plot_online_kmeans_s1.py
@@ -21,7 +21,7 @@ def main():
     data = circle.random_uniform(n_samples=1000)
 
     n_clusters = 5
-    clustering = OnlineKMeans(circle, n_clusters=n_clusters)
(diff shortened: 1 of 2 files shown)


ninamiolane:
@andreakiro this is not a bug, this how the line should be written with the latest version of geomstats, i.e. the master branch of the github repository. If you downloaded geomstats via pip, you might have a bug until we put the latest geomstats version on pypi.

@luisfpereira you confirm?

luisfpereira:
Exactly, @ninamiolane.

andreakiro:
Yes, indeed i've installed the package with pip!

luisfpereira:
You're fully right @andreakiro: last version of `pip` still does not have this. By last version we are thinking `master` branch.

@ninamiolane, do you think the recent changes are worthy enough to justify a new release soon? From my side, I think we could probably release given all the changes in `learning`... (I would suggest as release date somwhere before next gsi)

PR #1885, opened 2023-09-06
Fixes and improvements coming from test refactoring (part 4) 

luisfpereira:
Follows #1848. It consists of several small fixes that intend to make the library more consistent and remove bugs. Most of these fixes were identified while developing #1767.


ninamiolane (changes requested):
YESSSS! Excited for this. There is the linting error on the numerics file and some testing errors that should be fixed.

PR #1890, opened 2023-10-05
Fix and clean information geometry

luisfpereira:
This PR fixes several issues in information geometry (most of them related with vectorization) and cleans several parts of the code by taking more advantage of inheritance and/or new vectorization functions.

The main addition is the notion of `RandomVariable`, in particular `ScipyRandomVariable`. This comes from the fact that the `InformationManifold` method `sample` uses `scipy.stats.<distribution>.rvs`. We just need to "translate" the inputs from geomstats to `scipy`, which can be done in a generic way (or at least in two: one for univariate distributions, and another for multivariate). We also take advantage of the `scipy.stats.<distribution>.pdf`: for some manifolds we directly use this method, for others we use it to validate our implementations (in new tests framework - coming very soon).

PR #1894, opened 2023-10-24
Fix and clean learning module

luisfpereira:
Similar to #1885 and #1890, but for learning module. Makes things more consistent.


examples/learning_graph_embedding_and_predicting.py
```diff
@@ -56,7 +56,7 @@ def main():
     )
 
     kmeans.fit(X=embeddings)
-    centroids = kmeans.centroids_
+    centroids = kmeans.cluster_centers_
```
ninamiolane, line 59: NIT: in that case, maybe `cluster_centers = kmeans.cluster_centers_` would make more sense.

ninamiolane (changes requested):
Need to fix tests, e.g.

AttributeError: 'RiemannianKMeans' object has no attribute 'centroids_' from the test file, following the change of name.

PR #1896, opened 2023-10-25
Use "remainder" instead of "fmod" for mod for pytorch backend

FlorianPfaff:
Hi, 

I didn't check all boxes, but this is just a minor change. I have created an extended version of the _backend of geomstats for my [pyRecEst](https://github.com/FlorianPfaff/pyRecEst) library and now I want to see if you are interested in "mainlining" some fixes.

This fix concerns backend.mod. Currently, mod uses pytorch's fmod. However torch.fmod behaves like np.fmod. np.fmod and np.mod do not behave identically. A key difference is the handling of negative values. Hence, I changed torch.mod to use torch.remainder instead, which behaves like np.mod.

### DIFF
--- a/geomstats/_backend/pytorch/__init__.py
+++ b/geomstats/_backend/pytorch/__init__.py
@@ -107,7 +107,7 @@ tanh = _box_unary_scalar(target=_torch.tanh)
 
 
 arctan2 = _box_binary_scalar(target=_torch.atan2)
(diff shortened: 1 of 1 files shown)


luisfpereira:
Thanks @FlorianPfaff! We are very interested in "mainlining" your fixes.

P.S. some of the tests will fail, but it is probably unrelated with your pR (we're fixing it in #1896).

luisfpereira (approved):

PR #1897, opened 2023-10-26
backend: Added "endpoint" support to linspace #1

FlorianPfaff:
Hi. I just saw my other change got mainlined so I was looking into other changes. However, since the policy is apparently only commands that are used in geomstats should be supported on the backend (which I can understand), there are not so many other changes from my side. One additional thing I have is the support for endpoint=False in linspace. If you think that is any useful to you, feel free to merge.
Sidenote: I did not treat the "not (start_is_array or stop_is_array)" case differently. I think you are trying to see if start and stop are scalars - however, this case is also used when one inputs a list. For numpy, this works just fine; for pytorch, it doesn't. To make it work regardless, I just treated it as the (potentially) vector-valued case.

Considering some test cases fail anyway, I am not sure if I should run all tests for this PR.


luisfpereira:
Thanks again @FlorianPfaff. In fact, for now the philosophy is to support only what we need. If interest in the backend starts increasing (e.g. libraries like your's [pyRecEst](https://github.com/FlorianPfaff/pyRecEst) start using the backend), then we may need to reconsider this assumption.

P.S. tests are now working properly.

luisfpereira (approved):

PR #1904, opened 2023-11-21
Add Diffeo abstraction

luisfpereira:
Following the ideas started in #1863, this PR brings in the notion of `Diffeo`.

Additionally, it refactors `discrete_curves` (done with @alebrigant). This happens together, because this module is heavily dependent on the `Diffeo` abstraction: `SRVTransform`, and `FTransform`.

**Design of `Diffeo`**

(This part may interest you @ymontmarin.)

Most of it results from what was already implemented in `PullbackDiffeoMetric`.

In particular, `diffeomorphism`, and `inverse_diffeomorphism` keep the same signature, while `tangent_diffeomorphism`, and `inverse_tangent_diffeomorphism` have an additional `image_point` or `base_point` due to reasons specified in the docstrings of the abstract class.

We've also decided to split the autodiff part in a new class called `AutodiffDiffeo`. While `Diffeo` is purely abstract, `AutodiffDiffeo` (a child of `Diffeo`) brings in implementations for `tangent_diffeomorphism` and `inverse_tangent_diffeomorphism` given implementations of `diffeomorphism` and `inverse_diffeomorphism` through automatic differentiation. I think this split makes sense both from a conceptual and coding point of view, as `AutodiffDiffeo` defines several methods that are required almost exclusively for the automatic differentiation (e.g. `jacobian_diffeomorphism`) and are usually not considered outside of this scope. (This also simplifies implementations as the ones mentioned below)

Moving these methods out of `PullbackDiffeoMetric` to a new abstraction (`Diffeo`) also allows to bring in new implementations that build on top of the new class. In particular, `ReversedDiffeo` just converts a `Diffeo` from $M \to N$ in a diffeo from $N \to M$. More interestingly, `ComposedDiffeo` takes several diffeos (e.g. $M \to N \to P$) and creates a (composed) diffeo between the first and last manifold (e.g. $M \to P$). This may be useful to avoid creating unnecessary intermediate representations, while still keeping the possibility of using such representations without code duplication if wanted.

**Affected modules**

While several modules were affected by this change, the main changes were done in `spd_matrices`/`hpd_matrices`, and `discrete_curves`.

**`spd_matrices`/`hpd_matrices`**

(This part may interest you @YannCabanes, as `hpd_matrices` has changed similarly to `spd_matrices`.)

We've created the diffeomorphisms `LogDiffeo`, `PowerDiffeo`, and `CholeskyMap`. This hugely simplifies `SPDMatrices` API. I've also transformed some methods into functions due to wide use (`logmh`, `expm`, `powermh`). The intention is to move them to the backend afterwards.

Due to the new `LogDiffeo`, we've implemented `SPDLogEuclideanMetric`as a `PullbackDiffeoMetric`.

Due to the new `PowerDiffeo`, we've removed `power_affine`, and `power_euclidean` from `SPDAffineMetric`, and `SPDEuclideanMetric`, respectively, as this metric can be obtained with `SPDPowerMetric` (which takes advantage both from `PullbackDiffeoMetric` and `ScalarProductMetric`)  with much simpler (and complete) code, without any performance degradation. For example, `SPDEuclideanMetric.parallel_transport` would raise a `NotImplementedError` for `power_euclidean != 1`, whereas now it works properly.

**`discrete_curves`**

(This part may interest you @MMalikT)
 
(with @alebrigant )

We've decided, for the moment, to focus only on `DiscreteCurvesStartingAtOrigin`, which are discrete curves modulo translation.

The method `projection` takes a `DiscreteCurve` and removes translation and origin (@alebrigant, we may need to iterate on this, but for now, not having the origin in the point representation is much more convenient in order to take advantage of existing implementations, such as `NFoldManifold`). This method is widely used in the already existing notebooks to keep the previous behavior. Within the notebooks we've added back the origin before plotting.

We've implemented `SRVTransform` and `FTransform`. The corresponding pullback metrics (`ElasticTranslationMetric` and `SRVTranslationMetric`) simply inherit from `PullbackDiffeoMetric` without overriding any code.

For the quotienting by reparameterizations, we've divide the algorithms we have to align curves into two classes (`IterativeHorizontalGeodesic`, and `DynamicProgrammingAligner`). This follows an idea we were already pursuing in `graph_space` and deeply simplifies the implementation of `SRVTranslationReparametrizationBundle` (we simply need to pass an aligner - or accept the default one) and `SRVTranslationReparametrizationQuotientMetric` (the implementation is independent of the algorithm). Both algorithms are still fragile and their implementations need to be improved.

Several auxiliar methods were added to (conceptually) simplify the implementations, such as `forward_difference`, `centered_difference`, and so on. A next step is to extend this notions to ambient manifolds other than the Euclidean space.

We've fully removed `ClosedDiscreteCurves` because the implementation is not consistent with the changes. We need to revisit this afterwards.


ymontmarin:
@luisfpereira Hi Luis, thank you for letting me know ! I just read the new classes and think is great! It unify the structure of the notions. Also the subclass for autodiff behavior is way more elegant !

alebrigant (approved):
Great work, this makes discrete curves so much cleaner. The way you simplified the code for the fiber bundle structure in particular is very nice. Thanks !! I just put some comments on a few docstrings.

geomstats/geometry/diffeo.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Parameters
+        ----------
+        base_point : array-like, shape=[..., *space_shape]
+            Base point.
+
+        Returns
+        -------
+        image_point : array-like, shape=[..., *image_shape]
+            Inner-product matrix.
```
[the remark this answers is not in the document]
luisfpereira: Bad copy-paste. Fixing. Thanks!

geomstats/geometry/diffeo.py
```diff
@@ -0,0 +1,331 @@
...
+        """
+
+    @abc.abstractmethod
+    def inverse_diffeomorphism(self, image_point):
+        r"""Inverse diffeomorphism at base point.
+
+        :math:`f^-1: N \rightarrow M` of the manifold
```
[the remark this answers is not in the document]
luisfpereira: Clearly not.

geomstats/geometry/diffeo.py
```diff
@@ -0,0 +1,331 @@
...
+        image_tangent_vec : array-like, shape=[..., *image_shape]
+            Image tangent vector at image of the base point.
+        """
+        if base_point is None:
+            base_point = self.inverse_diffeomorphism(image_point)
+
+        batch_shape = get_batch_shape(self._space_point_ndim, tangent_vec, base_point)
+        flat_batch_shape = (-1,) if batch_shape else ()
+
+        J_flat = gs.reshape(
```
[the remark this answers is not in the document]
luisfpereira: Changed to lower.

geomstats/test_cases/geometry/fiber_bundle.py
```diff
@@ -477,15 +482,19 @@ def tangent_riemannian_submersion(self, tangent_vec, base_point):
...
         if base_point is None:
-            if fiber_point is not None:
-                base_point = self.riemannian_submersion(fiber_point)
-            else:
-                raise ValueError(
-                    "Either a point (of the total space) or a "
-                    "base point (of the base manifold) must be "
-                    "given."
-                )
+            base_point = self.riemannian_submersion(fiber_point)
```
ninamiolane, line 493: NIT: We could use total_point instead of fiber_point, so that the points mirror the names of the spaces:
- total_point belongs to total_space,
- base_point belongs to base_space,
and "fiber_point" can confuse people as it seems that it refers to a point on the fiber (i.e., to a group element), as opposed to a point in the fiber bundle (which is an element of the total space)

tests/tests_geomstats/test_geometry/data/discrete_curves.py
```diff
@@ -67,20 +56,21 @@ class SRVShapeBundleTestData(FiberBundleTestData):
...
-                [dict(n_reps=n_reps, n_times=n_times) for n_reps in self.N_VEC_REPS]
-            )
-        return self.generate_tests(data)
 
-    def horizontal_geodesic_has_horizontal_derivative_test_data(self):
-        data = []
-        for n_times in [20]:
-            data.extend([dict(n_points=n_points, n_times=n_times) for n_points in [1]])
-        return self.generate_tests(data)
+class AlignerCmpTestData(TestData):
```
ninamiolane, line 60: What is "Cmp"? Is AlignedCmp a class in the codebase? Can you leave a docstring explaining why it deserves to have a spcial TestData?

tests/tests_geomstats/test_geometry/data/hpd_matrices.py
```diff
@@ -58,42 +54,24 @@ def belongs_test_data(self):
...
-        "geodesic_ivp_belongs",
-        "exp_belongs",
-    )
-
     def exp_domain_vec_test_data(self):
         return self.generate_vec_data()
 
 
-class HPDLogEuclideanMetricTestData(ComplexRiemannianMetricTestData):
+class HPDLogEuclideanMetricTestData(PullbackDiffeoMetricTestData):
```
ninamiolane, line 75: 🎉

tests/tests_geomstats/test_geometry/data/positive_lower_triangular_matrices.py
```diff
@@ -3,33 +3,12 @@
...
-        return self.generate_random_data()
-
-    def differential_gram_belongs_to_symmetric_matrices_test_data(self):
-        return self.generate_random_data()
-
-    def inverse_differential_gram_belongs_to_lower_triangular_matrices_test_data(self):
-        return self.generate_random_data()
-
-
 class PositiveLowerTriangularMatrices2TestData(TestData):
```
ninamiolane, line 12: Maybe remove the `2` in the PositiveLowerTriangularMatrices2 here.

tests/tests_geomstats/test_geometry/test_discrete_curves.py
```diff
@@ -2,224 +2,191 @@
...
+
+
+@pytest.mark.smoke
+class TestAlignerCmp(TestCase, metaclass=DataBasedParametrizer):
+    total_space = DiscreteCurvesStartingAtOrigin(k_sampling_points=10)
+    bundle = SRVTranslationReparametrizationBundle(total_space)
 
-    space.equip_with_group_action("reparametrizations")
-    space.equip_with_quotient_structure()
+    aligner = IterativeHorizontalGeodesic()
```
ninamiolane, line 177: NIT: this class' name should probably end with Aligner, to match the logic of line 178.

tests/tests_geomstats/test_geometry/test_spd_matrices.py
```diff
@@ -2,49 +2,105 @@
...
     SPDEuclideanMetricTestData,
     SPDLogEuclideanMetricTestData,
     SPDMatrices2TestData,
     SPDMatrices3TestData,
     SPDMatricesTestData,
+    SymMatrixLogSmokeTestData,
 )
 
 # TODO: check affine invariance
+# TODO: add SPDPowerMetricTestCase against some power=1?
```
ninamiolane, line 56: Put TODOs in github issues, not in the code (otherwise the nb of TODOs  will blow up)

ninamiolane (approved):
This is fantastic, thank you very much! I've left minor comments.

tests/tests_geomstats/test_geometry/data/discrete_curves.py
luisfpereira: "Cmp" stands for comparison. This is a kind of test I've recently added that compares the outputs of two objects that do the same job. e.g. in this case it compares the outputs of the two algorithms we have for alignment.

I've implemented a similar comparison for metrics: if we have two different ways of implementing a given metric, it is very powerful to compare the results both provide as it would be very unlikely two independent ways of doing the same thing give the same results in random data.

tests/tests_geomstats/test_geometry/data/positive_lower_triangular_matrices.py
luisfpereira: The 2 here means `n=2`, as this corresponds to smoke data on a particular space.

tests/tests_geomstats/test_geometry/test_spd_matrices.py
luisfpereira: Done.

tests/tests_geomstats/test_geometry/test_discrete_curves.py
luisfpereira: Done.

PR #1907, opened 2023-11-28
Endow space with mathematical structure

luisfpereira:
Following some discussions we've been having about successively endowing a space with mathematical structures (especially during last year's hackathon), I suggest in this PR a technical way of accomplishing it. It allows to have much more flexible code with minimal changes. I use the square matrices and the corresponding `MatrixLieAlgebra` structure as example.

**Square matrices and matrix Lie algebra**

Currently, `SquareMatrices` inherit from `MatrixLieAlgebra`, which in turn inherits from `VectorSpace`. In my view, `SquareMatrices` plain implementation should be unaware of its role as Lie algebra for the general linear group.

This can be accomplished by (obviously) noticing that square matrices is a Manifold not because it is a Lie algebra, but because it is a vector space (and, of course, the fact it is a vector space allows it to be the Lie algebra of some space). Therefore, the plain implementation of square matrices should be a vector space with the possibility of adding `LieAlgebra` structure (meaning in this case bringing in methods such as `baker_campbell_hausdorff`). That is, `LieAlgebra` should not inherit from `Manifold`, but be something we can add to the  manifold.

In terms of plain `SquareMatrices` implementation, it is even enough to see it is a very particular case of `Matrices`, so we can simply implement it as a child where we set `m = n`.

For the `LieAlgebra` structure, I suggest we implement it in a separated object, which I'm calling, for the lack of a better name,  `SquareMatricesLieAlgebraStruct`. Notice, that similarly to the metric, a `MathStruct` is aware of the space it is put on (reason to receive `space` as first input).

This way, if someone is working with `SquareMatrices` out of a Lie groups context, they can instantiate an object that is simply a vector space. Within the context of Lie groups, they can add the Lie algebra structure with the following code:

```python
space = SquareMatrices(n=3)

space.endow_with_math_struct(SquareMatricesLieAlgebraStruct)
```

**Method calling**

When we equip with a metric, we call one of its methods with `space.metric.<method_name>`. Moreover, one space can only be associated with one metric at time.

For mathematical structures, I suggest `space.<method_name>` is desired, as the space **becomes** e.g. a `LieAlgebra`. Moreover, one space can be endowed with multiple (**compatible**) math structures. This translates into code in the following way:

```python
space = SquareMatrices(n=3)
point_a, point_b = space.random_point(2)

space.baker_campbell_hausdorff(point_a, point_b)
# raises AttributeError

space.endow_with_math_struct(SquareMatricesLieAlgebraStruct)
space.baker_campbell_hausdorff(point_a, point_b)
# works properly
```

**Name clashing**

Compatible math structures mean that there should be no name clashing. In case it happens, the current implementation chooses the method that corresponds to the mathematical structure added firstly (can be easily changed in the other direction).

**Remarks**

In my view, this solution makes the code more clean, readable, and closer to the mathematical literature. For educational purposes, I believe it makes it even better, as it becomes very clear why some methods are required (e.g. `baker_campbell_hausdorff` is only useful in the context of Lie groups, so a student can safely ignore its implementation until they start being concerned with the topic).

In terms of code organization, it can also make things nicer. For example, I suggest in the case of square matrices, we create `geomstats/geometry/square_matrices.py`, move `SquareMatrices` there, but keep `SquareMatricesLieAlgebraStruct` in `general_linear.py`, as it has a meaning in that context.

On the other side, we will end up with more smaller objects. I usually see it as a positive thing, as it means they're very easy to mantain, but I agree there's a negative side, as the relationships between objects need to be taken care of more carefully.

**Next steps**

Once you've reviewed this and we reach a satisfying solution for everyone, I'll extend the implementation to the other `LieAlgebra`'s. This should be enough to merge as a first step. Later, we should start thinking about additional mathematical structures to bring in.

### DIFF
--- a/geomstats/geometry/base.py
+++ b/geomstats/geometry/base.py
@@ -12,6 +12,24 @@ from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.pullback_metric import PullbackMetric
 
 
+class MathStruct(abc.ABC):
+    """Abstract mathematical structure.
+
+    Implements methods required to handle method/attribute calling.
+    """
+
+    def __init__(self, space):
+        self._space = space
+
+    def __getattr__(self, name):
+        """Get attribute.
+
+        Uses space to get attribute, as it is aware of mathematical structures.
+        This way preserves overriding expected behavior.
+        """
+        return getattr(self._space, name)
+
+
 class VectorSpace(Manifold, abc.ABC):
     """Abstract class for vector spaces.
 
--- a/geomstats/geometry/general_linear.py
+++ b/geomstats/geometry/general_linear.py
@@ -187,7 +187,7 @@ class GeneralLinear(MatrixLieGroup, OpenSet):
         return path
 
 
-class SquareMatrices(MatrixLieAlgebra):
+class SquareMatrices(Matrices):
     """Lie algebra of the general linear group.
 
     This is the space of matrices.
@@ -200,17 +200,14 @@ class SquareMatrices(MatrixLieAlgebra):
 
     def __init__(self, n, equip=True):
         self.n = n
-        super().__init__(dim=n**2, representation_dim=n, equip=equip)
-        self._mat_space = Matrices(n, n, equip=False)
+        super().__init__(n=n, m=n, equip=equip)
 
-    @staticmethod
-    def default_metric():
-        """Metric to equip the space with if equip is True."""
-        return MatricesMetric
 
-    def _create_basis(self):
-        """Create the canonical basis of the space of matrices."""
-        return self._mat_space.basis
+class SquareMatricesLieAlgebraStruct(MatrixLieAlgebra):
+    """Lie algebra structure for square matrices."""
+
+    def __init__(self, space):
+        super().__init__(space=space, representation_dim=space.n)
 
     def basis_representation(self, matrix_representation):
         """Compute the coefficient in the usual matrix basis.
@@ -227,7 +224,7 @@ class SquareMatrices(MatrixLieAlgebra):
         basis_representation : array-like, shape=[..., dim]
             Representation in the basis.
         """
-        return self._mat_space.flatten(matrix_representation)
+        return self.flatten(matrix_representation)
 
     def matrix_representation(self, basis_representation):
         """Compute the matrix representation for the given basis coefficients.
@@ -244,4 +241,4 @@ class SquareMatrices(MatrixLieAlgebra):
         matrix_representation : array-like, shape=[..., n, n]
             Matrix.
         """
-        return self._mat_space.reshape(basis_representation)
+        return self.reshape(basis_representation)
--- a/geomstats/geometry/lie_algebra.py
+++ b/geomstats/geometry/lie_algebra.py
@@ -12,14 +12,14 @@ Lead author: Stefan Heyder.
 import abc
 
 import geomstats.backend as gs
-import geomstats.errors
-from geomstats.geometry.base import VectorSpace
+from geomstats.errors import check_integer
+from geomstats.geometry.base import MathStruct
 from geomstats.geometry.matrices import Matrices
 
 from ._bch_coefficients import BCH_COEFFICIENTS
 
 
-class MatrixLieAlgebra(VectorSpace, abc.ABC):
+class MatrixLieAlgebra(MathStruct, abc.ABC):
     """Class implementing matrix Lie algebra related functions.
 
     Parameters
@@ -29,10 +29,10 @@ class MatrixLieAlgebra(VectorSpace, abc.ABC):
         Lie algebra.
     """
 
-    def __init__(self, representation_dim, **kwargs):
-        geomstats.errors.check_integer(representation_dim, "representation_dim")
-        super().__init__(shape=(representation_dim, representation_dim), **kwargs)
+    def __init__(self, space, representation_dim):
+        check_integer(representation_dim, "representation_dim")
         self.representation_dim = representation_dim
+        super().__init__(space=space)
 
     bracket = Matrices.bracket
 
@@ -118,7 +118,4 @@ class MatrixLieAlgebra(VectorSpace, abc.ABC):
         matrix_representation : array-like, shape=[..., *point_shape]
             Matrix.
         """
-        if self.basis is None:
-            raise NotImplementedError("basis not implemented")
-
         return gs.einsum("...i,ijk ->...jk", basis_representation, self.basis)
--- a/geomstats/geometry/manifold.py
+++ b/geomstats/geometry/manifold.py
@@ -63,6 +63,24 @@ class Manifold(abc.ABC):
         if equip:
             self.equip_with_metric()
 
+        self._math_structs = []
+
+    def __getattr__(self, name):
+        """Get attribute.
+
+        Looks for the attribute in the space and in the endowed mathematical
+        structures.
+        """
+        try:
+            return self.__getattribute__(name)
+        except AttributeError as e:
+            for math_struct in self._math_structs:
+                try:
+                    return getattr(math_struct, name)
+                except AttributeError:
+                    pass
+            raise e
+
     def equip_with_metric(self, Metric=None, **metric_kwargs):
         """Equip manifold with a Riemannian metric.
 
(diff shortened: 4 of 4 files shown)


geomstats/geometry/base.py
```diff
@@ -12,6 +12,24 @@
 from geomstats.geometry.pullback_metric import PullbackMetric
 
 
+class MathStruct(abc.ABC):
```
ninamiolane, line 15: Let's just class it Structure, to avoid concatenating abbreviations which end up being hard to read for external users & collaborators.

geomstats/geometry/general_linear.py
```diff
@@ -200,17 +200,14 @@ class SquareMatrices(MatrixLieAlgebra):
...
 
-    @staticmethod
-    def default_metric():
-        """Metric to equip the space with if equip is True."""
-        return MatricesMetric
 
-    def _create_basis(self):
-        """Create the canonical basis of the space of matrices."""
-        return self._mat_space.basis
+class SquareMatricesLieAlgebraStruct(MatrixLieAlgebra):
```
ninamiolane, line 206: No need for "Struct" IMO. If we want to append something, let's use Structure.

ninamiolane: Even better, we could give this the name of the structure, i.e.
`SquareMatricesLieBracket`, so that:
- `VectorSpace` + `LieBracket` = `LieAlgebra`.

geomstats/geometry/general_linear.py
```diff
@@ -200,17 +200,14 @@ class SquareMatrices(MatrixLieAlgebra):
...
+    """Lie algebra structure for square matrices."""
+
+    def __init__(self, space):
+        super().__init__(space=space, representation_dim=space.n)
 
     def basis_representation(self, matrix_representation):
```
ninamiolane, line 212: basis_representation and matrix_representation should be methods of the space, not methods of the LieBracket structure. Because they depend on a notion of basis, and the notion basis is a vector space property. We should be careful about which methods go in the Structure classes and which methods stay in the Spaces, as we did for the distinction between Metrics and Spaces.

geomstats/geometry/lie_algebra.py
```diff
@@ -12,14 +12,14 @@
...
-from geomstats.geometry.base import VectorSpace
+from geomstats.errors import check_integer
+from geomstats.geometry.base import MathStruct
 from geomstats.geometry.matrices import Matrices
 
 from ._bch_coefficients import BCH_COEFFICIENTS
 
 
-class MatrixLieAlgebra(VectorSpace, abc.ABC):
+class MatrixLieAlgebra(MathStruct, abc.ABC):
```
ninamiolane, line 22: The LieAlgebra is the space equipped with the LieBracket.
If this inherits from a Structure, it needs to have the name of a structure, i.e., a LieBracket.

geomstats/geometry/manifold.py
```diff
@@ -63,6 +63,24 @@ def __init__(
...
+    def __getattr__(self, name):
+        """Get attribute.
+
+        Looks for the attribute in the space and in the endowed mathematical
+        structures.
+        """
+        try:
+            return self.__getattribute__(name)
+        except AttributeError as e:
+            for math_struct in self._math_structs:
```
ninamiolane, line 77: `self._structures`

geomstats/geometry/manifold.py
```diff
@@ -121,6 +139,25 @@ def equip_with_quotient_structure(self):
         self.quotient = self.new(equip=False)
         self.quotient.equip_with_metric(QuotientMetric_, fiber_bundle=self.fiber_bundle)
 
+    def endow_with_math_struct(self, MathStruct, **kwargs):
```
ninamiolane, line 142: rename to `equip_with_structure(self, Structure, **kwargs)` so that it mirrors the metric case. this makes the API easier for users.

Otherwise users have to remember that it's `equip` for metric but `endow` which is cumbersome.

ninamiolane (changes requested):
Excellent! I've left comments about:
- which methods should go where, i.e., in Structure versus Space
- naming: no abbreviation nor synonyms (i.e. equip and endow is confusing -> equip everywhere).

Should a Metric be a Structure, then? I think it should.

geomstats/geometry/general_linear.py
luisfpereira: I fully agree here. This probably means we need to create an intermediate manifold structure inheriting from `VectorSpace` called e.g. `MatrixVectorSpace`, as this methods only make sense for the matrix case. Do you agree? (If yes, I'll do it in a different PR.)

geomstats/geometry/general_linear.py
luisfpereira: I've appended structure to avoid having someone thinking that this is the Lie algebra of square matrices (the concept does not make sense, but it could lead to thinking something like that).

Nevertheless, your last comment fully answers it: very nice naming strategy!

geomstats/geometry/base.py
luisfpereira: `MathematicalStructure` instead or would be too long?

luisfpereira:
Thanks for the suggestions @ninamiolane. Regarding your high-level comments:

> * I think metrics should inherit from the structure class. they are just one structure we put on the space.

I agree with you, though I think we should probably provide this behavior only later (I'm thinking about name clashes as `exp` - both used in `LieGroup`, `Connection` and `RiemannianMetric` - that we should think about first).

When we reach this point, then it would be very easy to add metric structures other than `LeviCivitaConnection`.

> * I think the spaces should be maximally equipped by default, i.e., we should give the LieBracket structure to SquareMatrices by default for the users who are not experts in differential geometry and will have trouble adding it themselves.

The difficulty here is that it is straightforward to add structure, but it is a bit harder to remove it.

A possible solution could be to have a flag like `equip` to choose between plain or maximally equipped. (Probably the same flag?)

luisfpereira:
@ninamiolane, I've address all your comments. I think this is now in a mature state. My only doubt is if we should search for the methods from the first added structures or from the last added structures (for now the former is implemented).

I've implemented a solution for your "maximally equipped" suggestion, that follows the same principle as `default_metric`.

I've added a cache mechanism to `get_attr` that ensures equivalent performance (cf. current state).

 If you agree with the changes, I'll apply it to the all the uses of `MatrixLieAlgebra` and merge.

I think we should apply it soon to `RiemannianMetric`, but probably do it in another iteration (maybe even do a new release first).

geomstats/geometry/general_linear.py
```diff
@@ -198,50 +198,9 @@ class SquareMatrices(MatrixLieAlgebra):
...
-        basis_representation : array-like, shape=[..., dim]
-            Coefficients in the basis.
-
-        Returns
-        -------
-        matrix_representation : array-like, shape=[..., n, n]
-            Matrix.
-        """
-        return self._mat_space.reshape(basis_representation)
+    def default_maximal_structures(self):
```
ninamiolane, line 205: --> `default_structures`

we don't need the "maximal" here.

Can you add the docstring and explain there that the default structures are the maximal structures we can equip this space with?

geomstats/geometry/general_linear.py
```diff
@@ -198,50 +198,9 @@ class SquareMatrices(MatrixLieAlgebra):
         Integer representing the shape of the matrices: n x n.
     """
 
-    def __init__(self, n, equip=True):
+    def __init__(self, n, equip=True, equip_maximally=True):
```
ninamiolane, line 201: So `equip` refers to equip with metrics and equip_maximally to equip with structures?

maybe use `equip_with_metric` and `equip_with_structures` then?

geomstats/geometry/manifold.py
```diff
@@ -121,6 +151,43 @@ def equip_with_quotient_structure(self):
         self.quotient = self.new(equip=False)
         self.quotient.equip_with_metric(QuotientMetric_, fiber_bundle=self.fiber_bundle)
 
+    def equip_with_structure(self, Structure, **kwargs):
+        """Endow manifold with mathematic structure.
```
ninamiolane, line 155: Endow -> Equip

geomstats/geometry/manifold.py
```diff
@@ -121,6 +151,43 @@ def equip_with_quotient_structure(self):
...
+
+        Parameters
+        ----------
+        Structure : Structure object
+        """
+        self._attr_cache = {}
+        self._structures.append(Structure(self, **kwargs))
+        return self
+
+    def equip_maximally(
```
ninamiolane, line 174: what is the difference between this method and the one above? Could you add an explanatory docstring?

ninamiolane (approved):
Looking good, thanks!

The distinction between "equip" and "equip_maximally" is too implicit. I suggest changing to equip_with_metric and equip_with_structure / equip_with_structures.

Otherwise, add the missing docstrings and good to go for me. Thank you!

geomstats/geometry/base.py
```diff
@@ -12,6 +12,34 @@
...
+        except AttributeError:
+            return NotImplemented
+
+    def __getattr__(self, name):
+        """Get attribute.
+
+        Uses space to get attribute, as it is aware of mathematical structures.
+        This way preserves overriding expected behavior.
+        """
+        attr = self.__getattr__(name)
```
johnharveymath, line 36: I think this should be `self._getattr` here? This whole PR seems like a great design direction to me.

PR #1912, opened 2023-11-30
Update codecov to work based on diff

luisfpereira:
codecov is failing recurrently for very small differences in coverage (e.g. -0.01%). This PR changes its behavior to work based on differences to the last commit, accepting failures up to 1%.


luisfpereira:
P.S. problem still persists.

PR #1920, opened 2023-12-22
Improve documentation rendering

qbarthelemy:
## Issue

Correct equations, references, bullet points, etc rendering in documentation.
No code modification.


agramfort (approved):
❤️

geomstats/geometry/connection.py
```diff
@@ -363,7 +363,7 @@ def riemann_tensor(self, base_point):
         :math:`R_{ijk}^l = dx^l(R(X_j, X_k)X_i)`
 
         which gives :math:`R_{ijk}^lk` as a sum of four terms:
-        :math:`R_{ijk}^l =
+        :math:`R_{ijk}^l =`
```
luisfpereira, line 366: delete this line and fix above ``R_{ijk}^lk` -> `R_{ijk}^l` (see https://en.wikipedia.org/wiki/Riemann_curvature_tensor#Coordinate_expression).

(Sorry for asking for this, I know it does not come from this PR.)

luisfpereira (changes requested):
Thank you very much! Can you please just address my comment?

geomstats/geometry/connection.py
qbarthelemy: Is the correction what you wanted?

qbarthelemy: Can you also indicate how to correct these equations?

https://geomstats.github.io/api/geomstats.geometry.html#geomstats.geometry.fiber_bundle.FiberBundle.integrability_tensor_derivative

https://geomstats.github.io/api/geomstats.geometry.html#geomstats.geometry.pre_shape.PreShapeSpaceBundle.integrability_tensor_derivative

luisfpereira: I think there is an extra `_` that needs to be removed, i.e. it should be `:math:`\nabla_X Y |_P = nabla_x y`, since I think upper/lower case refer to vectors in total/base spaces.

luisfpereira (approved):

Issue #1928, opened 2024-01-15
Bug: Incorrect (3, 1) Riemann Tensor computation in PyTorch backend due to PyTorch autodiff jacobian `create_graph` being turned off


yangshengaa:
### Describe the bug

To compute the (3, 1) Riemann tensor, we need to differentiate through the Christoffel symbol of the second kind, which in turn is computed by differentiating through the metric. Both diffferentiations rely on autodiff in `pytorch.autograd.jacobian` as inherited from [geomstats._backend.pytorch.autodiff](https://github.com/geomstats/geomstats/blob/0102ceaeac84c39a9806cb1bd2853dbbbc134e48/geomstats/_backend/pytorch/autodiff.py#L90), but the `create_graph` stays `False` by default. This means the second time we call autodiff, no gradient information is tracked and thus the first two terms in Riemann tensor are zeros regardless. 

We could expose the `create_graph` argument to users with default still set to `False`, and just change locally how we compute the Riemann tensor without affecting other codes. 

Similar issues has been brought up, see [Bug: Jacobian of Jacobian in pytorch produces null matrices #1801](https://github.com/geomstats/geomstats/issues/1801) also. 

### Steps/Code to Reproduce

Below demonstrate that recursive calling to pytroch jacobian yields zero. 

```
# treat a single hidden layer neural net feature map as an immersion 
d = 3  # input dim 
w = 10 # hidden dim 
W = torch.randn((w, d))
nl = nn.Sigmoid() # Sigmoid nonlinearity

class NNFeatMap(ImmersedSet):
    def __init__(self, equip=True):
        super().__init__(dim=d, equip=equip)

    def immersion(self, point):
        res = nl(point @ W.T)
        return res

    def _define_embedding_space(self):
        return Euclidean(dim=w)

gs_model = NNFeatMap(equip=False)
gs_model.equip_with_metric(PullbackMetric)

# random input 
X = torch.randn(4, d)
jacobian_christoffels = gs.autodiff.jacobian_vec(gs_model.metric.christoffels)(X)
```

With Sigmoid nonlinearity, `jacobian_christoffels` should not be zero under random input, but in fact they are all zero. This is due to the mechnism above. 

This suggests the first two terms in Riemann tensor computation is incorrect, since they call exactly `gs.autodiff.jacobian_vec(self.metric.christoffels)(base_point)`.

### Expected Behaviour

Expected nonzero values.

### Actual Behaviour

Only zeros are thrown. 

### Your environment

[...]

Issue #1801, opened 2023-01-27
Bug: Jacobian of Jacobian in pytorch produces null matrices


tramy1258:
### Describe the bug

Computing `inner_product_derivative_matrix` in `FisherRaoMetric` requires computing Jacobian and Hessian of `pdf` at given base points. We would like to use Jacobian of Jacobian to compute Hessian matrix to be more time efficient, instead of computing Hessian and Jacobian separately. This method works in `autograd` but not in `pytorch` where outputs are just null matrices. The error can be reproduced with the code below. 

Another point that might be worth looking into is that using nested Jacobian produces matrix of shape `(...,1,dim,dim)` whereas using Hessian produces matrix of shape `(...,dim,dim)`.

### Steps/Code to Reproduce

```
import geomstats as gs

def test(x):
    return gs.array([gs.sum(x**4)])

# Basic case
point = gs.array([0.1,0.5])
jj = gs.autodiff.jacobian(gs.autodiff.jacobian(test))(point)
print(jj,jj.shape)
print('-'*30)

# Vectorized case
point = gs.array([[0.1,0.5],[0.2,0.7]])
jj = gs.autodiff.jacobian_vec(gs.autodiff.jacobian_vec(test))(point)
print(jj,jj.shape)
```

### Expected Behaviour

Similar behaviour to `autograd`.

### Actual Behaviour

Outputs null matrices.

### Your environment

```shell
* pytorch, autograd
* Windows 10
```

yangshengaa:
From my understanding, currently we are not able to differentiate on top of `gs.autodiff.jacobian_vec` because by default pytorch `TORCH.AUTOGRAD.FUNCTIONAL.JACOBIAN(...)` puts `create_graph=False`, which means the gradient infomation is not tracked when computing the first order, and thus cannot generate second order when calling it again. 

Below is the source code from [geomstats._backend.pytorch.autodiff](https://github.com/geomstats/geomstats/blob/main/geomstats/_backend/pytorch/autodiff.py):

```
def jacobian_vec(func, point_ndim=1):
    """<some docstrings>
    """

    def _jacobian(point):
        if point.ndim == point_ndim:
            return _torch_jacobian(func=lambda x: func(x), inputs=point)
        return _torch.stack(
            [
                _torch_jacobian(func=lambda x: func(x), inputs=one_point)
                for one_point in point
            ],
            axis=0,
        )

    return _jacobian
```

you see that `create_graph=True` is not there.

[...]

PR #1929, opened 2024-01-16
set create_graph to True in `pytorch` jacobian and hessian computations to align with behaviors in `autograd` backend

yangshengaa:
By default, pytorch turns off the `create_graph` flag for runtime and memory concerns in its jacobian and hessian computations. However, this means we are not able to continue differentiate through the jacobian (see [Bug: Jacobian of Jacobian in pytorch produces null matrices #1801](https://github.com/geomstats/geomstats/issues/1801)) and may incur incorrect computations in geometric quantities such as Riemann tensor (see [Bug: Incorrect (3, 1) Riemann Tensor computation in PyTorch backend due to PyTorch autodiff jacobian create_graph being turned off #1928](https://github.com/geomstats/geomstats/issues/1928))

This PR merely sets the `create_graph` to True by default for simplicifty, potentially at the cost of runtime and memory. But with this we are able to fully align with the behaviors in `autograd` backend.

### DIFF
--- a/geomstats/_backend/pytorch/autodiff.py
+++ b/geomstats/_backend/pytorch/autodiff.py
@@ -82,7 +82,7 @@ def jacobian(func):
     """
 
     def _jacobian(point):
-        return _torch_jacobian(func=lambda x: func(x), inputs=point)
+        return _torch_jacobian(func=lambda x: func(x), inputs=point, create_graph=True)
 
     return _jacobian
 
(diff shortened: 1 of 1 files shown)


luisfpereira:
Thanks @yangshengaa, I can take it from here (I'll make some fixes and add tests to ensure we do not forget the reason to add the flag).

ninamiolane:
@luisfpereira should this PR be closed, now that https://github.com/geomstats/geomstats/pull/1930 is merged?

PR #1930, opened 2024-01-17
Fix and clean value_and_grad

luisfpereira:
Inspired by #1929, this PR is the first of a series of PRs that intends to fix and clean the `autodiff` backend module.

In this PR, the focus is the `value_and_grad` function. The following is done:

* vectorization is fixed, i.e. `value_and_grad` is now able to handle (correctly) batches of points

* `point_ndims` variable is added, following a similar behavior as `argnums` (used both in `autograd` and `pytorch`). This variable is required to handle vectorization properly

* custom gradient is fixed for `pytorch` to allow a custom computation of the squared distance gradient (see `SpecialEuclidean` and `Grassmannian`). Before this was only working for `autograd`

* custom gradient is improved for `pytorch` backend with potential gains in performance for the cases where only the gradient wrt to some variables is required for a multivariable function

* `to_numpy` flag is removed, as it is useless (this behavior must be controlled from outside)

* the tests are updated to new style and several additional use cases are added (in particular regarding vectorization). Mostly all `value_and_grad` use cases should now be being tested.  There's still a small proportion of  tests in the old style, but they will be updated in the following up PRs

* all the uses of `value_and_grad` are updated to ensure correct use of the updated API

### DIFF
--- a/geomstats/_backend/autograd/autodiff.py
+++ b/geomstats/_backend/autograd/autodiff.py
@@ -5,6 +5,56 @@ import autograd.numpy as _np
 from autograd import jacobian
 
 
+def _get_max_ndim_point(*points):
+    """Identify point with higher dimension.
+
+    Same as `geomstats.vectorization._get_max_ndim_point`.
+
+    Parameters
+    ----------
+    points : array-like
+
+    Returns
+    -------
+    max_ndim_point : array-like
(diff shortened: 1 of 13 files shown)


luisfpereira:
I'm merging @ninamiolane since everything is tested and passing.

Issue #1725, opened 2022-10-26
Change `default_coords_type` to boolean flag `intrinsic=True`


sophiaas:
### Where should the content be modified?

https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/manifold.py

### What should be modified?

As discussed at the IHP hackathon with @luisfpereira, I suggest changing the `default_coords_type` argument in the init of [Manifold](https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/manifold.py) (and objects that subclass it) to a boolean `intrinsic=True`.

PR #1938, opened 2024-01-25
Added Brownian motion class for manifolds

dralston78:
## Checklist

- [x] I added appropriate unit tests.
- [x] I made sure the code passes all unit tests. (refer to comment below)
- [x] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#coding-style-guidelines) and API.

## Description

I implemented a class to generate a realization of a Brownian motion walk on a manifold in intrinsic coordinates.  While Brownian motion on a manifold is often defined as a Stratonovich stochastic differential equation (SDE), I have used the converted form to an Ito SDE in order to use the Euler–Maruyama method.  I added a unit test to verify the axioms of Brownian motion in Euclidean space.  

## Additional context

This class requires the pytorch backend.

The method to generate a path requires two conditions on the underlying manifold.  First, the manifold must be described in terms of its intrinsic coordinates.  Second, the manifold must be equipped with a metric that has a `cometric_matrix` method and a `christoffels` method.  

I proposed this idea to Professor Miolane at the end of her seminar class earlier this year, and she encouraged me to create an implementation.  However, as I don't have a background in software engineering, I wouldn't be surprised if there are some bad practices in my code/tests, or other flaws.  I'm looking forward to getting your feedback and learning more!


luisfpereira (approved):
Nice @dralston78!

I've add a couple of commit changing the following (let me know if you disagree with something):

* `space.metric` instead of `.metric` because if space is equipped with another metric at a later stage, then it is taken into consideration by the code; otherwise it is ambiguous. (Same with `space.dim`)

* I've removed the check on pytorch because someone may have closed form solutions for `christoffels` and `cometric_matrix`. if they don't have, then the existing error messages should be enough for users understand they need to change to autograd/pytorch

* I've removed the check `space.default_coords_type != "intrinsic"` because unfortunately this flag is misused and sometimes some spaces have intrinsic coordinates, but have a string specifying which kind of coordinates it refers to (e.g. PoincareBall -> "ball"). It is related with #1725. I'll do such a change asap.

* I've removed the checks `not hasattr(space.metric, "christoffels")` and `not hasattr(space.metric, "cometric_matrix")` because these methods always exist, though usually raise `NotImplementedError`. This is a design issues to which we haven't yet found a solution

* I've removed the check on `christoffels` and `cometric_matrix` shapes because I don't want to pay the price of running both once only for shape checking.

* for the reasons above, we are not doing checks and assume people know when they can use `BrownianMotion` - after fixing `intrinsic`, we can to an `instantiation` time check that will be equivalent to all the current tests

* I've added a name to the reference in the docstring, so that we can reference it in the text in case we need.

* I've updated `sample_path` docstrings.

* I've collected paths in a list and then used stack so that we avoid instantiating an array of zeros

* I've vectorized `sample_path`, i.e. allowed it to receive multiple points, by adding `ellipses` to `gs.einsum` and ensuring we sample independently for each point.


Regarding tests:

* I've done small changes, but no update in the behavior (they're very nice!)

geomstats/distributions/brownian_motion.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, space):
+        self.space = space
+
+    def sample_path(self, end_time, n_steps, initial_point):
+        """Generate a sample path of Brownian motion.
+
+        Parameters
+        ----------
+        end_time : float
```
ninamiolane, line 44: Please add parameters description

geomstats/distributions/brownian_motion.py
```diff
@@ -0,0 +1,75 @@
...
+        path : array-like, shape=[..., n_steps, dim]
+        """
+        step_size = end_time / n_steps
+
+        paths = [initial_point]
+        for _ in range(1, n_steps):
+            paths.append(self._step(step_size, paths[-1]))
+        return gs.stack(paths)
+
+    def _step(self, step_size, current_point):
```
ninamiolane, line 59: Please add a docstring even for a private method (to help future collaborators understand what you did)

tests/tests_geomstats/test_distributions/test_brownian_motion.py
```diff
@@ -0,0 +1,75 @@
...
+
+import geomstats.backend as gs
+from geomstats.distributions.brownian_motion import BrownianMotion
+from geomstats.geometry.euclidean import Euclidean
+from geomstats.test.test_case import TestCase, autograd_and_torch_only
+
+
+@autograd_and_torch_only
+@pytest.mark.slow
+class TestBrownianMotion(TestCase):
```
ninamiolane, line 14: Unit-tests should be unit, ie we should have at least one unit test per function.

Missing `test_sample_path` and missing `test__step`

ninamiolane (commented):
LGTM! Just minor comments on the docstrings + need "unit" tests.

tests/tests_geomstats/test_distributions/test_brownian_motion.py
dralston78: Hi @ninamiolane, thank you for this clarification!  With regards to this comment, would the ideal structure be to have a test class `TestBrownianMotion`, and then have this class have two methods `test_sample_path` and `test__step`?

luisfpereira: I think tests are already very good here, @ninamiolane. Two reasons: 1) this computation is relatively slow, so it is better to not add too many redundant tests; 2) there's randomness involved (we are sampling from a normal), which mean we cannot do a "normal test" (the checks done by @dralston78 are pretty clever).

luisfpereira:
@dralston78, with #1943 merged, we can now check for `intrinsic` (a boolean; it should be `True` for `BrownianMotion` use) at instantiation time (we may still need to check if the flag is properly used everywhere, but that can be done afterwards).

dralston78:
> @dralston78, with #1943 merged, we can now check for `intrinsic` (a boolean; it should be `True` for `BrownianMotion` use) at instantiation time (we may still need to check if the flag is properly used everywhere, but that can be done afterwards).

Great!  Would you like me to add a private method `_instantiation_check` that accomplishes the same checks I originally had?  Or should we assume users are aware of the requirements?

luisfpereira:
@dralston78, adding back `check_coordinates` taking into account how `intrinsic` works now will do the job.

luisfpereira:
@ninamiolane, this looks ready to me. Can we merge?

PR #1941, opened 2024-01-26
Remove dead code from vectorization

luisfpereira:
This PR removes dead code from vectorization, i.e. code that is not being used anywhere and is not intended to be used by external users.


luisfpereira:
I'm merging, since test failures are unrelated.

PR #1942, opened 2024-01-26
Remove default_point_type from Manifold

luisfpereira:
This PR removes redundant `string` attribute `default_point_type` from `Manifold`. The `int` `point_ndim` is a more expressive attribute and is already being used extensively in the library.


ninamiolane (approved):
Great idea. Thanks !

PR #1943, opened 2024-01-26
Remove default_coords_type and use intrinsic boolean instead

### LINKED ISSUE #1725: Change `default_coords_type` to boolean flag `intrinsic=True`

luisfpereira:
Motivated by #1938 and following #1725 (closes #1725), this PR removes `default_coords_type` and uses a boolean `intrinsic` instead. In the hyperbolic space an attribute called `coords_type` is still kept due to need to transform coordinates (in any case, distinction between extrinsic case, and the two intrinsic implementations is now more clear).


ninamiolane (approved):
LGTM! Great idea.

Issue #1825, opened 2023-02-02
Bug: Left-invariant geodesics on SO(3) gives NaNs because of error in the optimization computing the log


psarahdactyl:
### Describe the bug

When making a left invariant metric on SO(3) and calling the geodesic function, I get NaNs (with autograd backend) and segfault (with pytorch background). In both backends, this bug doesn't happen everytime. After playing with the random seed I _think_ I found a version that always fails. Because of the segfault in pytorch, I did not get very far debugging it so all the information I have is done using the autograd backend.

### Steps/Code to Reproduce

In this code, I use the autograd background, make an instance of SO3, and make a left invariant metric with a specific metric matrix at identity.

I make a geodesic starting at the `initial_point` of the identity matrix, and going to the `end_point` of a generated rotation from an angle `theta`.

When I run this code, sometimes it will say `overflow encountered in matmul`

I've traced it back to calling the `scipy.optimize.minimize` function in [connection.py](https://github.com/geomstats/geomstats/blob/master/geomstats/geometry/connection.py#L180). Please see the stacktrace for more details.
In `connection.py` I've added this line fixing the random seed and changing the `disp` option to `True` to get more output.
This display option only works with the autograd backend; pytorch just segfaults.
```
  objective_with_grad = gs.autodiff.value_and_grad(objective, to_numpy=True)

  gs.random.seed(2)
  tangent_vec = gs.flatten(gs.random.rand(*max_shape))  #.numpy()
  print("connection::log - tangent_vec:", tangent_vec)
  print("connection::log - point:", point)
  print("connection::log - base_point:", base_point)
  print("connection::log minimize")
  res = minimize(
      objective_with_grad,
      tangent_vec,
      method="L-BFGS-B",
      jac=True,
      options={"disp": True, "maxiter": max_iter},
      tol=tol,
  )
```

main code:
```
import os
import numpy as np

os.environ["GEOMSTATS_BACKEND"] = "autograd"

import geomstats.backend as gs
gs.set_default_dtype("float64")
from geomstats.algebra_utils import from_vector_to_diagonal_matrix
from geomstats.geometry.invariant_metric import InvariantMetric
from geomstats.geometry.invariant_metric import BiInvariantMetric
from geomstats.geometry.special_orthogonal import SpecialOrthogonal

SO3 = SpecialOrthogonal(n=3, point_type="matrix")

initial_point = gs.eye(3)
print("initial point (identity)")
print(initial_point)

# testing a bunch of rotations
[...]

PR #1944, opened 2024-01-29
Improve exp and log solvers API

### LINKED ISSUE #1891: Bug: LogODESolverdoesn't handle coordinate dependent connection even in non batch mode
### Describe the bug

When a riemanian metric is instantiated with a non uniform metric matrix, the solver will end up calling methods metrix_matrix and cometric_matrix on a sequence of points (of size equal to n_nodes). 
gs.linalg.inv will end up raising an error because the metric_matrix is not invertible.

Can you adapt the code and stack the call of metric_matrix and geodesic_equation to fix the bug?

Thanks

### Steps/Code to Reproduce

-

### Expected Behaviour

-

### Actual Behaviour

-

### Your environment

```shell
Windows 10, geomstats 2.7
```

luisfpereira:
This PR improves `ExpSolver` and `LogSolver` APIs by adding a more refined control of solving strategies in `Connection`.

Regarding `numerics.geodesic`:

* a previously unknown bug in `LogODESolver` regarding wrapping and unwrapping of states is fixed (@alebrigant, this improves a lot the computations in `gamma`, `dirichlet`, and `beta`; it is probably the reason for some weird behaviors we were having last year when trying to  develop alternative solver solutions)

* tests in `test_numerics` are improved (all geodesic solvers run against closed-form solutions when existing, or against other solvers)

* numeric solvers are guaranteed to work with points represented as multidimensional tensors (e.g. matrices) (closes #1891)

Regarding `geometry.invariant_metric`:

* a vectorization fix is done in `_InvariantMetricMatrix.geodesic_equation` (found by a newly added vectorization test for the geodesic_equation)  (this may justify some problems we were having with this metric)

* specific exp and log solvers are created for `_InvariantMetricMatrix` because `exp` and `log` need a pre or post processing
   * this removes the need to overwrite `exp`, `log`, and `geodesic` (the ones in `Connection` work properly)

* default solvers are changed in `_InvariantMetricMatrix`:
  *  new defaults seem more robust (NB: some xfails could be removed from tests) and do not require `autodiff` (i.e. this metric can now be used with `numpy`! similar as in `information_geometry`)
  * this may interest you @psarahdactyl since it is deeply connected to #1825

Additionally:

* xfails are removed from information geometry and SE with invariant metric tests due to the fixes in the solvers


luisfpereira:
@ninamiolane, I'm merging because this fixes several issues. We can address any suggestion coming from you in a future PR.

geomstats/geometry/invariant_metric.py
```diff
@@ -7,13 +7,81 @@
...
+        metric. For a right-invariant metric, :math:`dR` is used instead of
+        :math:`dL` and :math:`ad^*` is replaced by :math:`-ad^*`. The
+        exponential map is approximated by numerical integration
+        of this equation, with initial conditions :math:`\dot{\gamma}(0)`
+        given by the argument `tangent_vec` and :math:`\gamma(0)` by
+        `base_point`.
+
+        Parameters
+        ----------
+        tangent_vec : array-like, shape=[..., n, n]
```
ninamiolane, line 42: missing space

geomstats/geometry/invariant_metric.py
```diff
@@ -7,13 +7,81 @@
...
+
+        References
+        ----------
+        https://en.wikipedia.org/wiki/Runge–Kutta_methods
+
+        .. [Kolev]   Kolev, Boris. “Lie Groups and Mechanics: An Introduction.”
+                     Journal of Nonlinear Mathematical Physics 11, no. 4, 2004:
+                     480–98. https://doi.org/10.2991/jnmp.2004.11.4.5.
+        """
+        base_point, left_angular_vel = space.metric._pre_exp(tangent_vec, base_point)
```
ninamiolane, line 62: What's pre_exp??

geomstats/geometry/invariant_metric.py
```diff
@@ -7,13 +7,81 @@
...
+        return super().exp(space, left_angular_vel, base_point)
+
+    def geodesic_ivp(self, space, tangent_vec, base_point):
```
ninamiolane, line 65: Add full docstring, with parameter description.

geomstats/geometry/invariant_metric.py
```diff
@@ -7,13 +7,81 @@
...
+        """Geodesic curve for initial value problem."""
+        base_point, left_angular_vel = space.metric._pre_exp(tangent_vec, base_point)
+        return super().geodesic_ivp(space, left_angular_vel, base_point)
+
+
+class InvariantMetricMatrixLogODESolver(LogODESolver):
+    """A log solver adapted to _InvariantMetricMatrix."""
+
+    def log(self, space, point, base_point):
+        """Logarithm map."""
```
ninamiolane, line 75: Docstring missing: can you explain why the super().log is not enough for this log function?
Also, which is the space.metric.left (left-metric) hardcoded: does it only work for the left invariant metrics? If so, it should be indicated in the docstring?

^^^Could you put these explanations in the docstrings?

geomstats/geometry/invariant_metric.py
```diff
@@ -55,9 +123,11 @@ def __init__(self, space, metric_mat_at_identity=None, left=True):
         self._instantiate_solvers()
 
     def _instantiate_solvers(self):
```
ninamiolane, line 125: Add docstring, even for a private function, to help readability (we are having problems using the solvers in the lab, because we are having problems understanding the code). In particular, explain the choice of the hardcoded parameters. n_nodes, use_jac, etc.

geomstats/geometry/invariant_metric.py
ninamiolane: why is it "left"? does it only works for left-invariant metric?

geomstats/geometry/invariant_metric.py
```diff
@@ -778,21 +745,28 @@ def parallel_transport(
 
         def acceleration(state, time):
             """Compute the right-hand-side of the parallel transport eq."""
```
ninamiolane, line 747: Add parameters description 🙏

geomstats/geometry/invariant_metric.py
```diff
@@ -778,21 +745,28 @@ def parallel_transport(
...
-            omega, zeta = state[1:]
-            gam_dot, omega_dot = self.geodesic_equation(state[:2], time)
+            omega = state[..., 1, :, :]
```
ninamiolane, line 748: omega and zeta are not very meaningful: any chance we could rename these variables by the concepts they represent (not by a greek letter found in one paper)?

geomstats/numerics/geodesic.py
```diff
@@ -62,26 +71,33 @@ class ExpODESolver(ExpSolver):
     """
 
     def __init__(self, integrator=None):
+        super().__init__(solves_ivp=False)
+
         if integrator is None:
             integrator = GSIVPIntegrator()
 
+        self._integrator = None
```
ninamiolane, line 79: what's the difference between _integrator and integrator? Could you add the description of the parameters to help readability?

geomstats/numerics/geodesic.py
```diff
@@ -62,26 +71,33 @@ class ExpODESolver(ExpSolver):
...
         self.integrator = integrator
 
+    @property
+    def integrator(self):
+        """An instance of ODEIVPSolver."""
```
ninamiolane, line 84: should we call it "solver" then? it would help readability, if Solver-type objects are indeed called solvers.

geomstats/numerics/geodesic.py
```diff
@@ -62,26 +71,33 @@ class ExpODESolver(ExpSolver):
...
+        return self._integrator
+
+    @integrator.setter
+    def integrator(self, integrator):
+        """Set integrator."""
```
ninamiolane, line 89: Parameter description

geomstats/numerics/geodesic.py
```diff
@@ -62,26 +71,33 @@ class ExpODESolver(ExpSolver):
...
+        self.solves_ivp = integrator.tchosen
+        self._integrator = integrator
+
     def _solve(self, space, tangent_vec, base_point, t_eval=None):
```
ninamiolane, line 93: Docstring + parameter descriptions

geomstats/numerics/geodesic.py
```diff
@@ -443,15 +404,12 @@ def log(self, space, point, base_point):
...
-        tangent_vec = gs.reshape(res.x, base_point.shape)
-
-        return tangent_vec
+        return gs.reshape(res.x, base_point.shape)
 
 
-class _LogShootingSolverUnflatten(
-    _LogBatchMixins, _GeodesicBVPFromExpMixins, LogSolver
-):
+class _LogShootingSolverUnflatten(_LogBatchMixins, LogSolver):
```
ninamiolane, line 410: Docstring? At this point, I'm confused about why there are so many log classes.

geomstats/numerics/geodesic.py
```diff
@@ -530,7 +490,11 @@ def _create_grid(self):
         return gs.linspace(0.0, 1.0, num=self.n_nodes)
 
     def _default_initialization(self, space, point, base_point):
-        point_0, point_1 = base_point, point
+        if point.ndim == 1:
```
ninamiolane, line 493: Docstring

ninamiolane: (not necessarily to explain what the function does, that's explicit from the code, but *why* we need that)

geomstats/numerics/geodesic.py
```diff
@@ -540,8 +504,8 @@ def _default_initialization(self, space, point, base_point):
         return gs.vstack([pos_init, vel_init])
 
     def _boundary_condition(self, state_0, state_1, space, point_0, point_1):
-        pos_0 = state_0[: space.dim]
-        pos_1 = state_1[: space.dim]
+        pos_0 = state_0[: point_0.shape[0]]
```
ninamiolane, line 507: Docstring

geomstats/numerics/geodesic.py
```diff
@@ -608,7 +572,7 @@ def _jacobian(self, _, raveled_state, space):
     def _solve(self, space, point, base_point):
         bvp = lambda t, state: self._bvp(t, state, space)
```
ninamiolane, line 573: Docstring

geomstats/numerics/geodesic.py
```diff
@@ -680,8 +644,9 @@ def path(t):
         return path
 
     def _simplify_log_result(self, result, space):
-        _, tangent_vec = gs.reshape(gs.transpose(result.y)[0], (2, space.dim))
-        return tangent_vec
+        return gs.reshape(result.y[..., 0], (2,) + space.shape)[1]
 
     def _simplify_result_t(self, result, space):
-        return gs.transpose(result[: space.dim, :])
+        return gs.moveaxis(
```
ninamiolane, line 650: Docstring? How is it simplified and why?

geomstats/numerics/geodesic.py
```diff
@@ -680,8 +644,9 @@ def path(t):
         return path
 
     def _simplify_log_result(self, result, space):
-        _, tangent_vec = gs.reshape(gs.transpose(result.y)[0], (2, space.dim))
-        return tangent_vec
+        return gs.reshape(result.y[..., 0], (2,) + space.shape)[1]
```
ninamiolane, line 647: Docstring? How is it simplified and why?

ninamiolane (commented):
Thanks! Good idea to merge to solve the bugs asap. Very exciting to see the solvers take a proper format!

My only thing is that I have the hardest time reading the code. There are many classes with similar names, many attributes with similar names, and missing docstrings. Sometimes, the code is somewhat explicit that I can understand *what* the code does, but does *why* we need that method/function, why we need another class, and navigating from one class to the other to try to patch things together really hinders readability (to me).

--> Could you add explanations to the docstrings? Even for the private functions. Possibly copy-pasting elements from the PR descriptions? Thank you!

geomstats/geometry/invariant_metric.py
luisfpereira: Where? It seems correct.

geomstats/geometry/invariant_metric.py
luisfpereira: This comes from the fact that the geodesic equation is "written" entirely in the Lie algebra". Therefore, before calling `geodesic_equation` we need to translate tangent vectors there. This probably just means we need to improve the design of this class such that we do not have to create "special solvers".

geomstats/geometry/invariant_metric.py
luisfpereira: I'll add it. (In any case, notice it is exactly the same as the parent docstring (the reason I haven't added at first place).)

geomstats/geometry/invariant_metric.py
luisfpereira: Relates with above answer: `geodesic_equation` is written in the `LieAlgebra`, we are finding a solution that lives on the `LieAlgebra`, we need to translate it to the `LieGroup`.

Notice `left` is not hard-coded. It is a boolean that if `True` means "left", else "right". This boolean is controlled at instantiation time.

geomstats/geometry/invariant_metric.py
luisfpereira: I'll do it. (Notice this was fixed in this PR, not added now.)

geomstats/geometry/invariant_metric.py
luisfpereira: Same as above.

geomstats/numerics/geodesic.py
luisfpereira: This is a `Python` thing. We want to make an action when doing `solver.integrator = SomeIntegrator()`. I'll check if I can simplify this part to avoid confusion.

geomstats/numerics/geodesic.py
luisfpereira: Bad naming here. I probably need to call `ODEIVPSolver` -> `ODEIVPIntegrator` instead, since this object `.integrate`.

geomstats/numerics/geodesic.py
luisfpereira: We have an algorithm: log shooting. From a user perspective we have one solver `LogShootingSolver`. We have two implementations of this algorithm that change only on the way they handle vectorization. The choice of the version from a user perspective is done with `flatten`:

>     flatten : bool
>        If True, the optimization problem is solved together for all the points.


If `True`, only one optimization problem is solved each time, even in batch mode. This is implemented by `_LogShootingSolverFlatten`. The name comes from the fact that the input is `flattened` before the optimization.

If `False`, an optimization problem is solved for each point/base_point in the batch.

The implementations are sufficiently different to deserve different classes (NB: this detail is hidden from the users.)

The flatten implementation (original implementation in geomstats) is less robust for the cases where the difficulty of each log problem in the batch is very different (number of iterations always depend on the worst case). It may be faster for similarly difficult logs (depends on the space).

geomstats/numerics/geodesic.py
luisfpereira: Can you be more explicit on the part that is not clear?

This relates with the algorithm we use for solving this problem. Since we solve the ODE as a boundary value problem, we need to provide an initialization for the values at each node.

PR #1946, opened 2024-02-05
Add FlatRiemannianMetric and fix EuclideanMetric inheritance

luisfpereira:
This PR splits `EuclideanMetric` into `FlatRiemannianMetric` and the usual `EuclideanMetric`. The first is a flat metric whose `metric_matrix` can be different from the usual Euclidean metric. The second assumes the metric matrix is `gs.eye(dim)`.

See next PRs for further justification for this split.


geomstats/geometry/euclidean.py
```diff
@@ -218,3 +195,195 @@ def parallel_transport(
...
+        -------
+        radius : array-like, shape=[...,]
+            Injectivity radius.
+        """
+        radius = gs.array(math.inf)
+        return repeat_out(self._space.point_ndim, radius, base_point)
+
+
+class StandardEuclideanMetric(EuclideanMetric):
+    """Class for the standard Euclidean metric."""
```
ninamiolane, line 312: Explain what is "standard", explain the difference with the EuclideanMetric, eg in a See Also section in the docstring.

ninamiolane (changes requested):
Could the distinction EuclideanMetric and StandardEuclideanMetric be hidden for the user? As a user, if I want the standard euclidean metric, I will search the library for EuclideanMetric, see that there is something, and try to use that.

Could one of them be private, and only a single class `EuclideanMetric` be accessible, such that this class dispatches between the Standard and the Matrix euclidean metrics?

At the very least the docstrings should explain the difference between the two.

luisfpereira:
@ninamiolane, I'm using `FlatRiemannianMetric` term and keeping `EuclideanMetric`. Now it should be easier to disambiguate.

PR #1947, opened 2024-02-05
Add new flat Matrices metrics

luisfpereira:
This PR builds on top of #1946 to add new metric matrices for `Matrices` (and other matrix vector spaces). Explicitly, we add `MatricesDiagMetric`.  With the addition of diffeomorphisms to the `EuclideanSpace`, we  implicitly add the ability to easily implement a `PullbackDiffeoMetric` from an `Euclidean` space equipped with a non-standard `EuclideanMetric`.

Next PR will show application of these concepts.

NB: merging this PR before #1946 also merges #1946.


geomstats/geometry/diffeo.py
```diff
@@ -332,3 +332,38 @@ def inverse_tangent_diffeomorphism(
             image_point = base_point
 
         return tangent_vec
+
+
+class VectorSpaceDiffeo(Diffeo):
```
ninamiolane, line 337: Missing parameter description

ninamiolane (changes requested):
Most docstrings are missing parameter descriptions. I'll review after they are added, because it'll be easier to read and understand the code with them.

luisfpereira:
@ninamiolane, docstrings updated. Could you please review now?

PR #1949, opened 2024-02-06
Add SymmetricHollowMatrices and corresponding permutation-invariant metric

luisfpereira:
This PR builds on top of #1946 to add the matrix vector space `SymmetricHollowMatrices` and a permutation-invariant metric on it `HollowMatricesPermutationInvariantMetric`.

It also fixes small inheritance issues in `geometry.base`. In particular, deletes an abstract method which is also defined at manifold level in order to allow proper multiple inheritance from `LevelSet` and `MatrixVectorSpace`.

This is a first step to bring in pullback metrics on `Cor+` via the off-log diffeomorphism.

NB: merging this PR before #1946 also merges it.

PR #1950, opened 2024-02-06
Add off-log metrics on Cor+

luisfpereira:
This PR builds on top of #1949 to add off-log metrics on full-rank correlation matrices `Cor+`. The diffeomorphism `OffLogDiffeo` is also added, since `OffLogMetric` is implemented as a `PullbackDiffeoMetric`.

NB: merging this PR also merges #1949.

### DIFF
--- a/geomstats/geometry/full_rank_correlation_matrices.py
+++ b/geomstats/geometry/full_rank_correlation_matrices.py
@@ -6,8 +6,8 @@ Lead authors: Yann Thanwerdas and Olivier Bisson.
 References
 ----------
 .. [T2022] Yann Thanwerdas. Riemannian and stratified
-geometries on covariance and correlation matrices. Differential
-Geometry [math.DG]. Université Côte d'Azur, 2022.
+    geometries on covariance and correlation matrices. Differential
+    Geometry [math.DG]. Université Côte d'Azur, 2022.
 """
 
 import logging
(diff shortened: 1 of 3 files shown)


geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -337,3 +350,187 @@ def __init__(self, space):
...
+def off_map(matrix):
+    """Subtract diagonal to a matrix."""
+    return matrix - Matrices.to_diagonal(matrix)
+
+
+class OffLogDiffeo(Diffeo):
+    """Off-log diffeomorphism from Cor+ to Hol.
+
+    A diffeomorphism between full-rank correlation matrices Cor+ and
+    symmetric hollow matrices Hol.
```
ninamiolane, line 364: Add a reference, or an equation, or an explanation of what off-log is.

Parameter description missing.

ninamiolane (changes requested):
Same: I feel that the code is missing crucial information for it to be readable by someone external to the research. Eg, what is "offlog"? what is "off"? Do we have an equation, a reference to a paper?

Functions are missing parameter descriptions in their docstrings, I'll review once this is added.

luisfpereira:
@ninamiolane, I've updated the docstrings. Hopefully the code is easier to understand now. I should also add we are writing a paper with these implementations, which justifies the missing references at some points. I'm pointing out to the relevant chapter of Yann's thesis when possible.

@olivierbisson, could you please think about the docstrings of the private methods of `OffLogDiffeo`?

geomstats/geometry/symmetric_matrices.py
```diff
@@ -133,3 +135,208 @@ def matrix_representation(basis_representation):
...
+        +\beta \operatorname{Sum}\left(X^2\right)
+        +\gamma \operatorname{Sum}(X)^2
+
+    Check out chapter 8 of [T2022]_ for more details.
+
+    References
+    ----------
+    .. [T2022] Yann Thanwerdas. Riemannian and stratified
+    geometries on covariance and correlation matrices. Differential
+    Geometry [math.DG]. Université Côte d'Azur, 2022.
```
ninamiolane, line 261: Thanks for the equations and ref, super helpful.

Need parameter descriptions

geomstats/geometry/symmetric_matrices.py
```diff
@@ -133,3 +135,208 @@ def matrix_representation(basis_representation):
...
+
+        cond_1 = 2 * alpha + (n - 2) * beta
+        cond_2 = alpha + (n - 1) * (beta + n * gamma)
+        if cond_1 < gs.atol or cond_2 < gs.atol:
+            raise ValueError(
+                f"Inequalities should be greater than 0, but: {cond_1} and {cond_2}."
+                "Check thanwerdas2022 theorem 8.7"
+            )
+
+    def _quadratic_form(self, tangent_vec):
```
ninamiolane, line 307: Need docstring

geomstats/geometry/symmetric_matrices.py
```diff
@@ -133,3 +135,208 @@ def matrix_representation(basis_representation):
...
+
+    def __init__(self, space, alpha=1.0, beta=1.0, gamma=1.0):
+        self._check_params(space, alpha, beta, gamma)
+        super().__init__(space=space)
+        self.alpha = alpha
+        self.beta = beta
+        self.gamma = gamma
+
+    @staticmethod
+    def _check_params(space, alpha, beta, gamma):
```
ninamiolane, line 272: Thx for the eqn, super helpful.

Need parameter descriptions (even though, I agree, it seems clear if we scroll up)

geomstats/test_cases/geometry/euclidean.py
```diff
@@ -21,35 +21,37 @@ def test_identity_belongs(self, atol):
         self.test_belongs(self.space.identity, gs.array(True), atol)
 
 
-class EuclideanMetricTestCase(RiemannianMetricTestCase):
+class FlatRiemannianMetricTestCase(RiemannianMetricTestCase):
```
ninamiolane, line 24: Awesome!

ninamiolane (approved):
Awesome! I really like the addition of the flat metric, but I would recommend putting it in a separate file, not in euclidean.py (because it's very confusing to hide it there, and it seems to say that we think that flat = euclidean).

I have a comment, regarding the design of MatrixVectorSpace, which I'd recommend to make private to avoid confusing users (it's bit of a pain to remember that there is MatrixVectorSpace and VectorSpace available).

Some docstrings are missing parameter descriptions, or details.

Otherwise, good to merge! Thank you so much, this is incredible!

geomstats/geometry/euclidean.py
```diff
@@ -86,9 +107,69 @@ def metric_matrix(self, base_point=None):
...
+            Point on the manifold.
+
+        Returns
+        -------
+        gamma : array-like, shape=[..., dim, dim, dim]
+            Christoffel symbols, with the contravariant index on
+            the first dimension.
+        """
+        if self._space.point_ndim > 1:
+            raise NotImplementedError("The Christoffel symbols are not implemented.")
```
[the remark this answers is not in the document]
luisfpereira: Are you sure about this? (Meanwhile I'll move this method to the `EuclideanMetric` to avoid the introduction of any conceptual error, as I'll merge #1946 now.)

According to our code, `christoffels` are obtained from a sum of products between `cometric_matrix` and `inner_product_derivative_matrix` (both obtained from `metric_matrix`). In a flat metric, by definition, the `metric_matrix` is independent of the `base_point`, so the `inner_product_derivative_matrix` is null. Therefore, the christoffels should be zero.

The geodesic equation implementation also suggests they should be zero.

geomstats/geometry/full_rank_correlation_matrices.py
```diff
@@ -26,7 +29,17 @@
 )
 from geomstats.geometry.pullback_metric import PullbackDiffeoMetric
 from geomstats.geometry.quotient_metric import QuotientMetric
-from geomstats.geometry.spd_matrices import CholeskyMap, SPDAffineMetric, SPDMatrices
+from geomstats.geometry.spd_matrices import (
+    CholeskyMap,
+    SPDAffineMetric,
+    SPDMatrices,
+    SymMatrixLog,
```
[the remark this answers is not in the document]
luisfpereira: It is a `Diffeo` instance.

PR #1953, opened 2024-02-13
Adding the ClosedDiscreteCurves class to the discr

MMalikT:
ete_curves.py file. This class is used to represent closed discrete curves.

## Checklist

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -176,6 +176,25 @@ def insert_zeros(array, axis=-1, end=False):
     return gs.concatenate((first, second), axis=-array_ndim)
 
 
+def insert_starting_point_at_end(point):
+    """Add a first point of curve as last point.
+
+    Adds first point of curve to end for closed curves case.
+
+    Parameters
+    ----------
+    point : array-like, shape=[..., k_sampling_points, ambient_dim]
+
+    Returns
+    -------
+    cld_point : array-like, shape=[..., k_sampling_points + 1, ambient_dim]
+    """
+    if gs.all(point[..., 0, :] == point[..., -1, :]):
+        return point
+    else:
+        return gs.concatenate((point, point[..., 0, :][..., None, :]), axis=-2)
+
+
 class DiscreteCurvesStartingAtOrigin(NFoldManifold):
     r"""Space of discrete curves modulo translations.
 
@@ -187,7 +206,7 @@ class DiscreteCurvesStartingAtOrigin(NFoldManifold):
     space M, with non-vanishing derivative.
 
     .. math::
-        Imm(I, M)=\{c \in C^{\infty}(I, M) \|c'(s)\|\neq 0 \forall s \in I \},
+        Imm(I, M)={c \in C^{\infty}(I, M) \|c'(s)\|\neq 0 \forall s \in I },
 
     where the interval of parameters I is taken to be I = [0, 1]
     without loss of generality.
@@ -204,6 +223,7 @@ class DiscreteCurvesStartingAtOrigin(NFoldManifold):
 
     def __init__(self, ambient_dim=2, k_sampling_points=10, equip=True):
         ambient_manifold = Euclidean(ambient_dim)
+        self.closed = False
         super().__init__(ambient_manifold, k_sampling_points - 1, equip=equip)
 
         self._quotient_map = {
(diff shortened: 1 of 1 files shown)


MMalikT:
Unfortunately, my description wasn't included in the PR. Sorry about that. Let me add it down here: The main contribution is to add the class ClosedDscreteCurvesStartingAtOrigin to the discrete_curves.py file. In my understanding, inthe closed curves case, the SRV transform as an extra point of information and hence I changed a little bit the SRV transform part as well to take this into account. @luisfpereira @alebrigant @ninamiolane Please let me know what you think.

geomstats/geometry/discrete_curves.py
```diff
@@ -356,29 +376,212 @@ def normalize(self, point):
...
+
+    def __init__(self, ambient_dim=2, k_sampling_points=10, equip=True):
+        ambient_manifold = Euclidean(ambient_dim)
+        self.closed = True
+        super().__init__(ambient_manifold, k_sampling_points - 1, equip=equip)
+
+        self._quotient_map = {
+            (SRVMetric, "rotations"): (
+                SRVRotationBundle,
+                SRVRotationQuotientMetric,
```
alebrigant, line 413: This is no longer the right way to define the quotient maps, I believe we no longer have new classes for each quotient metric. See how it is defined for (open) discrete curves.

geomstats/geometry/discrete_curves.py
```diff
@@ -204,6 +223,7 @@ class DiscreteCurvesStartingAtOrigin(NFoldManifold):
 
     def __init__(self, ambient_dim=2, k_sampling_points=10, equip=True):
         ambient_manifold = Euclidean(ambient_dim)
+        self.closed = False
```
alebrigant, line 226: Shouldn't this be used to call `ClosedDiscreteCurvesStartingAtOrigin` if True ?

geomstats/geometry/discrete_curves.py
```diff
@@ -356,29 +376,212 @@ def normalize(self, point):
...
+        """
+        sampling_times = gs.linspace(0.0, 1.0, self.k_sampling_points)
+
+        initial_point = self._sphere.random_point(n_samples)
+        initial_tangent_vec = self._sphere.random_tangent_vec(initial_point)
+
+        point = self._sphere.metric.geodesic(
+            initial_point, initial_tangent_vec=initial_tangent_vec
+        )(sampling_times)
+        return self.projection(point)
```
alebrigant, line 506: Does this really generate a curve that "looks" closed ? It seems like this is the same code as the one used for open curves.

geomstats/geometry/discrete_curves.py
```diff
@@ -404,11 +607,13 @@ def diffeomorphism(self, base_point):
             SRV representation.
         """
         ndim = self._point_ndim
-        base_point_with_origin = insert_zeros(base_point, axis=-ndim)
-
+        base_point_with_origin = self.space.insert_origin(base_point)
+        if self.space.closed:
+            base_point_with_origin = insert_starting_point_at_end(
+                base_point_with_origin
+            )
```
alebrigant, line 614: If this increases the number of sampling points of one, you may need to change some normalizations constants that depend on this number of sampling points.

geomstats/geometry/discrete_curves.py
```diff
@@ -444,15 +649,16 @@ def inverse_diffeomorphism(self, image_point):
         curve : array-like, shape=[..., k_sampling_points - 1, ambient_dim]
             Discrete curve.
         """
-        image_point_norm = self.ambient_manifold.metric.norm(image_point)
-
-        dt = 1 / (self.k_sampling_points - 1)
+        image_point_norm = self.space.ambient_manifold.metric.norm(image_point)
+        dt = 1 / (self.space.k_sampling_points - 1)
```
alebrigant, line 653: This might need to be updated, see previous comment.

alebrigant (changes requested):
Thanks for this PR ! I think the tests are not passing so these are just preliminary comments, waiting for other modifs to make the tests pass. 

A general comment on the structure: maybe it would be best to have 3 classes: `OpenDiscreteCurvesStartingAtOrigin` and `ClosedDiscreteCurvesStartingAtOrigin` and `DiscreteCurvesSartingAtOrigin` would call one or the other depending on the attribute `self.closed` ? It also seems unnecessary to copy a big portion of the code of open curves into the class of closed curves, maybe there should be some inheritance there. What do you think @luisfpereira ? In any case, there is still some work to make the tests pass !

luisfpereira:
@MMalikT, I've done some changes in the PR, to make everything consistent, though the results are disappointing (i.e. curves open when taking geodesics). In particular:

* I've changed the representation of closed curves to be of shape `(k_sampling_points - 2, ambient_dim)`. As I state in the docstrings, this is done because with this representation `delta = 1 / (k_sampling_points - 1)` is valid (both for open and closed curves)

* I've fixed small things in SRV to ensure we recover the same points and tangent vectors after using e.g. `inverse_diffeo` after `diffeo`


Please let me know what you think.

PR #1954, opened 2024-02-13
Add bregman divergence class

dralston78:
## Checklist

- [x] I added appropriate unit tests.
- [x] I made sure the code passes all unit tests. (refer to comment below)
- [x] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#coding-style-guidelines) and API.

## Description
I wrote a very basic implementation of Bregman divergence, that can hopefully be used to define dually flat manifolds from a given Bregman generator.  I also added an extremely simple unit test.  

I marked this pull request as a draft because I expect it will need to be largely refactored when implementing the more complex structures associated with statistical manifolds.

### DIFF
--- a/geomstats/information_geometry/bregman_divergence.py
+++ b/geomstats/information_geometry/bregman_divergence.py
@@ -1,0 +1,56 @@
+import geomstats.backend as gs
+
+class BregmanDivergence:
+    """Bregman divergence on a manifold."""
+
+    def __init__(self, space, potential_function):
+        r"""Initialize the Bregman divergence.
+
+        Parameters
+        ----------
+        space : Manifold
+            Manifold to compute the Bregman divergence on.
+        potential_function : callable
+            Scalar convex twice differentiable function.
+        """
+        self.space = space
+        self.potential_function = potential_function
+        self._check_coordinates(space)
+
+    def _check_coordinates(self, space):
+        """Check the manifold is defined in intrinsic coordinates."""
+        if not space.intrinsic:
+            raise ValueError(
+                "Space should be equipped with intrinsic coordinates for sampling"
+                "points to compute divergence."
+            )
+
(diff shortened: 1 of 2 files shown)


geomstats/information_geometry/bregman_divergence.py
```diff
@@ -0,0 +1,57 @@
...
+import geomstats.backend as gs
+
+
+class BregmanDivergence:
+    """Bregman divergence on a manifold."""
+
+    def __init__(self, space, potential_function):
+        r"""Initialize the Bregman divergence.
+
+        Parameters
```
ninamiolane, line 12: NIT: Parameters go in the class docstrings, by convention.

ninamiolane (commented):
LGTM! 

I'm trying to think about which objects would make the most sense.

Either what you have, ie, a class for the `BregmanDivergence`, as you do here and aswe originally discussed.
The problem with this one is that `divergence.divergence()` is not a very intuitive syntax.

Or, two classes:
- `class BregmanMetric(RiemannianMetric)` : a Riemannian metric that comes from a bregman divergence
- `class DualConnections(Connection)` : the dual connection that comes from a bregman divergence.
but the problem with this design is that the BregmanMetric and the DualConnections are not linked in any way, they're completely independent.

Or:
- `class BregmanManifold(Manifold)`: a manifold equipped with a bregman divergence, and associated bregman riemannian metric and dual connections.

We could call this manifold a `StatisticalManifold` or an `InformationManifold`. In the latter case, we could just put the machinery of the bregman divergence in the current implementation (Note: it's a mixin...not a class):
https://github.com/geomstats/geomstats/blob/main/geomstats/information_geometry/base.py

I think the `BregmanManifold/StatisticalManifold/InformationManifold` option makes the most sense.

PR #1955, opened 2024-02-14
Move finite differences to numerics

luisfpereira:
This PR moves some already-existing finite differences code to numerics (it has been implemented in the context of discrete curves, but the code is actually generic). This is motivated by the use of these functions in other context. New tests have been added (all functions are fully covered). Small fixes have been done in discrete curves due to a small bug in the finite differences code.


luisfpereira:
@ninamiolane, this PR just moves code around to the place it should be, fixes a bug, and adds tests. I'm merging as I need it to move on with discrete surfaces. Please let me know afterwards if you have any comment.

PR #1957, opened 2024-02-14
Add path-straightening algorithm

### LINKED ISSUE #1926: Add path-straightening algorithm
The path-straightening algorithm is already implicitly implement in the `discrete_surfaces`. Move it out, such that we can use it in another Riemannian manifolds and, more importantly, test it against closed-form solutions.

luisfpereira:
This PR generalizes implementation of  path-straightening algorithm already available on `discrete_surfaces`.

The algorithm minimizes the Riemannian geometry of a path.

Given the discretized nature of the path, I've created `numerics.path` with utils to compute the path energy of a uniformly-sampled path. Notice that there's a numerical analysis choice here: what finite-difference scheme to use? This implementation uses forward differences, but can easily be extended (not the same as @emmanuel-hartman original code).

Given the discretized nature of the output (the geodesic path is a collection of points), I've created `UniformlySampledPath` to interpolate between these points, such that `geodesic` is now a callable (and not an array). This relies on a 1d interpolator.

I've added a simple 1d linear interpolator in `numerics.interpolation`. Main reason: speed. We may use scipy machinery for more complex interpolation schemes.

I think may interest you @alebrigant and @MMalikT.

closes #1926

P.S. only missing tests, but already checked locally against closed-form solutions (works beautifully! though accuracy is not fantastic)

### DIFF
--- a/geomstats/numerics/path.py
+++ b/geomstats/numerics/path.py
@@ -2,7 +2,7 @@
 
 import geomstats.backend as gs
 from geomstats.numerics.finite_differences import forward_difference
-from geomstats.numerics.interpolation import LinearInterpolator1D
+from geomstats.numerics.interpolation import UniformUnitIntervalLinearInterpolator
 
 
 class UniformlySampledPathEnergy:
@@ -22,9 +22,9 @@ class UniformlySampledPathEnergy:
         energy : array-like, shape=[...,]
             Path energy.
         """
-        return self.path_energy(space, path)
+        return self.energy(space, path)
 
-    def path_energy_per_time(self, space, path):
+    def energy_per_time(self, space, path):
         """Compute Riemannian path enery per time.
 
         Parameters
@@ -46,9 +46,9 @@ class UniformlySampledPathEnergy:
         return space.metric.squared_norm(
             tangent_vecs,
             path[..., :-1, *point_ndim_slc],
-        ) / (2 * n_time)
+        ) / (2 * (n_time - 1))
 
-    def path_energy(self, space, path):
+    def energy(self, space, path):
         """Compute Riemannian path energy.
 
         Parameters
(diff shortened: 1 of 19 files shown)


ninamiolane (approved):
LGTM, thank you so much!

High-level comments:
- there are some docstring parameters' missing descriptions: could you add them?
- do we need two more classes in path, or would it be possible/interesting to make the code a bit more lightweight and use functions instead?
- need to add tests

geomstats/numerics/path.py
```diff
@@ -0,0 +1,94 @@
+"""Discrete-path related machinery."""
+
+import geomstats.backend as gs
+from geomstats.numerics.finite_differences import forward_difference
+from geomstats.numerics.interpolation import LinearInterpolator1D
+
+
+class UniformlySampledPathEnergy:
```
luisfpereira, line 8: Rename to `UnitUniformIntervalLinearInterpolator`?

ninamiolane:
The tests' failures are on discrete_surfaces:
ERROR tests/tests_geomstats/test_geometry/test_discrete_surfaces.py - geomstats.exceptions.AutodiffNotImplementedError: Minimization with 'autodiff' requires automatic differentiation.Change backend via the command export GEOMSTATS_BACKEND=pytorch in a terminal
ERROR tests/tests_geomstats/test_geometry/test_pullback_metric.py - geomstats.exceptions.AutodiffNotImplementedError: Minimization with 'autodiff' requires automatic differentiation.Change backend via the command export GEOMSTATS_BACKEND=pytorch in a terminal
ERROR tests/tests_geomstats/test_numerics/test_log.py - geomstats.exceptions.AutodiffNotImplementedError: Minimization with 'autodiff' requires automatic differentiation.Change backend via the command export GEOMSTATS_BACKEND=pytorch in a terminal

PR #1959, opened 2024-02-17
Adding statistical metric

dralston78:
## Checklist

## Description
I added a first attempt at a `StatisticalMetric` class.  Given a divergence, a metric tensor, connection, dual connection, and Amari cubic tensor can be induced to create a statistical manifold, a special type of Riemannian manifold with extra structure.  

## Additional context
This is just a first stab at how to code this structure within the Geomstats library, to share with collaborators.  Because of this, the PR is marked as a draft.  The documentation needs to be reformatted.  

Also, for some reason the `StatisticalMetric` class only works with the `autograd` back end -- I had a lot of trouble trying to get it to work with the `pytorch` backend.  

Looking forward to discussing here/on Slack!


geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
+"""Statistical metric to equip a statistical manifold."""
+
+import geomstats.backend as gs
+from geomstats.geometry.riemannian_metric import RiemannianMetric
+
+
+class StatisticalMetric(RiemannianMetric):
+    """
+    Defines statistical metric and connection induced from a divergence.
```
ninamiolane, line 9: Respect docstring conventions laid out here: https://geomstats.github.io/contributing/index.html#writing-docstrings

Eg: Imperativ form for verbs --> Define not Defines
Put verb directly after """

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Uses definitions provided in Nielson's An Elementary Introduction to
+    Information Geometry, Theorem 4 on page 15 (https://arxiv.org/abs/1808.08271)
```
ninamiolane, line 12: docstring missing parameter description (see other classes' docstring in geomstats)

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def __init__(self, dim, divergence):
+        self.dim = dim
+        self.divergence = self._unpack_tensor(divergence)
+
+    def _unpack_tensor(self, func):
```
ninamiolane, line 19: please write docstring even for private functions

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+        def wrapper(tensor):
+            return func(tensor[..., : self.dim], tensor[..., self.dim :])
+
+        return wrapper
+
+    def metric_matrix(self, base_point):
```
ninamiolane, line 25: docstring missing parameter descriptions

ninamiolane: Give reference of the paper you mention in each docstring, using

References
-----------

(see in other places in geomstats for how References is used)

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+        """Definition on line (51) on page 14."""
+        hess = gs.autodiff.hessian(self.divergence)
+        base_point_pair = gs.concatenate([base_point, base_point])
+        return -1 * hess(base_point_pair)[: self.dim, self.dim :]
+
+    def divergence_christoffels(self, base_point):
```
ninamiolane, line 31: Ditto

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+        """Definition on line (52) on page 14."""
+        hess = gs.autodiff.hessian(self.divergence)
+        jac_hess = gs.autodiff.jacobian(hess)
+        base_point_pair = gs.concatenate([base_point, base_point])
+        return -1 * jac_hess(base_point_pair)[:2, :2, 2:]
+
+    def dual_divergence_christoffels(self, base_point):
```
ninamiolane, line 38: Ditto

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,49 @@
...
+        """Definition on line (53) on page 14."""
+        hess = gs.autodiff.hessian(self.divergence)
+        jac_hess = gs.autodiff.jacobian(hess)
+        base_point_pair = gs.concatenate([base_point, base_point])
+        return -1 * jac_hess(base_point_pair)[:2, 2:, 2:]
+
+    def amari_divergence_tensor(self, base_point):
```
ninamiolane, line 45: Ditto.

geomstats/information_geometry/statistical_metric.py
ninamiolane: Also, instead of giving the page where we should look for the equation, just write the equation in the docstring, and start the docstring by r"""

--> on the website, the docstring will show the equation in latex!

Eg. https://geomstats.github.io/api/geomstats.information_geometry.html#geomstats.information_geometry.fisher_rao_metric.FisherRaoMetric

tests/tests_geomstats/test_information_geometry/test_statistical_metric.py
```diff
@@ -0,0 +1,72 @@
+import pytest
+
+import geomstats.backend as gs
+from geomstats.information_geometry.statistical_metric import StatisticalMetric
+from geomstats.test.test_case import TestCase, autograd_only
+
+
+@autograd_only
+@pytest.mark.slow
+class TestStatisticalMetric(TestCase):
```
ninamiolane, line 10: Should inherit from RiemannianMetricTestCase

ninamiolane: @luisfpereira can you help @dralston78 with the tests?

ninamiolane (changes requested):
Very nice! I made some cosmetic comments on the statistical_metric.py, but I think the main ideas are perfectly in place.

@luisfpereira can you have a look at the tests' code?

geomstats/information_geometry/statistical_metric.py
ninamiolane: Can the docstring also explain the difference between this metric and the Fisher-Rao metric?

And maybe add a section in the docstring:

See Also
--------
FisherRaoMetric

geomstats/information_geometry/statistical_metric.py
ninamiolane: Are these the Christoffel symbols associated with the metric_matrix? If not, let's keep the code like this. If yes, then they should just be called `christoffel`.

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,267 @@
...
+
+        self.dim = space.dim
+        self.divergence = self._unpack_inputs(divergence)
+
+    def __getattr__(self, attr):
+        r"""Built in method to delegate attribute access.
+
+        Delegate attribute access to the divergence conjugate connection.
+
+        Instanciated to avoid dimond inheritance problem with RiemannianMetric
```
psuarezserrato, line 172: typo , Instantiated.

geomstats/information_geometry/statistical_metric.py
psuarezserrato: Do you mean having this as a class too @ninamiolane ?

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,267 @@
...
+        ----------
+        base_point : array-like, shape=[..., dim]
+            Base point.
+
+        Returns
+        -------
+        tensor : array-like, shape=[..., dim, dim, dim]
+            Amari divergence tensor.
+        """
+        divergence_christoffels = self.divergence_christoffels(base_point)
```
psuarezserrato, line 265: update to new naming conventions, using classes

tests/tests_geomstats/test_information_geometry/test_statistical_metric.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@autograd_only
+# @pytest.mark.slow
+class TestStatisticalMetric(RiemannianMetricTestCase):
+    """
+    Test StatisticalMetric class by verifying expected induced metric, connection,
+    and dual connection properties in the case of Bregman divergence.
+
+    Using properties from Nielson's An Elementary Introduction to
```
psuarezserrato, line 17: Nielsen's

psuarezserrato (commented):
Great job @dralston78 , it's coming together nicely. Needs some help from @luisfpereira to check the tests pass.

geomstats/information_geometry/statistical_metric.py
luisfpereira: I think you are both right, if we make a slight change: by calling it `christoffels`, we will have access to all the machinery implemented in `Connection` (`geodesic_equation`, `exp`, `log`, etc). Though to make it work properly, I think we need to break this class in two, as both connections, when alone, are (affine) connections in their own terms.

So, I suggest to split this object into three objects (names to discuss). First, split it into `DivergenceConnection` and `DualDivergenceConnection`. Then, use these objects to define a third `AlphaConnection` that uses both of the above (this does not care about the way each of the divergence is defined, so no need to connect it with the divergence) e.g.:

```python
class AlphaConnection(Connection):
    def __init__(
        self,
        space,
        primal_connection,
        dual_connection,
        alpha=0,
    ):
        super().__init__(space)
        self.primal_connection = primal_connection
        self.dual_connection = dual_connection

    def christoffels(self, point):
        # see eq. 53
        first_term = (1 + self.alpha) / 2 * self.primal_connection.christoffels(point)
        second_term = (1 - self.alpha) / 2 * self.dual_connection.christoffels(point)

        return first_term - second_term
```


We probably still need to iterate on this, but I believe this `AlphaConnection` or some kind of `ConjugateConnectionPair` is the central object here, as e.g. the Amari-Chentsov tensor or other quantities required to define the statistical manifold come from this pair.


For easier use, in this case we can always create a `DivergenceAlphaConnection` that creates the necessary connections given a divergence.

luisfpereira (changes requested):
Please check my comment regarding `AlphaConnection` (we can speak about tests after the design is closed). @ninamiolane, @psuarezserrato, please take also a look to this and let me know what you think.

geomstats/information_geometry/statistical_metric.py
luisfpereira: So, I think here the main point is that any of these connections is also an (affine) connection when alone, so we definitely need to break this class into two (e.g. `DivergenceConnection`, `DualDivergenceConnection`). Then, by using `christoffels` as method name, we will have access to all the machinery already implemented in connection (e.g. `geodesic_equation`, `exp`, `log`, etc), as @ninamiolane suggested.


Then, we can always create appropriate objects that combine these (affine) connections (not by inheritance, but by composition). For example, the `AlphaConnection` can be created by:

```python
class AlphaConnection(Connection):
    def __init__(
        self,
        space,
        primal_connection,
        dual_connection,
        alpha=0,
    ):
        super().__init__(space)
        self.primal_connection = primal_connection
        self.dual_connection = dual_connection
        self.alpha = alpha 

    def christoffels(self, point):
        # see eq. 53
        first_term = (1 + self.alpha) / 2 * self.primal_connection.christoffels(point)
        second_term = (1 - self.alpha) / 2 * self.dual_connection.christoffels(point)

        return first_term - second_term
```

Alternatively, we could create the `AlphaConnection` by using the concepts of mean connection (which is itself a connection resulting from a combination of connections) and Amari-Chentsov tensor (eq. 52). 


We probably still need to iterate on this, but I believe this `AlphaConnection` or some kind of `ConjugateConnectionPair` is the central object here, as e.g. the Amari-Chentsov tensor or other quantities required to define the statistical manifold can be computed from this pair.


For the divergence case we can always create a `DivergenceAlphaConnection`/`DivergenceInducedConnection` later (for usability) that creates the necessary connections given a divergence.


The main message is that there's a lot of ways of defining all of this and we probably need to keep the idea of composing objects in the back of our mind all the time.

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def __init__(self, space, divergence):
+        super().__init__(space=space)
+        assert space.intrinsic, (
+            "The manifold must be parametrized by an intrinsic coordinate system."
+        )
+        self.dim = space.dim
+        self.divergence = self._unpack_inputs(divergence)
+
+    def _unpack_inputs(self, func):
```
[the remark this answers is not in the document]
dralston78: Just to clarify, this should just be a function outside of any class that is used throughout the script?

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,386 @@
...
+
+    def __init__(self, space, divergence):
+        super().__init__(space=space)
+        assert (
+            space.intrinsic
+        ), "The manifold must be parametrized by an intrinsic coordinate system."
+        self.dim = space.dim
+        self.divergence = unpack_inputs(divergence, self.dim)
+
+    def christoffels(self, base_point):
```
luisfpereira, line 85: Split into two methods: `first_kind_christoffels` and `christoffels`. Do the same below for the dual (I'll have some questions regarding the dual, but I'll leave them for the next iteration).

geomstats/information_geometry/statistical_metric.py
```diff
@@ -0,0 +1,386 @@
...
+        metric_matrix_base_point = self.metric_matrix(base_point)
+        firsk_kind_primal_christoffels = gs.einsum(
+            '...kij,...km->...mij',
+            second_primal_christoffels,
+            metric_matrix_base_point
+        )
+        firsk_kind_dual_christoffels = gs.einsum(
+            '...kij,...km->...mij', second_dual_christoffels, metric_matrix_base_point
+        )
+        return firsk_kind_dual_christoffels - firsk_kind_primal_christoffels
```
luisfpereira, line 386: After the split above, we can make this much simpler (and faster!).

luisfpereira (changes requested):
Let's do a new iteration on this @dralston78.

luisfpereira (changes requested):
After [this iteration](https://github.com/geomstats/geomstats/pull/1959#pullrequestreview-2011773957), let's tackle this more involved one @dralston78. (Please let me know if you disagree with something.) Names are not important for now and the ones I'm suggesting need to be improved.

The goal is to tackle the case of the statistical manifold induced by a Bregman generator (I'll be focusing on the potential function, and not in the divergence, though later we can look to things from the divergence perspective - so no need to delete existing objects relying on the divergence).

1 **Potential function**

I suggest to create a class called `PotentialFunction` with methods `__call__`, `hessian` (eq 63), `third_derivatives` (better naming?) (eq 65).

Reasoning: 
* allow users to hard-code derivatives (e.g. easy to do for `gs.sum(point**4)`)
* avoid making direct calls to autodiff when coding within connections/metrics (derivatives handled in one place)
* easier to test and ensure correctness
 
By default, we can implement those methods using `autodiff`. At init we pass the `potential_function` callable so that we don't have to inherit every time we want to create a new potential function for which we don't want to hard-code derivatives.


2 **Dummy flat-connection**

Create a dummy-flat connection with null (first-kind?) Christoffels (eq 64). (Maybe we don't need to explicitly create this object; living here to recall the potential function-connection is flat)


3 **Levi-Civita connection coming from potential function-metric tensor**

Similarly to already-existing `StatisticalMetric`, create `StatisticalConnectionFromPotentialFunction` inheriting from `RiemannianMetric` calling potential function hessian to define `metric_matrix` and implementing the totally symmetric cubic tensor $C$ (should we call it `cubic_tensor` or `amari_chentsov_tensor`? if mathematically correct, I would call it `amari_chentsov_tensor`; also change in `StatisticalMetric`) given by the third derivatives of the potential function (eq 65) (as in the metric tensor case, simply call method from potential function class). 

Notice that after this implementation, we have (more than) a fully-fledged Levi-Civita connection with geodesics. We also have flat geodesics from the "dummy" connection (2).


4 **Alpha-connection from statistical metric**

Create an alpha-connection class that takes a statistical metric (3) and computes the alpha-connection Christoffels using equation (50). Notice that though this is generic, it can be later used to get the conjugate connection of the potential function-connection (thinking about this result $\left({ }^F \nabla\right)^*={ }^F \nabla^1$). Since it is flat, we can use it to validate our implementation.


**Remarks**

All these objects will 1) provide users with different ways of  defining similar objects, so that they can choose the best for their use case, 2) give us different independent  ways of defining things, which is fantastic for testing.

Do not care much about interfaces: for the inits of these new objects, assume users are passing always the same potential function. We will iterate on this next time.


P.S. I may be missing something, please be critical with the suggestions.

dralston78:
Hi Luis,  as I start to get working on this next iteration, here are some comments/questions I have on your suggestions.  

**Overall comments:**  
- When I start to make these changes, should they be in a separate pull request and a separate branch?  Do we want to merge the current PR and then start a fresh one for the next iteration?  

- In this next iteration we will only consider Bregman-divergence generated metrics/connections.  Does this mean that we will return to a more general structure where we consider geometric structure from any divergence?  Or for the rest of the project will we only consider Bregman-divergence?  I think that the structure from a general divergence is much more powerful: for example, I have a test showing that we can recover the fisher information metric from the KL divergence.  Also, I was just reading about something called a Log-divergence, the structure induced by this divergence can have geometric consequences in portfolio theory.  

**1 Potential Function**
- Should this class inherit from any base class? 

**3 LC connection from potential function**
 - Should we also include a `dual_connection` method in this class?  Note, this `dual_connection` might not be flat (with the same coordinate system on the manifold)!

**4 Alpha Connection from statistical metric**
 - Very minor, but it seems that really this will be not only a connection structure, but more broadly a metric structure.  What if we call it `AlphaMetric`?

luisfpereira:
@dralston78,

> When I start to make these changes, should they be in a separate pull request and a separate branch?

> In this next iteration we will only consider Bregman-divergence generated metrics/connections. Does this mean that we will return to a more general structure where we consider geometric structure from any divergence? Or for the rest of the project will we only consider Bregman-divergence?


Before merging this one, what I would like to have is the (general) geometric structure induced by any divergence AND a particular implementation for the Bregman-divergence that takes advantage of the fact it results from a potential function.

The idea is 2-fold. On one hand, we show we can make computations with general divergences, and, if they have particular structures we can take advantage of, specialized implementations. On the other hand, we can use the specialized implementations to test the general ones. Specialized implementations are also expected to be faster.
In this particular case, I would like to test the statistical manifold implemented using the potential function against the statistical manifold implemented using the divergence.


> Should this class [Potential function] inherit from any base class?

At some point our goal is to create an abstract class from which it inherits from. We can start with just a particular class to check if the idea is worth pursuing. (If this works nicely, I think I'll ask you a similar object for the divergence)


> Should we also include a dual_connection method in this class [LC connection from potential function]?

Probably. Though we should notice the dual connection in this case comes from the alpha-connection, so we need to be careful instantiating the objects. (But for compatibility with the other `StatisticalManifold`, we should probably have it; we can add it later though)


> Very minor, but it seems that really this will be not only a connection structure, but more broadly a metric structure. What if we call it AlphaMetric?

It is a good point. Though I would not say it has a metric structure, but more like it requires the space to be endowed with a metric structure (i.e. it makes use of an existing structure). It is very subtle, but I would ask you soon a change on the computation of the Christoffels that is justified by this. Do you agree with the distinction?

geomstats/information_geometry/statistical_metric.py
```diff
@@ -422,3 +423,50 @@ def amari_divergence_tensor(self, base_point):
         first_primal = self.primal_connection.first_kind_christoffels(base_point)
         first_dual = self.dual_connection.first_kind_christoffels(base_point)
         return first_dual - first_primal
+
+class PotentialFunction:
+    def __init__(self, potential_function):
+        self.potential_function = potential_function
```
luisfpereira, line 429: I would make this private, as this is just a way of creating this object without having to inherit. An alternative would be to directly implement `PotentialFunction` as class and override `__call__` method.

geomstats/information_geometry/statistical_metric.py
```diff
@@ -422,3 +423,50 @@ def amari_divergence_tensor(self, base_point):
...
+        self.hessian_function = self.hessian()
+        self.third_derivative_function = self.third_derivative()
```
luisfpereira, line 431: Delete? having the methods is enough.

geomstats/information_geometry/statistical_metric.py
```diff
@@ -422,3 +423,50 @@ def amari_divergence_tensor(self, base_point):
...
+
+    def __call__(self, x):
```
luisfpereira, line 433: will need to find a better name for `x`.

geomstats/information_geometry/statistical_metric.py
```diff
@@ -422,3 +423,50 @@ def amari_divergence_tensor(self, base_point):
...
+        return self.potential_function(x)
+
+    def hessian(self):
```
luisfpereira, line 436: receive `x` and return result, not function. (Imagine we make an non autodiff-based implementation of the hessian)

geomstats/information_geometry/statistical_metric.py
```diff
@@ -422,3 +423,50 @@ def amari_divergence_tensor(self, base_point):
...
+        return gs.autodiff.hessian(self.potential_function)
+
+    def third_derivative(self):
```
luisfpereira, line 439: Same as for `hessian`: don't mind repeating `gs.autodiff.hessian(self.potential_function)` again.

luisfpereira (commented):
Quick review of the `PotentialFunction`.

PR #1964, opened 2024-02-23
Add method to change the lists in `scalar_product_metric.py`

### LINKED ISSUE #1760: Add method to change the lists in `scalar_product_metric.py`
Follows #1740.

Scaling factor for wrapping of methods of `underlying_metric` in `ScalarProducMetric` is given by verifying if method name belongs to `SQRT_LIST`, `LINEAR_LIST`, `QUADRATIC_LIST`, `INVERSE_LIST`, `INVERSE_QUADRATIC_LIST`, and mapping it with the right value.

Currently, there's no proper way of updating these lists. Something like this works, but looks more like hacking than programming:

```python

from geomstats.geometry import scalar_product_metric

scalar_product_metric.SQRT_LIST.append("my_new_method")
```

A better solution may be something like (with better naming):

```python
from geomstats.geometry.scalar_product_metric import add_scaled_method

add_scaled_method("my_new_method", scaling_type="sqrt")
```

This is relevant e.g. for "power" users that need to implement a new metric with a method whose output needs to be scaled.

(With @johnharveymath.)

johnharveymath:
A new class method is added to `ScalarProductMetric` called `add_scaled_method` which allows the user to configure how a method from RiemannianMetric should be scaled.  The lists become class variables to allow them to be edited.

This fixes #1760.

###Tests

The added feature does not seem amenable to unit testing, but the following code does demonstrate that it works:

```
from geomstats import geomstats
from geomstats.geometry.scalar_product_metric import ScalarProductMetric

sphere = Hypersphere(2)
point = sphere.random_point(1)
print(sphere.metric.injectivity_radius(point))

new_metric = 2.0 * sphere.metric
print(new_metric.injectivity_radius(point))

ScalarProductMetric.add_scaled_method('injectivity_radius', 'linear')
new_metric = 2.0 * sphere.metric
print(new_metric.injectivity_radius(point))
```

This has output

```
3.141592653589793
3.141592653589793
6.283185307179586
```

as expected.

### Questions

1. Should we add a functionality to remove items from the lists or to restore the 'factory default'?
2. Are we happy with allowing methods which are not explicitly handled to just silently go through untransformed? Maybe we should at least log that it's happening?

### DIFF
--- a/geomstats/geometry/scalar_product_metric.py
+++ b/geomstats/geometry/scalar_product_metric.py
@@ -10,25 +10,6 @@ from functools import wraps
 import geomstats.backend as gs
 import geomstats.errors
 
-SQRT_LIST = ["norm", "dist", "dist_broadcast", "dist_pairwise", "diameter"]
-LINEAR_LIST = [
-    "metric_matrix",
-    "inner_product",
-    "inner_product_derivative_matrix",
-    "squared_norm",
-    "squared_dist",
-    "covariant_riemann_tensor",
-]
-QUADRATIC_LIST = []
-INVERSE_LIST = [
-    "cometric_matrix",
-    "inner_coproduct",
-    "hamiltonian",
-    "sectional_curvature",
-    "scalar_curvature",
-]
-INVERSE_SQRT_LIST = ["normalize", "random_unit_tangent_vec", "normal_basis"]
-
 
 def _wrap_attr(scaling_factor, func):
     @wraps(func)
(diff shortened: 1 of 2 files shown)


luisfpereira (changes requested):
Thanks @johnharveymath!

The idea is very good. I'm just wondering if we can keep the previous function-based approach and add  a method called e.g. `register_scaled_method` that does exactly what `add_scaled_method` is doing, but in a more transparent way (by being outside of the class). Is there any technical reason for such a solution not work?

luisfpereira:
Exactly that @johnharveymath, very nice! 

I like your singleton class idea very much (even though threading is not necessarily something we care about). The other advantage of this strategy I can see is that everything becomes encapsulated in an object, so we know clearly where to look for. I sugggest to make this class private (`_ScaledMethodsRegistry`).

johnharveymath:
Here we go. It seemed to make sense to put `_get_scaling_factor` into the new class as well. Conceptually, this class now holds all information to do with which methods can be handled and how to handle them. I put `_RESERVED_NAMES` in here as well.
The result is that `ScalarProductMetric` now goes to `_Scaled_Methods_Registry` every time it wants to know anything about a method, either accessing one of its private methods or directly accessing its private variable `_RESERVED_NAMES`.
I'm not 100% of whether that's a nice structure.
Then, as for where the function `register_scaled_method` should go, I realised that we usually keep everything a little deep in the namespace and you have to go to the correct module to pull it out. Doing `from geomstats.geometry import register_scaled_method` would be a bit out of the ordinary. So it is going to be `from geomstats.geometry.scalar_product_metric import register_scaled_method`. I think that's fine, as it's not going to see a lot of use. I think the most use it will get will be internal. Once we're happy with this I can add it to a few classes where I see the need, either in this PR or in a new one.

luisfpereira (approved):
I think the design is very clean @johnharveymath. From my side we can merge as it it. (I suggest to register methods in another PR.)

PR #1967, opened 2024-02-27
Klein bottle parametrization conversions

sofiagonga:
Add Klein bottle parametrization conversion from instrinsic coordinates (dimension 2) to:
- Extrinsic coordinates (4 dimensional full parametrization)
- Klein bagel coordinates (3 dimension R^3 immersion)
- Klein bottle classical parametrization (3 dimension R^3 immersion)

### DIFF
--- a/geomstats/geometry/klein_bottle.py
+++ b/geomstats/geometry/klein_bottle.py
@@ -186,7 +186,7 @@ class KleinBottle(Manifold):
         point_odd = gs.stack([x_canonical, y_canonical_odd], axis=-1)
         return gs.where(gs.mod(num_steps, 2) == 0, point_even, point_odd)
 
-    def to_coords(self, point, coords_type):
+    def to_coords(self, point, coords_type, **kwargs):
         """Convert point from intrinsic to coordinates type.
 
         Parameters
@@ -202,7 +202,7 @@ class KleinBottle(Manifold):
             Point in new representation.
         """
         coords_func = getattr(self, f"to_{coords_type}_coords")
-        return coords_func(point)
+        return coords_func(point, **kwargs)
 
     @staticmethod
     def to_extrinsic_coords(point):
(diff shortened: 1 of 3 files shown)


sofiagonga:
Hi @luisfpereira , thank you so much for the feedback and the changes. They look great! I don't know we can get 'smoke data' for this problem. I took the different parametrizations from  Wikipedia (https://en.wikipedia.org/wiki/Klein_bottle). The only thing I haven't accounted for the variability of the different factors in the parametrizations. There are so many that I thought it was better to leave it 'unit size'.

ninamiolane:
WOW! Thank you so much, Sofia and Luis.

For smoke data, @sofiagonga , that would be examples of points on the klein bottle where you calculate (by hand) what the result of the change of coordinates should be. Then, the smoke test checks that the code arrives at the same result.

sofiagonga:
Hi @luisfpereira ! 
I have added some manually generated smoke data, which can be found in _test/test_geomstats/test_geometry/data/klein_bottle.py_ under the class _KleinBottleParametrizationTestData()_. I don't know if this is the best way to do this. 
I then tried using this data in lines 47-53 of tests/tests_geomstats/test_geometry/test_klein_bottle.py in a class called  _TestKleinBottleCoordChange()_ but commented it out since I wasn't sure how to do this.

I really appreciate all the help!

Sofia

luisfpereira:
Fantastic @sofiagonga! I was looking for exactly what you've added.

Notice the little changes I've made to make it work: the test was already defined (`test_to_coords`), so I've simply transformed the data you've added in a list of dicts, each with one example (`point`, `coords_type`, `expected`). 
Please let me know if there's something in this process that is not clear.

Additionally, I've changed `r` to `radius` in `to_bagel` because we try to avoid one letter variables.

I'm merging. Looking forward for more contributions!

luisfpereira (approved):

PR #1969, opened 2024-03-01
Create exact Principal Geodesic Analysis on the hyberbolic plane

alebrigant:
This PR:
- creates exact principal geodesic analysis (PGA) in the hyperbolic space
- adds changes of coordinates for tangent vectors in the hyperbolic space
- updates the information geometry notebook by adding clustering and PGA in the space of normal distributions (=Poincaré half plane)

TODO:
- extend hyperbolic PGA to higher dimensions (should be straightforward)
- add exact PGA in the hypersphere (formulas available in the same reference)


geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+    .. [CSV2016] R. Chakraborty, D. Seo, and B. C. Vemuri,
+        "An efficient exact-pga algorithm for constant curvature manifolds."
+        Proceedings of the IEEE conference on computer vision and pattern
+        recognition. 2016.
+    """
+
+    def __init__(self, space, n_vec=100):
+        self.space = space
+        self.half_space = Hyperbolic(2, coords_type="half-space")
+        self.space_ext = Hyperbolic(2, coords_type="extrinsic")
```
luisfpereira, line 397: Make private?

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+            the Poincare half-space or Poincare ball, n_features is
+            2. If it is the hyperboloid, n_features is 3.
+        y : Ignored (Compliance with scikit-learn interface)
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        estimator = FrechetMean(space=self.space).fit(X)
```
luisfpereira, line 430: instantiate at init, since we can then e.g. update some numerical parameters before calling fit. call it e.g. `mean_estimator` (already used in other learning methods).

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+        ]
+        axis_1 = vectors_half_space[gs.argmax(costs)]
+        axis_2 = gs.array([-axis_1[1], axis_1[0]])
+        components_half_space = gs.stack((axis_1, axis_2))
+        self.components_ = self.space.from_tangent_coordinates(
+            components_half_space, mean_half_space, "half-space"
+        )
+        return self
+
+    def fit(self, X, y=None):
```
luisfpereira, line 464: why `fit` and `_fit` if `fit` only call `_fit`?

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+
+        return gs.stack(
+            [
+                self.space.from_coordinates(proj1_ext, "extrinsic"),
+                self.space.from_coordinates(proj2_ext, "extrinsic"),
+            ]
+        )
+
+
+class ExactPGA(_BasePCA):
```
luisfpereira, line 522: Here we don't need to inherit, as this is a simple interface.

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+        vectors_half_space = gs.hstack(
+            (gs.cos(angles_half_space), gs.sin(angles_half_space))
+        )
+        norms = self.half_space.metric.norm(vectors_half_space, mean_half_space)
+        vectors_half_space = gs.einsum("ij,i->ij", vectors_half_space, 1 / norms)
+        vectors_ext = self.space.half_space_to_extrinsic_tangent(
+            vectors_half_space, mean_half_space
+        )
+
+        def variance_of_projections(pt_ext, mn_ext, vec_ext):
```
luisfpereira, line 448: private method instead of inner function?

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+            projections = self._project_on_geodesic_extrinsic(pt_ext, mn_ext, vec_ext)
+            costs = self.space_ext.metric.dist(mn_ext, projections) ** 2
+            return gs.sum(costs)
+
+        costs = [
```
luisfpereira, line 453: we can probably vectorize here. (I can help you out)

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+    References
+    ----------
+    .. [CSV2016] R. Chakraborty, D. Seo, and B. C. Vemuri,
+        "An efficient exact-pga algorithm for constant curvature manifolds."
+        Proceedings of the IEEE conference on computer vision and pattern
+        recognition. 2016.
+    """
+
+    def __init__(self, space, n_vec=100):
+        self.space = space
```
luisfpereira, line 395: add check if dimension is correct.

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+        angles_half_space = gs.linspace(0.0, 2 * gs.pi, self.n_vec)
+        angles_half_space = gs.expand_dims(angles_half_space, axis=1)
+        vectors_half_space = gs.hstack(
+            (gs.cos(angles_half_space), gs.sin(angles_half_space))
+        )
+        norms = self.half_space.metric.norm(vectors_half_space, mean_half_space)
+        vectors_half_space = gs.einsum("ij,i->ij", vectors_half_space, 1 / norms)
+        vectors_ext = self.space.half_space_to_extrinsic_tangent(
+            vectors_half_space, mean_half_space
+        )
```
luisfpereira, line 446: What would be the difference if we get only the last tangent vector from `vectors_half_space`, transform it to extrinsic, get the geodesic function directly from extrinsic, and then sample it uniformly between 0 and 1 with `n_vec`?

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+        recognition. 2016.
+    """
+
+    def __init__(self, space, n_vec=100):
+        self.space = space
+        self.half_space = Hyperbolic(2, coords_type="half-space")
+        self.space_ext = Hyperbolic(2, coords_type="extrinsic")
+        self.n_vec = n_vec
+
+    def _project_on_geodesic_extrinsic(self, point_ext, mean_ext, vector_ext):
```
luisfpereira, line 400: Shouldn't we move this method to `Hyperboloid`? It looks like an interest method to have there.

geomstats/learning/pca.py
```diff
@@ -353,3 +355,186 @@ def _fit(self, X, base_point=None):
...
+        "An efficient exact-pga algorithm for constant curvature manifolds."
+        Proceedings of the IEEE conference on computer vision and pattern
+        recognition. 2016.
+    """
+
+    def __init__(self, space, n_vec=100):
+        self.space = space
+        self.half_space = Hyperbolic(2, coords_type="half-space")
+        self.space_ext = Hyperbolic(2, coords_type="extrinsic")
+        self.n_vec = n_vec
```
luisfpereira, line 398: maybe `n_grid` instead? for me it is a choice of how many points the geodesic is discretized on for the projection.

luisfpereira (changes requested):
Thanks @alebrigant! Looking forward for your comments on the questions I've raised.

geomstats/learning/pca.py
alebrigant: The difference is that you need to sample from tangent vectors in the extrinsic tangent space and I wanted to go fast so I didn't write down the formula (no need to look for one in the half-space). But it should certainly be changed at some point.

geomstats/learning/pca.py
alebrigant: Thanks, I just copied the structure of the existing class and forgot to change it.

geomstats/learning/pca.py
alebrigant: Sure, thanks !

luisfpereira: Can't vectorize actually. I think this is the best implementation.

luisfpereira (changes requested):
@alebrigant, a couple of very small changes due to naming and ready to go from my side.

geomstats/geometry/hyperboloid.py
```diff
@@ -174,6 +178,45 @@ def extrinsic_to_intrinsic_coords(self, point_extrinsic):
         """
         return self.change_coordinates_system(point_extrinsic, "extrinsic", "intrinsic")
 
+    def project_on_geodesic(self, point, mean, vector):
```
luisfpereira, line 181: Just to confirm: is this valid in any dimension?

alebrigant: Yes, according to the paper in reference.

alebrigant (commented):
Thanks for your help Luis, it's all good for me !

luisfpereira (approved):
Thanks @alebrigant! Merging.

PR #1971, opened 2024-03-07
Added natural gradient descent optimizer method

SammySuliman:
## Checklist

### DIFF
--- a/geomstats/learning/natural_gradient_descent.py
+++ b/geomstats/learning/natural_gradient_descent.py
@@ -1,0 +1,81 @@
+import geomstats.backend as gs
+import numpy as np
+from scipy.integrate import quad_vec
+import torch
+from torch.optim import Optimizer
+import math
+
+class NaturalGradientDescent(Optimizer):
+    """Implements Natural Gradient Descent using Amari's anaytic solution
+    for the inverse Fisher information matrix
+
+    Parameters
+    ----------
+    params : array-like, shape=[3, dim]
+        The current layer of the ResNet, consisting of a vector of nodes, 
+        a vector of weights, and a bias scalar 
+
+    lr : float
+        The learning rate to update the weights
+
+    wd : float
+        Penalization on weights to prevent overfitting
+
+    bias : Boolean
+        Whether or not to include bias term in training
+
+    References
+    ----------
+    .. [AS1985] Amari S., Karakida R., Oizumi M. (2018)
+        Fisher Information and Natural Gradient Learning of Random Deep Networks.
+    """
+    def __init__(self, params, lr=0.05, wd = 0.0, bias=True):
+        defaults = dict(lr=lr, wd=wd, bias=bias)
+        self.params = params
+        # inherits all default methods / attributes of Optimizer
+        super(NaturalGradientDescent, self).__init__(params, defaults)
+    def step(self):
+        """Performs one step of Natural Gradient Descent"""
+        w_list = []  
+        w_star_list = [] 
+        w_tilde_list = []
+        for group in self.param_groups: # only 1 group for now
+            for i, param in enumerate(group['params']):
+                if i % 2 == 0: # only want to capture the weight parameters
+                    param2 = param.detach().numpy()
+                    w_list.append(param2)
+                    # Append each bias value to the corresponding weight vector
+                    bias_param = group['params'][i+1]
+                    param_concat_bias = torch.cat((param.data, bias_param.view(-1, 1)), dim=1)
+                    param2 = param_concat_bias.detach().numpy()
+                    w_star_list.append(param2)
+
+            for w_layer in w_list:
+                for w in w_layer:
+                    w2 = np.append(w, 0.0)
+                    w_tilde_list.append(w2)
+
+            for n_l, weight_layer in enumerate(w_list):
+                for n_u, w in enumerate(weight_layer):
+                    n = len(w)
+                    x = torch.randn(n) # approximate value of current node since we assume x ~ N(0, I)
+                    w_star = w_star_list[n_l][n_u]
+                    w_tilde = w_tilde_list[n_l][n_u]
+                    e_star_0 = np.zeros(n+1)
+                    e_star_0[-1] = 1
+                    A_00 = (1/2*gs.sqrt(2)) * math.erf(w_star[-1] / gs.sqrt(gs.dot(w,w))) # see Appendix II, eq. 107
+                    A_0n = (1/gs.sqrt(2)) * (math.exp((-1/2) * (w_star[-1] / gs.sqrt(gs.dot(w,w)))**2)) # see Appendix II, eq. 109
+                    A_nn = ((1/2*gs.sqrt(2)) * A_00) -  ((w_star[-1] / gs.sqrt(gs.dot(w,w))) * A_0n) # see Appendix II, eq. 110
+                    D = A_00 * A_nn - A_0n**2 # pp. 15, eq. 79
+                    X = (1/D) * A_00 - (1 / A_00) # pp. 15, eq. 78
+                    Y = (-1) * A_0n / D # pp. 15, eq. 78
+                    Z = A_nn / D - (1 / A_00) # pp. 15, eq. 78
+                    if param.grad is None:
+                        continue
+                    if i % 2 == 0: # updating the gradients for the weights only
+                        d_p = (-1) * param.grad.data * group['lr'] * ((1/A_00) * x + \
+                            (X / gs.dot(w,w) * gs.dot(w,x) + (Y / gs.sqrt(gs.dot(w,w))) * w)) # pp. 15, eq. 81
+                    else: # updating the gradients for the bias only
+                        d_p = (-1) * param.grad.data * group['lr'] * \
+                            ((1/A_00) + Z + (Y * gs.dot(w, x) / gs.sqrt(gs.dot(w,w))) * w_star[-1]) # pp. 15, eq. 82
+                    param.data.add_(d_p)


geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+        self.params = params
+
+        super().__init__(params, defaults)
+
+    def step(self):
+        """Perform one step of Natural Gradient Descent."""
+        w_list = []
+        w_star_list = []
+        w_tilde_list = []
+        for group in self.param_groups:  # only 1 group for now
```
luisfpereira, line 52: what does this comment mean?

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+                    w_tilde_list.append(w2)
+
+            for n_l, weight_layer in enumerate(w_list):
+                for n_u, w in enumerate(weight_layer):
+                    n = len(w)
+                    x = torch.randn(
+                        n
+                    )  # approximate value of current node since we assume x ~ N(0, I)
+                    w_star = w_star_list[n_l][n_u]
+                    w_tilde = w_tilde_list[n_l][n_u]
```
luisfpereira, line 77: this is not being used on purpose? delete if the case.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+            for i, param in enumerate(group["params"]):
+                if i % 2 == 0:  # only want to capture the weight parameters
+                    param2 = param.detach().numpy()
+                    w_list.append(param2)
+                    # Append each bias value to the corresponding weight vector
+                    bias_param = group["params"][i + 1]
+                    param_concat_bias = torch.cat(
+                        (param.data, bias_param.view(-1, 1)), dim=1
+                    )
+                    param2 = param_concat_bias.detach().numpy()
```
luisfpereira, line 62: as above.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def step(self):
+        """Perform one step of Natural Gradient Descent."""
+        w_list = []
+        w_star_list = []
+        w_tilde_list = []
+        for group in self.param_groups:  # only 1 group for now
+            for i, param in enumerate(group["params"]):
+                if i % 2 == 0:  # only want to capture the weight parameters
+                    param2 = param.detach().numpy()
```
luisfpereira, line 55: why the need to detach? notice this will only run with `pytorch` within geomstats.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__(params, defaults)
+
+    def step(self):
+        """Perform one step of Natural Gradient Descent."""
+        w_list = []
+        w_star_list = []
+        w_tilde_list = []
+        for group in self.param_groups:  # only 1 group for now
+            for i, param in enumerate(group["params"]):
+                if i % 2 == 0:  # only want to capture the weight parameters
```
luisfpereira, line 54: is this always the case?

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+                    )
+                    param2 = param_concat_bias.detach().numpy()
+                    w_star_list.append(param2)
+
+            for w_layer in w_list:
+                for w in w_layer:
+                    w2 = np.append(w, 0.0)
+                    w_tilde_list.append(w2)
+
+            for n_l, weight_layer in enumerate(w_list):
```
luisfpereira, line 70: call `n_l` e.g. `layer_index` for readability?

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+                for n_u, w in enumerate(weight_layer):
```
luisfpereira, line 71: better naming: `weight` instead of `w`?

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+                    A_nn = ((1 / 2 * gs.sqrt(2)) * A_00) - (
+                        (w_star[-1] / gs.sqrt(gs.dot(w, w))) * A_0n
+                    )  # see Appendix II, eq. 110
+                    D = A_00 * A_nn - A_0n**2  # pp. 15, eq. 79
+                    X = (1 / D) * A_00 - (1 / A_00)  # pp. 15, eq. 78
+                    Y = (-1) * A_0n / D  # pp. 15, eq. 78
+                    Z = A_nn / D - (1 / A_00)  # pp. 15, eq. 78
+                    if param.grad is None:
+                        continue
+                    if i % 2 == 0:  # updating the gradients for the weights only
```
luisfpereira, line 95: probably a bug here: `i` comes from another for, it is always fixed within this for.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+                        math.exp((-1 / 2) * (w_star[-1] / gs.sqrt(gs.dot(w, w))) ** 2)
+                    )  # see Appendix II, eq. 109
+                    A_nn = ((1 / 2 * gs.sqrt(2)) * A_00) - (
+                        (w_star[-1] / gs.sqrt(gs.dot(w, w))) * A_0n
+                    )  # see Appendix II, eq. 110
+                    D = A_00 * A_nn - A_0n**2  # pp. 15, eq. 79
+                    X = (1 / D) * A_00 - (1 / A_00)  # pp. 15, eq. 78
+                    Y = (-1) * A_0n / D  # pp. 15, eq. 78
+                    Z = A_nn / D - (1 / A_00)  # pp. 15, eq. 78
+                    if param.grad is None:
```
luisfpereira, line 93: probably bug here (as below): `param` comes from a previous for.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,120 @@
...
+        Fisher Information and Natural Gradient Learning of Random Deep Networks.
+    """
+
+    def __init__(self, params, lr=0.05, wd=0.0, bias=True):
+        defaults = dict(lr=lr, wd=wd, bias=bias)
+        self.params = params
+
+        super().__init__(params, defaults)
+
+    def step(self):
```
luisfpereira, line 47: I think we should probably split this into smaller functions, where it is easier to understand what are the inputs and outputs: e.g. it seems we collect weight layers for each param group, maybe a `_collect_weight_layers` method helps on code readability. Also, creating a `_step_param_group` may help avoiding so much nesting (which, as I comment below, seem to be bug prone).

luisfpereira (changes requested):
Thanks @SammySuliman! Please take a look to my comments below.

On the general side, I think we should transform your test function in a notebook (some kind of a how-to showing how to use `NaturalGradientDescent`), and add a minimal test that shows we can do one `step` properly (in the most basic model we can think of).

For the `step` method, please see my comments below (there's probably small bugs to fix).

@ninamiolane, on a more backend philosophical side: this code only works with `pytorch`, as it assumes the NN structure of `pytorch` and inherits from its `Optimizer` class. It is not trivial (neither relevant?) to make it work with other backends. Is this direction something we're happy pursuing?

geomstats/learning/natural_gradient_descent.py
SammySuliman: for the purpose of the succeeding code, i need to treat param2 as a regular array, that is, detach it from its computation graph.

geomstats/learning/natural_gradient_descent.py
SammySuliman: there is by default only one parameter group, unless you choose to add more. I can delete the comment

geomstats/learning/natural_gradient_descent.py
SammySuliman: yes the even indices are always associated with the weights for that step, the odd indices with the bias

geomstats/learning/natural_gradient_descent.py
SammySuliman: thanks for pointing this out! I have since corrected the error

geomstats/learning/test_NaturalGradientDescent.py
SammySuliman, line 1: The original paper by Amari makes assumptions that only hold in the case of a randomly-connected resnet, so i have changed the test case.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,176 @@
...
+<<<<<<< HEAD
```
luisfpereira, line 49: Notice there's some issues with merge.

SammySuliman: sorry there was some strange issue with git. I believe it should have been corrected.

geomstats/learning/natural_gradient_descent.py
```diff
@@ -0,0 +1,110 @@
...
+                    # Append each bias value to the corresponding weight vector
+                    bias_param = group["params"][i + 1]
+                    param_concat_bias = torch.cat(
+                        (param.data, bias_param.view(-1, 1)), dim=1
+                    )
+                    param2 = param_concat_bias.detach().numpy()
+                    w_star_list.append(param2)
+
+            for layer_index, weight_layer in enumerate(w_list):
+                A_00_list = []
```
ninamiolane, line 66: For readability to future users/contributors, let's make the variable names understandable. 
In particular:
- no uppercase letters for variables, unless they are classes.
As per python coding style guidelines: https://peps.python.org/pep-0008/
- to "single letter" names, instead, prefer english names.

geomstats/learning/test_NaturalGradientDescent.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # Create an instance of the 2-layer model
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    model = SimpleModel().to(device)
+
+    # Define loss function and optimizer
+    criterion = nn.CrossEntropyLoss()
+    parameters = list(model.parameters())
+    optimizer = NaturalGradientDescent(parameters)
+    #for p in parameters:
```
ninamiolane, line 45: Remove leftover comments

PR #1974, opened 2024-03-08
Scalar product metric API

### LINKED ISSUE #1968: Make ScalarProductMetric API consistent with Riemannian metric
Our design choices sent us in a direction where we do not use metrics in isolation, i.e. they're always equipping some space. This is achieved by asking for `space` at metric initialization and using `space.equip_with_metric` to endow a given space with a metric.

This is true for all metrics, but `ScalarProductMetric`.  This metric receives `underlying_metric` as input and forces us to use `space.metric = ScalarProductMetric(underlying_metric)` or `space.metric = 2*underlying_metric`.

To eliminate this inconsistency, I suggest to change the signature of `ScalarProductMetric` to receive `space` instead of `underlying_metric`. Then, at init, we can do something like:

```python
self._space = space
self._underlying_metric = space.metric
```

and keep the rest equal.

I also suggest to equip the space directly with the scaled metric when using `ScalarProductMetric.__mul__`, by using `equip_with_metric` within that method. This will result in e.g.:

```python
space.equip_with_metric(MyMetric)  # space equipped with MyMetric

2.*space.metric  # space becomes equipped with ScalarProductMetric with underlying metric MyMetric
```

I don't fully like how implicit the last line is, but seems to me that this is the only consistent way of keeping such syntax sugar with our current design. The alternative would be to remove this method.

What are your thoughts on this @johnharveymath?

johnharveymath:
## Description

Changes the signature of `ScalarProductMetric` to take `space` instead of `metric` as suggested in #1968.

This means that `Manifold` and `ComplexManifold` needed changing: `equip_with_metric` must be able to take an instantiated metric, and not just a class. In particular, they can now be equipped with the specific `ScalarProductMetric`.

The `RiemannianMetric.__mul__` and `ScalarProductMetric.__mul__` methods can now only accept metrics that have been equipped by a particular space. I included a check here, as the user would probably expect `metric` to be scaled by this operation but it is `metric._space.metric` that is scaled. It could be quite confusing.

A number of other classes and tests use `ScalarProductMetric` and these have been refactored.

## Issue

Closes #1968

## Next steps

One issue with putting this together was that `ScalarProductMetric` can be used silently through `RiemannianMetric.__mul__`. It might be reasonable to remove this functionality and insist on explicitly creating these metrics for the sake of maintainability.

Another thing that came up was that there are two `equip_with_metric` methods with identical content, which both needed changing. Maybe there should be a parent `Manifold` with `RealManifold` and `ComplexManifold` as children?

### DIFF
--- a/geomstats/geometry/complex_manifold.py
+++ b/geomstats/geometry/complex_manifold.py
@@ -64,8 +64,12 @@ class ComplexManifold(abc.ABC):
         if inspect.isclass(Metric):
             self.metric = Metric(self, **metric_kwargs)
         else:
+            if self.metric._space is not self:
+                raise ValueError(
+                    "Cannot equip space with metric instantiated with another space."
+                )
+
             self.metric = Metric
-            self.metric._space = self
 
         return self
 
--- a/geomstats/geometry/complex_manifold.py
+++ b/geomstats/geometry/complex_manifold.py
@@ -7,6 +7,7 @@ Lead author: Yann Cabanes.
 """
 
 import abc
+import inspect
 
 import geomstats.backend as gs
 
@@ -48,7 +49,7 @@ class ComplexManifold(abc.ABC):
 
         Parameters
         ----------
-        Metric : Connection object
+        Metric : Connection object or instance or ScalarProductMetric instance
             If None, default metric will be used.
         """
         if Metric is None:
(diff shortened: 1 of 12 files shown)


johnharveymath:
Marking this as draft because I would like to add more test coverage to `scalar_product_metric.py`. Otherwise I think it is ready to go so can be reviewed.

johnharveymath:
I have no idea why `test_tutorial_matplotlib()` in `tests/tests_geomstats/test_visualization.py` is suddenly broken, but I don't think it is anything to do with this PR. Ready to go now.

PR #1975, opened 2024-03-13
Clean discrete surfaces exp solver

luisfpereira:
This PR cleans the exponential solver in discrete surfaces by taking advantage of machinery introduced in #1957. It also fixes small bugs in path straightening and interpolation. Performance is expected to be the same (no relevant improvement done), except for geodesic coming from the use of a tangent vector and a base point (which now takes advantage of the `path` module). Discrete geodesic ivp is also made public (following path straightening).


luisfpereira:
I'm merging because this may be helpful for some of the experiments Trey and @MMalikT are doing. We can open a PR later to address any comment on this.

PR #1978, opened 2024-03-19
Use aligner abstraction within fiber bundle and rely on group action on total space

luisfpereira:
This PR updates `FiberBundle` to:

1. use an `AlignerAlgorithm`: follows what was already being done in discrete curves and graph space (notice the abstraction was initially defined within the graph space context)

2. definitely moves `group_action` to the total space by removing it from fiber bundle

For 1., the main advantages are i) an easier control of the numerical parameters, ii) homogenization of the code among all the uses of quotient spaces (including graph space), iii) it becomes easier and more natural to plugin different aligner algorithms (this has already being taken advantage of within discrete curves).

For 2., we use `space.equip_with_group_action` method and make everything consistent, i.e. a group action endows the total space. Notice that a group action can simply be a callable (see e.g. `full_rank_correlation_matrices.CorrelationMatricesBundle`) or a child of `GroupAction`. The aligner algorithm works with just a little extra information regarding group dimensionality.

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -13,7 +13,7 @@ import geomstats.backend as gs
 from geomstats.algebra_utils import from_vector_to_diagonal_matrix
 from geomstats.geometry.diffeo import AutodiffDiffeo, Diffeo
 from geomstats.geometry.euclidean import Euclidean
-from geomstats.geometry.fiber_bundle import FiberBundle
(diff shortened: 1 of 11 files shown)


luisfpereira:
@ninamiolane, I'm merging because, as we've discussed in our last meeting, the API becomes much cleaner (we need with some urgency for discrete curves). There's some naming we may need to improve regarding the group action (let's take a look again to it at some point).

PR #1980, opened 2024-03-25
Add LogScaled metrics on Cor+

luisfpereira:
This PR implements the log scaled metrics on full-rank correlation matrices (see "[Permutation-invariant log-Euclidean geometries on full-rank correlation matrices](https://hal.science/hal-03878729)").

To accomplish the goal, other spaces, metrics, and diffeos are created: 

* diffeos: `LogScalingDiffeo`, `ConstantValueRowSumDiffeo`

* spaces: `NullRowSumsSymmetricMatrices`

* abstract spaces: `DiffeomorphicVectorSpace`, `DiffeomorphicMatrixVectorSpace`

* metrics: `LogScaledMetric`, `NullRowSumsPermutationInvariantMetric`

`projection` is also moved to `Manifold` to avoid inheritance problems.

Done in collaboration with @olivierbisson.

### DIFF
--- a/geomstats/geometry/symmetric_matrices.py
+++ b/geomstats/geometry/symmetric_matrices.py
@@ -119,7 +119,7 @@ class SymmetricMatrices(MatrixVectorSpace):
         mat_dim = (gs.sqrt(8.0 * vec_dim + 1) - 1) / 2
         if mat_dim != int(mat_dim):
             raise ValueError(
-                "Invalid input dimension, it must be of the form"
+                "Invalid input dimension, it must be of the form "
                 "(n_samples, n * (n + 1) / 2)"
             )
         mat_dim = int(mat_dim)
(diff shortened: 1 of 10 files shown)


geomstats/geometry/symmetric_matrices.py
```diff
@@ -341,11 +346,390 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
...
+
+        pre_basis = gs.array_from_sparse(indices, values, (k + 1, self.n, self.n))
+        return self.matrix_representation(
+            self.basis_representation(
+                pre_basis,
+            )
+        )
+
+
+class NullRowSumsPermutationInvariantMetric(EuclideanMetric):
```
olivierbisson, line 622: check the default case for n=3

olivierbisson (approved):
check high dimension?

geomstats/geometry/symmetric_matrices.py
luisfpereira: Spot on! Fixed.

luisfpereira:
> check high dimension?

For the future, this comment comes from some numerical instability related with shooting with large tangent vectors (for any metric in Cor+). We may need to check this at some point, though it seems to be stable for reasonable tangent vectors.

PR #1981, opened 2024-03-26
Implement PoincareHalfSpaceMetric as PullbackDiffeoMetric

luisfpereira:
This PR takes advantage of some objects introduced in #1969 (in particular, `HyperbolicDiffeo`), to implement the `PoincareHalfSpaceMetric` as a `PullbackDiffeoMetric` (it was already being implicitly done as such).

To achieve that:

* `HyperbolicDiffeo` is moved to the main code (before it was in test cases, as it was used only for tests)
* `PoincareHalfSpaceMetric` inherits from `PullbackDiffeoMetric`, and redundant methods were removed

For consistency, also the `HyperbolicMetric` was removed, as it was not really playing any role.

@alebrigant, can you please take a quick look and let me know if you agree with the changes?


alebrigant (approved):
Very nice !!

PR #1982, opened 2024-03-26
Add NaturalToStandardDiffeo

luisfpereira:
This PR adds `NaturalToStandardDiffeo` for gamma distributions. It simply moves code that was within `GammaDistributions` to its own object.

There several advantages of doing this:

* testing is hugely simplified (see the amount of testing code deleted, though we are actually testing this code more thoroughly)

* we can take advantage of existing structures to delete redundant code (in this case by using `InvolutionDiffeomorphism`)

* we can e.g. define a manifold `StandardGammaDistributions` (notice it is already done for testing) if someone wants to directly work with standard coordinates (the metric then is just a `PullbackDiffeoMetric`).

Above all, these changes make the code easier to maintain.

@alebrigant, can you please take a quick look and let me know if you agree with the changes?


alebrigant (approved):
Seems perfect, thanks !

PR #1984, opened 2024-04-02
Implement PreShapeMetric as PullbackDiffeoMetric

luisfpereira:
Similarly to #1981, this PR implements `PreShapeMetric` as a `PullbackDiffeoMetric`. With the implementation done this way, we remove unnecessary and duplicate code and make more clear the relationship between the pre-shape space and the hypersphere.


ninamiolane (approved):
Awesome!!

PR #1986, opened 2024-04-04
Clean discrete surfaces elastic metric

luisfpereira:
This PR cleans the discrete surfaces elastic metric:

* einsums are replaced by proper backend functions

* docstrings are fixed and simplified

* default solvers are only instantiated if an autodiff-based backend is used (this way we can compute inner products with numpy)

* a potential bug is solved in `_inner_prod_a2` (@emmanuel-hartman, could you please confirm we should multiply by vertex areas instead of divide? this comes from the original implementation)

### DIFF
--- a/geomstats/geometry/discrete_surfaces.py
+++ b/geomstats/geometry/discrete_surfaces.py
@@ -319,7 +319,7 @@ class DiscreteSurfaces(Manifold):
         return gs.sqrt(gs.linalg.det(surface_metrics_bp))
 
     @staticmethod
-    def _surface_metric_matrices_from_one_forms(one_forms):
+    def surface_metric_matrices_from_one_forms(one_forms):
         """Compute the surface metric matrices directly from the one_forms.
 
         This function is useful for efficiency purposes.
@@ -358,7 +358,7 @@ class DiscreteSurfaces(Manifold):
             the triangulated surface.
         """
         one_forms = self.surface_one_forms(point)
-        return self._surface_metric_matrices_from_one_forms(one_forms)
+        return self.surface_metric_matrices_from_one_forms(one_forms)
 
     def laplacian(self, point):
         r"""Compute the mesh Laplacian operator of a triangulated surface.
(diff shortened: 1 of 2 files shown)


geomstats/geometry/discrete_surfaces.py
```diff
@@ -514,26 +514,23 @@ def __init__(self, space, a0=1.0, a1=1.0, b1=1.0, c1=1.0, d1=1.0, a2=1.0):
         self.d1 = d1
         self.a2 = a2
 
-        self.exp_solver = DiscreteSurfacesExpSolver(space, n_steps=10)
+        if not gs.__name__.endswith("numpy"):
```
ninamiolane, line 517: Isn't there a less hacky way to test the backend?

geomstats/geometry/discrete_surfaces.py
```diff
@@ -514,26 +514,23 @@ def __init__(self, space, a0=1.0, a1=1.0, b1=1.0, c1=1.0, d1=1.0, a2=1.0):
...
 
     def _inner_product_a0(self, tangent_vec_a, tangent_vec_b, vertex_areas_bp):
         r"""Compute term of order 0 within the inner-product.
 
         Denote h and k the tangent vectors a and b respectively.
-        Denote q the base point, i.e. the surface.
-
-        The equation of the inner-product is:
-        :math:`\int_M (G_{a_0} + G_{a_1} + G_{b_1} + G_{c_1} + G_{d_1} + G_{a_2})vol_q`.
+        Denote q the base point.
```
ninamiolane, line 531: Why did you remove the rest of this docstring? I felt that the additional explanations were helping.

ninamiolane (approved):
Awesome, thanks!

geomstats/geometry/discrete_surfaces.py
luisfpereira: for the name no... probably something to add.

luisfpereira:
Changing the division to a multiplication in `_inner_prod_a2` impacts a lot the results (as expected). Probably a closer look to this is important to ensure I'm not missing out something.

geomstats/geometry/discrete_surfaces.py
ninamiolane: Ah yes, all good, let's keep it like this

luisfpereira:
After looking carefully to some equations, I believe `_inner_prod_a2` was properly implemented (it should be a division, not a multiplication). In any case, a third-party confirmation of this is welcome.

Good to merge from my side.

PR #1987, opened 2024-04-06
Update SPDBuresWassersteinMetric.squared_dist to fix numerical stability issues

### LINKED ISSUE #1882: Bug: Error due to complex number distance in Bures-Wasserstein metric, using pytorch backend
The Bures-Wasserstein distance function gives the following error for some pairs of spd matrices, when using the Pytorch backend:

> File ~/anaconda3/envs/my-torch/lib/python3.10/site-packages/geomstats/geometry/spd_matrices.py:854, in SPDBuresWassersteinMetric.squared_dist(self, point_a, point_b)
>     850 trace_prod = gs.trace(sqrt_product)
>     852 squared_dist = trace_a + trace_b - 2.0 * trace_prod
> 854 return gs.where(squared_dist < 0.0, 0.0, squared_dist)
>  RuntimeError: "lt_cpu" not implemented for 'ComplexFloat'

The function returning the error has the following code:

`product = gs.matmul(point_a, point_b)`
`sqrt_product = gs.linalg.sqrtm(product)`
`trace_a = gs.trace(point_a)`
`trace_b = gs.trace(point_b)`
`trace_prod = gs.trace(sqrt_product)`
`squared_dist = trace_a + trace_b - 2.0 * trace_prod`
`return gs.where(squared_dist < 0.0, 0.0, squared_dist)`

The problem in my case is that sqrt_product has complex entries, and so trace_prod is also a complex number. All the imaginary parts are on the order of e-10, so it's probably just precision error.

Then, error comes from the statement, `squared_dist < 0.0`, due to squared_dist being a complex number. The difference between numpy and pytorch is that the former will still perform the comparison and given, while the latter will return an error.

I checked that the input matrices of the function belong to the manifold.

### Steps/Code to Reproduce

I don't know how to obtain two spd matrices that will give the behavior of giving a complex matrix in the squre root of their product, other than copying the entries of my two big matrices, so I can't provide code to reproduce right now. I do notice that I only get the complex matrix with float32 tensors and not with the same matrix of type float64.

### Expected Behaviour

No error is thrown.

### Actual Behaviour

`File ~/anaconda3/envs/my-torch/lib/python3.10/site-packages/geomstats/geometry/spd_matrices.py:854, in SPDBur
esWassersteinMetric.squared_dist(self, point_a, point_b)
    850 trace_prod = gs.trace(sqrt_product)
    852 squared_dist = trace_a + trace_b - 2.0 * trace_prod
--> 854 return gs.where(squared_dist < 0.0, 0.0, squared_dist)

RuntimeError: "lt_cpu" not implemented for 'ComplexFloat'
`

### Your environment

```shell
geomstats version 2.5.0, running on Ubuntu 20.04
```

mwilson221:
## Checklist

## Description

This is a response to [Bug: Error due to complex number distance in Bures-Wasserstein metric, using pytorch backend #1882 ](https://github.com/geomstats/geomstats/issues/1882) (closes #1882)

I've been working with Bures-Wasserstein for spd matrices. The issue is the result of the sqrtm function being used. The solution I have found is to use the fact that spd matrices are orthogonally diagonalizble (due to being symmetric) and full rank (due to positive definite) to calculate square roots via diagonalization. This both fixes the issue of imaginary outputs (even for matrices with spd.belongs(sigma)==True -- see Bures_distance_check), and increases the speed. I also have a torch implementation that allows for batch computation on the gpu (of space.metric.dist_pairwise), which greatly speeds up the computation, though it is not backend agnostic.

To be clear, this only works for spd-matrices. Bures-Wasserstein is technically a distance for symmetric positve semidefinite matrices. So, my approach is not actually valid for the Bures metric in general, but is correct (as far as I know) as the Riemannian metric for spd's.

## Issue

sqrtm in Bures-Wasserstein metric for spd's sometimes returns imaginary numbers

I wrote this code months ago, now that I'm looking at it again I think the 

gs.sqrt(Lx*(Lx>0))

might actually be intended to work for non-spd matrices.. so this might not be completely correct...

## Additional context

I'm not really sure how to make a pull request. 

# Reproducible example:

import numpy as np
import geomstats
from geomstats.geometry.spd_matrices import SPDMatrices, SPDBuresWassersteinMetric

spd = SPDMatrices(3)
spd.equip_with_metric(SPDBuresWassersteinMetric)

X = np.random.randn(3,1)
Sigma_x = X@X.T

Y = np.random.randn(3,1)
Sigma_y = Y@Y.T

#matrices not spd - should return false
print(spd.belongs(Sigma_x))
print(spd.belongs(Sigma_y))

Sigma_x_star = spd.projection(Sigma_x)
Sigma_y_star = spd.projection(Sigma_y)

#should return True
print(spd.belongs(Sigma_x_star))
print(spd.belongs(Sigma_y_star))

#will sometimes return imaginary numbers due to sqrtm
spd.metric.dist(Sigma_x_star, Sigma_y_star)

# Solutions:
Accuracy check: https://github.com/mwilson221/dtmrpy/blob/main/Bures_distance_check.ipynb
Speed check: https://github.com/mwilson221/dtmrpy/blob/main/Bures_distance_speed_check.ipynb

Thanks

### DIFF
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -835,12 +835,12 @@ class SPDBuresWassersteinMetric(RiemannianMetric):
             Riemannian squared distance.
         """
         La, Qa = gs.linalg.eigh(point_a)
-        point_a_sqrt = Qx @ gs.diag(gs.sqrt(Lx*(Lx>0))) @ Qx.T
+        point_a_sqrt = Qa @ gs.diag(gs.sqrt(La*(La>0))) @ Qa.T
     
         Lc,Qc = gs.linalg.eigh(point_a_sqrt@point_b@point_a_sqrt)
         cross_term = Qc @ gs.diag(gs.sqrt((Lc*(Lc>0)))) @ Qc.T
     
-        return np.trace(point_a + point_b - 2*cross_term)
+        return gs.trace(point_a + point_b - 2*cross_term)
 
     def parallel_transport(
         self,
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -833,14 +833,29 @@ class SPDBuresWassersteinMetric(RiemannianMetric):
         -------
         squared_dist : array-like, shape=[...]
             Riemannian squared distance.
+
+        Notes
+        -----
+        Use of `abs` in the output prevents nan when calling
+        `sqrt` in very small negative outputs (e.g. -1e-16).
         """
         La, Qa = gs.linalg.eigh(point_a)
-        point_a_sqrt = Qa @ gs.diag(gs.sqrt(La*(La>0))) @ Qa.T
-    
-        Lc,Qc = gs.linalg.eigh(point_a_sqrt@point_b@point_a_sqrt)
-        cross_term = Qc @ gs.diag(gs.sqrt((Lc*(Lc>0)))) @ Qc.T
-    
-        return gs.trace(point_a + point_b - 2*cross_term)
+        point_a_sqrt = Matrices.mul(
+            Qa,
+            gs.vec_to_diag(
+                gs.sqrt(La * (La > 0)),
+            ),
+            Matrices.transpose(Qa),
+        )
+
+        Lc, Qc = gs.linalg.eigh(point_a_sqrt @ point_b @ point_a_sqrt)
+        cross_term = Matrices.mul(
+            Qc,
+            gs.vec_to_diag(gs.sqrt(Lc * (Lc > 0))),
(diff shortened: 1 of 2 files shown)


luisfpereira:
Very nice, thanks for the contribution @mwilson221! The notebooks with the benchmarks are spot on (we may need to make some changes in geomstats to make the use of gpu more straightforward).

I've fixed the linting and the vectorization. The more "controversial" change I've done is to add an `abs` to the output, because will testing I've realized sometimes the output is something like `-1e16`, which creates `nan` when calling `sqrt` to get the distance.

After running the tests several times, it looks the code is quite stable now.

luisfpereira (approved):

PR #1990, opened 2024-04-09
Correct number of samples for `VectorSpace.random_tangent_vec`

### LINKED ISSUE #1989: Bug: `random_tangent_vec` has different behaviours depending on class
### Describe the bug

If a manifold inherits from `VectorSpace`, the method `random_tangent_vec` will only produce `n_samples` tangent vectors if `base_point` has length `n_samples`. This seems to have been introduced in #1949. This does not match the method in `Manifold`. The method in `VectorSpace` also samples tangent vectors uniformly over a hypercube. This means they are not uniformly distributed in terms of their direction.

### Steps/Code to Reproduce

```from geomstats.geometry.hypersphere import Hypersphere
from geomstats.geometry.product_manifold import ProductManifold

base = Euclidean(1)
fiber = Hypersphere(1)
product_manifold = ProductManifold([base, fiber])

point = product_manifold.random_point()

base_random_point = product_manifold.project_from_product(point)[0]
base_random_vectors = base.random_tangent_vec(base_point=base_random_point, n_samples=2)
print(len(base_random_vectors))     # Expect 2, get 1

fiber_random_point = product_manifold.project_from_product(point)[1]
fiber_random_vectors = fiber.random_tangent_vec(base_point=fiber_random_point, n_samples=2)
print(len(fiber_random_vectors))     # Expect 2, get 2

vectors = product_manifold.random_tangent_vec(point, 2)    # Fails
```

### Expected Behaviour

I expect the printed values to both be `2`. I expect the variable `vectors` to be computed.

### Actual Behaviour

The first printed value is `1`. The variable `vectors` cannot be computed because the two factors have different numbers of vectors, and so they cannot be pooled.

### Your environment

```shell
main branch, numpy backend.
```

johnharveymath:
## Description

When there are multiple basepoints, both implementations of `random_tangent_vec` (the implementations in `manifold.py` and `base.py`) should now check that `n_samples` matches the number of basepoints. If the values do not match, then in case `n_samples == 1` (default option), it resets it to be equal to the number of basepoints. Otherwise an error is raised.

Docstrings have been updated to clarify that the vectors generated have no statistical value.

## Issue

Fixes #1989 insofar as the bug is no longer present. The reference to uniformity of direction understated the difficulty of that problem, so it should probably be ignored.

### DIFF
--- a/geomstats/geometry/base.py
+++ b/geomstats/geometry/base.py
@@ -125,8 +125,9 @@ class VectorSpace(Manifold, abc.ABC):
     def random_tangent_vec(self, base_point=None, n_samples=1):
         """Generate random tangent vec.
 
-        This method is not recommended for statistical purposes, as the tangent vectors generated are not
-        drawn from a distribution related to the Riemannian metric.
+        This method is not recommended for statistical purposes, as the
+        tangent vectors generated are not drawn from a distribution related
+        to the Riemannian metric.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


luisfpereira (approved):
Thanks @johnharveymath! Merging.

PR #1991, opened 2024-04-10
Fix dynamic programming algorithm

luisfpereira:
This PR fixes the dynamic programming algorithm. In particular, we ensure the end point of the aligned curve corresponds to the end point of the curve. I'm afraid we are not there yet, but hopefully this is a first step towards correctness.

Moreover, I've tried to improve naming in order to make the code more readable, but I'm afraid it is still too hard to follow.

**Known issues** (also occur in previous implementation)

The dynamic programming outputs a squared dist. After alignment, we can also compute a distance using the total space metric (this is the distance quotient metric outputs).

For more complex cases (thinking about birds), I've realized these distances are similar, but slightly off. Moreover, when we swap `point` and `base_point`, the dynamic programming squared dist is the same, but the distance we compute with the total space is not necessarily symmetric (though close).

For simpler cases as the curves generated by `random_point`, the results are better (correspondence between DP and total space; almost symmetric distance).

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -1584,10 +1584,13 @@ class SRVReparametrizationBundle(FiberBundle):
         Space of discrete curves starting at the origin
     """
 
-    def __init__(self, total_space):
+    def __init__(self, total_space, aligner=None):
+        if aligner is None:
+            aligner = IterativeHorizontalGeodesicAligner(total_space)
+
         super().__init__(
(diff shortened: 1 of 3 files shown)


ninamiolane (approved):
Thank you so catching this! LGTM.

@alebrigant does this look good to you?

luisfpereira:
I'm also merging this one, as it is a bug fix. @alebrigant, please let me know if there's something you disagree with, so that I can address it in a later PR.

PR #1992, opened 2024-04-11
Update discrete curves elastic metric for ratio equal to 1

luisfpereira:
This PR updates discrete curves elastic metric to handle the case where `a / 2b = 1`, for any ambient dimension, by using theorem 2.1 of [(Bauer, 2022)](https://arxiv.org/abs/2209.09862). It also uses the theorem 2.1 equation to improve testing (validates our diffeomorphism-based approach).

Notice how easy this implementation is, suggesting our abstraction are not that bad! Here we use: `Diffeo`, `PullbackDiffeoMetric`, `ScalarProductMetric`, `NFoldManifold`, `NFoldMetric`. Besides the specific implementation for the diffeos and the metric in the image space, we are simply reusing code without any specialization.

Next steps: extend implementation for ratios other than 1. Only need to implement a different metric for the image space.


geomstats/geometry/discrete_curves.py
```diff
@@ -715,12 +715,18 @@ def inner_product(self, tangent_vec_a, tangent_vec_b, base_point=None):
...
+    a and b.
 
-    See [NK2018]_ for details.
+    For curves in :math:`\mathbb{R}^2`, it pullbacks the L2 metric by
+    the F-transform (see [NK2018]_).
+
+    For curves in :math:`\mathbb{R}^d`, it pullbacks a scaled L2 metric by the SRV
+    transform (see [BCKKNP2022_). It only works for ratio :math:`a / 2b = 1`.
+
+    When a=1, and b=1/2, it is equivalent to the SRV metric.
```
ninamiolane, line 729: Very helpful, thank you so much for this docstring.

geomstats/test_cases/geometry/discrete_curves.py
```diff
@@ -1,13 +1,92 @@
...
+from geomstats.numerics.finite_differences import forward_difference
 from geomstats.test.random import get_random_times
 from geomstats.test.vectorization import generate_vectorization_data
 from geomstats.test_cases.geometry.fiber_bundle import FiberBundleTestCase
 from geomstats.test_cases.geometry.nfold_manifold import NFoldManifoldTestCase
+from geomstats.test_cases.geometry.pullback_metric import PullbackDiffeoMetricTestCase
 from geomstats.vectorization import get_batch_shape
 
 
+def elastic_distance(space, point_a, point_b, a=1.0, b=0.5):
```
ninamiolane, line 13: Should this overwrite `elastic_metric.dist` ? I guess this depends on the answer to the question: which one is faster 1) our current implementation of `dist` that uses our classes and the diffeo abstraction, or 2) this function?

ninamiolane (approved):
LOVE IT! Indeed, the abstractions work beautifully.

I just have one question, otherwise good to merge. Thank you!

geomstats/test_cases/geometry/discrete_curves.py
luisfpereira: Spot on!

A quick test has shown `elastic_distance` to be faster (though both are fast).
Still, I have a refactoring in mind for `NFoldMetric` that is likely to put both on pair or even make the current one faster (it will take advantage of the fact most operations are vectorized for multiple batch dimensions).

I suggest to merge as is, and check this again after the refactoring I'm suggesting for the `NFoldMetric`.


(We can actually make `elastic_distance` even faster by computing the length directly from the velocity (since we use the point, we need to do finite differences again inside).)

luisfpereira: We'll check this again later @ninamiolane. I'm merging now as it is useful.

PR #1993, opened 2024-04-15
Add horizontal projection from alignment

luisfpereira:
This PR adds the possibility of obtaining the horizontal projection of a tangent vector from the alignment method.

Assuming `horizontal_projection` is not overriden, this method tries to obtain the horizontal projection from:

1) the vertical projection

2) the horizontal lift of the tangent submersion

3) (new) the alignment method

These different possibilities come from the fact that different methods are easier to define in certain bundles.

This is motivated by the `SRVRotationBundle`, where `align` is straighforward (and we didn't have an implementation for horizontal projection; now we use the generic one).

### DIFF
--- a/geomstats/geometry/fiber_bundle.py
+++ b/geomstats/geometry/fiber_bundle.py
@@ -4,12 +4,17 @@ Lead author: Nicolas Guigui.
 """
 
 import math
-import sys
 from abc import ABC, abstractmethod
 
 import geomstats.backend as gs
 from geomstats.numerics.optimizers import ScipyMinimize
 from geomstats.vectorization import get_batch_shape
+
+
+def _from_base(method):
+    """Decorate method in order to avoid recursive calls."""
+    method._from_base = True
+    return method
 
 
 class AlignerAlgorithm(ABC):
@@ -200,6 +205,7 @@ class FiberBundle(ABC):
         """
         return gs.copy(point)
 
+    @_from_base
     def tangent_riemannian_submersion(self, tangent_vec, base_point):
         """Project a tangent vector to base manifold.
 
@@ -222,8 +228,7 @@ class FiberBundle(ABC):
         projection: array-like, shape=[..., {base_dim, [n, m]}]
             Tangent vector to the base manifold.
         """
-        caller_name = sys._getframe().f_back.f_code.co_name
-        if caller_name == "horizontal_projection":
+        if hasattr(self.horizontal_projection, "_from_base"):
             raise NotImplementedError(
                 "Tangent Riemannian submersion is not implemented."
             )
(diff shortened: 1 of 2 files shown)


geomstats/geometry/fiber_bundle.py
```diff
@@ -222,6 +222,12 @@ def tangent_riemannian_submersion(self, tangent_vec, base_point):
         projection: array-like, shape=[..., {base_dim, [n, m]}]
             Tangent vector to the base manifold.
         """
+        caller_name = sys._getframe().f_back.f_code.co_name
```
ninamiolane, line 225: Why do we need a syntax like this? Why not a NotImplementedError in the parent class, that does not appear when the class is indeed overridden?

The current syntax with == "horizontal_projection" does not seem robust to changes of names in methods, so I'm not sure it's a good practice?

geomstats/geometry/fiber_bundle.py
```diff
@@ -269,10 +275,21 @@ def horizontal_projection(self, tangent_vec, base_point):
...
 
-        return self.horizontal_lift(
-            self.tangent_riemannian_submersion(tangent_vec, base_point),
-            fiber_point=base_point,
-        )
+        if not (
+            caller_name == "horizontal_lift"
+            or caller_name == "tangent_riemannian_submersion"
+        ):
+            try:
```
ninamiolane, line 282: try/except is hacky. Any chance we could do this another way?

geomstats/geometry/fiber_bundle.py
```diff
@@ -294,7 +311,7 @@ def vertical_projection(self, tangent_vec, base_point):
         """
         caller_name = sys._getframe().f_back.f_code.co_name
         if caller_name == "horizontal_projection":
-            raise NotImplementedError
+            raise NotImplementedError("Vertical projection is not implemented.")
```
ninamiolane, line 314: I read this as if horizontal_projection is the method calling vertical_projection, then we raise the error.

Is this because, otherwise, we have vertical_projection calling horizontal_projection calling vertical_projection, etc... ? Any chance we could use a more friendly syntax?

ninamiolane (approved):
LGTM! I'm not a fan of the "hacks" with the caller names, but _I think_ I understand why we'd need them. Any chance we could use a more readable / less hacky syntax to achieve the same goal?

if not, good to merge!

geomstats/geometry/fiber_bundle.py
luisfpereira: Definitely not a good practice, though it is hard to find a better alternative that is as general.

No problems with maintaining this in any case, as our tests will scream at us if we change these names (also, these names are quite stable, we're probably not changing them at any time soon).

luisfpereira: Above comment is already updated. New solution seems to be more clear.

PR #1994, opened 2024-04-15
Simplify SRVRotationnReparametrizationBundle

luisfpereira:
This PR simplifies `SRVRotationReparametrizationBundle` by making it more general (results are the same). This way, it works independently of the aligner we use on `SRVReparametrizationBundle` (before it was only working for `IterativeHorizontalGeodesicAligner`).

Moreover, it opens a precedent for a more general bundle we can implement: `ComposedGroupActionFiberBundle`. The name is not great (suggestions?), but the idea is that we can create a `ComposedGroupAction` and, when a total space is equipped with it, we associate it with this bundle that aligns points by iteratively aligning wrt each group action (there's probably some compatibility condition that needs to be ensured).
In other words, this may be a simple way of avoiding implementing a fiber bundle for any combination of group actions, assuming there's no specific algorithms to find the optimal alignment wrt all group actions (we avoid a combinatorial explosion!).
If you agree with this idea, I can move on with the implementation.

We should probably also consider the case of commutative group actions, as I believe we don't need to go through an iterative process, but only compute the alignment wrt each group action once (this applies for sure in the discrete curves with ambient dim = 2).

P.S. I removed `return_rotation` because I'm already thinking about deleting this bundle if we move on with the implementation of `ComposedGroupActionFiberBundle`, but notice it is straightforward to obtain it (e.g. use the total space equipped with rotations).


geomstats/geometry/discrete_curves.py
```diff
@@ -1842,75 +1842,21 @@ def _init_space_with_reparametrizations(self, total_space):
...
-        bundle = self._total_space_with_reparametrizations.fiber_bundle
-        return bundle.aligner.discrete_horizontal_geodesic(base_point, point, spline)[
-            ..., -1, :, :
-        ]
-
     def _align_single(self, point, base_point):
-        """Align point to base point, non vectorized."""
-        aligned_point = gs.copy(point)
-        rotation = gs.eye(self._total_space.ambient_manifold.dim)
+        """Align point to base point."""
```
ninamiolane, line 1846: Add docstring and parameter check.

Or better: add automatic `numpydoc` validation checks in another PR?

See how we did it in TopoModelX?
https://github.com/search?q=repo%3Apyt-team%2FTopoModelX%20numpydoc&type=code

ninamiolane (approved):
Love it! Just one comment on the mathematical correctness, versus our (numerically equivalent, IMO) implementation + docstrings

geomstats/geometry/discrete_curves.py
luisfpereira: Added `numpydoc` to the todo list and updated docstring.

PR #1995, opened 2024-04-18
Add quotient structure registry

luisfpereira:
This PR implements a similar strategy to register quotient structures as done in #1964 to register methods of the scalar product metric.

Though the code is a bit verbose, the main advantages of this strategy are:

1) adding the quotient structure is decoupled from the __init__ method of the class. As a consequence, we (both developers and users) can register quotient structures at any moment. Moreover, overriding quotient structures to e.g. experiment new fiber bundle/quotient metric implementations becomes very easy and natural (instead of hacky)

2) nice error messages are raised when using `equip_with_quotient_structure`, in particular, showing which are the quotient structures available for a given space or combination of space/metric or space/group action


geomstats/geometry/group_action.py
```diff
@@ -69,6 +70,23 @@ def __call__(self, group_elem, point):
         return self._group.compose(point, group_elt)
 
 
+class LieAlgebraBasedSpecialOrthogonalAction(LieAlgebraBasedGroupAction):
```
ninamiolane, line 73: I don't think the `point_type` (or `coords_type`? 🤔 ) of the group's elements should appear in the name of the class.

Because, in the code base, we usually make the distinction between the math and the way the math is implemented.

Class names <-> Math
The name of the class should refer to the mathematical construct: here, the fundamental action of SOn.

Parameters in class <-> Implementation
We can have a parameter called `point_type`, `coords_type` or else (not sure point_type and coords_type are what we want) that says how the group's points are parameterized.

Similarly to matrix_representation and basis_representation for the points in matrix-spaces.

geomstats/geometry/manifold.py
```diff
@@ -264,3 +274,161 @@ def projection(self, point):
...
+
+
+class _QuotientStructureRegistry:
+    """Registry for quotient structures."""
+
+    STRUCTURES = {}
+
+    @classmethod
+    def _as_key(self, Obj):
+        """Transform an instance of a class into a key type.
```
ninamiolane, line 286: What is a "key type"? Could you give an example or explanation in the docstring?
For future contributors

geomstats/geometry/discrete_curves.py
```diff
@@ -15,6 +15,7 @@
 from geomstats.geometry.fiber_bundle import AlignerAlgorithm, FiberBundle
 from geomstats.geometry.hypersphere import Hypersphere
 from geomstats.geometry.landmarks import Landmarks
+from geomstats.geometry.manifold import register_quotient_structure
```
ninamiolane, line 18: NIT: I wonder if register_quotient_structure should be "hidden" in manifold.

More generally, I feel that manifold.py and base.py might be hidden important code. base.py has LevelSet, ImmersedSet, etc if I remember correctly.

I would recommend having: manifold.py, level_set.py, immersed_set.py, quotient.py, etc visible.

Not for this PR though, but to be discussed?

ninamiolane (approved):
LGTM! Thank you, I like the design! Just see my minor comments.

geomstats/geometry/group_action.py
luisfpereira: Great point. I'll change accordingly.

geomstats/geometry/manifold.py
luisfpereira: Not a good name. It refers to the key that is used in the `STRUCTURES` `dict`. I'll update the docstrings.

geomstats/geometry/discrete_curves.py
luisfpereira: What do you mean by 

> register_quotient_structure should be "hidden" in manifold"

I kept it public because the method can potentially be used by users without having to import `Manifold` (a bit like the similar method @johnharveymath developed in #1964).


I think the other points are relevant. We should probably split `base` into multiple files in a future PR. We may need to check if there will be no circular imports (though I believe not).

geomstats/geometry/group_action.py
luisfpereira: @ninamiolane, I've added a boolean (`lie_algebra_repr`) to control representation type (if `True`, it uses considers group elements to be represented as vectors of the Lie algebra).

I've also renamed it to `ComposeAction` (probably not a great name). You have a different name to it, right? What is it? Should we rename?

geomstats/geometry/discrete_curves.py
ninamiolane: Sorry I wasn't clear. I didn't mean public/private, but rather separated in a different file. I feel that manifold.py and base.py contain too many important classes, and we should probably split both manifold.py and base.py into several files.

Also, we should look into our init.py and import strategies, i.e., starting using `__all__` and other syntaxes like that to ease imports.

Not a priority, some next PR!

geomstats/geometry/discrete_curves.py
```diff
@@ -1905,3 +1892,26 @@ def align(self, point, base_point):
                 for point_, base_point_ in zip(point, base_point)
             ]
         )
+
+
+register_quotient_structure(
```
ninamiolane, line 1897: NIT: See my comment on the PR on alternate aligner about naming the parameters for readbility.

geomstats/geometry/group_action.py
```diff
@@ -29,33 +30,37 @@ def __call__(self, group_elem, point):
         """
 
 
-class LieAlgebraBasedGroupAction(GroupAction):
+class ComposeAction(GroupAction):
```
ninamiolane, line 33: Thanks! It makes it easier to read.

But now it makes me wonder: why do we need this class in addition to the already existing `Group` (or `LieGroup` or `MatrixLieGroup`?) class?

I think we should rename `ComposeAction` to `Group` (or `LieGroup` or `MatrixLieGroup`?) and import here the code from `Group` (or `LieGroup` or `MatrixLieGroup`?).

A group is basically a set with a group action acting on itself...

ninamiolane (commented):
Thanks! One comment on the fact that we should probably merge ComposeGroup and Group... possibly answering your question about finding a better name.

geomstats/geometry/group_action.py
ninamiolane: Yes, I agree that GroupAction is different than Group.

Yes, our current ComposeGroupAction is, in effeect,  a group. A group is G x G -> G

PR #1996, opened 2024-04-19
Add alternate alignment algorithm

luisfpereira:
This PR builds on top of #1995 to add the alternate alignment algorithm  as discussed in #1994 (EDIT: first we thought about implement it as `ProductFiberBundle`). It also removes `SRVRotationsReparametrizations`.

Only doubt here is if we want to use the total space metric to check convergence or simply an Euclidean difference. For now using total space.

NB: click on last commit to see changes coming from this PR.


geomstats/geometry/fiber_bundle.py
```diff
@@ -489,3 +490,120 @@ def integrability_tensor_derivative(
             in Kendall shape spaces. Unpublished.
         """
         raise NotImplementedError
+
+
+class ProductFiberBundle(FiberBundle):
```
ninamiolane, line 495: I think "ProductFiberBundle" seems to be saying that we have a product: total_space_1 x total_space_2 since "fiber bundle" often refers to the underlying total_space in the math literature.

This could be confusing since what we want is actually one single total_space with two group actions, and not the product of two total_spaces each with its group action.

My question: why having a new class ProductFiberBundle at all? Can't we put this into FiberBundle directly?

The Rationale being that:

Class Names <-> only refer to the Mathematics: here, we have one single fiberbundle, with one single action, of a more complicated group (a product group) that we _implement_ as a product.

But "Product" should not appear in the class name, because it's purely an implementation choice (and an approximation performed in the alignment procedure). Or, make that class name private.

ninamiolane (approved):
Very nice!

My only comment is about *not* to make our implementation choices appear in the class names, and keep class names purely Mathematics-based.

Thank you!

geomstats/geometry/fiber_bundle.py
luisfpereira: Here `Product` comes from a product of groups, not product of total spaces. I can see why looking to the implementation may point towards the second.

Since what defines this fiber bundle is the alignment algorithm, to avoid creating this new bundle, I suggest implementing it as an `AlignerAlgorithm` (a bit like e.g. `DynamicProgramming`) and then use the same bundle. Would this be preferable to you? Probably is what has more logical sense.

Thanks for raising this point!

luisfpereira: @ninamiolane, I've addressed this comment, please let me know what you think. Implementation hasn't changed, it is really just a matter of code organization. I think this version makes more math/logical sense.

geomstats/geometry/fiber_bundle.py
```diff
@@ -153,11 +271,14 @@ class FiberBundle(ABC):
     def __init__(self, total_space, aligner=None):
         self._total_space = total_space
         if aligner is True:
-            aligner = (
-                DistanceMinimizationBasedAligner(total_space)
-                if not gs.__name__.endswith("numpy")
-                else None
-            )
+            if isinstance(total_space.group_action, tuple):
+                aligner = AlternateAlignmentAlgorithm(total_space)
```
ninamiolane, line 275: NIT: AlternateAlignmentAlgorithm --> AlternatingAligner (shorter, and closer to the other names we use that end with "Aligner" and not "Alignment". The "Algorithm" part is not needed.)

geomstats/geometry/fiber_bundle.py
```diff
@@ -153,11 +271,14 @@ class FiberBundle(ABC):
...
+            else:
+                aligner = (
+                    DistanceMinimizationBasedAligner(total_space)
```
ninamiolane, line 278: DistanceMinimizingAligner? shorter?

geomstats/geometry/rank_k_psd_matrices.py
```diff
@@ -338,7 +330,20 @@ def __init__(self, space, total_space=None):
...
+                SpecialOrthogonalComposeAction(total_space.k)
+            )
+
+        if not hasattr(total_space, "quotient"):
+            total_space.equip_with_quotient_structure()
 
         super().__init__(space=space, total_space=total_space)
+
+
+register_quotient_structure(
```
ninamiolane, line 344: Ditto: name parameters for readability for other contributors

ninamiolane (approved):
LGTM! Nit on names! Thank you so much

PR #2000, opened 2024-06-10
Add (naive) unparameterized surfaces

luisfpereira:
This PR adds (naive) unparameterized surfaces. Naive comes from the fact we only implement a very non-performant relaxed alignment algorithm (needs more sophisticated version to be of use in practice).

 Implementation of  the quotient metric simply relies on `RelaxedPathStraightening` and existing quotient structure. `ReparametrizationBundle` is only created for the instantiation of a proper aligner.

`RelaxedPathStraightening` is both an aligner algorithm (implements `align`) and a `LogSolver` (implements `geodesic_bvp` and `log`). `QuotientMetric` is updated to handle these objects (it will also have impact on curves - probably of your interest @alebrigant; avoids repeating computations).
This because algorithms that are based on a discrete path to compute the alignment also compute a discrete geodesic during the process (and therefore a log).

`Surface` is required because unparameterized surfaces cannot be represented simply by the vertices (we need both the vertices and the faces). Output of any method (e.g. `exp`, `log`, `geodesic`) is an array (the vertices wrt the faces of the total space).
If input is array, we assume faces of total space.
Combinations of array-surface as input are handled (see `discrete_geodesic_bvp`).

The relaxation term is implemented generically. As default, we use a `VarifoldMetric`. Imports are done in a way that avoids importing from `geomstats.varifold`, as I've added `pykeops` as an OPTIONAL dependency.
This justifies `Surfaces` being defined in `geomstats._mesh`.

To avoid an overdependency in `pykeops` and make our code flexible, we have `SurfacesKernel`. It controls how to create a kernel and how to pass arguments to the kernel defined by `pykeops`.

I've added a limited number of kernels, but it is something easy to extend (more concerned with overall structure for now).

`VarifoldMetric` is a simple implementation that relies on `SurfaceKernel`.

Tests are still missing.

Next steps (for later): add multiresolution methods for alignment. (It should be easy with this structure, as long as we have good ways to decimate and upsample meshes -> need to rely on external libraries for this.)


geomstats/_mesh.py
```diff
@@ -0,0 +1,44 @@
+"""Mesh-related data structures."""
+
+import geomstats.backend as gs
+
+
+class Surface:
```
ninamiolane, line 6: I understand why we need the faces-vertices tuple in the computations... though could we avoid creating another class?

Because:
- we have a class explosion in Geomstats: so many classes of different types... the architecture diagrams start to be messy. I would be in favor of only using classes when we leverage inheritance.
- I remember other packages (like scikit-timeseries) who made the choice of encompassing their data into objects, and had many problems for compatibility with other python packages afterwards, possibly regretting this choice?
- doesn't it break automatic differentiation to have data encompassed into objects?

geomstats/geometry/discrete_surfaces.py
```diff
@@ -67,6 +73,10 @@ def __init__(
             equip=equip,
         )
 
+    def new(self, equip=True):
```
ninamiolane, line 76: Why having `new` in DiscreteSurfaces but not in any other Manifold?

geomstats/geometry/discrete_surfaces.py
```diff
@@ -67,6 +73,10 @@ def __init__(
...
+        """Create manifold with same parameters."""
```
ninamiolane, line 77: Need Parameters description. Esp since equip is actually not used?

Explain what "same" parameters mean in the docstring.

geomstats/geometry/discrete_surfaces.py
```diff
@@ -1127,3 +1139,254 @@ def geodesic_ivp(self, tangent_vec, base_point):
...
+                options={"disp": False},
+            )
+
+        self.lambda_ = lambda_
+        self.discrepancy_loss = discrepancy_loss
+        self.path_energy = path_energy
+        self.optimizer = optimizer
+        self.initialization = initialization
+
+    def _default_discrepancy_loss(self):
```
ninamiolane, line 1209: Missing Return in the docstring

geomstats/geometry/discrete_surfaces.py
```diff
@@ -1127,3 +1139,254 @@ def geodesic_ivp(self, tangent_vec, base_point):
...
+        kernel = SurfacesKernel(
+            position_kernel,
+            tangent_kernel,
+        )
+        varifold_metric = VarifoldMetric(kernel)
+        return lambda point: varifold_metric.loss(
+            point, target_faces=self._total_space.faces
+        )
+
+    def _default_initialization(self, point, base_point):
```
ninamiolane, line 1229: Missing Return in the docstring

geomstats/geometry/discrete_surfaces.py
```diff
@@ -1127,3 +1139,254 @@ def geodesic_ivp(self, tangent_vec, base_point):
...
+        -------
+        aligned_point : array-like, shape=[..., *total_space.shape]
+            Aligned point.
+        """
+        discr_geod_path = self.discrete_geodesic_bvp(point, base_point)
+        point_ndim_slc = (slice(None),) * self._total_space.point_ndim
+        return discr_geod_path[..., -1, *point_ndim_slc]
+
+
+class ReparametrizationBundle(FiberBundle):
```
ninamiolane, line 1372: Missing elements in the docstrings

geomstats/varifold.py
```diff
@@ -0,0 +1,226 @@
...
+        K(u, v)=e^{-2 \langle u, v \rangle ^2 / \sigma^2}
+
+    Generates the expression: `Exp(IntCst(2)*b*((u|v)-IntCst(1)))`
+    """
+    u, v = Vi(init_index, dim), Vj(init_index + 1, dim)
+    b = 1 / (sigma * sigma)
+    return (2 * b * ((u * v).sum() - 1)).exp()
+
+
+class VarifoldMetric:
```
ninamiolane, line 154: Doesn't inherit from RiemannianMetric? Explain in the docstring why not.

geomstats/varifold.py
```diff
@@ -0,0 +1,226 @@
...
+
+        def squared_dist(vertices):
+            point = Surface(vertices, target_faces)
+            return (
+                kernel_target
+                - 2 * self.kernel(target_point, point)
+                + self.kernel(point, point)
+            )
+
+        return squared_dist
```
ninamiolane, line 226: Fill missing descriptions in docstrings

pyproject.toml
```diff
@@ -51,10 +51,11 @@ lint = [
 ]
 test = ["pytest", "pytest-cov", "coverage", "jupyter", "ipython < 8.23"]
 graph = ["networkx"]
+varifold = ["pykeops"]
 autograd = ["autograd >= 1.3"]
 pytorch = ["torch >= 1.9.1"]
 backends = ["geomstats[autograd, pytorch]"]
-opt = ["geomstats[graph]"]
+opt = ["geomstats[graph, varifold]"]
```
ninamiolane, line 58: name it "pykeops" instead of varifold, since we use the name of packages in the pyproject.toml (not the name of the math...)

geomstats/geometry/discrete_surfaces.py
luisfpereira: All the spaces that are used as total space in the context we do not have explicit representation of base need to define this method (see e.g. `DiscreteCurvesStartingAtOrigin`).

geomstats/_mesh.py
luisfpereira: `Surface` has two purposes:

1. it allows to keep our api simple: e.g. `log(point, base_point)`, `geodesic(initial_point, end_point=end_point)`

2. it caches information to avoid repeated computation of surface information (e.g. faces normals, areas, and centroids); this is especially relevant when computing varifold loss of a point against the same point (`end_point`) during optimization


**More on the first point**

Some packages choose to use two arrays to represent a surface mesh (`vertices`, `faces`). We can't do it without changing our API: e.g. `log(vertices, faces, base_vertices, base_faces)`. Besides being annoying, we couldn't use the methods for e.g. `QuotientMetric` as is currently the case (notice the only method we specialize is `discrete_geodesic_bvp`).

We could choose to represent points as a `tuple` instead of a class. `(vertices, faces)`. This keeps our API valid, though 1) we do not have anymore the possibility of caching (important for second point); 2) any time we need information on a mesh we need to have a function taking `vertices` and `faces`, which we cannot override (right now we do `mesh.face_centroids`; we could provide a different method to compute centroids by defining a different `Surface` object); 3) we could only handle batches with lists `list[Surface]`, as tuples would be reserved for the surface representation (with current implementation, batch can be `list[Surface]` or `tuple[Surface]`). 


**Other comments**

Compatibility should not be an issue, as this object is very simple. Anyone could even use an encapsulated version of their surface representation by creating a new mesh object with the following attributes (`vertices`, `faces`, `face_centroids`, `face_normals`, `face_areas`). In this case, they would be the ones taking advantage of inheritance.
It can be annoying, but I can't see any better alternative in this case.

No problem with automatic differentiation (notice we use it to solve the optimization problem).

geomstats/geometry/discrete_surfaces.py
luisfpereira: `equip` is used. A mistake on my side. Fixed.

luisfpereira:
I'm merging. Tests for the unparametrized surfaces are very naive (they just ensure the code runs), but I think it is the best we can do given the current naive implementation of the relaxed alignment problem.

PR #2003, opened 2024-06-10
Homogenize use of reparametrization

luisfpereira:
This PR replaces uses of "reparameterization" by "reparametrization". Though both are valid, the goal is to reduce confusion, as `space.equip_with_group_action("reparametrizations")` expects this exact spelling.


ninamiolane (approved):
Thanks!

PR #2004, opened 2024-06-10
Update curves bundles

luisfpereira:
This PR updates SRVReparametrizationBundle to handle any elastic parameter (the code was already prepared for that) and renames it `ReparametrizationBundle`.

@alebrigant, regarding `SRVRotationBundle`, the `SRV` stands for the SRV transformation or for the elastic parameters of  the total space? In other words, if I use different a, b, is the alignment wrt rotations still proper?

EDIT: `SRVRotationBundle` -> `RotationBundle`. It uses the diffeo available in the `PullbackDiffeoMetric` that endows the space.

### DIFF
--- a/geomstats/geometry/discrete_curves.py
+++ b/geomstats/geometry/discrete_curves.py
@@ -1706,11 +1706,11 @@ class ReparametrizationBundle(FiberBundle):
         return tangent_vec_ver
 
 
-class SRVRotationBundle(FiberBundle):
-    """Principal bundle of curves modulo rotations with the SRV metric.
+class RotationBundle(FiberBundle):
+    """Principal bundle of curves modulo rotations with an elastic metric.
 
     This is the fiber bundle where the total space is the space of parameterized
-    curves equipped with the SRV metric, the action is given by rotations, and
+    curves equipped with an elastic metric, the action is given by rotations, and
     the base space is the shape space of curves modulo rotations.
 
     Parameters
@@ -1743,9 +1743,9 @@ class SRVRotationBundle(FiberBundle):
         aligned : array-like, shape=[..., k_sampling_points - 1, ambient_dim
             Curve optimally rotated with respect to reference curve.
         """
-        srv_transform = self._total_space.metric.diffeo
-        initial_srv = srv_transform.diffeomorphism(base_point)
-        end_srv = srv_transform.diffeomorphism(point)
+        transform = self._total_space.metric.diffeo
+        initial_srv = transform.diffeomorphism(base_point)
+        end_srv = transform.diffeomorphism(point)
 
         mat = gs.matmul(Matrices.transpose(initial_srv), end_srv)
         u_svd, _, vt_svd = gs.linalg.svd(mat)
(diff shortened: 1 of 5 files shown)


alebrigant (approved):
Great ! `SRVRotationBundle` should be extendable to elastic metrics by replacing the SRV transform by the more general F-transform. This can be put in the docstring of the align method:
```
Find optimal rotation of curve with respect to base curve, by
using Procrustes analysis on the F-transforms. More precisely,
find the matrix A in SO(n) that minimizes

.. math::
        \langle F(c_0), AF(c_1)\rangle_{L^2}

This is acheived by an SVD decomposition.
```

luisfpereira:
Thanks @alebrigant! What about `ambient_dim > 2`? That is, since we're pulling back a scaled metric on the image of SRV, would this still work? If that the case, I imagine it should also work for `ambient_dim = 2`.

alebrigant:
> Thanks @alebrigant! What about `ambient_dim > 2`? That is, since we're pulling back a scaled metric on the image of SRV, would this (i.e. the current implementation) still work? If that the case, I imagine it should also work for `ambient_dim = 2`.

The current implementation should work in any dimension: obtain the optimal rotation matrix A in SO(n) by an SVD decomposition of $\int q_0(t)q_1(t)^T dt$, where $q_0$, $q_1$ are the F-transforms, i.e. curves in $\R^n$ when `ambient_dim` is n.

luisfpereira:
@alebrigant, the issue is the F-transform only works for `dim=2`, right?

alebrigant:
> @alebrigant, the issue is the F-transform only works for `dim=2`, right?

Ooh yes you are right. So I guess for now, the SRV is working in any dimension, but the elastic framework only for plane curves ? To make it work in any dimension we will really need to implement the cone metric. Then alignment w.r.t. rotations will also need to be modified.

luisfpereira:
@alebrigant, current status is:

1. we can use any elastic parameter when `dim = 2` because we rely on `FTransform`

2. for `dim> 2`, we can choose `a`, `b` such that `a / 2 b = 1` (that is, we have it already working for cases other than SRV, though, as you mention we still need the cone metric to extend for other ratios)


That being said, I see how we can make `RotationBundle` work for 1. What would be the changes needed to case 2 (for a, b other than SRV)?

alebrigant:
> @alebrigant, current status is:
> 
>     1. we can use any elastic parameter when `dim = 2` because we rely on `FTransform`
> 
>     2. for `dim> 2`, we can choose `a`, `b` such that `a / 2 b = 1` (that is, we have it already working for cases other than SRV, though, as you mention we still need the cone metric to extend for other ratios)
> 
> 
> That being said, I see how we can make `RotationBundle` work for 1. What would be the changes needed to case 2 (for a, b other than SRV)?

For the second case, the algorithm should be the same if the metric being pulled back by the SRV transform is just a scaled version of the L2 metric. But I need to check, and confirm this to you.

luisfpereira:
@alebrigant, I'm merging. For rotations, for now (and as before), it makes use of the diffeo that is in the metric, which is probably the definite solution will end up given your comment:

> For the second case, the algorithm should be the same if the metric being pulled back by the SRV transform is just a scaled version of the L2 metric.

PR #2005, opened 2024-06-11
Fix issues with sklearn 1.5.0

luisfpereira:
This PR is a quick fix for issues that appear with [sklearn 1.5.0. release](https://pypi.org/project/scikit-learn/1.5.0/) when  using `torch` as backend. `sklearn` now checks the device, in a way that invalidates passing `torch` tensors for some functions. 

For now, a quick and dirty wrapper does the job. We should think about a strategy for the long term though. I believe that changes in `sklearn` can actually benefit us in the long run, but we should confirm where they're aiming towards.

### DIFF
--- a/geomstats/learning/_sklearn.py
+++ b/geomstats/learning/_sklearn.py
@@ -1,12 +1,34 @@
-from sklearn.decomposition import PCA
-from sklearn.linear_model import LinearRegression
+"""Collections of wrappers around scikit-learn.
+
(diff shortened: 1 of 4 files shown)


luisfpereira:
@ninamiolane, I'm merging has this is kind of urgent to prevent testing failures/allow users to use latest sklearn version. At some point we need to revisit our uses of `sklearn`.

PR #2011, opened 2024-06-13
Add ruff

### LINKED ISSUE #1927: Use ruff for linting in the CI
[ruff](https://github.com/astral-sh/ruff) is definitely faster than `flake8` (see repo README) and supports "Drop-in parity with [Flake8](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-flake8), isort, and Black".

I think we should replace them in our CI and update pre-commit to be based on `ruff`.

NB: it is already used in some libraries in our ecosystem (e.g. [SciPy](https://github.com/scipy/scipy)).

Green light here, @ninamiolane? (I'm thinking about also using it in [neurometry](https://github.com/geometric-intelligence/neurometry).)

luisfpereira:
This PR adds `ruff` and removes `black`, `flake8`, and `isort` (while keeping similar behavior). Github action is updated accordlingly. Several files have been also updated to fix linting.

closes #1927.


ninamiolane (approved):
*Excellent* idea. Thanks for taking the time to do this! Good to merge when the last test passes.

PR #2016, opened 2024-06-14
Add L2SurfacesMetric

luisfpereira:
This PR adds a very basic `L2SurfacesMetric` to put on `DiscreteSurfaces` that is basically a standard L2 metric on landmarks with inner products being weighted by number of vertices.


geomstats/geometry/discrete_surfaces.py
```diff
@@ -1154,6 +1155,43 @@ def geodesic_ivp(self, tangent_vec, base_point):
         )
 
 
+class L2SurfacesMetric(NFoldMetric):
+    """L2 metric on the space of discrete surfaces."""
```
ninamiolane, line 1159: NIT: missing description of base_manifold and n_copies attributes.

ninamiolane (approved):
Nice, thank you!

geomstats/geometry/discrete_surfaces.py
luisfpereira: Improved docstrings as a note (as these are not attributes of the metric).

Issue #2028, opened 2024-09-13
New implementation of Affine Invariant squred distance for SPD matrices


dherrera1911:
I'm thinking of doing my first pull request, so asking a few questions here before I do that.

I wish to contribute an alternative implementation of the Affine Invariant SPD distance that is 2x to 5x faster than geomstats implementation, both with numpy as pytorch backends. Since this is the most used metric in SPD matrix manifold, it seemed like a valuable contribution. My implementation uses the generalized eigenvalues formula for the AI distance. I just want to check how I would go about adding it, regarding the questions below:

1. I implemented a self-contained function `generalized_eigenvalues(A,B)` using the gs backend that is called for computing the AI distance. I wonder where I should include this function. As a separate function in the same spd_matrices.py module? in the algebra_utils.py file?
2. Taking the SPDBuresWassersteinMetric class as an example, it seems that what I need to do is define a function `squared_dist(self, point_a, point_b)` for the `SPDAffineMetric` class using my implementation. Does this sound right?
3. Do I need to do anything other than checking that the tests run for sending a pull request? (I tested manually that the two implementations give the same result)

Thanks

Issue #1082, opened 2021-08-10
Error for operations on different devices: CPU versus CUDA


ninamiolane:
Issue reported by Nicolas Legendre (https://github.com/NicolasLegendre1)

For this code:
![image](https://user-images.githubusercontent.com/8267869/128795411-3d82704f-0be6-4be1-810b-cd9444dbd179.png)
![image](https://user-images.githubusercontent.com/8267869/128795420-9a7b6dc4-a861-4184-827a-b521bd49b841.png)

Nicolas gets the following error:
![image](https://user-images.githubusercontent.com/8267869/128795429-fde101d8-b017-429a-baf4-b758ca43370e.png)

And for this code:
![image](https://user-images.githubusercontent.com/8267869/128795447-c517ea00-61f5-44a3-8495-a530ae035bab.png)

He gets:
![image](https://user-images.githubusercontent.com/8267869/128795456-6912fb8c-74cd-49af-807b-3f0f908d05dc.png)

We should test geomstats on cpu + cuda devices, as there seem to be several incompatibilities.

mvdoc:
Hi, first of all thank you so much for this package. I'm exploring riemaniann geometry for my work and this package and the learning materials are amazing.

I hit the same problem when trying to compute the frechet mean on a torch tensor on GPU. Here's a minimal reproducible example and the resulting traceback.

```python
import os
os.environ["GEOMSTATS_BACKEND"] = "pytorch"
import torch

import geomstats.backend as gs
import geomstats.geometry.spd_matrices as spd

manifold = spd.SPDMatrices(20)

# Generate some random data
data = gs.array([manifold.random_point() for _ in range(10)])
# Push to GPU
data = data.to("cuda")

# Compute frechet mean
from geomstats.learning.frechet_mean import FrechetMean
mean = FrechetMean(metric=manifold.metric)
mean.fit(data)
``` 

<details>
<summary>Full traceback</summary>

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-7-abb08f84a022> in <module>
     16 from geomstats.learning.frechet_mean import FrechetMean
     17 mean = FrechetMean(metric=manifold.metric)
---> 18 mean.fit(data)

src/lib/geomstats/geomstats/learning/frechet_mean.py in fit(self, X, y, weights)
    544                 epsilon=self.epsilon,
    545                 verbose=self.verbose,
--> 546                 init_point=self.init_point,
    547             )
[...]

PR #2029, opened 2024-09-24
Faster more stable implementation of Affine Invariante SPD squared_dist

dherrera1911:
## Checklist

- [X] If necessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [X] I made sure the code passes all unit tests. (refer to comment below)
- [X] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#coding-style-guidelines) and API.
- [X] My code is properly documented and I made sure the documentation renders properly. ([Link](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#documentation))
- [X] I linked to issues and PRs that are relevant to this PR.

## Description

I implemented an explicit formula for the squared Affine Invariant distance in the SPD manifold. The implementation uses the generalized-eigenvalue formula for the distance. Generalized eigenvalues are computed with the method of section 7.2 in the paper [Eigenvalue and Generalized Eigenvalue Problems: Tutorial](https://arxiv.org/pdf/1903.11240).

## Issue

The implementation can be up to 4 or 5 times faster than the current implementation, and it also seems more numerically stable in some tests I ran (where current implementation returned NaN's while the new implementation didn't). Also, it doesn't involve creating arrays with the gs backend. This means that if called for PyTorch Tensors that are in the GPU, the new implementation will work, while the current one will throw an error of arrays in different devices (https://github.com/geomstats/geomstats/issues/1082#issuecomment-2349334826).

I started Issue  #2028 with some questions about where to put the Generalized Eigenvalue function, but that could be closed now, since it will be reviewed.

## Additional context

I wrote a script (attached as .txt) that contains the new implementations as a stand-alone function, and compares both the results and the time of the stand-alone function to the geomstats implementation. I could incorporate this script as a benchmark or similar if guided as to where/how to do that.
[test_ai_distance.txt](https://github.com/user-attachments/files/17118489/test_ai_distance.txt).

I didn't add unit tests because it doesn't seem like it's a new functionality and there's already tests for Affine Invariant distance I suppose? But I can add tests if needed.

I also added the package pykeops to the list of packages for geomstats with tests, since it seemed to be a missing dependency when I was running the tests preparing this contribution.

### DIFF
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -96,6 +96,36 @@ def _aux_differential_power(power, tangent_vec, base_point):
     return (eigvectors, transp_eigvectors, numerator, denominator, temp_result)
 
 
+def generalized_eigenvalues(point_a, point_b):
+    """Compute the generalized eigenvalues of SPD matrix pair.
+
+    point_auxiliary function to the functions differential_power and
+    inverse_differential_power.
+
+    Parameters
+    ----------
+    point_a : array_like, shape=[..., n, n]
+        Symmetric positive definite matrix.
+    point_b : array_like, shape=[..., n, n]
+        Symmetric positive definite matrix.
+
+    Returns
+    -------
+    generalized_eigenvalues : array-like, shape=[...]
+    """
+    # Compute eigendecomposition of point_b
+    eigvals_b, eigvecs_b = gs.linalg.eigh(point_b)
+    # Get matrix turning point_b into identity by congruence
+    inv_sqrt_eigvals_b = gs.sqrt(1.0 / eigvals_b)
+    inv_sqrt_eigvals_b = gs.expand_dims(inv_sqrt_eigvals_b, -2)
+    scaled_b_eigvecs = eigvecs_b * inv_sqrt_eigvals_b
+
+    # Apply congruence to point_a and get generalized eigenvalues
+    point_a_scaled = gs.moveaxis(scaled_b_eigvecs, -1, -2) @ point_a @ scaled_b_eigvecs
+    generalized_eigenvalues = gs.linalg.eigvalsh(point_a_scaled)
+    return generalized_eigenvalues
+
+
 class SymMatrixLog(Diffeo):
     """Matrix logarithm diffeomorphism.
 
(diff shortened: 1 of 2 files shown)


geomstats/geometry/spd_matrices.py
```diff
@@ -644,6 +675,31 @@ def log(self, point, base_point):
...
+        squared_dist : array-like, shape=[...]
+            Riemannian squared distance.
+
+        Notes
+        -----
+        Use of `abs` in the output prevents nan when calling
+        `sqrt` in very small negative outputs (e.g. -1e-16).
+        """
+        gen_eigvals = generalized_eigenvalues(point_a, point_b)
+        return gs.abs(gs.sum(gs.log(gen_eigvals) ** 2, axis=-1))
```
luisfpereira, line 701: Have you checked this happens often? What would be the cases? Distance of a point against itself?

luisfpereira (approved):
Thanks @dherrera1911!

I've done small cleanups to the code, but no dramatic change.

I've removed `pykeops` from test dependencies because that dependencies are more related with test-related packages than packages we need to run the code. For example, notice you can do `pytest tests/tests_geomstats/test_geometry/test_spd_matrices.py::TestSPDAffineMetric` and will not need `pykeops`.
You can all install everything you need with `pip install geomstats[all]`.

I've checked the new code is slightly faster, both for multiple points and high dimensions. For now we are not running benchmarks, but I'll keep your suggestion in mind: an interesting general benchmark is to compute the distance using the log and with the particular implementations (the particular implementation must be faster, otherwise we should probably remove it).

Accuracy is a bit harder to evaluate, but I imagine you're right and it is more accurate, since it does not involve computing `logm`, which may be problematic in some cases.

Great it works with the GPU out of the box!

I think for now we can leave `generalized_eigenvalues` where you've put it. 

Please let me know if you agree with everything.

(Sorry for the very late reply.)

geomstats/geometry/spd_matrices.py
dherrera1911: I haven't actually, I just copied it (code and motivation) from the Bures Wasserstein squared distance down below. I guessed that if that distance had this issue, the Affine Invariant would too.

dherrera1911: All changes look good. It's nice to get a feel for the style for potential future contributions.

Small note: I mentioned the new procedure is more stable, not more accurate (probably related though). If you generate the spd matrices by generating n-by-b matrix `A` with randn and taking `SPD = A @ A.t()`, a non-trivial proportion of times the original implementation will generate Nan's, while the new one won't. In the script I shared, SPD matrices are sampled with the geomstats built in method and the original method doesn't generate NaN's.

luisfpereira: I've tested without `abs` and it seems to work properly. For now I'll merge without it, and if we have problems later I'll add it back in (I prefer to have the less "hacking" possible in the code if that works fine).

Thanks for the clarification on stable vs accurate. I agree with you that they're probably very much correlated.

luisfpereira:
Thanks for the contribution @dherrera1911. I'm merging now. Looking forward for new contributions!

PR #2036, opened 2024-12-04
More efficient Bures Wasserstein distance computation in SPD manifold

dherrera1911:
- [X] If necessary, my code is [vectorized](https://www.geeksforgeeks.org/vectorization-in-python/).
- [X] I made sure the code passes all unit tests. (refer to comment below)
- [X] My PR follows [geomstats coding style](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#coding-style-guidelines) and API.
- [X] My code is properly documented and I made sure the documentation renders properly. ([Link](https://github.com/geomstats/geomstats/blob/main/docs/contributing.rst#documentation))

## Description

I implemented a more efficient version of the Bures Wasserstein squared distance. The new method is 2 to 4 times faster than the original one.

The improved efficiency comes from the `cross_term`, which is the trace of matrix C, which is the square root of matrix C_sq = (A^{1/2} @ B @ A^{1/2}). Matrix C_sq is symmetric positive definite, so the eigenvalues of C are the square root of the eigenvalues of C_sq. Because tr(C) equals the sum of the eigenvalues of C, we compute tr(C) by:

- Computing eigenvalues of C2
- Taking their square root
- Adding the resulting eigenvalues

This procedure does not require actually computing C, but only the eigenvalues of C_sq.

The original method actually computed C, which requires computing both eigenvalues and eigenvectors of C_sq.

I also changed the computing of `point_a_sqrt` to use `apply_func_to_eigvalsh`, which seems to be preferred geomstats style.

### DIFF
--- a/geomstats/geometry/spd_matrices.py
+++ b/geomstats/geometry/spd_matrices.py
@@ -892,16 +892,17 @@ class SPDBuresWassersteinMetric(RiemannianMetric):
         Use of `abs` in the output prevents nan when calling
         `sqrt` in very small negative outputs (e.g. -1e-16).
         """
+        tr_a = gs.trace(point_a)
+        tr_b = gs.trace(point_b)
+
         point_a_sqrt = apply_func_to_eigvalsh(point_a, gs.sqrt)
 
-        Lc, Qc = gs.linalg.eigh(point_a_sqrt @ point_b @ point_a_sqrt)
-        cross_term = Matrices.mul(
-            Qc,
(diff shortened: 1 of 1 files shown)


luisfpereira (approved):
Thank you @dherrera1911!

I confirm that this new implementation is faster (got 1.5x speedup in some quick tests I've performed).

I've just updated `c_eigvals` -> `c_sq_eigvals` for more clarity.

PR #2037, opened 2024-12-14
Simplify MultinomialMetric implementation by moving transform methods out

luisfpereira:
This PR simplifies `MultinomialMetric` implementation by moving transform methods (i.e. simplex <-> hypersphere, and respective tangents) out.

Main gains (for now) are in the testing code, where particular tests for the transforms where fully removed and more robust `Diffeo`-based tests are now used (NB: diffeo is considered between simplex and its image under componentwise square root; naming may be a bit misleading).

Initially I though I could implement this metric as a pullback diffeo metric. Later realized it is not the case.
Though `exp`, `log`, and `geodesic` are pulled back methods, which makes me wonder if we're missing an intermediate abstraction (is there any notion of pullback connection? - initially quick search seems to suggest [there is](https://math.stackexchange.com/a/2458292/663177)).

I should probably read the reference in the docstrings, but if you have any thoughts on it we can probably move faster.

Edit: test failures are related to latest release of `scikit-learn`. Fixing it in another PR.


alebrigant (commented):
The MultinomialMetric is actually a pullback metric: it is the pullback of the Euclidean metric on the positive part of the sphere by a diffeo that takes square roots, (see p.10 in the [paper the information geometry module of geomstats](https://dl.acm.org/doi/pdf/10.1145/3627538?casa_token=gu92hTKyhggAAAAA:pdH_RSRvtbGFDTU-omp4D1Wr5Xgy5O_AsIUehtTQhtCaw9fVQ1nnj6wEb1469XL5UFvlCpO3IWQngQ)). So it would be very relevant to use the pullback implementation here.

luisfpereira:
Thanks @alebrigant, this is the information I was really looking for.

Two extra questions, before moving forward with this:

1. if we consider to implement the positive part of the hypersphere for this purpose, how should we handle the boundary? what I really want to ask is: should we avoid implementing the "positive sphere" and for implementation purposes see the "positive sphere" as the image of the multinomial under this diffeomorphism? (that I can now rename to `SimplexToPositiveHypersphere`) in other words, would it be confusing if I use the `Hypersphere` as image space of the diffeomorphism? if yes, maybe renaming `image_space` to `codomain` helps?

2. instead of implementing the (positive) `Hypersphere` of radius `r`, I should be able to use a scaled product metric on the `Hypersphere` of radius `r=1`, right?

alebrigant:
> Thanks @alebrigant, this is the information I was really looking for.
> 
> Two extra questions, before moving forward with this:
> 
>     1. if we consider to implement the positive part of the hypersphere for this purpose, how should we handle the boundary? what I really want to ask is: should we avoid implementing the "positive sphere" and for implementation purposes see the "positive sphere" as the image of the multinomial under this diffeomorphism? (that I can now rename to `SimplexToPositiveHypersphere`) in other words, would it be confusing if I use the `Hypersphere` as image space of the diffeomorphism? if yes, maybe renaming `image_space` to `codomain` helps?
> 
>     2. instead of implementing the (positive) `Hypersphere` of radius `r`, I should be able to use a scaled product metric on the `Hypersphere` of radius `r=1`, right?

1. I think it's fine to use `Hypersphere` as the image space, as far as I can tell.
2. Exactly, the inner product on the `Hypersphere` of radius 1 should just be multiplied by `r ** 2` (or equivalently, the norm should be multiplied by `r`).

luisfpereira:
Great, @alebrigant, thanks! I've updated the code based on your comments.

Just one last question: is the sectional curvature given by i) `2*sqrt(n_draws)` or ii) `1 / (2*sqrt(n_draws))^2`?

If i), then our `ScalarProductMetric` has an issue: we consider `sectional_curvature` to be part of `_INVERSE_LIST`, i.e. we divide the curvature of the original space (`Hypersphere`) by the scale (`(2*sqrt(n_draws))^2`).

If ii), then there's a typo/mistake in the paper.

alebrigant:
It's ii), there is indeed a mistake in the paper. Thanks for spotting this @luisfpereira !

alebrigant:
It's ii), there is indeed a mistake in the paper. Thanks for spotting this @luisfpereira !

PR #2057, opened 2025-09-17
Quick backend update and code cleaning

luisfpereira:
This PR updates `torch.array` and improves uses of array and stack.


luisfpereira:
Merging now. There's some warnings that must be tackled soon.

PR #2068, opened 2025-12-10
Clarify definition of phylogenetic tree and adjust TreeTopology, Tree, TreeSpace, and BHVMetric accordingly

### LINKED ISSUE #2070: Tree at origin of BHV space (star tree) cannot be initialised
bhv_space = TreeSpace(n_labels=5)

lengths = gs.array([3, 1])
top_tree = Tree(gs.array([Split({3, 4}, {0, 1, 2}), Split({0, 1}, {2, 3, 4})]), lengths)
bottom_tree = Tree(
    gs.array([Split({0, 4}, {1, 2, 3}), Split({2, 3}, {0, 1, 4})]), lengths
)

gg = bhv_space.metric.geodesic(top_tree, bottom_tree)
print(gg)
print(gg(.501))
print(gg(.5))

IndexError                                Traceback (most recent call last)
Cell In[10], line 12
     10 print(gg)
     11 print(gg(.501))
---> 12 print(gg(.5))

File ~/Documents/ucsb/geomstats/geomstats/geomstats/geometry/stratified/bhv_space.py:633, in GTPSolver.geodesic.<locals>.<lambda>(t)
    626     return [fnc(t) for fnc in fncs]
    628 fncs = [
    629     self._geodesic_single(initial_point_, end_point_)
    630     for initial_point_, end_point_ in zip(initial_point, end_point)
    631 ]
--> 633 return lambda t: _vec(t, fncs=fncs)

File ~/Documents/ucsb/geomstats/geomstats/geomstats/geometry/stratified/bhv_space.py:624, in GTPSolver.geodesic.<locals>._vec(t, fncs)
    622 def _vec(t, fncs):
    623     if len(fncs) == 1:
--> 624         return fncs[0](t)
    626     return [fnc(t) for fnc in fncs]

File ~/Documents/ucsb/geomstats/geomstats/geomstats/geometry/stratified/bhv_space.py:599, in GTPSolver._geodesic_single.<locals>.geodesic_(t)
    596 if isinstance(t, (float, int)):
    597     t = gs.array([t])
--> 599 return TreeBatch([geodesic_t(t_) for t_ in t])
...
--> 108         partition=(tuple(splits[0].part1.union(splits[0].part2)),),
    109         split_sets=(splits,),
    110     )

IndexError: list index out of range

abertics:
## Checklist

## Description

The `Tree` class uses splits and edge lengths to represent unrooted phylogenetic trees without pendant edges.

Following Semple and Steel 2003 and [Lueg, 2023](https://ediss.uni-goettingen.de/handle/11858/14443), an unrooted phylogenetic tree is a tuple of edge splits and edge lengths; $T = (E, l)$ where $E \subset e\_compatible$ and $l \in (0, \infty)^{|E|}$.

Such a tree has N labels, $L_0=\{0, 1, ..., N-1\}$. Those N labels yield a set of possible edges, $\ e = \{A\|B : L_0 = A \bigcup B, \|A\|, \|B\| ≥ 2\}$. Note that this definition excludes pendant edges (those touching a leaf), yielding $\|e\| = 2^{N-1} - N - 3$ possible interior edges of which a maximum of $N-3$ are compatible at once.

BHV space is defined for $N\ge 4$, as any less will not yield any interior edges and thus be boringly Euclidean.

People working with phylogenetic trees might have additional requirements, and the way the code is now set up allows these to be added on top of the existing class structure, rather than needing to set up weird internal modifications. The possible additional constraints concern pendant edges and roots.

Pendant edges: No pendant edges is now the default, because often it is the distance between splits that matters in phylogenetic trees, not the distance between the split and the leaf, because the leaf is at an arbitrary time, the present. Too add lengths to the pendant edges we will create a space that is the Cartesian product of BHV space and Euclidean space of dimension $N$. Geodesics in this new space are just geodesics in pendant-less BHV space, with extra straight line segments in $(0, \infty)^{N}$ added on. This will be done in a subsequent PR. (add issue and tag?)

Roots: Unrooted is the default. There is a bijection between unrooted phylogenetic trees with $N$ labels and rooted ones with $N-1$ labels (just add a leaf, label it 0, and call it a day). We can add more explicit support for this during class construction in a subsequent PR. (add issue and tag?)

## Issue

Solves the following issues:
- TreeTopology allowed empty splits.
- Tree allowed non-positive lengths.
- Impossible to create star tree at origin -- Closes #2070  
- The generate_random_tree function produced trees that are not phylogenetic trees. It also allowed for trees with an empty set on one side of the split. That split, while a legal split, does not make sense in the context of edge length and should not exist.

## Additional context

TODO:
- I dislike that there is a separate trees.py file. Feels v unnecessary.
- I dislike generate_splits/delete_splits being separate. Maybe should be part of ForestTopology?
- Unsure how to write tests for this. Should add one verifying that only_internal_edges=True and only_internal_edges=False both work, but the current test for bhv_space and tree generation is not clear.
- Would like to do Cartesian product to handle pendant edges. This should be another PR.
- Should random_point be able to generate the star ever?? The probability of that is technically zero? Or is it? (Seems different if I think in terms of topologies or in terms of spaces)


geomstats/geometry/stratified/bhv_space.py
```diff
@@ -56,6 +62,13 @@ def generate_random_tree(n_labels, p_keep=0.9, btol=1e-8):
     initial_splits = generate_splits(labels)
     splits = delete_splits(initial_splits, labels, p_keep, check=False)
 
+    if only_internal_edges:
```
luisfpereira: I think this should go before l63.

Better yet, would be to have it within `generate_splits`, as it would also solve the bug for `wald`.

A last question is: do we want to provide a boolean option or should we assume vertices of degree 1 are forbidden?

luisfpereira (changes requested):
Almost there @abertics.

Just a couple of comments to address (sorry the pedantism).

geomstats/geometry/stratified/bhv_space.py
```diff
@@ -56,6 +62,13 @@ def generate_random_tree(n_labels, p_keep=0.9, btol=1e-8):
...
+        splits[:] = [
```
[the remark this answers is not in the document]
abertics: This is a weird Python-specific operation called slice assignment, used to modify the contents of a list in-place, also changing the original object. I'll remove it, because I'm not sure it matters.

geomstats/geometry/stratified/bhv_space.py
abertics: I agree it should go within generate_splits. But I would prefer to add an argument like excluded_degrees, that is a list of positive integers. Then in the logic of creation we exclude the possibility of generating splits resulting in degrees we dislike. The problem is it is doable but non-trivial to find degree of node from the splits list. In practice, we will probably only care about excluding degree 1 and degree 2 nodes (xtree). 

This whole tree debacle should be an issue so that we can address it in more depth.

geomstats/geometry/stratified/bhv_space.py
```diff
@@ -94,11 +72,33 @@ class TreeTopology(ForestTopology):
         ``False``.
     """
 
-    def __init__(self, splits):
-        super().__init__(
-            partition=(tuple(splits[0].part1.union(splits[0].part2)),),
-            split_sets=(splits,),
-        )
+    def __init__(self, splits, n_labels=None):
+        if not splits:
```
luisfpereira, line 77: undefined behavior if `n_labels=None`.

geomstats/geometry/stratified/bhv_space.py
```diff
@@ -94,11 +72,33 @@ class TreeTopology(ForestTopology):
...
+            super().__init__(
+                partition=(set(range(n_labels)),),
+                split_sets=(splits,),
+            )
+        else:
+            TreeTopology._check_no_pendant_edges(splits)
```
luisfpereira: why not `self`?

geomstats/geometry/stratified/bhv_space.py
```diff
@@ -94,11 +72,33 @@ class TreeTopology(ForestTopology):
...
-            partition=(tuple(splits[0].part1.union(splits[0].part2)),),
-            split_sets=(splits,),
-        )
+    def __init__(self, splits, n_labels=None):
+        if not splits:
+            super().__init__(
+                partition=(set(range(n_labels)),),
+                split_sets=(splits,),
+            )
+        else:
```
luisfpereira, line 85: should check `n_labels` and splits compatibility if `n_labels is not None`?

luisfpereira (changes requested):
Thanks @abertics. The docstrings are really good!

Just an initially set of comments to fix something I've found while translating the notebook tests to pytest tests.

luisfpereira:
Also, can we add the beautiful description of this PR to the docstrings of the appropriate classes?

geomstats/geometry/stratified/bhv_space.py
luisfpereira: this applies in other parts of the code too

my recommendation is to use `self`, even if it is a `staticmethod`

luisfpereira:
Just translated the notebook into tests.

There's something failing sometimes. Run `pytest tests/tests_geomstats/test_geometry/test_stratified/test_bhv_space.py::TestBHVMetric`.

After we identify the issue and fix it, it is mostly ready to merge (great if we can update the docstrings as mentioned above before).

geomstats/geometry/stratified/bhv_space.py
abertics: do you want me to assert it is not None?

abertics: I'll try something and see if it is satisfactory

geomstats/geometry/stratified/bhv_space.py
abertics: i did it static bc i was copying 

    def __init__(self, splits, lengths, n_labels=None):
        Tree._check_valid_lengths(lengths, len(splits))

abertics: I think it makes more sense to leave as Tree. or Tksladfj. rather than self, so its clear it isn't using an instance data. id say self makes sense if we want subclasses to be able to override, but in this case it's just helping us privately so meh.

abertics:
I have added n_labels None logic. I've added stuff from PR description to doc string.

I have not changed static method call. I have also not fixed the test, because I can't seem to run the test with it failing.

luisfpereira (approved):

PR #2088, opened 2026-02-05
Skip BHV space tests when networkx is missing

adithya17032006:
This PR updates the BHV space tests to be skipped when the optional
dependency `network` is not installed.

This avoids ImportError during pytest collection on minimal or
Windows environments where optional dependencies are not present
by default.


luisfpereira:
Thanks for the contribution @adithya17032006.

I'm skipping merging this for now because `networkx` is central to the implementation of BHV space and tests need to have it installed to run properly.

PR #2106, opened 2026-06-03
Rename `geomstats.stratified` to `metric_geometry`

luisfpereira:
Due to changes in scope, it seems sensible to rename `geometry.stratified` to `metric_geometry`, which is the umbrella term that contains geodesic metric spaces.


luisfpereira:
@abertics, what do you think?

PR #2124, opened 2026-08-13
bug: raise clear error in log for antipodal points

### LINKED ISSUE #1218: In Hypersphere, SOn and Stiefel, Log of antipodal points should throw an error

sdoygb:
## Summary

Fixes #1218: the logarithm is not well-defined for antipodal points, but `Hypersphere.log` silently returned zero vectors and `Stiefel.log` failed with an obscure complex-matrix error.

Now both raise a clear `ValueError`, consistent with the existing `SOn.log` behaviour (which already checks `are_antipodals`).

## Changes

- **Hypersphere.log**: check `inner_product(base_point, point) ~ -1` before the Taylor expansion (which diverges at antipodal points: `sinc(pi) = tanc(pi) = 0`)
- **Stiefel.log**: check `X^T Y ~ -I_p`, complementing the existing determinant/sheet check (which only covers the `p == n` disconnected case)
- **Tests**: `pytest.raises(ValueError, match="antipodal")` for both spaces

## Verification

- Manual repro before fix: `Hypersphere(2).log(-x, x)` -> `[0, 0, 0]` (silent); `Stiefel(3, 2).log(-X, X)` -> `ValueError: Non-neglible imaginary part` (obscure)
- After fix: both raise `ValueError: The Logarithm is not well-defined for antipodal points`
- Non-antipodal regression: `exp(log(y, x)) == y` preserved
- Full test files: **654 passed, 8 skipped** (test_hypersphere.py + test_stiefel.py)

### DIFF
--- a/geomstats/geometry/hypersphere.py
+++ b/geomstats/geometry/hypersphere.py
@@ -766,6 +766,11 @@ class HypersphereMetric(RiemannianMetric):
             of point at the base point.
         """
         inner_prod = self._space.embedding_space.metric.inner_product(base_point, point)
+        if gs.any(gs.isclose(inner_prod, -1.0)):
+            raise ValueError(
+                "The Logarithm is not well-defined for antipodal points: "
+                f"{point} and {base_point}."
+            )
         cos_angle = gs.clip(inner_prod, -1.0, 1.0)
         squared_angle = gs.arccos(cos_angle) ** 2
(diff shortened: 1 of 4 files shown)


luisfpereira:
Thanks @sdoygb! Merging.

PR #2125, opened 2026-08-28
Add WarpedProductMetric and WarpedProductManifold

### LINKED ISSUE #1977: Introduce warped product metrics
Warped product metrics should be fairly straightforward, building on the work already done in `ProductRiemannianMetric`.

Since `ProductRiemannianMetric` already has the machinery needed to gather the correct information from the different factors and, rather than implementing the `_pool_outputs_from_function` method for `_IterateOverFactorsMixins`, the class pools the information using a specific algorithm for each method, it will mostly be about making a slight change to those pooling methods to rescale the result in certain directions.

The `geodesic` method is one area that calls for its own code. We can just leave it to use `Connection.geodesic` but it should be possible to get it to solve a modified equation on the base and pool that with the standard geodesic equation on the fibre. This would be better, if we have an exact solution in place for the fibre already.

Given the similarity with `ProductRiemannianMetric`, it seems best to add `WarpedProductMetric` to `product_manifold.py`.

I plan to work on this over the next month or so. I have a student studying a particular example for an undergraduate dissertation and, all going well, he should be able to use this to enhance his project. The result might be his own contribution to the package or perhaps a nice notebook to add to our examples.

sdoygb:
### Description

Implements the warped product metric `g = g_B + f² g_F` on `M = B ×_f F`, following the approach outlined by @johnharveymath in #1977.

**New classes** in `geomstats/geometry/product_manifold.py`:

- `WarpedProductMetric(ProductRiemannianMetric)`:
  - `metric_matrix` / `inner_product` / `squared_norm` rescale the fibre block by the square of the warping function evaluated at the base point;
  - `exp`, `log`, `dist` and `geodesic` are computed through the generic connection machinery (Christoffel symbols derived from `metric_matrix` by automatic differentiation, ODE integration for exp, shooting for log), so the base–fibre coupling of the warped geodesic equation is correctly accounted for;
  - the warping function is a callable `f: B → ℝ₊` that may depend on the base point (constant and smooth functions supported).
- `WarpedProductManifold(base, fiber, warping_function)`: convenience class equipping the product with the warped metric.

**Tests** (`tests/tests_geomstats/test_geometry/test_warped_product_metric.py`):

- metric structure (inner product, metric matrix, norm scaling, batching) valid on any backend;
- under an autodiff backend (`autograd`/`pytorch`), closed-form validations:
  - the Euclidean plane in polar-like coordinates `dr² + r²dy²` (warping `f(r) = r`): exp/log/geodesics match straight lines in Cartesian coordinates;
  - the round-sphere metric `dθ² + sin²θ dφ²` (warping `f(θ) = sin θ`): meridian and equator geodesics;
  - constant warping reduces to a product (linear geodesics);
  - exp∘log roundtrips, geodesic length = initial speed, and the base–fibre coupling (a pure-fibre velocity curves the base component, unlike a product metric).

### Notes / limitations

- Requires an autodiff backend for exp/log/geodesics (clear error otherwise); only vector points (`point_ndim == 1`) are supported, as for the generic geodesic equation.
- The connection is derived from the factor metrics' `metric_matrix`, which must be expressed in intrinsic coordinates of the factors (e.g. the extrinsic `HypersphereMetric` is not suitable as a factor).
- The `geodesic` method uses the generic solver; a dedicated solver exploiting the split base/fibre structure (as suggested in #1977) is left as a follow-up.

### Checklist

- [x] Clear and explanatory title
- [x] Vectorized implementation
- [x] Appropriate unit tests added
- [x] Tests pass on `autograd` and `numpy` backends (18 passed / 9 passed + 9 skipped)
- [x] PEP8 / ruff clean (only pre-existing SIM103 notes untouched)
- [x] Properly documented

Closes #1977

### DIFF
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -20,22 +20,15 @@ COMPLEX_OBJECTS = (ComplexRiemannianMetric, ComplexManifold)
 
 
 def _factor_is_complex(factor):
-    if (
-        isinstance(factor, COMPLEX_OBJECTS)
-        or hasattr(factor, "underlying_metric")
+    return isinstance(factor, COMPLEX_OBJECTS) or (
+        hasattr(factor, "underlying_metric")
         and isinstance(factor.underlying_metric, COMPLEX_OBJECTS)
-    ):
-        return True
-
-    return False
+    )
 
 
 def _has_mixed_fields(factors):
     bools = [_factor_is_complex(factor) for factor in factors]
-    if len(set(bools)) == 2:
-        return True
-
-    return False
+    return len(set(bools)) == 2
 
 
 def _all_equal(arg):
--- a/geomstats/geometry/product_manifold.py
+++ b/geomstats/geometry/product_manifold.py
@@ -9,8 +9,11 @@ import geomstats.backend as gs
 import geomstats.errors
 from geomstats.geometry.complex_manifold import ComplexManifold
 from geomstats.geometry.complex_riemannian_metric import ComplexRiemannianMetric
+from geomstats.geometry.connection import Connection
 from geomstats.geometry.manifold import Manifold
 from geomstats.geometry.riemannian_metric import RiemannianMetric
+from geomstats.numerics.geodesic import ExpODESolver, LogShootingSolver
+from geomstats.numerics.ivp import GSIVPIntegrator
 from geomstats.vectorization import get_batch_shape
 
 COMPLEX_OBJECTS = (ComplexRiemannianMetric, ComplexManifold)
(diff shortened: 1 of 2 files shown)


sdoygb:
CI summary for reviewers (head `ac1835f`):

**Green** ✅
- numpy + pytorch `tests/tests_geomstats/` (incl. the new warped product tests)
- all `tests/tests_scripts` jobs, all codecov checks
- The earlier pytorch failure in `test_optimization.py` (TorchLBFGS/Adam minimize) was flaky — it passes on re-run (14586 passed, 0 failed).

**Red** ❌ (both pre-existing on `main`, unrelated to this PR)
- `lint`: repo-wide SIM103 flags from the current ruff version (1298 errors across historical files; `main` fails identically). The files touched by this PR are lint-clean.
- autograd `tests/tests_geomstats/`: single failure in `test_backend/test_dtype.py` (`AttributeError: 'list' object has no attribute 'dtype'`) — a numpy-2/autograd compatibility issue that also fails on `main`. The new warped product tests pass under autograd (14393 passed).

sdoygb:
### Upstream issues analysis (why the red checks are pre-existing)

I dug into the three red items so reviewers don't have to. **All three also fail (or flake) on `main` and are unrelated to this PR's diff.**

#### 1. `lint` — ruff version drift (1298–1300 errors, also failing on `main`)

- The lint job runs `chartboost/ruff-action@v1` with **no pinned ruff version**, so it uses the latest ruff. Recent ruff releases expanded the **default rule set**: e.g. C408 ×974, LOG015 ×92, RUF012 ×75, PLR0402 ×42, plus UP/TRY/SIM/RUF100 flags — rules the repo's historical code was never written against.
- The repo config only does `extend-select = ["I", "D"]` and never restricts `select`, so the expanded defaults apply.
- Verified locally: `ruff check --isolated --select E,F .` → **0 errors in this PR's files**; `ruff check .` (repo config, ruff 0.16.3) → **1300 errors** including files untouched by anyone recently.
- Suggested upstream fixes: pin the ruff version (e.g. `version: 0.8.x` in the action or a `ruff` dev pin), and/or set an explicit `select` in `pyproject.toml`. Happy to open a follow-up PR for either.

#### 2. autograd `tests/tests_geomstats/` — `test_dtype.py` `where` (also failing on `main`)

- Single failure: `test_func_out_dtype['where'-[True, False]-20.0-20.0-{}-<class 'numpy.float64'>]` → `AttributeError: 'list' object has no attribute 'dtype'`.
- Root cause: under numpy 2.x, `autograd.numpy.where([True, False], 20.0, 20.0)` returns a **plain Python list** instead of an ndarray (verified locally with numpy 2.5.2: `type(...) is list`, `isinstance(..., np.ndarray) is False`). The autograd backend inherits this from `autograd` 1.7.0 (2023), which predates numpy 2.0 and was never updated.
- This is an upstream `autograd` incompatibility; the geomstats autograd backend needs either a `where` shim or an autograd fork/update.

#### 3. pytorch `test_optimization.py` — flaky (passed on re-run)

- On the first CI run, `test_minimize` failed for `TorchLBFGS` and `TorchAdam` (convergence `assertAllClose(res.x, expected, atol=atol)` on `_sum_squares`). On the re-run (head `ac1835f`) the entire pytorch suite passed: **14586 passed, 0 failed**. The failures do not touch this PR's code paths (product_manifold / warped tests) and are best explained by optimizer-convergence variance in the CI environment.

**Bottom line:** everything attributable to this PR passes on all three backends (numpy/pytorch fully green; autograd 14393 passed with the single pre-existing `where` failure). The 2 stable red checks reproduce on `main` and have clear upstream root causes above.

